PR #46, opened 2023-05-10
Fix repository links in Cargo.toml files

lucajone:
The repository links as shown on `crates.io` (e.g. on https://crates.io/crates/cedar-policy) are broken at the moment (they point under `github.com/cedar` instead of `github.com/cedar-policy`). This change fixes that!


aaronjeline (commented):
Can you increment the patch number on these by one so I can re push to crates?

lucajone:
I can indeed!

aaronjeline (approved):

PR #50, opened 2023-05-12
Add method to extract action entities from schema

john-h-kastner-aws:
The schema contains definitions for all action entities, it was previously necessary to duplicate action entities in an entity store before using the action groups in policies.

* Add a ` pub fn action_entities(&self) -> cedar_policy_core::entities::Result<Entities>` to the internal `ValidatorSchema` type
* Expose this publicly by adding `pub fn action_entities(&self) -> Result<Entities, entities::EntitiesError>` to the public `Schema` type
* Update the CLI to use this new API to load actions form the schema into the entity store before evaluating policies

### DIFF
--- a/cedar-policy-validator/src/schema.rs
+++ b/cedar-policy-validator/src/schema.rs
@@ -23,8 +23,8 @@
 use std::collections::{hash_map::Entry, HashMap, HashSet};
 
 use cedar_policy_core::{
-    ast::{Eid, EntityType, EntityUID, Id, Name},
-    entities::JSONValue,
+    ast::{Eid, Entity, EntityType, EntityUID, Id, Name, RestrictedExpr},
+    entities::{Entities, JSONValue, TCComputation},
     parser::{err::ParseError, parse_name, parse_namespace},
     transitive_closure::{compute_tc, TCNode},
 };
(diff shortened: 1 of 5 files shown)


cedar-policy/src/api.rs
```diff
@@ -589,6 +589,12 @@ impl Schema {
             file,
         )?))
     }
+
+    /// Extract from the schema an `Entities` containing the aciton entities
```
khieta, line 593: aciton -> action

aaronjeline (approved):

cedar-policy-validator/src/schema.rs
```diff
@@ -1182,6 +1215,36 @@ impl ValidatorSchema {
...
+        // Invert the action hierarchy to get the ancestor relation expected for
+        // the `Entity` datatype instead of descendent as stored by the schema.
+        // We could store the un-inverted `memberOf` relation for each action,
+        // but I judge that the current implementation is actually less error
+        // prone, as it minimizes the threading of data structures through some
+        // complicated bits of schema construction code, and avoids computing
+        // the TC twice.
+        let mut action_ancestors: HashMap<&EntityUID, HashSet<EntityUID>> = HashMap::new();
+        for (action_euid, action_def) in &self.action_ids {
+            for descendent in &action_def.descendants {
```
khieta, line 1229: Let's consistently use one of descendent or descendant 😄 I personally prefer the "a" but I think both are technically valid.

cedar-policy-validator/src/schema.rs
```diff
@@ -1182,6 +1215,36 @@ impl ValidatorSchema {
...
+                    .insert(action_euid.clone());
+            }
+        }
+
+        Entities::from_entities(
+            self.action_ids.iter().map(|(action_id, action)| {
+                Entity::new(
+                    action_id.clone(),
+                    action.attributes.clone(),
+                    action_ancestors.remove(action_id).unwrap_or_default(),
```
khieta, line 1242: Do you expect `action_id` to end up in `action_ancestors`? I thought it was already excluded from `action_def.descendants`?

khieta (approved):

cedar-policy-validator/src/schema.rs
john-h-kastner-aws: `remove` is used here to move the ancestors set for `action_id` or of the hash map so that I can avoid a clone.

khieta (approved):

PR #56, opened 2023-05-18
Arithmetic overflow validation

mattmccutchen-amazon:
*Issue #, if available:* #63

*Description of changes:*
Enhance the validator to detect possible arithmetic overflow errors in policies based on bounds on `Long` inputs declared in the schema.  (Draft; more details will be provided here when the work is further along.)

*Known issues (not an exhaustive list; see also the TODO and similar comments added to the code):*

- [ ] Probably change the term "arithmetic overflow" to "integer overflow" to match the existing evaluation error message (even if I personally think that "arithmetic overflow" is more descriptive).
- [ ] Add unit tests.


mattmccutchen-amazon:
Anwar asked me to mark this PR "ready for review" because I am seeking review even though I haven't fixed the known issues to the degree I think would normally be expected for a "final" review.  I guess it's up to our team what we use GitHub's "ready for review" status to mean.

cedar-policy-core/src/entities/json/schema_types.rs
```diff
@@ -166,7 +185,18 @@ impl std::fmt::Display for SchemaType {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Self::Bool => write!(f, "bool"),
-            Self::Long => write!(f, "long"),
+            Self::Long { min, max } =>
```
andrewmwells-amazon, line 188: I think you could do:
`Self::Long{LONG_MIN, LONG_MAX} => write!(f, "long with no bounds")`

cedar-policy-validator/overflow-checking-example/policies.txt
```diff
@@ -0,0 +1,34 @@
+// Note: 2^63 == 9223372036854775808; Long can be [-2^63, 2^63 - 1]
+// 1 <= principal.x8 <= 8
+
```
andrewmwells-amazon, line 3: I'd use constants. Something like `MAX_LONG_MINUS_EIGHT`, `MAX_LONG_MINUS_NINE`, `MIN_LONG_PLUS_EIGHT` and `MAX_LONG_PLUS_NINE`.

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -275,6 +275,107 @@ impl SchemaType {
...
+            bounds_opt: Some(SchemaLongBounds {
+                min: i64::MIN,
+                max: i64::MAX,
+            }),
+        })
+    }
+}
+
+// In the schema file, we want the bounds of a Long attribute to be given as
+// "min" and "max" fields at the same level as `"type": "Long"` with both or
```
andrewmwells-amazon, line 290: I'm not convinced they need to be at the same level. I think the code would be cleaner and the interface would be fine if the schema were written as something like:
```
... "foo": {
  "type": "Long",
  "bounds": {
    "min": 0,
    "max": 8
  }
}

File: cedar-policy-validator/src/schema_file_format.rs [hunk 4]
```diff
@@ -275,6 +275,107 @@ impl SchemaType {
...
+            min: i64::min(bound1, bound2),
+            max: i64::max(bound1, bound2),
+        })
+    } else {
+        None
+    };
+    Ok(SchemaTypeVariant::Long(SchemaLongDetails { bounds_opt }))
+}
+
+// TODO: It looks like this is unused except by arbitrary_schematype_size_hint.
```
andrewmwells-amazon, line 377: Tagging @cdisselkoen, @shaobo-he-aws and @amzn-aeline

cedar-policy-validator/src/typecheck.rs
```diff
@@ -1857,22 +1960,89 @@ impl<'a> Typechecker<'a> {
...
+    // was convenient to factor out of the overflow checks as they are currently
+    // structured. The code structure will probably have to change as we polish
+    // the overflow analysis.
+    fn arith_overflow_check_finish(
+        new_min_opt: Option<i64>,
+        new_max_opt: Option<i64>,
+        any_arg_can_be_any: bool,
+        type_errors: &mut Vec<TypeError>,
+        arith_expr: &Expr,
+    ) -> (
```
andrewmwells-amazon, line 2022: [nit] I prefer returning a struct with named fields to returning a tuple that requires comments to explain the fields.

andrewmwells-amazon:
> Though granted, removing this particular piece of code may be easy enough that I could just do it in this PR.
This is already a largish PR so I'd vote for doing that separately.

cedar-policy-validator/src/schema_file_format.rs
mwhicks1: We should follow what JSON Schema does: https://json-schema.org/understanding-json-schema/reference/numeric.html#range

cedar-policy-validator/src/types.rs
```diff
@@ -401,7 +480,8 @@ impl Type {
                 primitive_type: Primitive::Bool,
             } => Type::json_type("Boolean"),
             Type::Primitive {
-                primitive_type: Primitive::Long,
+                // FIXME: Preserve bounds? What uses this?
```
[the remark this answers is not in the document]
mattmccutchen-amazon: I added the bounds here, but it looks like for the error messages we currently have, printing the bounds may just be distracting and not helpful.  For example, for an erroneous expression `true > 1`, instead of:
```
Unexpected type. Expected one of [{"type":"Long"}] but saw {"type":"True"}
```
we now get:
```
Unexpected type. Expected one of [{"type":"Long","min":-9223372036854775808,"max":9223372036854775807,"can_be_any":true}] but saw {"type":"True"}
```
(I see two other error messages that include types, but I'm not aware of a way to trigger either of them with a `Long` type.)  So I'm thinking the right thing to do for the user is leave out the bounds, even though it seems inconsistent design-wise.  If in the future, this function is used in a context in which the bounds are helpful, we would reevaluate.

cedar-policy/src/api.rs
```diff
@@ -3904,6 +3904,127 @@ mod schema_based_parsing_tests {
...
+                    }
+                },
+            },
+            "actions": { "view": { } }
+        }}
+        ))
+        .expect("should be a valid schema");
+
+        // We should be able to parse a set of longs. There was previously a bug
+        // where the set of rejected as heterogenous because the type Long with
```
andrewmwells-amazon, line 4630:
```suggestion
        // where the set was rejected as heterogenous because the type Long with
```

cedar-policy/src/api.rs
```diff
@@ -3904,6 +3904,127 @@ mod schema_based_parsing_tests {
...
+    fn long_in_bounds() {
+        let schema = Schema::from_json_value(json!(
+        {"": {
+            "entityTypes": {
+                "Employee": {
+                    "memberOfTypes": [],
+                    "shape": {
+                        "type": "Record",
+                        "attributes": {
+                            "num": { "type": "Long", "min": 3, "max": 4 }
```
andrewmwells-amazon, line 4546: Not related to this PR, but what did we do regarding making `min` and `max` optional but requiring both if one is given?

andrewmwells-amazon (approved):

shaobo-he-aws:
Can we move the examples to the CLI so that users know how to use it? Otherwise it's kind of a hidden feature.

shaobo-he-aws (approved):

Issue #57, opened 2023-05-19
Support partial evaluation in CLI


Neio:
### Category

Other

### Describe the feature you'd like to request

Be able to do partial evaluation in the CLI

### Describe the solution you'd like

Introduce an optional parameter "--partial" in the CLI, so to change the entity mode to partial and be able to do partial authorization and give residual result.

### Describe alternatives you've considered

None

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

shaobo-he-aws:
Hi @Neio, thank you for your interest in Cedar. Partial evaluation is currently an experimental feature and not ready for production use. Hence, we cannot accept any PR related to this feature. I hope you can understand it.

PR #58, opened 2023-05-19
Support partial evaluation in cli

Neio:
*Issue #, if available:* 

Support partial evaluation in CLI #57

*Description of changes:*

Introduce an optional parameter "--partial" in the CLI, so to change the entity mode to partial and be able to do partial authorization and give residual result.

Example of when missing principal in the request:

In default mode:
```
while evaluating policy prototypes access policy, encountered the following error: cannot access attribute of unspecified entity: department
```

In partial mode:
```
while evaluating policy prototypes access policy, encountered the following error: The expression evaluated to a residual: true && (true && (((unknown(<Unspecified>::"principal")["department"]) == "HardwareEngineering") && (5 <= (unknown(<Unspecified>::"principal")["jobLevel"]))))
```

### DIFF
--- a/cedar-policy-cli/src/lib.rs
+++ b/cedar-policy-cli/src/lib.rs
@@ -204,6 +204,8 @@ pub struct AuthorizeArgs {
     /// Time authorization and report timing information
     #[arg(short, long)]
     pub timing: bool,
+    #[arg(long)]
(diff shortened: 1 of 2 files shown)


Neio:
Closing the PR since this feature is not ready for production yet

PR #68, opened 2023-05-24
Don't convert to EST if CST -> AST parse fails

aaronjeline:


shaobo-he-aws (approved):
LGTM. Just a nit-picking comment like #69.

PR #69, opened 2023-05-24
Don't convert to EST if CST -> AST parse fails

aaronjeline:


cedar-policy-core/src/parser.rs
```diff
@@ -56,6 +56,9 @@ pub fn parse_policyset_to_ests_and_pset(
     let mut errs = Vec::new();
     let cst = text_to_cst::parse_policies(text).map_err(err::ParseErrors)?;
     let pset = cst.to_policyset(&mut errs);
+    if pset.is_none() {
```
shaobo-he-aws, line 59: Nit picking here: I think we can use a `if let` here to simplify the logic of this function. For instance,

```Rust
if let Some(pset) = cst.to_policyset(&mut errs) {
  let ests = ...;
  Ok(ests, pset)
} else {
  Err(err::ParseErrors(errs));
}
```

shaobo-he-aws (approved):

PR #70, opened 2023-05-25
allow `check-parse` command to read from stdin

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-policy-cli/src/lib.rs
+++ b/cedar-policy-cli/src/lib.rs
@@ -71,7 +71,7 @@ pub struct ValidateArgs {
 pub struct CheckParseArgs {
     /// File containing the policy set
     #[clap(short, long = "policies", value_name = "FILE")]
-    pub policies_file: String,
+    pub policies_file: Option<String>,
 }
 
 /// This struct contains the arguments that together specify a request.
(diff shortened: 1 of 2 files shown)


aaronjeline (approved):

cedar-policy-cli/src/lib.rs
```diff
@@ -688,30 +680,51 @@ fn read_policy_and_links(
...
 
-fn read_policy_file(filename: impl AsRef<Path>) -> Result<PolicySet> {
-    let src = std::fs::read_to_string(filename.as_ref()).context(format!(
-        "failed to open policy file {}",
-        filename.as_ref().display()
-    ))?;
-    let ps = PolicySet::from_str(&src).context(format!(
-        "failed to parse policies from file {}",
-        filename.as_ref().display()
+fn read_from_file_or_stdin(filename: Option<impl AsRef<Path>>, context: &str) -> Result<String> {
```
andrewmwells-amazon, line 690: Can we add some docs so users can hover over `read_from_file_or_stdin` and `read_from_file` and see why one would be used vs the other? Personally I find it confusing that `read_from_file` calls `read_from_file_or_stdin`

andrewmwells-amazon (changes requested):

andrewmwells-amazon (approved):

cedar-policy-cli/src/lib.rs
```diff
@@ -71,7 +71,7 @@ pub struct ValidateArgs {
 pub struct CheckParseArgs {
     /// File containing the policy set
     #[clap(short, long = "policies", value_name = "FILE")]
-    pub policies_file: String,
+    pub policies_file: Option<String>,
```
cdisselkoen, line 74: A comment about what `None` implies would be helpful here.  (The comment gets incorporated into the `--help` by `clap`)

PR #71, opened 2023-05-25
Removed `ExtensionFunctionOp`

shaobo-he-aws:
*Issue #, if available:*

*Description of changes:*
`ExtensionFunctionOp` used to contain just a `Name` and didn't do anything special. So, I replaced it with just a `Name`.

### DIFF
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -115,7 +115,7 @@ pub enum ExprKind<T = ()> {
     /// Ideally, we find some way to make this non-representable.
     ExtensionFunctionApp {
         /// Extension function to apply
-        function_name: Name,
+        fn_name: Name,
         /// Args to apply the function to
         args: Arc<Vec<Expr<T>>>,
     },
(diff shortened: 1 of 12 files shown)


cedar-policy-core/src/ast/expr.rs
```diff
@@ -530,13 +530,16 @@ impl Expr {
                 arg1.substitute(definitions)?,
                 arg2.substitute(definitions)?,
             )),
-            ExprKind::ExtensionFunctionApp { op, args } => {
+            ExprKind::ExtensionFunctionApp {
+                function_name,
```
andrewmwells-amazon, line 534: [nit] Personally I prefer longer names but to be consistent with internal style I think this should be "fn_name".
Otherwise, maybe rename "args" to "arguments".

andrewmwells-amazon (changes requested):

cedar-policy-core/src/ast/expr.rs
shaobo-he-aws: Done.

aaronjeline (approved):

andrewmwells-amazon (approved):

PR #79, opened 2023-05-25
Partial schema validation

john-h-kastner-aws:
Experimental implementation of partial schema validation as described in https://github.com/cedar-policy/rfcs/pull/8.

The experimental implementation does not include the `additionalMemberOfTypes` and `additionalMemberOf` fields discussed in the RFC. I'm not convinced these are neccecary for the feature as a whole to be useful, and they require additional changes inside existing validation code.

I'm marking this ready for review as an experimental feature, so reviewers do not need to review the semantics of partial schema validation. They should be sure to review:

1. Changes which ensure partial schema validation is in fact experimental, i.e., it's properly placed behind the `partial-schema` feature.
2. That the code changes are should not effect the behavior of the strict validator.

### DIFF
--- a/cedar-policy-validator/src/typecheck/test_utils.rs
+++ b/cedar-policy-validator/src/typecheck/test_utils.rs
@@ -30,7 +30,7 @@ use cedar_policy_core::ast::{EntityType, EntityUID, Expr, ExprShapeOnly, StaticP
 use crate::{
     schema::ACTION_ENTITY_TYPE,
     type_error::TypeError,
-    types::{Attributes, EffectSet, RequestEnv, Type},
+    types::{EffectSet, OpenTag, RequestEnv, Type},
     NamespaceDefinition, ValidationMode, ValidatorSchema,
 };
 
@@ -78,7 +78,7 @@ impl Typechecker<'_> {
                     .parse()
                     .expect("Placeholder type \"Resource\" failed to parse as valid type name."),
             ),
-            context: &Attributes::with_attributes(None),
+            context: &Type::record_with_attributes(None, OpenTag::ClosedAttributes),
             principal_slot: None,
             resource_slot: None,
         };
@@ -153,11 +153,27 @@ pub(crate) fn assert_policy_typechecks_for_mode(
     mode: ValidationMode,
 ) {
     with_typechecker_from_schema(schema, |mut typechecker| {
+        let policy: Arc<Template> = policy.into();
         typechecker.mode = mode;
         let mut type_errors: HashSet<TypeError> = HashSet::new();
-        let typechecked = typechecker.typecheck_policy(&policy.into(), &mut type_errors);
+        let typechecked = typechecker.typecheck_policy(&policy, &mut type_errors);
         assert_eq!(type_errors, HashSet::new(), "Did not expect any errors.");
         assert!(typechecked, "Expected that policy would typecheck.");
+
+        // Ensure that partial schema validation doesn't cause any policy that
+        // should validate with a complete schema to no longer validate with the
+        // same complete schema.
+        typechecker.mode = ValidationMode::Permissive;
+        let typechecked = typechecker.typecheck_policy(&policy, &mut type_errors);
+        assert_eq!(
+            type_errors,
+            HashSet::new(),
+            "Did not expect any errors under partial schema validation."
+        );
+        assert!(
+            typechecked,
+            "Expected that policy would typecheck under partial schema validation."
+        );
     });
 }
 
--- a/cedar-policy-validator/src/typecheck/test_utils.rs
+++ b/cedar-policy-validator/src/typecheck/test_utils.rs
@@ -65,7 +65,7 @@ impl Typechecker<'_> {
     ) -> TypecheckAnswer<'a> {
         // Using bogus entity type names here for testing. They'll be treated as
         // having empty attribute records, so tests will behave as expected.
-        let request_env = RequestEnv::Known {
+        let request_env = RequestEnv::DeclaredAction {
             principal: &EntityType::Specified(
                 "Principal"
                     .parse()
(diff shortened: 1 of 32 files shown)


mwhicks1:
Some comments on the writeup:

> `RequestEnv` is now an enumeration which of either the request context as it was previously defined, or `RequestEnv::Unknown`.

This is a single catchall for handling missing actions, and also missing principals/resources associated with  a particular action. Would it be worthwhile to include a specific action in the cross-product, but associated with an _unknown_ principal or resource, in case the listed `principalTypes` or `resourceTypes` are incomplete?

> An undeclared entity type has an open attributes record, but a declared entity type without an explicitly defined attributes record has a closed attributes record by default. This is a bit weird. 

Why is it weird? Seems natural to me: The fact that you are explicitly declaring the entity in your schema, with an empty record, means you are taking a stand on it. If you want it to be open, you have to declare it as open, in the schema.

Same comment about not including a `memberOfTypes` element -- what you have seems right to me.

cedar-policy-validator/src/types.rs
```diff
@@ -349,27 +424,46 @@ impl Type {
...
+            // UBs of ActionEntities are AnyEntity. So if we have an
+            // ActionEntity here its attrs are known.
             Type::EntityOrRecord(EntityRecordKind::ActionEntity { attrs, .. }) => {
                 attrs.iter().any(|(found_attr, _)| attr.eq(found_attr))
             }
-            // Other record or entity types always might have some attribute
-            // since the type may be the result of a least upper bound which
-            // could have dropped attributes from the type.
-            Type::EntityOrRecord(_) => true,
+            // A record will have an attribute if the attribute is in it's
```
mwhicks1, line 449: it's --> its

cedar-policy-validator/src/types.rs
```diff
@@ -349,27 +424,46 @@ impl Type {
...
+            // attributes map. Records computed as a LUB may have an open
+            // attributes record, but that is handled by the first match case.
+            Type::EntityOrRecord(EntityRecordKind::Record { attrs, .. }) => {
+                attrs.get_attr(attr).is_some()
+            }
+            _ => false,
```
mwhicks1, line 455: What about the `Never` type?

cedar-policy-validator/src/types.rs
```diff
@@ -904,9 +1022,40 @@ impl EntityRecordKind {
         }
     }
 
+    /// Return `true` if this entity or record may have additional undeclared
+    /// attributes.
+    pub(crate) fn has_open_attributes_record(&self, schema: &ValidatorSchema) -> bool {
+        match self {
```
mwhicks1, line 1028: Redundant with code that seems to be doing the same thing above, i.e., `may_have_attr` ? Repeating comment from there: What about `Never` ?

cedar-policy-validator/src/types.rs
```diff
@@ -904,9 +1022,40 @@ impl EntityRecordKind {
...
+                    // additional attributes.
+                    .unwrap_or(true)
+            }),
+        }
+    }
+
     pub(crate) fn get_attr(&self, schema: &ValidatorSchema, attr: &str) -> Option<AttributeType> {
         match self {
-            EntityRecordKind::Record { attrs } => attrs.get_attr(attr).cloned(),
+            EntityRecordKind::Record { attrs, .. } => attrs.get_attr(attr).cloned(),
```
mwhicks1, line 1058: But if you have additional field types, shouldn't you possibly return `Never` for the attribute?

cedar-policy-validator/src/types.rs
```diff
@@ -934,8 +1083,13 @@ impl EntityRecordKind {
...
         use EntityRecordKind::*;
         match (rk0, rk1) {
-            (Record { attrs: attrs0 }, Record { attrs: attrs1 }) => Some(Record {
+            (Record { attrs: attrs0, .. }, Record { attrs: attrs1, .. }) => Some(Record {
                 attrs: Attributes::least_upper_bound(schema, attrs0, attrs1),
+                // We could do an explicit check here to see if `attr0 == attr1`
+                // to keep `false` for `open_attributes`, but this result
+                // is already achieved due to the subtype check done by
+                // `Type::least_upper_bound`.
+                open_attributes: true,
```
mwhicks1, line 1092: I'm not sure why you don't want to be precise. If I have an `if/then/else` and both branches return a record literal, I'd like to know that the record literal is closed. That way I'll catch `(if * then { f: 1 } else { f: 2 }).g` in partial schema validation as an error.

mwhicks1 (commented):
I didn't review all the testing code, just to get this posted sooner. Will do that on the next review.

john-h-kastner-aws (commented):
I'm making some tweaks to `in`, so I'll look at the comments there once I'm done with that

cedar-policy-validator/src/typecheck/test_utils.rs
```diff
@@ -153,11 +153,27 @@ pub(crate) fn assert_policy_typechecks_for_mode(
...
         typechecker.mode = mode;
         let mut type_errors: HashSet<TypeError> = HashSet::new();
-        let typechecked = typechecker.typecheck_policy(&policy.into(), &mut type_errors);
+        let typechecked = typechecker.typecheck_policy(&policy, &mut type_errors);
         assert_eq!(type_errors, HashSet::new(), "Did not expect any errors.");
         assert!(typechecked, "Expected that policy would typecheck.");
+
+        // Ensure that partial schema validation doesn't cause any policy that
+        // should validate with a complete schema to no longer validate with the
+        // same complete schema.
```
khieta, line 165: Useful property. We should set up a target for partial validation in `cedar-spec` that checks properties like this.

khieta (approved):
High-level comment: it would be helpful for review to split this PR into two: one for incidental refactoring and one for changes that need to be behind the partial validation flag.

But overall looks fine, so approving as-is.

shaobo-he-aws (approved):

Issue #73, opened 2023-05-25
Better error message for EntityType not specified in Schema (due to namespace)


andrewmwells-amazon:
### Category

Cedar validation features

### Describe the feature you'd like to request

In this slack thread: https://cedar-policy.slack.com/archives/C0547KH7R19/p1684968294619939

A user found a confusing error when he had a namespaced schema but his entities file didn't use the namespaced type.

### Describe the solution you'd like

We should warn if types aren't declared in the schema and we should suggest a user likely meant a namespaced type if a matching namespaced type is declared in the schema.

### Describe alternatives you've considered

We could keep the current behavior and update documentation to clarify behavior.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

cdisselkoen:
In the slack thread, the user got the error 
```
failed to parse entities from file mini-entities.json: error while deserializing entities: Attribute "color" on Environment::"env-123" shouldn't exist according to the schema:
```
Notably, they didn't get an error that `Environment` isn't even a type declared in the schema.  Arguably they should.  Any objections to adding this check during schema-based parsing of entities?  Namely, that declaring an entity with a type that isn't declared in the schema should be an error?  Then, it can be that error that we improve the message on in the presence of namespaces

john-h-kastner-aws:
We were talking about how schema based parsing should handle action entities in a another thread. It seems to me that action entities shouldn't be rejected when using schema based parsing because the only valid actions are those defined in the schema. Those should be extracted from the schema and not be duplicated elsewhere and parsed with other entities.

The `Action` entity types for various namespaces aren't explicitly defined in the schema, so emitting an error when an entity type is not defined will have this effect, but we should be delibrate with the error message. Claiming that the `Action` entity type does not exist would clearly be incorrect. It does exist, we just think it's incorrect to try to parse it.

cdisselkoen:
> we should be deliberate with the error message

I'll make sure to have a separate, more informative, error message for `Action` entity types (including namespace+`Action`) vs for all other entity types not declared in the schema.

> action entities should be rejected when using schema based parsing

[...]

PR #110, opened 2023-06-05
schema-based-parsing: Report errors if parents have incorrect type

### LINKED ISSUE #74: Parsing entities with schema file errors for incorrect attributes but not for incorrect parent types
### Category

Cedar language features

In this slack thread: https://cedar-policy.slack.com/archives/C0547KH7R19/p1684968294619939

If we use a schema file when parsing an entities file, we give errors for incorrect attributes. We don't give similar errors for incorrect parent types.

When we use a schema file to parse entities, we should give errors for incorrect parent types.

N/A

### Additional context

_No response_

- [X] ⚠️ This feature might incur a breaking change

cdisselkoen:
Fixes #74.

### DIFF
--- a/cedar-policy-core/src/entities/json/entities.rs
+++ b/cedar-policy-core/src/entities/json/entities.rs
@@ -23,7 +23,7 @@ use crate::entities::{Entities, EntitiesError, TCComputation};
 use crate::extensions::Extensions;
 use serde::{Deserialize, Serialize};
 use smol_str::SmolStr;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 /// Serde JSON format for a single entity
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
@@ -51,9 +51,12 @@ pub trait Schema {
         &'s self,
         entity_type: &EntityType,
     ) -> Box<dyn Iterator<Item = SmolStr> + 's>;
+
+    /// Get the entity types which are allowed to be parents of the given entity type.
+    fn allowed_parent_types<'s>(&'s self, entity_type: &EntityType) -> HashSet<EntityType>;
 }
 
-/// Simple type that implements `Schema` by expecting no attributes to exist
+/// Simple type that implements `Schema` by expecting no attributes or parents to exist
 #[derive(Debug, Clone)]
 pub struct NullSchema;
 impl Schema for NullSchema {
@@ -62,6 +65,9 @@ impl Schema for NullSchema {
     }
     fn required_attrs(&self, _entity_type: &EntityType) -> Box<dyn Iterator<Item = SmolStr>> {
         Box::new(std::iter::empty())
+    }
+    fn allowed_parent_types(&self, _entity_type: &EntityType) -> HashSet<EntityType> {
+        HashSet::new()
     }
 }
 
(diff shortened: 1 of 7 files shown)


aaronjeline (approved):

cedar-policy-core/src/entities/json/entities.rs
```diff
@@ -229,6 +235,11 @@ impl<'e, 's, S: Schema> EntityJsonParser<'e, 's, S> {
                 }
             })
             .collect::<Result<_, JsonDeserializationError>>()?;
+        // if this `allowed_parent_types` is `None`, then there are no restrictions
+        // (any parent type is allowed)
+        let allowed_parent_types = self
+            .schema
+            .map(|schema| schema.allowed_parent_types(uid.entity_type()));
```
khieta, line 242: I haven't looked at all the surrounding code, so this is an uninformed question, but: why would `allowed_parent_types` be None here, and why does that mean that any parent type should be allowed? The corresponding function in `schema.rs` doesn't return an optional type.

cdisselkoen: It would be `None` if `self.schema` is `None`, i.e., if we're not doing schema-based parsing.  Perhaps this can be clearer somehow?

john-h-kastner-aws: Thinking ahead, it could make sense to return `None` from the impl in `schema.rs` if we wanted to implemented schema based parsing using a partial schema (once that's finished)

cedar-policy-validator/src/schema.rs
```diff
@@ -998,7 +998,7 @@ impl ValidatorSchema {
             for p_entity in action.applies_to.applicable_principal_types() {
                 match p_entity {
                     EntityType::Concrete(p_entity) => {
-                        if !entity_types.contains_key(p_entity) {
+                        if !entity_types.contains_key(&p_entity) {
```
john-h-kastner-aws, line 1001: Clippy and rust-analyzer seem to disagree on if this should have an `&`.

cdisselkoen: maybe the recommendation changed in a recent Rust version?

cedar-policy-core/src/entities/json/entities.rs
cdisselkoen: @khieta I'll update the comment.
@john-h-kastner-aws This is a good point, but I think it makes sense to refactor then / in that PR rather than right now.

cedar-policy-validator/src/schema.rs
```diff
@@ -1238,6 +1238,31 @@ impl cedar_policy_core::entities::Schema for ValidatorSchema {
             }
         }
     }
+
+    fn allowed_parent_types<'s>(
+        &'s self,
+        entity_type: &cedar_policy_core::ast::EntityType,
+    ) -> HashSet<cedar_policy_core::ast::EntityType> {
```
john-h-kastner-aws, line 1245: I don't think this will let an action entity be a member of a any other entity.

It's a bit strange to need to handle action entities at all since they should be extracted from the schema instead of parsed separately. Not sure how best to deal with them.

cdisselkoen: is it an error to include actions in `Entities`?  If so, schema-based parsing doesn't need to handle Action entities, and I think we're good here (except maybe improving the error message for that case)?

john-h-kastner-aws: I think it can be an error. You shouldn't be providing any actions that weren't defined in the schema.

john-h-kastner-aws (approved):

cedar-policy-validator/src/schema.rs
cdisselkoen: I think #73 will cover improving the error message in that case

PR #115, opened 2023-06-08
Provide errors for undeclared entity types and inconsistent action declarations

cdisselkoen:
Addresses many of the issues discussed in comments on #73, but not the namespace part (which will come in a separate PR later).

This PR adds the following to the parsing of entity data:
* During schema-based parsing, error if an entity in the entity data has a type which wasn't declared in the schema.
* During parsing with or without a schema, error if an action entity has a parent which is not an action entity.
* During schema-based parsing, error if an action is provided in the entity data which was not declared in the schema.
* During schema-based parsing, error if an action is provided in both the entity data and the schema, but with conflicting definitions (different attribute values, different parents, missing or extra attributes).

### DIFF
--- a/cedar-policy-core/src/entities/json/entities.rs
+++ b/cedar-policy-core/src/entities/json/entities.rs
@@ -15,15 +15,17 @@
  */
 
 use super::{
-    EntityUidJSON, JSONValue, JsonDeserializationError, JsonDeserializationErrorContext,
-    JsonSerializationError, NullSchema, Schema, TypeAndId, ValueParser,
+    EntityTypeDescription, EntityUidJSON, JSONValue, JsonDeserializationError,
+    JsonDeserializationErrorContext, JsonSerializationError, NoEntitiesSchema, Schema, TypeAndId,
+    ValueParser,
 };
-use crate::ast::{Entity, RestrictedExpr};
+use crate::ast::{Entity, EntityUID, RestrictedExpr};
 use crate::entities::{Entities, EntitiesError, TCComputation};
 use crate::extensions::Extensions;
 use serde::{Deserialize, Serialize};
 use smol_str::SmolStr;
 use std::collections::HashMap;
+use std::sync::Arc;
 
 /// Serde JSON format for a single entity
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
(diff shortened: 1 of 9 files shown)


cedar-policy-core/src/entities/json/entities.rs
```diff
@@ -121,13 +135,40 @@ impl<'e, 's, S: Schema> EntityJsonParser<'e, 's, S> {
...
+                }
+            };
+        match &entity_schema_info {
+            EntitySchemaInfo::NoSchema => {} // no checks to do
+            EntitySchemaInfo::Action(action) => {
+                // here, we ensure that all the attributes on the schema's copy of the
+                // action do exist in `ejson.attrs`. Later when consuming `ejson.attrs`,
+                // we'll do the rest of the checks for attribute agreement.
+                for (schema_attr, _) in action.attrs() {
+                    if ejson.attrs.contains_key(schema_attr) {
```
andrewmwells-amazon, line 160: [nit] why not this:
```
if !ejson.attrs.contains_key(schema_attr) {
    return Err(JsonDeserializationError::ActionDeclarationMismatch { uid });
}
```

cedar-policy-core/src/entities/json/entities.rs
```diff
@@ -153,11 +194,10 @@ impl<'e, 's, S: Schema> EntityJsonParser<'e, 's, S> {
...
                 )),
-                Some(schema) => {
-                    // query the schema to get the expected type. Depending on
-                    // the expected type, we may parse the contents of the
-                    // attribute differently.
-                    let (rexpr, expected_ty) = match schema.attr_type(etype, &k) {
+                EntitySchemaInfo::NonAction(desc) => {
+                    // Depending on the expected type, we may parse the contents
+                    // of the attribute differently.
+                    let (rexpr, expected_ty) = match desc.attr_type(&k) {
```
andrewmwells-amazon, line 200: [nit] I'd prefer `restricted_expr` here (and below). `rexpr` makes me think `regular_expr`

cedar-policy-core/src/entities/json/err.rs
```diff
@@ -68,15 +68,46 @@ pub enum JsonDeserializationError {
         /// Expression we got instead
         got: Box<RestrictedExpr>,
     },
+    /// Parents of actions should be actions, but this action has a non-action parent
+    #[error("{uid} is an action, so it should not have a parent {parent}, which is not an action")]
+    ActionParentIsNotAnAction {
```
andrewmwells-amazon, line 73: [nit] I'd shorten to `ActionParentIsNonAction`

cedar-policy-validator/src/schema.rs
```diff
@@ -1166,11 +1167,12 @@ impl ValidatorSchema {
...
-    pub fn action_entities(&self) -> cedar_policy_core::entities::Result<Entities> {
-        // Invert the action hierarchy to get the ancestor relation expected for
-        // the `Entity` datatype instead of descendant as stored by the schema.
+    /// Construct an `Entity` object for each action in the schema
+    fn action_entities_iter<'s>(
+        &'s self,
+    ) -> impl Iterator<Item = cedar_policy_core::ast::Entity> + 's {
         // We could store the un-inverted `memberOf` relation for each action,
-        // but I judge that the current implementation is actually less error
+        // but I [John] judge that the current implementation is actually less error
```
andrewmwells-amazon, line 1175: maybe use github username?

andrewmwells-amazon (approved):

cedar-policy-core/src/entities/json/entities.rs
cdisselkoen: We also have the `val_into_rexpr` and `type_of_rexpr` functions.  This would be a larger refactor if we were to be consistent -- can we do that in a separate PR?

PR #117, opened 2023-06-12
Update record types to allow for open and closed record

john-h-kastner-aws:
This splits off a change to record types made as part of partial schema validation

* Adds `open_attributes` to the record type struct. When true, we assume a record with that type may have other attributes. When false it will have exactly the attributes specified in the type.
* Record literals and records in schema are now closed.
* Update `has` to match use this new field. A `has` on a closed record will have type `False` if the attribute is not present.
* Update subtyping and LUB computation

I've update the Dafny Spec (https://github.com/cedar-policy/cedar-spec/pull/44) ~~, but don't have the working proofs~~ (Kesha finished them off for me). Diff testing and PBT passed an overnight run. 

### DIFF
--- a/cedar-policy-validator/src/type_error.rs
+++ b/cedar-policy-validator/src/type_error.rs
@@ -106,17 +106,19 @@ impl TypeError {
         }
     }
 
-    pub(crate) fn missing_attribute(
+    pub(crate) fn unsafe_attribute_access(
         on_expr: Expr,
         missing: String,
         suggestion: Option<String>,
+        may_exist: bool,
     ) -> Self {
         Self {
             on_expr: Some(on_expr),
             source_location: None,
-            kind: TypeErrorKind::MissingAttribute(MissingAttribute {
+            kind: TypeErrorKind::UnsafeAttributeAccess(UnsafeAttributeAccess {
                 missing,
                 suggestion,
+                may_exist,
             }),
         }
     }
@@ -212,8 +214,16 @@ pub enum TypeErrorKind {
     IncompatibleTypes(IncompatibleTypes),
     /// The typechecker detected an access to a record or entity attribute
     /// that it could not statically guarantee would be present.
-    #[error("Attribute not found in record or entity {}", .0.missing)]
-    MissingAttribute(MissingAttribute),
+    #[error(
+        "Attribute not found in record or entity: {}{}",
+        .0.missing,
+        if .0.may_exist {
+            ". There may be additional attributes that the validator is not able to reason about."
+        } else {
+            ""
+        }
+    )]
+    UnsafeAttributeAccess(UnsafeAttributeAccess),
     /// The typechecker could not conclude that an access to an optional
     /// attribute was safe.
     #[error("Unable to guarantee safety of access to optional attribute {}", .0.optional)]
@@ -267,9 +277,10 @@ pub struct TypesMustMatch {
 
 /// Structure containing details about a missing attribute error.
 #[derive(Debug, Hash, Eq, PartialEq)]
-pub struct MissingAttribute {
+pub struct UnsafeAttributeAccess {
     missing: String,
     suggestion: Option<String>,
+    may_exist: bool,
 }
 
 /// Structure containing details about an unsafe optional attribute error.
--- a/cedar-policy-validator/src/type_error.rs
+++ b/cedar-policy-validator/src/type_error.rs
@@ -108,7 +108,7 @@ impl TypeError {
 
     pub(crate) fn unsafe_attribute_access(
         on_expr: Expr,
-        missing: String,
+        attribute: String,
         suggestion: Option<String>,
         may_exist: bool,
     ) -> Self {
@@ -116,7 +116,7 @@ impl TypeError {
             on_expr: Some(on_expr),
             source_location: None,
             kind: TypeErrorKind::UnsafeAttributeAccess(UnsafeAttributeAccess {
-                missing,
+                attribute,
                 suggestion,
                 may_exist,
             }),
@@ -216,7 +216,7 @@ pub enum TypeErrorKind {
     /// that it could not statically guarantee would be present.
     #[error(
         "Attribute not found in record or entity: {}{}",
-        .0.missing,
+        .0.attribute,
         if .0.may_exist {
             ". There may be additional attributes that the validator is not able to reason about."
         } else {
@@ -278,8 +278,10 @@ pub struct TypesMustMatch {
 /// Structure containing details about a missing attribute error.
 #[derive(Debug, Hash, Eq, PartialEq)]
 pub struct UnsafeAttributeAccess {
-    missing: String,
+    attribute: String,
     suggestion: Option<String>,
+    /// When this is true, the attribute might still exist, but the validator
+    /// cannot guarantee that it will.
     may_exist: bool,
 }
 
--- a/cedar-policy-validator/src/types.rs
+++ b/cedar-policy-validator/src/types.rs
@@ -134,22 +134,31 @@ impl Type {
     }
 
     pub(crate) fn any_record() -> Type {
-        Type::record_with_required_attributes(None)
+        Type::EntityOrRecord(EntityRecordKind::Record {
+            attrs: Attributes::with_required_attributes(None),
+            // (open_attributes = false) <: (open_attributes = true),
+            // so this makes `any_record` a super type of all records.
+            open_attributes: OpenTag::OpenAttributes,
+        })
     }
 
     pub(crate) fn record_with_required_attributes(
         required_attrs: impl IntoIterator<Item = (SmolStr, Type)>,
+        open_attributes: OpenTag,
     ) -> Type {
         Type::EntityOrRecord(EntityRecordKind::Record {
             attrs: Attributes::with_required_attributes(required_attrs),
+            open_attributes,
         })
     }
 
     pub(crate) fn record_with_attributes(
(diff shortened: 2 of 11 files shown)


cedar-policy-validator/src/types.rs
```diff
@@ -1273,27 +1385,82 @@ mod test {
...
+            )
+        );
+        println!(
+            "{}",
+            Type::is_subtype(
+                &empty_schema(),
+                &Type::record_with_attributes_open_attributes(None, false),
+                &Type::record_with_attributes_open_attributes(None, false)
+            )
+        );
```
khieta, line 1442: Rather than printing here, you might want actual tests. Or at least add comments with what the expected output is.

john-h-kastner-aws: outdated

cedar-policy-validator/src/types.rs
```diff
@@ -349,12 +363,19 @@ impl Type {
...
         match ty {
-            // If the type of the expression is a entity reference type, and the
-            // attribute is not declared to exist in the schema, then we know
-            // the entity may not have the attribute. For an entity least upper
-            // bound resulting from multiple entity types, the type might have
-            // the attribute if any of the constituent entity types have the
-            // attribute in the schema.
+            // Never, being the bottom type, is a subtype of EntityOrRecord, so
+            // it could have any attributes.
+            Type::Never => true,
```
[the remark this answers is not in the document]
john-h-kastner-aws: Yes, there would have been an error elsewhere before an expression could have type `Never`. I think this behavior makes the most sense regardless, and it will become the necessary behavior with partial schema validation.

cedar-policy-validator/src/types.rs
```diff
@@ -968,8 +1051,28 @@ impl EntityRecordKind {
...
+                },
+            ) => {
+                // Closed attributes subtype open attributes. A record type with
+                // open attributes will always contain a value that is not in a
+                // record type with closed attributes, so open attribute record
+                // types can never subtype closed attribute record types.
+                (!open_attr0 || *open_attr1)
+                // When `rk1` has open attributes, width subtyping applies since
+                // there may be attributes in `rk0` that are not listed in
+                // `rk1`.  When `rk1` is closed, a subtype of `rk1` may not have
```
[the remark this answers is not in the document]
john-h-kastner-aws: I don't think so

cedar-policy-validator/src/type_error.rs
```diff
@@ -267,9 +277,10 @@ pub struct TypesMustMatch {
 
 /// Structure containing details about a missing attribute error.
 #[derive(Debug, Hash, Eq, PartialEq)]
-pub struct MissingAttribute {
+pub struct UnsafeAttributeAccess {
     missing: String,
     suggestion: Option<String>,
+    may_exist: bool,
```
cdisselkoen, line 283: docs on this?  I presume the interpretation is that, if `may_exist` is true, the `missing` attribute might actually exist we just don't know for sure?

cedar-policy-validator/src/types.rs
```diff
@@ -868,14 +907,42 @@ impl IntoIterator for Attributes {
...
+    // exactly match the attributes listed in the type.
+    ClosedAttributes,
+}
+
+impl OpenTag {
+    pub(crate) fn is_open(self) -> bool {
+        match self {
+            OpenTag::OpenAttributes => true,
+            OpenTag::ClosedAttributes => false,
+        }
```
cdisselkoen, line 927: this could be written `matches!(self, OpenTag::OpenAttributes)`, or even `self == OpenTag::OpenAttributes`.  Not sure if either of those are actually better, just pointing it out.

cedar-policy-validator/src/types.rs
```diff
@@ -904,9 +971,31 @@ impl EntityRecordKind {
...
+            // because the upper bound for any two action entities is
+            // `AnyEntity`, so if we have an ActionEntity here its attributes
+            // are known precisely.
+            EntityRecordKind::ActionEntity { .. } => false,
+            // The `AnyEntity` type has no declared attributes, but it is a
+            // super type of all other entity types which may have attributes,
+            // so it clearly may have additional attributes.
+            EntityRecordKind::AnyEntity => true,
+            // An entity LUB may not have additional attributes.
+            EntityRecordKind::Entity(_) => false,
```
cdisselkoen, line 992: what if one of the components of the LUB has additional attributes?

cedar-policy-validator/src/types.rs
```diff
@@ -934,8 +1023,15 @@ impl EntityRecordKind {
     ) -> Option<EntityRecordKind> {
         use EntityRecordKind::*;
         match (rk0, rk1) {
-            (Record { attrs: attrs0 }, Record { attrs: attrs1 }) => Some(Record {
+            (Record { attrs: attrs0, .. }, Record { attrs: attrs1, .. }) => Some(Record {
                 attrs: Attributes::least_upper_bound(schema, attrs0, attrs1),
+                // If attrs0 <: attrs1 (or attrs1 <: attr0) without width
+                // subtyping, we could return the supertype for a LUB preserving
+                // `open_attributes: false`,  but this result is already
```
cdisselkoen, line 1030: `open_attributes: false` ==> `ClosedAttributes`

cdisselkoen (approved):
just some minor comments

cedar-policy-validator/src/types.rs
john-h-kastner-aws: The types are organized a bit differently in Rust than in Dafny. In the Dafny it would have been difficult to let records be closed without letting entity types also have the open/closed flag, so entity types got the flag even though they don't really use it. For the Rust entity types don't have the flag.

john-h-kastner-aws: There's also a bit of subtlety because the record type returned by `get_attributes_type` might be open, but full entity LUB is still closed because it has access to all it's constituent entity types - which are closed - so it can account for every attribute that could possibly exist.

cdisselkoen: Ack.  If it's subtle, it may be worth putting what you just wrote in a comment in the code here

cedar-policy-validator/src/types.rs
```diff
@@ -366,10 +387,13 @@ impl Type {
...
-            // Other record or entity types always might have some attribute
-            // since the type may be the result of a least upper bound which
-            // could have dropped attributes from the type.
-            Type::EntityOrRecord(_) => true,
+            // A record will have an attribute if the attribute is in its
+            // attributes map. Records computed as a LUB may have an open
+            // attributes record, but that is handled by the first match case.
+            Type::EntityOrRecord(EntityRecordKind::Record { attrs, .. }) => {
+                attrs.get_attr(attr).is_some()
+            }
```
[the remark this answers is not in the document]
cdisselkoen: If that's an important invariant that correctness depends on, might be worth documenting in a comment on the function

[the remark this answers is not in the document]
john-h-kastner-aws: This function doesn't depend on it for correctness. The observation is that correctness of the validator actually doesn't depend on the correctness of this case.

cedar-policy-validator/src/type_error.rs
```diff
@@ -267,9 +277,12 @@ pub struct TypesMustMatch {
 
 /// Structure containing details about a missing attribute error.
 #[derive(Debug, Hash, Eq, PartialEq)]
-pub struct MissingAttribute {
-    missing: String,
+pub struct UnsafeAttributeAccess {
```
andrewmwells-amazon, line 280: Can we pass in this struct to `fn unsafe_attribute_access(` instead of passing the `attribute`, `suggestion` and `may_exist` arguments?

andrewmwells-amazon (approved):

cedar-policy-validator/src/type_error.rs
john-h-kastner-aws: That functions is supposed to be a constructor for this struct, so it wouldn't make sense IMO

khieta:
This PR and [cedar-spec/PR#44](https://github.com/cedar-policy/cedar-spec/pull/44) assume that the context and entities are "closed", meaning that they are not allowed to have any extra fields. IIRC this is not what is enforced by schema-based parsing (@cdisselkoen can correct me if I'm wrong).  I think we will want these two to match: that may involve changing the assumptions in this PR, or creating an issue to update schema-based parsing in the future.

cdisselkoen:
> this is not what is enforced by schema-based parsing

Actually, it is, since April 6 (internal CR 88757947). Schema-based parsing will error on additional attributes being present, and on required attributes being missing, for both entities and records.  See the errors split out [here](https://github.com/cedar-policy/cedar/blob/66e79a4f76c569f1ab93a8e8069180413ba52027/cedar-policy-core/src/entities/json/err.rs#L123)

cdisselkoen:
As of now, though, the schema-based parsing has no concept of open/closed entities/records, effectively assuming that all entities/records are closed.  Should this PR also adjust schema-based parsing to be aware of the possibility of open records?

john-h-kastner-aws:
> Should this PR also adjust schema-based parsing to be aware of the possibility of open records?

Schema format isn't updated with this PR, so no.

khieta:
So is [this comment](https://github.com/cedar-policy/cedar/blob/main/cedar-policy-core/src/entities/json/context.rs#L45) (and another similar one in entities/json/entities.rs)  out of date?

cdisselkoen:
Yep.  Will fix

andrewmwells-amazon (approved):

PR #123, opened 2023-06-15
Resolve: Issue 114 Expose-EvaluationError

### LINKED ISSUE #114: Expose `AuthorizationError`
### Category

User level API changes

`EvaluationError` is currently exposed through `cedar/api.rs` as a string message (code copied below). This may encourage users to pattern match on our error messages, which are subject to change. We should consider updating this to match cedar-policy-core's `EvaluationError` (per the TODO), or construct some new enum. 
```
/// Errors encountered while evaluating policies or expressions, or making
/// authorization decisions.
#[derive(Debug, Clone, PartialEq, Eq, Error)]
pub enum EvaluationError {
    /// Error message, as string.
    /// TODO in the future this can/should be the actual Core `EvaluationError`
    #[error("{0}")]
    StringMessage(String),
}
```

See above.

N/A

### Additional context

_No response_

- [x] ⚠️ This feature might incur a breaking change

NetwCodeProjects:
closes #114 

## Reviewed by Kesha Hietala (khieta)

So, after some digging, I found the EvaluationError enum in `cedar-policy-core/src/evaluator/err.rs`‎ which is imported by the `cedar/cedar-policy-core/src/evaluator.rs` on lines 25-27.

I haven't tested this Idea but using that information and the TODO I arrived at a solution.
> /// TODO in the future this can/should be the actual Core `EvaluationError

[!] There's an unfinished comment on line 447 of api.rs I meant to say something like `[+] Modified to Resolve TODO `

### Solution
`cedar/cedar-policy/src/api.rs`
Import the EvaluationError enum from `cedar-policy-core` and include it as a variant called CoreError in the ApiEvaluationError enum.

The #[from] attribute allows for automatic conversion from EvaluationError to ApiEvaluationError::CoreError.

Consistent error handling experience leverages the error types and patterns defined in cedar-policy-core. 
This ensures that any changes made to cedar-policy-core's EvaluationError are reflected in the ceder-policy/src API module without any need for manual synchronization.

```rust
use cedar_policy_core::evaluator::err::EvaluationError;

#[derive(Debug, Clone, PartialEq, Eq, Error)]
pub enum ApiEvaluationError {
    #[error("{0}")]
    CoreError(#[from] EvaluationError),
}
````

### Additional changes

There are 8 references to EvaluationError that would need to be changed.
Here's an example using line 121-132 from cedar-policy/src/api.rs.

```rust
pub fn attr(&self, attr: &str) -> Option<Result<EvalResult, ApiEvaluationError>> {
    let expr = self.0.get(attr)?;
    let all_ext = Extensions::all_available();
    let evaluator = RestrictedEvaluator::new(&all_ext);
    Some(
        evaluator
            .interpret(expr.as_borrowed())
            .map(EvalResult::from)
            .map_err(ApiEvaluationError::from),
    )
}
```
## Further changes
`cedar-policy-core/src/authorizer.rs`
There are a few `Vec<string>` calls that need to be updated.
For example, in the Diagnostics struct.
- We need to update the Diagnostics struct to use `Vec<EvaluationError>` for the errors field instead of `Vec<String>`
```rust
#[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
pub struct Diagnostics {
    pub reason: HashSet<PolicyID>,
    pub errors: Vec<EvaluationError>,
}
```
- Modify the Response struct's 'new' function to accept a `Vec<EvaluationError>` for the errors parameter and update the construction of diagnostics. 
```rust
impl Response {
    pub fn new(decision: Decision, reason: HashSet<PolicyID>, errors: Vec<EvaluationError>) -> Self {
        Response {
            decision,
            diagnostics: Diagnostics { reason, errors },
        }
    }
}

```
The Response struct will now use a `Vec<EvaluationError>` for the errors field within the Diagnostics struct. It's important to 
update any code that creates a new Response instance to pass a `Vec<EvaluationError>` for the errors parameter.

### DIFF
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -29,6 +29,8 @@ use cedar_policy_core::entities::JsonDeserializationErrorContext;
 use cedar_policy_core::entities::{ContextSchema, Dereference, JsonDeserializationError};
 use cedar_policy_core::est;
 use cedar_policy_core::evaluator::{Evaluator, RestrictedEvaluator};
+pub use cedar_policy_core::evaluator::EvaluationError;
+pub use cedar_policy_core::authorizer::AuthorizationError;
 pub use cedar_policy_core::extensions;
 use cedar_policy_core::extensions::Extensions;
 use cedar_policy_core::parser;
@@ -127,7 +129,7 @@ impl Entity {
             evaluator
                 .interpret(expr.as_borrowed())
                 .map(EvalResult::from)
-                .map_err(|e| EvaluationError::StringMessage(e.to_string())),
+                .map_err(EvaluationError::from),
         )
     }
 }
@@ -360,7 +362,7 @@ pub struct Diagnostics {
     /// If no policies applied to the request, this set will be empty.
     reason: HashSet<PolicyId>,
     /// list of error messages which occurred
-    errors: HashSet<String>,
+    errors: Vec<EvaluationError>,
 }
 
 impl From<authorizer::Diagnostics> for Diagnostics {
@@ -380,10 +382,7 @@ impl Diagnostics {
 
     /// Get the error messages
     pub fn errors(&self) -> impl Iterator<Item = EvaluationError> + '_ {
-        self.errors
-            .iter()
-            .cloned()
-            .map(EvaluationError::StringMessage)
+        self.errors.iter()
     }
 }
 
@@ -445,16 +444,6 @@ impl From<authorizer::PartialResponse> for ResidualResponse {
             diagnostics: p.diagnostics.into(),
         }
     }
-}
-
-/// Errors encountered while evaluating policies or expressions, or making
-/// authorization decisions.
-#[derive(Debug, Clone, PartialEq, Eq, Error)]
-pub enum EvaluationError {
-    /// Error message, as string.
-    /// TODO in the future this can/should be the actual Core `EvaluationError`
-    #[error("{0}")]
-    StringMessage(String),
 }
 
 /// Used to select how a policy will be validated.
--- a/cedar-policy-core/src/ast/restricted_expr.rs
+++ b/cedar-policy-core/src/ast/restricted_expr.rs
@@ -315,7 +315,7 @@ impl<'a> Hash for RestrictedExprShapeOnly<'a> {
 }
 
 /// Errors generated in the restricted_expr module
-#[derive(Debug, Clone, PartialEq, Hash, Error)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, Error)]
 pub enum RestrictedExpressionError {
     /// A "restricted" expression contained a feature that is not allowed
     /// in "restricted" expressions. The `feature` is just a string description
--- a/cedar-policy-core/src/entities/json/schema_types.rs
+++ b/cedar-policy-core/src/entities/json/schema_types.rs
@@ -17,9 +17,10 @@
 use crate::ast::{EntityType, Name, Type};
 use smol_str::SmolStr;
 use std::collections::HashMap;
+use serde::{Serialize, Deserialize};
 
 /// Possible types that schema-based parsing can expect for Cedar values.
-#[derive(Debug, PartialEq, Eq, Clone)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 pub enum SchemaType {
     /// Boolean
     Bool,
@@ -55,7 +56,7 @@ pub enum SchemaType {
 }
 
 /// Attribute type structure used in [`SchemaType`]
-#[derive(Debug, PartialEq, Eq, Clone)]
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize, Clone)]
 pub struct AttributeType {
     /// Type of the attribute
     attr_type: SchemaType,
--- a/cedar-policy-core/src/evaluator/err.rs
+++ b/cedar-policy-core/src/evaluator/err.rs
@@ -18,9 +18,10 @@ use crate::ast::*;
 use smol_str::SmolStr;
 use std::sync::Arc;
 use thiserror::Error;
+use serde::{Serialize, Deserialize};
 
 /// Errors that can occur during evaluation
-#[derive(Debug, PartialEq, Clone, Error)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Error)]
 pub enum EvaluationError {
     /// Tried to lookup this entity UID, but it didn't exist in the provided
     /// entities
@@ -125,7 +126,7 @@ fn pretty_type_error(expected: &[Type], actual: &Type) -> String {
     }
 }
 
-#[derive(Debug, PartialEq, Clone, Error)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Error)]
 pub enum IntegerOverflowError {
     #[error("integer overflow while attempting to {} the values {arg1} and {arg2}", match .op { BinaryOp::Add => "add", BinaryOp::Sub => "subtract", _ => "perform an operation on" })]
     BinaryOp {
--- a/cedar-policy-core/src/extensions.rs
+++ b/cedar-policy-core/src/extensions.rs
@@ -25,6 +25,7 @@ pub mod partial_evaluation;
 
 use crate::ast::{Extension, ExtensionFunction, Name};
 use crate::entities::SchemaType;
+use serde::{Serialize, Deserialize};
 use thiserror::Error;
 
 lazy_static::lazy_static! {
@@ -133,7 +134,7 @@ impl<'a> Extensions<'a> {
 }
 
 /// Errors thrown when looking up an extension function in [`Extensions`].
-#[derive(Debug, PartialEq, Clone, Error)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Error)]
 pub enum ExtensionFunctionLookupError {
     /// Tried to call a function that doesn't exist
     #[error("extension function does not exist: {name}")]


cedar-policy/src/api.rs
```diff
@@ -440,11 +441,12 @@ impl ResidualResponse {
 /// Errors encountered while evaluating policies or expressions, or making
 /// authorization decisions.
 #[derive(Debug, Clone, PartialEq, Eq, Error)]
-pub enum EvaluationError {
+pub enum ApiEvaluationError {
```
cdisselkoen, line 444: I'm not sure this name change makes things clearer for users.  I think we should endeavor to make the public type called `EvaluationError`, regardless of what gymnastics that requires under the hood here.  (Probably means we can't `use` Core's `EvaluationError` unqualified in this file, and have to refer to it as something like `core::EvaluationError`.)

cedar-policy/src/api.rs
```diff
@@ -440,11 +441,12 @@ impl ResidualResponse {
...
     /// Error message, as string.
     /// TODO in the future this can/should be the actual Core `EvaluationError`
+    /// [+] Modifi
```
cdisselkoen, line 447: Nit: comment is unfinished

cedar-policy/src/api.rs
```diff
@@ -440,11 +441,12 @@ impl ResidualResponse {
 /// Errors encountered while evaluating policies or expressions, or making
 /// authorization decisions.
 #[derive(Debug, Clone, PartialEq, Eq, Error)]
-pub enum EvaluationError {
+pub enum ApiEvaluationError {
     /// Error message, as string.
     /// TODO in the future this can/should be the actual Core `EvaluationError`
```
cdisselkoen, line 446: These two lines of comments need adjustment with these changes

cedar-policy/src/api.rs
```diff
@@ -440,11 +441,12 @@ impl ResidualResponse {
...
+    /// [+] Modifi
     #[error("{0}")]
-    StringMessage(String),
+    CoreError(#[from] EvaluationError),
```
cdisselkoen, line 449: This variant name is public.  We generally more-or-less try to hide the distinction between Core and this crate from this crate's consumers.  Could we name the variant `EvaluationError` or something instead?

That also raises the question, do we need this enum at all vs. just using Core's `EvaluationError` directly?

cdisselkoen (changes requested):
Thanks for the PR!  A few minor comments.

cdisselkoen:
For posterity, noting that this closes #114.  (Used one of the [valid magic incantations](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) there, that lets GitHub know the issue and PR are linked.)

NetwCodeProjects:
Updated the PR description to use a magic incantation. If there are no objections, I will begin to make the changes outlined. 

Should I worry about pacifying `Cargo check`. My current plan is to just make the changes outlined.

shaobo-he-aws:
> Updated the PR description to use a magic incantation. If there are no objections, I will begin to make the changes outlined.
> 
> Should I worry about pacifying `Cargo check`. My current plan is to just make the changes outlined.

We can always link the issue when we merge the PR. Please go ahead to make the changes request.

khieta:
Hi @NetwCodeProjects! There has been no activity on this PR for a week or so -- are you ok with me taking over?

Also, I see that I didn't answer your question above (sorry!) 
> Should I worry about pacifying `cargo check`?

We will want our CI to pass before merging a PR, so yes, you will have to pacify to some extent.

NetwCodeProjects:
# Requested Changes
## 1.
> Could we name the variant EvaluationError or something instead?

In this updated code, the CoreError variant has been renamed to EvaluationError

```rust
pub enum EvaluationError {

    #[error("{0}")]
    EvaluationError(#[from] core::EvaluationError),
}
```
## 2.
> (Probably means we can't use Core's EvaluationError unqualified in this file, and have to refer to it as something like core::EvaluationError.)

 The core::EvaluationError type is directly wrapped in the EvaluationError variant.
```rust
use cedar_policy_core::evaluator::err as core;

   #[error("{0}")]
    EvaluationError(#[from] core::EvaluationError),
}
```

> That also raises the question, do we need this enum at all vs. just using Core's EvaluationError directly?

If you want to provide a distinct error type for your crate and potentially have custom variants specific to your crate's functionality, then keeping the EvaluationError enum can be useful. However, if the core::EvaluationError provides all the necessary error variants and is sufficient for your use case, you can directly use it without the need for an additional enum.

Consider the trade-offs.  Does it makes sense to use the existing EvaluationError enum or directly expose core::EvaluationError to consumers.

## 3.
Lines 122 and 130 changed from ApiEvaluation... to EvaluationError
the same change was made on various lines which called ApiEvaluationError.

```rust
    pub fn attr(&self, attr: &str) -> Option<Result<EvalResult, EvaluationError>> {
        let expr = self.0.get(attr)?;
        let all_ext = Extensions::all_available();
        let evaluator = RestrictedEvaluator::new(&all_ext);
        Some(
            evaluator
                .interpret(expr.as_borrowed())
                .map(EvalResult::from)
                .map_err(EvaluationError::from),
        )
    }
}
```

## 4.
comments updated.

```rust
pub enum EvaluationError {
    /// Represents an evaluation error originating from the `cedar-policy-core` crate.
    /// EvaluationError variant wraps the core::EvaluationError using the #[from] attribute.
    #[error("{0}")]
    EvaluationError(#[from] core::EvaluationError),

    // Add your custom variants here, if needed.
    // Custom variants can represent errors specific to the crate's functionality.
}

```

khieta:
Thanks for the update!

It looks like the CI is still failing. To prevent this, you can check the build locally by running `cargo test` and `cargo fmt` in the top-level `cedar` directory.

To answer your question
> Does it makes sense to use the existing EvaluationError enum or directly expose core::EvaluationError to consumers.

At present, I think that exposing `EvaluationError` directly is sufficient. If we decide to add new error variants in the future, we can revisit then.

NetwCodeProjects:
# Exposing EvaluationError directly
> At present, I think that exposing EvaluationError directly is sufficient. If we decide to add new error variants in the future, we can revisit then.

To expose directly, the `pub use` statement allows you to directly expose cedar_policy_core::evaluator::err::EvaluationError as EvaluationError

The EvaluationError enum can be deleted without introducing an additional enum in your crate. I would leave a comment to show where the enum was in case you decide to add new error variants in the future.

```rust
pub use cedar_policy_core::evaluator::err::EvaluationError;

/// to add new error variants in the future create pub enum here

pub fn attr(&self, attr: &str) -> Option<Result<EvalResult, EvaluationError>> {
    let expr = self.0.get(attr)?;
    let all_ext = Extensions::all_available();
    let evaluator = RestrictedEvaluator::new(&all_ext);
    Some(
        evaluator
            .interpret(expr.as_borrowed())
            .map(EvalResult::from)
            .map_err(EvaluationError::from),
    )
}
```

NetwCodeProjects:
# Changes to successfully build the program.

## 1. return an EvaluationError directly from the closure.

Error message:

expected `[closure@authorizer.rs:101:64]` to be a closure that returns `EvaluationError`, but it returns `String`

<details>
<summary>error message</summary>

```rust
error[E0271]: expected `[closure@authorizer.rs:101:64]` to be a closure that returns `EvaluationError`, but it returns `String`
   --> cedar-policy-core\src\authorizer.rs:101:31
    |
101 |                   errors.extend(partial.residuals.policies().map(|p| {
    |  ________________________------_^
    | |                        |
    | |                        required by a bound introduced by this call
102 | |                     format!(
103 | |                         "while evaluating policy {}, encountered the following error: {}",
104 | |                         p.id(),
105 | |                         EvaluationError::NonValue(p.condition())
106 | |                     )
107 | |                 }));
    | |__________________^ expected `EvaluationError`, found `String`
    |
    = note: required for `std::iter::Map<impl Iterator<Item = &policy::Policy>, [closure@cedar-policy-core\src\authorizer.rs:101:64: 101:67]>` to implement `Iterator`
note: required by a bound in `extend`
```

</details>

### solution

I deleted the format! line and now EvaluationError::NonValue(p.condition()) is returned, Error gone.
BUT the original code wanted to return "while evaluating policy {}, encountered the following error: {}"
with the values for p.id() and EvaluationError::NonValue(p.condition()). 

It seems to me like creating a variant in evaluator/err.rs would simplify correcting the error. The variant could convert a string into an EvaluationError object but that seems unspecific for my preferences, so then the next idea is to create a PolicyError variant which returns the error message as it was defined in the original code. "while evaluating policy {}, encountered the following error: {}"

There is policy.rs in cedar-policy-core/src/ast which is imported to construct enum EvaluationError, so policy types and methods might be defined already.

```rust
#[error("The expression evaluated to a residual: {0}")]
    NonValue(Expr),
```


## 2. mismatched types

Error message:

expected `EvaluationError`, found `String`

<details>
<summary>error message</summary>

```rust
error[E0308]: mismatched types
   --> cedar-policy-core\src\authorizer.rs:181:26
    |
181 |                     vec![msg],
    |                          ^^^ expected `EvaluationError`, found `String`
```

</details>

### solution
Similar error this time the error message is 
"while initializing the Evaluator, encountered the following error: {e}"

```rust
let msg = format!(
                    "while initializing the Evaluator, encountered the following error: {e}"
                );
                return ResponseKind::FullyEvaluated(Response::new(
                    Decision::Deny,
                    HashSet::new(),
                    vec![msg],
                ));
            }
        };
```
# Other errors
<details>
<summary>errors</summary>
```rust
error[E0277]: the trait bound `EvaluationError: Serialize` is not satisfied
    --> cedar-policy-core\src\authorizer.rs:845:17
     |
845  | #[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
     |                 ^^^^^^^^^ the trait `Serialize` is not implemented for `EvaluationError`
...
850  |     /// list of error messages which occurred
     |     ----------------------------------------- required by a bound introduced by this call
     |
     = help: the following other types implement trait `Serialize`:
               &'a T
               &'a mut T
               ()
               (T0, T1)
               (T0, T1, T2)
               (T0, T1, T2, T3)
               (T0, T1, T2, T3, T4)
               (T0, T1, T2, T3, T4, T5)
             and 192 others
     = note: required for `Vec<EvaluationError>` to implement `Serialize`
note: required by a bound in `authorizer::_::_serde::ser::SerializeStruct::serialize_field`
```

</details>

<details>
<summary>errors</summary>
```rust
error[E0277]: the trait bound `EvaluationError: Deserialize<'_>` is not satisfied
    --> cedar-policy-core\src\authorizer.rs:851:17
     |
851  |     pub errors: Vec<EvaluationError>,
     |                 ^^^^^^^^^^^^^^^^^^^^ the trait `Deserialize<'_>` is not implemented for `EvaluationError`
     |
     = help: the following other types implement trait `Deserialize<'de>`:
               &'a Path
               &'a [u8]
               &'a str
               ()
               (T0, T1)
               (T0, T1, T2)
               (T0, T1, T2, T3)
               (T0, T1, T2, T3, T4)
             and 192 others
     = note: required for `Vec<EvaluationError>` to implement `Deserialize<'_>`
note: required by a bound in `next_element`
```

</details>

The road to compatibility isn't far there's two categories which represent all of the problems.

NetwCodeProjects:
# Re: 1-2
Hello, I will push a passing build today. 

I made a comment the other day about Exposing EvaluationError directly. Which can be found here https://github.com/cedar-policy/cedar/pull/123#issuecomment-1610506147

I can delete the unused EvaluationError enum and leave comment
`"// if needed in future insert EvaluationError enum here"`.

To expose EvaluationError directly as EvaluationError,  
I will use `pub use` to cedar_policy_core::evaluator::err::EvaluationError.

<details>

# Exposing EvaluationError directly

> At present, I think that exposing EvaluationError directly is sufficient. If we decide to add new error variants in the future, we can revisit then.

To expose directly, the `pub use` statement allows you to directly expose cedar_policy_core::evaluator::err::EvaluationError as EvaluationError

The EvaluationError enum can be deleted without introducing an additional enum in your crate. I would leave a comment to show where the enum was in case you decide to add new error variants in the future.

```rust
pub use cedar_policy_core::evaluator::err::EvaluationError;

/// to add new error variants in the future create pub enum here

pub fn attr(&self, attr: &str) -> Option<Result<EvalResult, EvaluationError>> {
    let expr = self.0.get(attr)?;
    let all_ext = Extensions::all_available();
    let evaluator = RestrictedEvaluator::new(&all_ext);
    Some(
        evaluator
            .interpret(expr.as_borrowed())
            .map(EvalResult::from)
            .map_err(EvaluationError::from),
    )
}
```

</details>

khieta:
Your plan to `pub use` cedar_policy_core::evaluator::err::EvaluationError (or `AuthorizationError` per my comment above) sounds fine. I don't think the comment is necessary.

NetwCodeProjects:
Could I ask for some clarity?

>I think a good approach would be to make a new AuthorizationError struct that can be one of two things

Are you asking for somthing like this?
<details>

```rust
#[derive(Debug, Error)]
pub struct AuthorizationError {
    #[error("while initializing the Evaluator, encountered the following error: {0}")]
    attribute_evaluation_error: Option<EvaluationError>,

    #[error("while evaluating policy {id}, encountered the following error: {error}")]
    policy_evaluation_error: Option<PolicyEvaluationError>,
}

#[derive(Debug, Error)]
pub enum EvaluationError {
    // ...
}

#[derive(Debug)]
pub struct PolicyEvaluationError {
    id: PolicyID,
    error: EvaluationError,
}
```

</details>

or 

<details>

```rust
#[derive(Debug, Error)]
pub struct AuthorizationError {
    #[source]
    cause: Box<dyn std::error::Error + Send + Sync>,
}

impl AuthorizationError {
    pub fn attribute_evaluation_error<E>(error: E) -> Self
    where
        E: std::error::Error + Send + Sync + 'static,
    {
        AuthorizationError {
            cause: Box::new(error),
        }
    }

    pub fn policy_evaluation_error<E>(id: PolicyID, error: E) -> Self
    where
        E: std::error::Error + Send + Sync + 'static,
    {
        AuthorizationError {
            cause: Box::new(PolicyEvaluationError { id, error }),
        }
    }
}

#[derive(Debug, Error)]
#[error("while evaluating policy {id}, encountered the following error: {error}")]
pub struct PolicyEvaluationError {
    id: PolicyID,
    error: EvaluationError,
}
```

</details>

You provided an enum so I just added that to the err.rs and started to make the changes needed for that to work. 

Since I use the official rust-analyzer extension in vscode I get errors and suggestions in advance. From my experience with compiled languages, you could have a passing build but still introduce an error.  Anyway, my point is there are unlisted challenges I'm solving. I spent 7 hours today, which was actually my shortest day because I have a statistics assignment due at midnight good ol' probabilities, permutations and linear regression calculations etc. I spend about 9-12 hours a day when I'm not busy with other things. Some days are quicker than others. 

Might be silly to ask but do you want me to add a struct or enum i've been using the enum you specified, with a named lifetime parameter `&'a PolicyID`, `enum AuthorizationError<'a>` also where should this enum/struct be located? evaluator/err.rs or authorizer.rs? or evalutor.rs ?

Ok so when i was talking about pub use cedar_policy_core::evaluator::err::EvaluationError that was to replace the EvaluationError enum inside the /api.rs.


dbug pic

khieta:
Apologies that the issue was underspecified. My intention was to flesh out the design while working on the fix 🙂 Through our back and forth, I have a clearer picture now of what I think the design should look like.

If I were making the changes myself, here is what I would do:

### Step 1

Make a new file in `cedar-policy-core` called `authorizer/err.rs`. Similar to `evaluator/err.rs`, this file will define the errors reported specifically by the authorizer. In that file, I would define the enum I proposed a few comments back (copied below).
```rust
/// Error type for errors that can be returned by the authorizer.
pub enum AuthorizationError {
    /// Failed to eagerly evaluate entity attributes when initializing the Evaluator
    #[error("while initializing the Evaluator, encountered the following error: {0}")]
    AttributeEvaluationError(EvaluationError),

    /// An error occurred when evaluating a policy
    #[error("while evaluating policy {}, encountered the following error: {}", &.id, &.error)]
    PolicyEvaluationError {
        id: &PolicyID,
        error: EvaluationError,
    },
}
```

### Step 2

Update the `cedar-policy-core` `Diagnostics` type as follows.

```rust
pub struct Diagnostics {
    /// `PolicyID`s of the policies that contributed to the decision. If no
    /// policies applied to the request, this set will be empty.
    pub reason: HashSet<PolicyID>,
    /// list of errors that occurred
    pub errors: Vec<AuthorizationError>,
}
```

This will break other definitions in `cedar-policy-score`, as you've already pointed out in this thread. One example is on line 101 of authorizer.rs. To fix this, you'll need to replace
```rust
errors.extend(partial.residuals.policies().map(|p| {
    format!(
        "while evaluating policy {}, encountered the following error: {}",
        p.id(),
        EvaluationError::NonValue(p.condition())
    )
}));
```
with something like
```rust
errors.extend(partial.residuals.policies().map(|p| { 
    AuthorizationError::PolicyEvaluationError(p.id(), EvaluationError::NonValue(p.condition()))
}));
```
(Warning: syntax may be incorrect since I haven't tried compiling this myself)

### Step 3

After all breaks in `cedar-policy-core` are fixed, switch to working on `cedar-policy`. First, `pub use` both `cedar_policy_core::evaluator::err::EvaluationError` and `cedar_policy_core::authorizer::err::AuthorizationError`. This will require some small updates (which you've mostly already made) to return an `EvaluationError` value in cases where we were previously returning a formatted string.

### Step 4

Similar to step 2, update the `cedar-policy` `Diagnostics` type to store `AuthorizationError`s instead of `String`s. This will again break things, but the fixes should be similar to previous steps (e.g., returning an error value instead of a formatted string).

----

Anyways, those are my suggestions. It might be useful to commit after each step so I can review your progress incrementally. Also possible: I could do steps 1 and 2 separately first, so that you could do steps 3 and 4 (which are more aligned with the original description of #114). Let me know what you think!

Aside: most of our team also uses the rust-analyzer extension of VSCode. We think it's great! I would recommend turning on the "format on save" option (instructions [here](https://stackoverflow.com/questions/67859926/how-to-run-cargo-fmt-on-save-in-vscode)) so you can avoid having to run `cargo fmt` to format your code.

NetwCodeProjects:
# Re: authorizer/err
I can handle steps 1- 4. I'll let you know if I run into any problems. Most of these steps are what I already have done, the biggest change is creating the authorizer/err.rs which at some point I thought of doing too.

khieta:
Hi @NetwCodeProjects, we recently began conducting a review of our error types and error messages, and this one is a pain point, so I took the liberty of starting the fix myself -- PR #186 corresponds to Steps 1 and 2 outlined above. Steps 3 and 4 are still not done, so you are welcome to continue working on them in this PR (apologies in advance for the merge conflict when/if my PR is merged).

Please leave comments on PR #186 if you see ways that my solution can be improved!

khieta:
#186 is merged! 

Also, as a heads up: we're currently working on improving our error reporting (you may have noticed several recent PRs & issues with "error" in the title). At some point (likely late next week), issue #114 will be on our critical path, so please let me know whether you still intend to work on this & if possible, an estimated timeline.

And thank you for your interest in contributing! If you'd like to find a different issue to work on, then take a look at our issues tagged as `good-first-issue` [[link](https://github.com/cedar-policy/cedar/issues?q=is%3Aopen+is%3Aissue+label%3Agood-first-issue)].

NetwCodeProjects:
Congratz on the merge. Timeline 24 hours or less. I'll send an update tomorrow either I finish or forfeit. The MOVEit hack has me busy on top of building software for clients. In 24 hrs you'll have an answer.

NetwCodeProjects:
# all changes made to cedar-policy api
### 1. The updates I made:

using pub use ...evaluator::EvaluationError 
compared to ...::err::EvaluationError

the reason why -> error[E0603]: module `err` is private.

I did try a version making the mod public
<details>

```rust
// IN cedar-policy-core/src/evaluator.rs
pub mod err;
```

</details>

 which worked but felt unnecessary since the goal is to expose eval-err. So I moved forward based on the cyber security principle of least privilege.

### 2. Next 

from
.map_err(|e| EvaluationError::StringMessage(e.to_string())),
to
.map_err(EvaluationError::from),

<details>

```rust
    /// Get the value for the given attribute, or `None` if not present.
    ///
    /// This can also return Some(Err) if the attribute had an illegal value.
    pub fn attr(&self, attr: &str) -> Option<Result<EvalResult, EvaluationError>> {
        let expr = self.0.get(attr)?;
        let all_ext = Extensions::all_available();
        let evaluator = RestrictedEvaluator::new(&all_ext);
        Some(
            evaluator
                .interpret(expr.as_borrowed())
                .map(EvalResult::from)
                .map_err(EvaluationError::from),
        )
    }
}
```

</details>

### 3. 

Change in api 
```rust 
pub struct Diagnostics {
    reason: HashSet<PolicyId>,
    errors: HashSet<String>,
```
to
```rust
errors: Vec<EvaluationError>
```

<details>

```rust
#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
pub struct Diagnostics {
    /// `PolicyId`s of the policies that contributed to the decision.
    /// If no policies applied to the request, this set will be empty.
    reason: HashSet<PolicyId>,
    /// list of error messages which occurred
    errors: Vec<EvaluationError>,
}
```

</details>

### 4. impl diagnostics
from this 

<details>

```rust
impl Diagnostics {
    ...
    /// Get the error messages
    pub fn errors(&self) -> impl Iterator<Item = EvaluationError> + '_ {
        self.errors
            .iter()
            .cloned()
            .map(EvaluationError::StringMessage)
    }
}
```

</details>

to this 

<details>

```rust

    /// Get the error messages
    pub fn errors(&self) -> impl Iterator<Item = EvaluationError> + '_ {
        self.errors.iter()
    }
}
```
</details>

we don't need to clone errors when returning them: we can just return references. The map function that was formatting the errors into strings is also removed, because we are now storing the errors directly.

With these changes, you can retrieve the EvaluationError values stored in the Diagnostics struct directly. You can then handle these error values however you wish, for example by matching on the error type.

## some thoughts 
serialize/deserialize errors became a rabbit hole, I followed it, finished, then had to do the same thing but adding Eq this time.

im about here the errors: .collect line giving me problems.

<details>

```rust
impl From<authorizer::Diagnostics> for Diagnostics {
    fn from(diagnostics: authorizer::Diagnostics) -> Self {
        Self {
            reason: diagnostics.reason.into_iter().map(PolicyId).collect(),
            errors: diagnostics.errors.iter().map(ToString::to_string).collect(),
        }
    }
}
```

</details>

<details>

```rust
error
a value of type `Vec<cedar_policy_core::evaluator::EvaluationError>` cannot be built from an iterator over elements of type `std::string::String`
the trait `FromIterator<std::string::String>` is not implemented for `Vec<cedar_policy_core::evaluator::EvaluationError>`
the trait `FromIterator<T>` is implemented for `Vec<T>`

// further error detail

error[E0277]: a value of type `Vec<cedar_policy_core::evaluator::EvaluationError>` cannot be built from an iterator over elements of type `std::string::String`
    --> cedar-policy\src\api.rs:372:72
     |
372  |             errors: diagnostics.errors.iter().map(ToString::to_string).collect(),
     |                                                                        ^^^^^^^ value of type `Vec<cedar_policy_core::evaluator::EvaluationError>` cannot be built from `std::iter::Iterator<Item=std::string::String>`
     |
     = help: the trait `FromIterator<std::string::String>` is not implemented for `Vec<cedar_policy_core::evaluator::EvaluationError>`
     = help: the trait `FromIterator<T>` is implemented for `Vec<T>`
note: the method call chain might not have had the expected associated types
    --> cedar-policy\src\api.rs:372:47
     |
372  |             errors: diagnostics.errors.iter().map(ToString::to_string).collect(),
     |                     ------------------ ------ ^^^^^^^^^^^^^^^^^^^^^^^^ `Iterator::Item` changed to `String` here
     |                     |                  |
     |                     |                  `Iterator::Item` is `&AuthorizationError` here
     |                     this expression has type `Vec<AuthorizationError>`
note: required by a bound in `collect`
    --> C:\Users\root\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib/rustlib/src/rust\library\core\src\iter\traits\iterator.rs:1892:19
     |
1892 |     fn collect<B: FromIterator<Self::Item>>(self) -> B
     |                   ^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Iterator::collect`


```

</details>

NetwCodeProjects:
I accidentally closed my pull request. I apologize for any confusion this may have caused.

NetwCodeProjects:
# Stage 3 info
The commit named "Stage 3 code preview" is to demonstrate changes. It is not merge grade yet. only to show the changes made so far.

cedar-policy/src/api.rs
```diff
@@ -127,7 +129,7 @@ impl Entity {
             evaluator
                 .interpret(expr.as_borrowed())
                 .map(EvalResult::from)
-                .map_err(|e| EvaluationError::StringMessage(e.to_string())),
+                .map_err(EvaluationError::from),
```
khieta, line 132: I think you can just drop the `map_err` here

khieta:
Serialize/deserialize is indeed a rabbit hole. That's why I bailed in #186 (see this [comment thread](https://github.com/cedar-policy/cedar/pull/186#discussion_r1265853890)). Unfortunately, I think Serialize & Deserialize are required for `cedar-policy-core`'s types since it's the top-level API.

Re the error message in your last comment, I think you can change this line:
```rust
errors: diagnostics.errors.iter().map(ToString::to_string).collect(),
```
to the following:
```rust
errors: diagnostics.errors,
```

cedar-policy/src/api.rs
```diff
@@ -360,7 +362,7 @@ pub struct Diagnostics {
     /// If no policies applied to the request, this set will be empty.
     reason: HashSet<PolicyId>,
     /// list of error messages which occurred
-    errors: HashSet<String>,
+    errors: Vec<EvaluationError>,
```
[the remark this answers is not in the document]
NetwCodeProjects: line 33 contains pub use auth..
new change ->  errors: Vec<AuthorizationError>

cedar-policy/src/api.rs
NetwCodeProjects: Dropped map_err.

NetwCodeProjects:
It might help to discuss with me where you see AuthorizationError fitting in, the current goal of the PR, where we are and what we want to return instead of string. 

The goal is clarity.
> In software engineering, rubber duck debugging (or rubberducking) is a method of debugging code by articulating a problem in spoken or written natural language.

cedar-policy/src/api.rs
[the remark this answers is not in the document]
khieta: Oops, missed line 33. That's fine then.

Since the `is_authorized` function in `cedar-policy-core` returns an `AuthorizationError`, we'll also want to use that here, instead of an `EvaluationError`.

khieta:
#186 added a new `AuthorizationError` type, which is a more accurate error type for `is_authorized` than `EvaluationError`. It also changed the `Diagnostics` type in `cedar-policy-core` to store `AuthorizationError`s. The goal of this PR is to update the `Diagnostics` type in `cedar-policy` to match (i.e., return `AuthorizationError`s instead of string error messages).

So: the original title and description of #114 is now incorrect. I left them as is because I wanted to preserve the original discussion... but I think I'll go back and change the title at least.

Does that help?

NetwCodeProjects:
It does help. I can see why certain decisions were made in the original and in this update. 
The choice between HashSet and Vec depends on the specific use case. 

HashSet is used when you want to quickly check for the presence of an item and do not want any duplicates. It is implemented as a hash table, providing constant-time average complexity for search, insert, and delete operations, assuming a good hash function.

Vec is a growable array type and allows duplicates. It's used when the order of elements matters and/or when you need to be able to access elements by their index.

In your original Diagnostics struct, HashSet was used for both reason and errors. The choice of HashSet implies that:

You don't want duplicate PolicyIds or error messages.
The order of the PolicyIds or error messages doesn't matter.

The update to Vec implies: 
you want to preserve the order of occurrence for the PolicyIds or error messages or allow identical error messages coming from different parts of the system and you want to record all of them.

This explains how the original code was handling strings, what was built to support that functionality and what we might need to change or remove to support the new Vec.

khieta:
Although `cedar-policy-core` uses a `Vec`, I think a `HashSet` would be preferred because the order of errors shouldn't matter -- the Cedar authorization engine makes no guarantees about the order in which it evaluates policies, because evaluating them in any order will produce the same authorization result. I imagine this was why the `cedar-policy` `Diagnostics` type originally used a `HashSet`. (The public API in `cedar-policy` allows us to hide any weirdness in `cedar-policy-core` 😉) 

So: as long as it doesn't present additional technical difficulties, it would be better to stick with a `HashSet` in this PR.

khieta:
I am planning to start work on this now, so I will be closing this PR.

@NetwCodeProjects thanks for your interest in contributing! Conversing with you helped me flesh out the details for how I want to implement this. It has also been a lesson to me to provide more clear descriptions in issues to make it easier for external contributors to contribute 🙂 

We hope to see contributions from you in the future!

NetwCodeProjects:
It's worth noting that there is a significant vulnerability associated with deserialization called "Deserialization Vulnerability" or "Serialization Vulnerability." This vulnerability occurs when an attacker crafts malicious data that, when deserialized, can lead to code execution, denial of service, or other security issues. 

It can lead to RCE (remote code execution) which is a critical vulnerability. 
(Typically, a $20k bug bounty)
https://www.baeldung.com/java-deserialization-vulnerabilities

A hacker can feed data to the server with code that does something malicious and then, when the server receives the code and "deserializes" it (changing it from binary into an object), that code gets executed on the server side. Similarly, if the server itself is infected, it could feed binary data back to a client (a player) whose PC deserializes it locally and executes the code.

Deserialization vulns were recently exploited in July 2023 on Minecraft servers.

https://www.yahoo.com/lifestyle/security-hole-minecraft-mods-lets-192354367.html

To mitigate deserialization vulnerabilities, it's essential to follow some best practices:

<details>

- Trustworthy sources: Only deserialize data from trusted sources. Avoid deserializing data from untrusted inputs, such as user inputs, external files, or network streams without proper validation.

- !!!Data validation!!!: Always validate and sanitize the input data before deserializing it. Check for unexpected or malicious content and reject data that doesn't adhere to expected formats or ranges.

- Use strong types: Use strong typing and avoid using generic deserialization functions when possible. Strongly-typed deserialization can provide an extra layer of security by limiting the scope of what data can be deserialized.

- Limit access privileges: Ensure that the code handling the deserialization has the least possible privileges required for its operation. This reduces the impact if an attacker successfully exploits a vulnerability.

- Beware of default behavior: Be cautious when using default deserialization settings, as they might not provide the necessary security measures. Always configure and customize the deserialization process according to your specific use case.

- Security Audits: Regularly review and audit the code handling deserialization to identify and fix potential vulnerabilities.

</details>

I began to review the code handling deserialization quietly because it might not be an issue but you should know about this now.

cdisselkoen:
For security issues, see our [policy](https://github.com/cedar-policy/cedar/blob/main/SECURITY.md) -- please do not discuss on public GitHub issues, and use our dedicated reporting mechanisms.

If you're just concerned about security in general, we do have an AppSec (application security) team at AWS that has reviewed this code and has not found vulnerabilities.  But, if you believe a vulnerability may exist, please don't hesitate to report via the mechanisms linked in our policy above.

PR #124, opened 2023-06-16
hide `Arbitrary` implementations behind crate features, not `cfg(fuzzing)`

cdisselkoen:
In support of https://github.com/cedar-policy/cedar-spec/issues/46, this lets downstream crates use these `Arbitrary` implementations even when not `cfg(fuzzing)`, just by enabling the new `arbitrary` feature.  As a bonus, `arbitrary` itself becomes an optional dependency, rather than required.

### DIFF
--- a/cedar-policy-core/Cargo.toml
+++ b/cedar-policy-core/Cargo.toml
@@ -23,6 +23,7 @@ rustc_lexer = "0.1"
 thiserror = "1.0"
 smol_str = { version = "0.2", features = ["serde"] }
 stacker = "0.1.15"
+arbitrary = { version = "1", features = ["derive"], optional = true }
 
 # ipaddr extension requires ipnet
 ipnet = { version = "2.5.0", optional = true }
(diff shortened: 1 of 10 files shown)


aaronjeline (approved):

john-h-kastner-aws (approved):

cdisselkoen:
Unsurprisingly, we see that this has broken the build of downstream `cedar-drt`, which needs to be changed to enable the new `arbitrary` feature.  This is a little bit of chicken-and-egg, in that we can't make the `cedar-drt` change first, because the `arbitrary` feature doesn't exist yet.  Technically we could get around this by first adding `arbitrary` as a no-op feature in these crates, then adjusting `cedar-drt` to enable it, and finally changing the `Arbitrary` implementations to be gated on the feature instead of `cfg(fuzzing)`, as in this PR.  I propose to not bother, and just override-and-merge here and then make the change to `cedar-drt` shortly.  Any objections?

andrewmwells-amazon:
> Unsurprisingly, we see that this has broken the build of downstream `cedar-drt`, which needs to be changed to enable the new `arbitrary` feature. This is a little bit of chicken-and-egg, in that we can't make the `cedar-drt` change first, because the `arbitrary` feature doesn't exist yet. Technically we could get around this by first adding `arbitrary` as a no-op feature in these crates, then adjusting `cedar-drt` to enable it, and finally changing the `Arbitrary` implementations to be gated on the feature instead of `cfg(fuzzing)`, as in this PR. I propose to not bother, and just override-and-merge here and then make the change to `cedar-drt` shortly. Any objections?

Sounds good to me.

PR #131, opened 2023-06-26
Generate ESTs on demand, not eagerly

### LINKED ISSUE #129: Remove EST conversion from critical path of processing text policies
### Category

Cedar language features

Currently, if you provide policies using `Policy::from_str()` or similar, the policy is converted to EST "eagerly" even if the EST is never demanded via `Policy::to_json()`.  This slows down the critical path of important operations such as `is_authorized()`.  The conversion to EST should instead be done lazily, on-demand.

The `cedar_policy::Policy` type does need to carry some "lossless" representation (see discussion on #125 for context) in order to properly support `Policy::to_json()`.  Currently, that's the EST, but it could instead be the CST or policy text format, which could be converted to EST on demand.  If we just swapped it out directly, though, we'd have the opposite problem, where if you provide that policy as EST (via `Policy::from_json()`), it would have to be converted to CST or text on the critical path.  The most flexible solution, that keeps both paths fast, would be to store either the EST or the CST/text, whichever is provided, and convert to the other on demand.

Discussed above.

### Additional context

Requested by a customer

- [X] 👋 I may be able to implement this feature request

cdisselkoen:
Fixes #129; see discussion there


cedar-policy-core/src/parser.rs
```diff
@@ -47,6 +47,28 @@ pub fn parse_policyset(text: &str) -> Result<ast::PolicySet, Vec<err::ParseError
...
+    text: &str,
+) -> Result<(HashMap<ast::PolicyID, &str>, ast::PolicySet), err::ParseErrors> {
+    let mut errs = Vec::new();
+    let cst = text_to_cst::parse_policies(text).map_err(err::ParseErrors)?;
+    let pset = cst
+        .to_policyset(&mut errs)
+        .ok_or_else(|| err::ParseErrors(errs.clone()))?;
+    let texts = cst
+        .with_generated_policyids()
+        .expect("shouldn't be None since parse_policies() and to_policyset() didn't return Err")
```
aaronjeline, line 62: Are we sure on this?

cdisselkoen: This `expect` is also present in the function below, with the `allow(clippy::expect_used)` annotation. If we're confident there, I'm confident here

aaronjeline (approved):

shaobo-he-aws (approved):
LGTM.

andrewmwells-amazon (approved):

PR #132, opened 2023-06-26
Improve error messages for malformed `SchemaType`s

john-h-kastner-aws:
A trick used to automatically parse these with `serde` resulted in inscrutable error messages when a  required field was omitted or an unrecognized field was included. These issues are avoided by manually implementing `Derive` for this struct.

The deserialization code is not very nice, so I'm open to code review suggestion on that.

### DIFF
--- a/cedar-policy-validator/src/schema_file_format.rs
+++ b/cedar-policy-validator/src/schema_file_format.rs
@@ -15,7 +15,10 @@
  */
 
 use cedar_policy_core::entities::JSONValue;
-use serde::{Deserialize, Serialize};
+use serde::{
+    de::{MapAccess, Visitor},
+    Deserialize, Serialize,
+};
 use serde_with::serde_as;
 use smol_str::SmolStr;
 use std::collections::{BTreeMap, HashMap};
@@ -189,7 +192,7 @@ impl std::fmt::Display for ActionEntityUID {
 
 /// A restricted version of the `Type` enum containing only the types which are
 /// exposed to users.
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize)]
 // This enum is `untagged` with these variants as a workaround to a serde
 // limitation. It is not possible to have the known variants on one enum, and
 // then, have catch-all variant for any unrecognized tag in the same enum that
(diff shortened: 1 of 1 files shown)


cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -203,15 +206,219 @@ pub enum SchemaType {
...
+                        return Err(serde::de::Error::duplicate_field("name"));
+                    }
+                    name = Some(map.next_value()?)
+                }
+            }
+        }
+
+        match type_name.as_ref().map(|s| s.as_str()) {
+            Some("String") => {
+                if element.is_some() {
```
andrewmwells-amazon, line 284: [nit] I think these 3 and `type_name` could be refactored to use a function that takes `SchemaType`.

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -203,15 +206,219 @@ pub enum SchemaType {
...
+                Fields::Name => {
+                    if name.is_some() {
+                        return Err(serde::de::Error::duplicate_field("name"));
+                    }
+                    name = Some(map.next_value()?)
+                }
+            }
+        }
+
+        match type_name.as_ref().map(|s| s.as_str()) {
```
andrewmwells-amazon, line 282: [nit] I think this match statement could be `fn try_string_to_schema_type` or similar.

shaobo-he-aws:
Could you share why the existing implementation is not good enough? Is it because we don't or can't enforce error on unknown fields of `SchemaType`?

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -203,15 +206,219 @@ pub enum SchemaType {
...
+            Name,
+        }
+
+        let mut type_name: Option<SmolStr> = None;
+        let mut element: Option<SchemaType> = None;
+        let mut attributes: Option<BTreeMap<SmolStr, TypeOfAttribute>> = None;
+        let mut additional_attributes: Option<bool> = None;
+        let mut name: Option<SmolStr> = None;
+
+        while let Some(key) = map.next_key()? {
```
shaobo-he-aws, line 247: IIUC, we early deserialize the values even if their corresponding keys are deemed to be problematic after we know all keys after this loop. For instance, we still deserialize the value corresponding to `attributes` in `{"attributes": ..., "type": "Bool"}`. I wonder if we can do it lazily. For instance, store the JSON representations of the values and deserialize them if we know the key combination is correct.

john-h-kastner-aws: I can do this by making the variables `Option<Result<T, M::Err>>`.

shaobo-he-aws:
> The problem solved by the original implementation was that I wanted a "catch all" case in an enum. The `SchemaType` enum should have some known variants for each built-in type tagged by the `type` field, each with some additional data specific to that variant. There should also be an additional case which captures any case where the `type` field is not one of the built-in values.

Just to double check. Is it the "catch call" case used for `TypeDef`? If it is, I was wondering if we can inline `SchemaTypeVariant` into `SchemaType`.

> 2. The tag field `type` was not present. Neither the tagged enum case nor the catch-all case captures this  since both expect that `type` will be present. This is correct, but the intermediated struct used in the implementation gives the error message `data did not match any variant of untagged enum SchemaType` rather than specifically complaining about the lack of a `type` tag.

Yes, it seems we don't have a workaround for this if we stick to the `untagged` attribute (https://github.com/serde-rs/serde/issues/773). Maybe we should change the error message in the PR from `serde::de::Error::missing_field("type")` to something like `serde::de::Error::custom("a `type` field is required")`?

cedar-policy-validator/src/err.rs
```diff
@@ -134,6 +134,23 @@ impl SchemaError {
     }
 }
 
+#[derive(Debug)]
```
shaobo-he-aws, line 137: Not sure if it makes sense to replace it with `Either<EntityUID, Name>`.

shaobo-he-aws (approved):
LGTM.

cedar-policy-validator/src/err.rs
cdisselkoen: I prefer helpful variant names (in this case, `ActionContext` and `EntityTypeShape`) to the bland ones `Left` and `Right`, particularly in this case where they provide additional information (that the `EntityUID` is for an action, for instance)

cdisselkoen: We could even go farther and name the `EntityUID` field of the `ActionContext` variant, eg, `action_uid`.

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -203,15 +206,266 @@ pub enum SchemaType {
...
+    fn visit_map<M>(self, mut map: M) -> std::result::Result<Self::Value, M::Error>
+    where
+        M: MapAccess<'de>,
+    {
+        use TypeFields::*;
+
+        // We keep field values wrapped in a `Result` initially so that we do
+        // not report errors due the contents of a field when the field is not
+        // expected for a particular type variant. We instead report that the
+        // field so not exist at all, so that the schema author can delete the
```
cdisselkoen, line 280: minor typo: "field so not exist"

cdisselkoen (approved):
LGTM, the custom deserializer will give us more flexibility in the future even if it might be more of a pain to maintain

PR #134, opened 2023-06-27
Implement RFC 9

cdisselkoen:
Fixes #133 and implements [RFC 9](https://github.com/cedar-policy/rfcs/blob/main/text/0009-disallow-whitespace-in-entityuid.md); see comments there.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #108, opened 2023-06-01
Simplify schema construction

john-h-kastner-aws:
The simplification mainly comes from postponing inverting the `memberOf` relation into a `descendants` relation until the last step of construction instead of inverting it immediately and threading the inverted relation through the rest of the computation.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #141, opened 2023-06-28
[2.3.x] Implement RFC 9

cdisselkoen:
Cherry-picks #134 onto the 2.3.x branch.  This cherry-pick was not trivial, in particular I think the absence of #108 on the 2.3.x branch caused a bunch of conflicts in the validator code.  Requesting review from @john-h-kastner-aws in particular as a result, as I made the fixes that "looked sensible" to me but it would be good to get John's signoff. 

### DIFF
--- a/cedar-policy-core/src/ast/entity.rs
+++ b/cedar-policy-core/src/ast/entity.rs
@@ -17,6 +17,7 @@
 use crate::ast::*;
 use crate::parser::err::ParseError;
 use crate::transitive_closure::TCNode;
+use crate::FromNormalizedStr;
 use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 use smol_str::SmolStr;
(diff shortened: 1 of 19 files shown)


cdisselkoen:
Downstream CI fails because the current DRT version assumes the presence of #124, so it doesn't work with this 2.3.x branch.  @aaronjeline or others, I assume we can get DRT to work with the 2.3.x branch by using the version of the DRT library we use for 2.2.x rather than `main`?  I.e. this is just a CI problem not a fundamental problem?

shaobo-he-aws (approved):

john-h-kastner-aws (commented):
Looks alright to me.

john-h-kastner-aws (approved):

andrewmwells-amazon:
> Downstream CI fails because the current DRT version assumes the presence of #124, so it doesn't work with this 2.3.x branch. @aaronjeline or others, I assume we can get DRT to work with the 2.3.x branch by using the version of the DRT library we use for 2.2.x rather than `main`? I.e. this is just a CI problem not a fundamental problem?

Currently we only build `cedar-spec` against the latest commit in the `main` branch of `cedar`, so this won't actually break anything. We could skip the downstream build check for PRs that aren't going into `main` or we could cut branches of `cedar-spec` for each `cedar` release.

Issue #67, opened 2023-05-24
Beautify CLI error messages


shaobo-he-aws:
### Category

Other

### Describe the feature you'd like to request

We can deploy rustc-style error message reporting (i.e., with colors, highlighting, and explanatory links).

### Describe the solution you'd like

We can use existing libraries like [miette](https://github.com/zkat/miette).

### Describe alternatives you've considered

Or we can perform better formatting of the error messages.

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

shaobo-he-aws:
Closed by #148.

PR #148, opened 2023-07-03
Revamp parse errors

spinda:
*Issue #, if available:* #67

*Description of changes:*

This transforms parse errors from something that looks like this:

image

into something that looks like this:

image

The pretty rendering is powered by the [`miette` crate](https://crates.io/crates/miette/5.1.0), which I've swapped in place of `anyhow`. I've also done a bit of `lalrpop` plumbing to make user-facing token names nice and track source locations for `ParseError::User` messages, made improvements to the underlying `ParseError`/`ParseErrors` types and parsing infrastructure, and added error message format options to the CLI.

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

### DIFF
--- a/cedar-policy-cli/src/lib.rs
+++ b/cedar-policy-cli/src/lib.rs
@@ -19,10 +19,8 @@
 // omitted.
 #![allow(clippy::needless_return)]
 
-use anyhow::{Context as _, Error, Result};
-use cedar_policy::*;
-use cedar_policy_formatter::{policies_str_to_pretty, Config};
 use clap::{Args, Parser, Subcommand};
+use miette::{miette, IntoDiagnostic, NamedSource, Report, Result, WrapErr};
 use serde::{Deserialize, Serialize};
 use std::{
     collections::HashMap,
@@ -32,6 +30,9 @@ use std::{
     str::FromStr,
     time::Instant,
 };
+
+use cedar_policy::*;
+use cedar_policy_formatter::{policies_str_to_pretty, Config};
 
 /// Basic Cedar CLI for evaluating authorization queries
 #[derive(Parser)]
@@ -104,8 +105,10 @@ impl RequestArgs {
         match &self.request_json_file {
             Some(jsonfile) => {
                 let jsonstring = std::fs::read_to_string(jsonfile)
-                    .context(format!("failed to open request-json file {jsonfile}"))?;
+                    .into_diagnostic()
+                    .wrap_err_with(|| format!("failed to open request-json file {jsonfile}"))?;
                 let qjson: RequestJSON = serde_json::from_str(&jsonstring)
+                    .into_diagnostic()
                     .context(format!("failed to parse request-json file {jsonfile}"))?;
                 let principal = qjson
                     .principal
@@ -134,7 +137,9 @@ impl RequestArgs {
                 let context = Context::from_json_value(
                     qjson.context,
                     schema.and_then(|s| Some((s, action.as_ref()?))),
-                )?;
+                )
+                .into_diagnostic()
+                .wrap_err_with(|| format!("failed to create a context from {jsonfile}"))?;
                 Ok(Request::new(principal, action, resource, context))
             }
             None => {
@@ -168,9 +173,12 @@ impl RequestArgs {
                         Ok(f) => Context::from_json_file(
                             f,
                             schema.and_then(|s| Some((s, action.as_ref()?))),
-                        )?,
-                        Err(e) => Err(Error::from(e)
-                            .context(format!("error while loading context from {jsonfile}")))?,
+                        )
+                        .into_diagnostic()
+                        .wrap_err_with(|| format!("failed to create a context from {jsonfile}"))?,
+                        Err(e) => Err(e).into_diagnostic().wrap_err_with(|| {
+                            format!("error while loading context from {jsonfile}")
+                        })?,
                     },
                 };
                 Ok(Request::new(principal, action, resource, context))
@@ -333,7 +341,7 @@ pub fn check_parse(args: &CheckParseArgs) -> CedarExitCode {
     match read_policy_set(args.policies_file.as_ref()) {
         Ok(_) => CedarExitCode::Success,
         Err(e) => {
-            println!("{:#}", e);
+            println!("{:?}", e);
             CedarExitCode::Failure
         }
     }
@@ -460,14 +468,16 @@ fn create_slot_env(data: &HashMap<SlotId, String>) -> Result<HashMap<SlotId, Ent
 fn link_inner(args: &LinkArgs) -> Result<()> {
     let mut policies = read_policy_set(Some(&args.policies_file))?;
     let slotenv = create_slot_env(&args.arguments.data)?;
-    policies.link(
-        PolicyId::from_str(&args.template_id)?,
-        PolicyId::from_str(&args.new_id)?,
-        slotenv,
-    )?;
+    policies
+        .link(
+            PolicyId::from_str(&args.template_id)?,
+            PolicyId::from_str(&args.new_id)?,
+            slotenv,
+        )
+        .into_diagnostic()?;
     let linked = policies
         .policy(&PolicyId::from_str(&args.new_id)?)
-        .context("Failed to add template-linked policy")?;
+        .ok_or_else(|| miette!("Failed to add template-linked policy"))?;
     println!("Template Linked Policy Added: {linked}");
     let linked = TemplateLinked {
         template_id: args.template_id.clone(),
@@ -539,11 +549,13 @@ impl From<TemplateLinked> for LiteralTemplateLinked {
 fn add_template_links_to_set(path: impl AsRef<Path>, policy_set: &mut PolicySet) -> Result<()> {
     for template_linked in load_liked_file(path)? {
         let slot_env = create_slot_env(&template_linked.args)?;
-        policy_set.link(
-            PolicyId::from_str(&template_linked.template_id)?,
-            PolicyId::from_str(&template_linked.link_id)?,
-            slot_env,
-        )?;
+        policy_set
+            .link(
+                PolicyId::from_str(&template_linked.template_id)?,
+                PolicyId::from_str(&template_linked.link_id)?,
+                slot_env,
+            )
+            .into_diagnostic()?;
     }
     Ok(())
 }
@@ -557,12 +569,19 @@ fn load_liked_file(path: impl AsRef<Path>) -> Result<Vec<TemplateLinked>> {
             return Ok(vec![]);
         }
     };
-    if f.metadata().context("Failed to read metadata")?.len() == 0 {
+    if f.metadata()
+        .into_diagnostic()
+        .wrap_err("Failed to read metadata")?
+        .len()
+        == 0
+    {
         // File is empty, return empty set
         Ok(vec![])
     } else {
         // File has contents, deserialize
-        serde_json::from_reader(f).context("Deserialization error")
+        serde_json::from_reader(f)
+            .into_diagnostic()
+            .wrap_err("Deserialization error")
     }
 }
 
@@ -579,8 +598,9 @@ fn write_template_linked_file(linked: &[TemplateLinked], path: impl AsRef<Path>)
         .write(true)
         .truncate(true)
         .create(true)
-        .open(path)?;
-    Ok(serde_json::to_writer(f, linked)?)
+        .open(path)
+        .into_diagnostic()?;
+    Ok(serde_json::to_writer(f, linked).into_diagnostic()?)
 }
 
 pub fn authorize(args: &AuthorizeArgs) -> CedarExitCode {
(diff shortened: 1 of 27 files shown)


spinda:
There's room for future work here.

- `ParseError::ToASTError` is still just a string and carries no source location info with it.
- Other types error types in the Cedar the pipeline could be given [`Diagnostic`](https://docs.rs/miette/latest/miette/trait.Diagnostic.html) instances (see https://docs.rs/miette/latest/miette/index.html#example for to derive those).
- Location information could be threaded through the AST structures Cedar uses beyond the parsing stage, which could then be attached to those further-along-the-pipeline error messages. A permission-denied error could point to the corresponding `forbid` policy rule, for example.
- `miette` allows us to attach documentation URLs to errors, which sounds like a big UX win. The syntax for that is `#[diagnostic(url(https://...))]`, or `#[diagnostic(url(docsrs))]` to automatically generate a [docs.rs](https://docs.rs) link to the error type.

shaobo-he-aws:
> There's room for future work here.
> 
> * `ParseError::ToASTError` is still just a string and carries no source location info with it.
> * Other types error types in the Cedar the pipeline could be given [`Diagnostic`](https://docs.rs/miette/latest/miette/trait.Diagnostic.html) instances (see https://docs.rs/miette/latest/miette/index.html#example for how to derive those).
> * Location information could be threaded through the AST structures Cedar uses beyond the parsing stage, which could then be attached to those further-along-the-pipeline error messages. A permission-denied error could point to the corresponding `forbid` policy rule, for example.
> * `miette` allows us to attach documentation URLs to errors with `Diagnostic` instances, which sounds like a big UX win. The syntax for that is `#[diagnostic(url(https://...))]`, or `#[diagnostic(url(docsrs))]` to automatically generate a [docs.rs](https://docs.rs) link to the error type.

These are great suggestions. Maybe we should create individual issues for them?

shaobo-he-aws:
Btw, the second error message in your example seems confusing to me (or it is when combined with the first error message). Maybe we should also create an issue for it?

shaobo-he-aws:
QQ: there are a lot of changes about `Vec<ParseError>` to `ParseErrors` in cedar-policy-core. Why do we want them?

spinda:
> Btw, the second error message in your example seems confusing to me (or it is when combined with the first error message). Maybe we should also create an issue for it?

Yeah, I noticed that. I think it's because of error recovery: the parser is attempting to ignore the unexpected `&&` token and continue parsing, but then it comes to a `::` token it doesn't expect because it doesn't know it's on the right-hand-side of a binary operator.

> QQ: there are a lot of changes about `Vec<ParseError>` to `ParseErrors` in cedar-policy-core. Why do we want them?

`ParseErrors` is set up with `miette` to give nice error messages and group related parse errors together. And most of the `Vec<ParseError>`s we were returning ended up wrapped in `ParseErrors` anyway, so this eliminates a lot of conversion boilerplate.

cedar-policy-core/Cargo.toml
```diff
@@ -24,6 +24,8 @@ thiserror = "1.0"
 smol_str = { version = "0.2", features = ["serde"] }
 stacker = "0.1.15"
 arbitrary = { version = "1", features = ["derive"], optional = true }
+miette = "5.9.0"
+phf = { version = "0.11.2", features = ["macros"] }
```
aaronjeline, line 28: How essential is this dep?

spinda: Not very essential. It just makes the parse-token-to-friendly-name mapping faster for error message generation, but the difference vs a regular `match` probably isn't noticeable.

shaobo-he-aws:
> `ParseErrors` is set up with `miette` to give nice error messages and group related parse errors together. And most of the `Vec<ParseError>`s we were returning ended up wrapped in `ParseErrors` anyway, so this eliminates a lot of conversion boilerplate.

I could easily be wrong but I think cedar-policy-core shouldn't depend on `miette` as its functionalities should be limited to provide essential APIs to use Cedar programatically. Pretty-printing error messages should be a feature built upon it.

aaronjeline:
I think this looks great overall, but we need a really high testing bar for this change.

aaronjeline:
> I could easily be wrong but I think cedar-policy-core shouldn't depend on miette as its functionalities should be limited to providing essential APIs to use Cedar programatically. Pretty-printing error messages should be a feature built upon it.

I think it's fine, providing high quality parse errors should be a core feature of the Cedar parser.

mwhicks1:
> I think this looks great overall, but we need a really high testing bar for this change.

What do you suggest? Is property-testing of the parser for a few days sufficient? What else?

cedar-policy-core/src/parser.rs
```diff
@@ -268,23 +275,78 @@ pub(crate) fn parse_literal(val: &str) -> Result<ast::Literal, Vec<err::ParseErr
...
 /// parse an identifier
 ///
 /// Private to this crate. Users outside Core should use `Id`'s `FromStr` impl
-/// or its constructors
-pub(crate) fn parse_ident(id: &str) -> Result<ast::Id, Vec<err::ParseError>> {
-    let mut errs = Vec::new();
-    text_to_cst::parse_ident(id)?
-        .to_valid_ident(&mut errs)
-        .ok_or(errs)
+/// or its constructorsVec<err::Error>Vec<err::Error>Vec<err::Error>
```
cdisselkoen, line 293: typo here

cedar-policy-core/src/parser.rs
```diff
@@ -268,23 +275,78 @@ pub(crate) fn parse_literal(val: &str) -> Result<ast::Literal, Vec<err::ParseErr
...
+    };
+    match (principal, action, resource, errs.as_slice()) {
+        (Some(p), Some(a), Some(r), &[]) => Ok(ast::Request {
+            principal: ast::EntityUIDEntry::concrete(p),
+            action: ast::EntityUIDEntry::concrete(a),
+            resource: ast::EntityUIDEntry::concrete(r),
+            context,
+        }),
+        _ => Err(errs),
+    }
```
cdisselkoen, line 349: this was recently removed -- do we need to add it back?

cedar-policy-core/src/parser/err.rs
```diff
@@ -14,51 +14,74 @@
...
-    #[error("error while {context}: {}", MultipleParseErrors(errs))]
+    #[error("error while {context}")]
     WithContext {
         /// What we were trying to do
         context: String,
         /// Error(s) we encountered while doing it
-        errs: Vec<ParseError>,
+        #[source]
+        #[diagnostic_source]
+        errs: ParseErrors,
```
cdisselkoen, line 62: will this hide the actual parse error and only show "error while parsing an EUID" or similar?  Shouldn't the `#[error]` include `{errs}`?

cedar-policy-core/src/parser/err.rs
```diff
@@ -68,31 +91,275 @@ impl ParseErrors {
...
 }
 
-impl std::fmt::Display for ParseErrors {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", MultipleParseErrors(&self.0))
+impl Display for ParseErrors {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match &self.0[..] {
+            [err] => Display::fmt(err, f),
+            _ => write!(f, "multiple parse errors"),
```
cdisselkoen, line 98: is this ok? don't we want to display what the errors were, in the multiple-errors case?

cdisselkoen (commented):
Changes overall look great to me.

cdisselkoen:
> I think cedar-policy-core shouldn't depend on `miette` ... Pretty-printing error messages should be a feature built upon it.

I think I disagree with this.  Error messages, which includes "good" / "pretty" error messages, are a fundamental part of a compiler or runtime.

cedar-policy-cli/src/lib.rs
```diff
@@ -104,8 +143,10 @@ impl RequestArgs {
         match &self.request_json_file {
             Some(jsonfile) => {
                 let jsonstring = std::fs::read_to_string(jsonfile)
-                    .context(format!("failed to open request-json file {jsonfile}"))?;
+                    .into_diagnostic()
+                    .wrap_err_with(|| format!("failed to open request-json file {jsonfile}"))?;
                 let qjson: RequestJSON = serde_json::from_str(&jsonstring)
+                    .into_diagnostic()
                     .context(format!("failed to parse request-json file {jsonfile}"))?;
```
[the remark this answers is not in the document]
spinda: Oops, thanks for spotting this! `wrap_err` (and `wrap_err_with`) is more idiomatic for `miette`. `context` and (`context_with`) is defined as an alias for porting code over from a`anyhow`. I originally had code left using the `context` wrapper when I had compatibility things in place for SemVer reasons, and missed this when ripping them out.

Additionally, `wrap_err_with` (`context_with` for `anyhow`) should be used here because `format!` allocates a new string. Using the `_with` variant and wrapping that up in a lambda means the allocation only happens on the error path.

cedar-policy-core/src/ast/literal.rs
```diff
@@ -71,7 +71,7 @@ impl std::fmt::Display for Literal {
 }
 
 impl std::str::FromStr for Literal {
-    type Err = Vec<parser::err::ParseError>;
+    type Err = crate::parser::err::ParseErrors;
```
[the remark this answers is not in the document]
spinda: Well spotted, will fix!

cedar-policy-core/src/parser.rs
spinda: Nope, must be an artifact of rebasing from `2.2` -> `2.3`.

cedar-policy-core/src/parser/err.rs
spinda: The `#[source]` and `#[diagnostic_source]` attributes will cause the inner `ParseErrors` error to be printed out when formatting the `WithContext`. It'll look like:

```
error while doing a thing

caused by:
  multiple parse errors:
    some parse error
    some other parse error
```

cedar-policy-core/src/parser/err.rs
spinda: The `Diagnostic::related` implementation for `ParseErrors` takes care of that: https://github.com/cedar-policy/cedar/pull/148/files/bd323f02f4b36018b5e3e5484a81896825d7bd04#diff-a0c6854c9e052ae63e9897a05331af59064d413ab7f360f54b41e762af89e862R106-R111

The sample image from the PR has an example of how this looks: https://user-images.githubusercontent.com/7958605/250595855-bbd5ac88-b993-4497-a67b-c8823cd72bd0.png

Including the source errors in the `Display` instance would print them out where `multiple parse errors` is, then duplicate them when the source errors are printed out below that. In general, the `Display` trait for an error should cover that error type itself, and let `Error`/`Diagnostic` instances direct error formatters how to dig deeper into root causes.

cedar-policy-core/src/parser/text_to_cst.rs
```diff
@@ -29,92 +29,89 @@ lalrpop_mod!(
...
-                }
-                Err(e) => {
-                    errors.push(e);
-                    Err(errors)
-                }
-            }
-        })
+// This helper function calls a generated parser, collects errors that could be
+// generated multiple ways, and returns a single Result where the error type is
+// `err::ParseError`.
```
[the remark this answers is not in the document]
spinda: Yes it is, thanks! I wrote this comment before swapping that around.

spinda:
@shaobo-he-aws
> I could easily be wrong but I think cedar-policy-core shouldn't depend on miette as its functionalities should be limited to providing essential APIs to use Cedar programatically. Pretty-printing error messages should be a feature built upon it.

The way I have things set up, `cedar-policy-cli` uses `miette` to pretty-print errors from `cedar-policy-core` to the console (c.f. ["Using `miette` in application code"](https://docs.rs/miette/latest/miette/#-in-application-code)).

`cedar-policy-core` also depends on `miette` to implement the [`Diagnostic`](https://docs.rs/miette/latest/miette/trait.Diagnostic.html) trait on its error types, which is an extension of the standard library `Error` trait (c.f. ["Using `miette` in libraries"](https://docs.rs/miette/latest/miette/#-in-libraries)). The `Diagnostic` protocol expresses a common way to extract metadata from an error type, like error codes, documentation URLs, source information, etc. Importing `miette` for the library use-case pulls in a [minimal set of dependencies](https://github.com/zkat/miette/blob/2e3e5c9d15e234495369e9b47d032644dd5664ad/Cargo.toml#L15-L19) to enable this.

Then on the application side, you can register various error formatters which consume the `Error`/`Diagnostic` traits to render error messages, like I do [here](https://github.com/spinda/cedar/blob/feature/spinda/better-parse-errors/cedar-policy-cli/src/main.rs#L30-L39) in the CLI. Out of the box, you get the fancy console-pretty-printing, a plain-text printer, and a machine-readable JSON printer, which I'm making available through a CLI arg. Other dependents of `cedar-policy` could use those same formatters, or consume the traits directly to surface that metadata in another way, e.g., via a GUI of some sort or perhaps LSP for IDEs.

cedar-policy-core/src/parser/err.rs
spinda: Notable, if there's only a single parse error in `errs`, it'll look like:

```
error while doing a thing

caused by:
  some parse error
```

since I set up `ParseErrors` to pass everything through transparently if there's only one error in the collection.

aaronjeline (approved):

aaronjeline (approved):

cedar-policy-core/src/parser/fmt.rs
```diff
@@ -427,6 +428,39 @@ impl std::fmt::Display for Slot {
     }
 }
 
+pub(crate) fn join_with_conjunction<T, W: Write>(
```
cdisselkoen, line 431: maybe a short doc string describing what this does?

cdisselkoen (approved):
Looks great!  Thanks for these changes

aaronjeline (approved):

spinda:
Rebased to resolve the `Cargo.toml` merge conflict; no other changes with the above push.

shaobo-he-aws (approved):

Issue #150, opened 2023-07-04
Action groups must be fully qualified


mwhicks1:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Schemas and Validation

### Describe the bug

An action's group membership requires a fully qualified name even when the specified group is within the same namespace.

### Expected behavior

When a referenced entity is defined in the same namespace in which it is referenced, I'd expect it not to require namespace qualification.

### Reproduction steps

Use the Cedar CLI to validate the policy `permit(principal,action,resource);` using the schema given below.

### Code Snippet

If you attempt to validate a policy using the following schema, you will see the error.
```
{
  "ExampleCo::Personnel": {
    "entityTypes": {},
    "actions": {
      "read": {},
      "viewPhoto": {
          "memberOf": [ 
              {
                "id": "read",
                "type": "Action"
             }
          ]
      }
    }
  }
}
```
The error goes away if you change the value of `type` to be `ExampleCo::Personnel::Action` instead.

### Log output

```
failed to parse schema from file dumb.schema: Undeclared actions: {"Action::\"read\""}
```

john-h-kastner-aws:
As a workaround, the `type` field is not required here, and, when it is not present, the action takes the default namespace.  You can write `{"id": "read"}` and it will be interpreted as `{"type": "ExampleCo::Personnel::Action", "id": "read"}`.

We should still update the behavior of an unqualified `"type"` to match how to behaves in an `memberOfTypes` list and elsewhere.

mwhicks1:
> As a workaround, the type field is not required here

Ack. Will update the docs to say this.

john-h-kastner-aws:
We should be careful releasing this change since it's a breaking change for anyone relying on the current behavior: 

This schema would become an error since `NS::Action::"group"` isn't defined, or, if it is defined, a policy with scope constraint `action in NS::Action::"group"` would start to apply to `NS::Action::"act"` where it did not before, potentially causing validation errors 

```json
{
  "": {
    "entityTypes": { },
    "actions": {
      "group": {
        "appliesTo": {"principalTypes": "User"},
      }
    }
  },
[...]

PR #151, opened 2023-07-05
Fixing bug #150 (Implicit namespaces in actions)

aaronjeline:


cedar-policy-validator/src/lib.rs
```diff
@@ -319,4 +319,71 @@ mod test {
 
         Ok(())
     }
+
+    #[test]
+    fn test_action_namespace_inference_multi() {
```
john-h-kastner-aws, line 324: I'd rather move this test to `schema.rs`, and then update it to specifically assert that schema that's built contains an action with the expected `memberOf`. Also the schema can use `json!` macro.

cedar-policy-validator/src/schema.rs
```diff
@@ -1013,7 +1019,7 @@ impl ValidatorSchema {
             for p_entity in action.applies_to.applicable_principal_types() {
                 match p_entity {
                     EntityType::Concrete(p_entity) => {
-                        if !entity_types.contains_key(p_entity) {
+                        if !entity_types.contains_key(&p_entity) {
```
john-h-kastner-aws, line 1022: No `&` here (and below). vscode's rust-analyzer and clippy disagree on this, and this line has churned a couple times already

john-h-kastner-aws (changes requested):

cedar-policy-validator/src/schema.rs
```diff
@@ -621,7 +621,12 @@ impl ValidatorNamespaceDef {
         namespace: Option<&Name>,
     ) -> Result<EntityUID> {
         let namespaced_action_type = if let Some(action_ty) = &action_id.ty {
-            Name::from_normalized_str(action_ty).map_err(SchemaError::EntityTypeParseError)?
+            let name =
```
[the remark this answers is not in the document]
aaronjeline: That is different behavior than I implemented, but we want to go with that behavior I assume?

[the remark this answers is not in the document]
john-h-kastner-aws: Yes. Entity types should be qualified according to the same rules everywhere

cedar-policy-core/src/ast/name.rs
```diff
@@ -304,6 +304,8 @@ impl<'a> arbitrary::Arbitrary<'a> for Id {
 
 #[cfg(test)]
 mod test {
+    use std::str::FromStr;
```
john-h-kastner-aws, line 307: nit: leftover import

john-h-kastner-aws (approved):

cedar-policy-core/src/ast/name.rs
john-h-kastner-aws:
```suggestion
```

john-h-kastner-aws (approved):

cedar-policy-validator/src/schema.rs
```diff
@@ -621,7 +621,8 @@ impl ValidatorNamespaceDef {
         namespace: Option<&Name>,
     ) -> Result<EntityUID> {
         let namespaced_action_type = if let Some(action_ty) = &action_id.ty {
-            Name::from_normalized_str(action_ty).map_err(SchemaError::EntityTypeParseError)?
+            Self::parse_possibly_qualified_name_with_default_namespace(action_ty, namespace)
+                .map_err(SchemaError::NamespaceParseError)?
```
cdisselkoen, line 625: Error type changed from `EntityTypeParseError` to `NamespaceParseError`, was that intentional?

cdisselkoen (approved):

cedar-policy-validator/src/schema.rs
john-h-kastner-aws: Good catch. That does not look right.

aaronjeline: Sure, I wasn't sure which one we wanted, since we are parsing a namespace as well?

john-h-kastner-aws (approved):

PR #152, opened 2023-07-05
Error on schema with attribute names duplicated in types

john-h-kastner-aws:
This was the original behavior, but it regressed when schema type parsing was rewritten in #132. The regression was not noticed due to a faulty test case. This commit also fixes the test case, and adds new test cases for other places where a duplicate key might go unnoticed.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/cedar-policy-validator/src/schema_file_format.rs
+++ b/cedar-policy-validator/src/schema_file_format.rs
@@ -1016,5 +1016,6 @@ mod test_duplicates_error {
               "actions": { }
             }
         }"#;
+        serde_json::from_str::<SchemaFragment>(src).unwrap();
     }
 }
--- a/cedar-policy-validator/src/schema_file_format.rs
+++ b/cedar-policy-validator/src/schema_file_format.rs
@@ -259,6 +259,16 @@ impl TypeFields {
     }
 }
 
+/// Used during deserialization to deserialize the attributes type map while
+/// reporting an error if there are any duplicate keys in the map. I could not
+/// find a way to do the `serde_with` conversion inline without introducing this
+/// struct.
+#[derive(Deserialize)]
+struct AttributesTypeMap(
+    #[serde(with = "serde_with::rust::maps_duplicate_key_is_error")]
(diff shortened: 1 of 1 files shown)


aaronjeline (changes requested):
Could you put a note in the CHANGELOG?

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -474,10 +482,8 @@ pub enum SchemaTypeVariant {
         element: Box<SchemaType>,
     },
     Record {
-        #[serde(with = "serde_with::rust::maps_duplicate_key_is_error")]
         attributes: BTreeMap<SmolStr, TypeOfAttribute>,
```
cdisselkoen, line 485: does this need to be `AttributesTypeMap` instead of the `BTreeMap`?

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -474,10 +482,8 @@ pub enum SchemaTypeVariant {
...
         #[serde(rename = "additionalAttributes")]
-        #[serde(default = "additional_attributes_default")]
```
cdisselkoen, line 480: why remove this?  and if we're removing this, should we also remove the function it refers to?

cdisselkoen (approved):
Tests look good

cedar-policy-validator/src/schema_file_format.rs
john-h-kastner-aws: It's still called by the manual deserialization implementation. It's removed because the annotation is dead code since were not deriving deserialize.

cedar-policy-validator/src/schema_file_format.rs
john-h-kastner-aws: It could, but I think I prefer leaving that struct as just an implementation detail for the custom deserializer. I'd rather not have the struct at all, so this makes it possible to delete it later if I figure out has serde_with works without messing with any other code.

PR #157, opened 2023-07-05
Fixes #153 (Stricter versioning)

### LINKED ISSUE #153: `cedar-policy` crate should fix the versions of `cedar-policy-core` and `cedar-policy-validator` it depends on

### Bug Category

Other

### Describe the bug

Currently, `cedar-policy` uses the implicit `^` dependency for Core and Validator.  This means that, e.g., `cedar-policy` 2.2 will accept Core and Validator 2.3 as compatible.  Unfortunately, since we don't currently follow semver for Core and Validator (only for `cedar-policy`), this could result in build problems or other breakage if Cargo actually tries to give Core/Validator 2.3 to `cedar-policy` 2.2 -- e.g., because someone tries to depend on `cedar-policy` `"=2.2"` in their `Cargo.toml`.  To prevent this, we should change the `Cargo.toml` of `cedar-policy` to use the `=` constraint in its dependencies on Core and Validator.  Probably also the same change in Validator's `Cargo.toml` for its Core dependency.

aaronjeline:
Partial fix to #153

If this change is approved I will backport to other 2.x versions

### DIFF
--- a/cedar-policy-formatter/Cargo.toml
+++ b/cedar-policy-formatter/Cargo.toml
@@ -10,7 +10,7 @@ homepage = "https://cedarpolicy.com"
 repository = "https://github.com/cedar-policy/cedar"
 
 [dependencies]
-cedar-policy-core = { version = "=2.3", path = "../cedar-policy-core" }
+cedar-policy-core = { version = "=2.3.0", path = "../cedar-policy-core" }
 pretty = "0.11"
 anyhow = "1.0"
 logos = "0.12.1"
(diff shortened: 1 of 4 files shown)


cedar-policy-formatter/Cargo.toml
```diff
@@ -10,7 +10,7 @@ homepage = "https://cedarpolicy.com"
 repository = "https://github.com/cedar-policy/cedar"
 
 [dependencies]
-cedar-policy-core = { version = "2.3", path = "../cedar-policy-core" }
+cedar-policy-core = { version = "=2.3", path = "../cedar-policy-core" }
```
cdisselkoen, line 13: probably `=2.3.0` is better to be even safer

cdisselkoen (approved):

anwarmamat (approved):

PR #167, opened 2023-07-06
Lossless display for policies

### LINKED ISSUE #125: Make the from_json displayed policy statement drop the "when {true}" clause if the when clause is empty
### Category

User level API changes

Problem: When providing an EST to the `Policy::from_json` method, then displaying that policy, the displayed policy contains `when {true}`.

Not sure about best implementation, but a blanket update to `Policy`'s `Display` trait might not be what we want. This would swallow `when {true}` clauses even if the customer wants it there (because they're using a different flow).

Expectation: if an EST has no conditions, we should not render a when clause. 

string parsing - not ideal

### Additional context

This is used in building gui tools for policy authoring

cdisselkoen:
## Description of changes

Adjusts the `Display` trait on toplevel `Policy` and `PolicySet` objects to display the lossless form of the policy.  (It turns out `Template` didn't actually have a `Display` trait implementation, so I added one for `Template` as well.)

This involves creating new functionality to directly pretty-print the EST (ie EST -> text).  To avoid too much code duplication, I refactored the AST `Display` trait to use this code as well, so it actually does AST -> EST -> text.  (AST -> EST is lossless and infallible, so the output should be identical.)

It's very possible there are bugs in v1 of this PR, I am posting it now but intend to run some DRT on it to make sure the pretty-print roundtrip property for AST still holds (despite going through my new EST-printing code now).  Do folks agree that the AST pretty-print property is sufficient and we don't need a separate EST pretty-print property in DRT?

## Issue #, if available

Fixes #125

(do folks feel that this change warrants a changelog entry?)

(except as discussed above)

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/est/head_constraints.rs
+++ b/cedar-policy-core/src/est/head_constraints.rs
@@ -302,7 +302,7 @@ impl std::fmt::Display for ActionInConstraint {
                     .into_euid(|| JsonDeserializationErrorContext::EntityUid)
                 {
                     Ok(euid) => write!(f, "in {euid}"),
-                    Err(e) => write!(f, "in (invalid entity uid: {e}"),
+                    Err(e) => write!(f, "in (invalid entity uid: {e})"),
                 }
             }
             Self::Set { entities } => {
--- a/cedar-policy-core/src/est/head_constraints.rs
+++ b/cedar-policy-core/src/est/head_constraints.rs
@@ -204,17 +204,15 @@ impl ActionConstraint {
 impl std::fmt::Display for PrincipalConstraint {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            Self::All => write!(f, "principal, "),
+            Self::All => write!(f, "principal"),
             Self::Eq(ec) => {
                 write!(f, "principal ")?;
                 std::fmt::Display::fmt(ec, f)?;
-                write!(f, ", ")?;
                 Ok(())
             }
             Self::In(ic) => {
                 write!(f, "principal ")?;
                 std::fmt::Display::fmt(ic, f)?;
-                write!(f, ", ")?;
                 Ok(())
             }
         }
@@ -224,17 +222,15 @@ impl std::fmt::Display for PrincipalConstraint {
 impl std::fmt::Display for ActionConstraint {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            Self::All => write!(f, "action, "),
+            Self::All => write!(f, "action"),
             Self::Eq(ec) => {
                 write!(f, "action ")?;
                 std::fmt::Display::fmt(ec, f)?;
-                write!(f, ", ")?;
                 Ok(())
             }
             Self::In(aic) => {
                 write!(f, "action ")?;
                 std::fmt::Display::fmt(aic, f)?;
-                write!(f, ", ")?;
                 Ok(())
             }
         }
--- a/cedar-policy-core/src/est/head_constraints.rs
+++ b/cedar-policy-core/src/est/head_constraints.rs
@@ -201,6 +201,131 @@ impl ActionConstraint {
     }
 }
 
+impl std::fmt::Display for PrincipalConstraint {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::All => write!(f, "principal, "),
+            Self::Eq(ec) => {
+                write!(f, "principal ")?;
+                std::fmt::Display::fmt(ec, f)?;
+                write!(f, ", ")?;
+                Ok(())
+            }
+            Self::In(ic) => {
+                write!(f, "principal ")?;
+                std::fmt::Display::fmt(ic, f)?;
+                write!(f, ", ")?;
+                Ok(())
+            }
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


aaronjeline (approved):
Approval, but agree on need for DRT.

shaobo-he-aws (approved):
LGTM.

shaobo-he-aws:
I was wondering if we should also test AST to EST is lossless and infallible. Namely, a round-tripping of AST/EST conversions.

cdisselkoen:
Although AST->EST is lossless and infallible, EST->AST is not, so the roundtrip doesn't hold in general.  There are a bunch of unit tests in the EST module that test roundtripping or pseudo-roundtripping.

shaobo-he-aws:
> Although AST->EST is lossless and infallible, EST->AST is not, so the roundtrip doesn't hold in general. There are a bunch of unit tests in the EST module that test roundtripping or pseudo-roundtripping.

Would AST->EST->AST always succeed and satisfy round-tripping property?

cdisselkoen:
Yes; good point

cedar-policy/src/api.rs
```diff
@@ -1894,6 +1903,29 @@ impl LosslessPolicy {
...
+        match self {
+            Self::Est(est) => write!(f, "{est}"),
+            Self::Text { text, slots } => {
+                if slots.is_empty() {
+                    write!(f, "{text}")
+                } else {
+                    // need to replace placeholders according to `slots`.
+                    // just find-and-replace wouldn't be safe/perfect, we
+                    // want to use the actual parser; right now we reuse
+                    // other implementation by just converting to EST and
```
andrewmwells-amazon, line 1917: [nit] other implementation

cedar-policy-core/src/est/head_constraints.rs
```diff
@@ -201,6 +201,131 @@ impl ActionConstraint {
...
+impl std::fmt::Display for ActionInConstraint {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Single { entity } => {
+                match entity
+                    .clone()
+                    .into_euid(|| JsonDeserializationErrorContext::EntityUid)
+                {
+                    Ok(euid) => write!(f, "in {euid}"),
+                    Err(e) => write!(f, "in (invalid entity uid: {e}"),
```
andrewmwells-amazon, line 305: Intentionally printed with just opening paren?

cedar-policy-core/src/est/head_constraints.rs
```diff
@@ -201,6 +201,131 @@ impl ActionConstraint {
     }
 }
 
+impl std::fmt::Display for PrincipalConstraint {
```
andrewmwells-amazon, line 204: To reduce code duplication, maybe these could share a function that takes `all_str`, `eq_str` and `in_str` as args?

cedar-policy-core/src/est/head_constraints.rs
```diff
@@ -201,6 +201,131 @@ impl ActionConstraint {
...
+                write!(f, "resource ")?;
+                std::fmt::Display::fmt(ic, f)?;
+                Ok(())
+            }
+        }
+    }
+}
+
+impl std::fmt::Display for EqConstraint {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
andrewmwells-amazon, line 263: Maybe these could share a fn that takes `ok_str` and `err_str` as args?

cedar-policy-core/src/est/head_constraints.rs
cdisselkoen: Good catch

shaobo-he-aws (approved):

cedar-policy-core/src/est/head_constraints.rs
cdisselkoen: Not quite sure what you mean.  There are more different types here than you might think at first glance (for instance, for `PrincipalConstraint`, the `In` variant contains `PrincipalOrResourceInConstraint`, while for `ActionConstraint`, the `In` variant contains `ActionInConstraint`)

cdisselkoen:
Running DRT `pp` target on my cloud desktop ran into a failure.  What it comes down to is that the AST involving `Expr::Record` where pairs have duplicate keys, does not roundtrip because in the process of roundtripping we end up keeping only the last key.  This is arguably the intended semantics IMO.  Inside a Cedar policy, the expression `{ foo: 2, foo: 3 } == { foo: 3 }` is `true` and not an error; but the EST for that expression will look like `{ foo: 3 } == { foo: 3 }`.  This is certainly a behavior change vs current `main`, in that now printing the AST for the former drops the duplicate keys and prints out the latter.  Not sure whether to call this a "regression" and/or "breaking change", or whether the new behavior is acceptable.

cedar-policy-core/src/est/head_constraints.rs
andrewmwells-amazon: Makes sense.

cdisselkoen:
Now that [RFC 20](cedar-policy/rfcs#20) has landed (#375), we can actually make the change in this PR without breaking things.  Will update shortly

cdisselkoen:
Alright, we are now passing tests, up-to-date with `main`, and I ran 500K tests locally for DRT `pp` without finding an error.  Can I get a re-review?

shaobo-he-aws (approved):

PR #169, opened 2023-07-10
Refactor AST Record case to use `BTreeMap` instead of vec-of-pairs

cdisselkoen:
## Description of changes

Having a `BTreeMap` here is both more ergonomic, and makes the representation simpler (because it disallows duplicate keys at the AST level).  I suspect the `Vec` is a holdover from a long-ago version that allowed computed keys.

One possible objection is that the vec-of-pairs representation would be necessary if we supported computed keys.  I counter that supporting computed keys would require a change to this API anyway; and also a change to the EST format.  Computed keys doesn't seem to be coming anytime soon and I propose that we cross that bridge when we get there.  Until then, this new representation is simpler and has nicer properties.

## Issue #, if available

This will fix AST->EST not being lossless; see comments on #167.

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -23,8 +23,7 @@ use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 use smol_str::SmolStr;
 use std::{
-    collections::HashMap,
-    collections::HashSet,
+    collections::{BTreeMap, HashMap},
     hash::{Hash, Hasher},
     mem,
     sync::Arc,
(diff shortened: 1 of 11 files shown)


aaronjeline:
Hmm, what if you have a runtime error in the first definition of key
```
ie: 
  { " foo" : 1 + "hello", 
    "foo" : 3 }
    ```

**Comment by cdisselkoen** [comment]
Is the Cedar semantics clear on this?  Or are we basically deciding here what the semantics of that should be?

What is the current behavior for that?

**Comment by aaronjeline** [comment]
> Is the Cedar semantics clear on this?
Yes. Each repeated key is evaluated and may error. Only the last definition of the key makes it into the record. 

```        
    // We allow repeated keys in Record literal definitions, and take the last
    // value given for a field. For example, { a: 1, b: 2, a: 3 } evaluates to
    // the record {b: 2, a: 3 }. We do not throw an error in the case of
    // repeated field IDs, this allows for duplicate keys in map literals
    function interpretRecord(es: seq<(Attr,Expr)>): Result<map<Attr, Value>> {
      if es == [] then
        Ok(map[])
      else
        var k := es[0].0;
        var v :- interpret(es[0].1);
        var m :- interpretRecord(es[1..]);
        if k in m.Keys then // If the same field is repeated later in the record,
          Ok(m)             // we give that occurrence priority and ignore this one.
        else
          Ok(m[k := v])
    }
   ```

cdisselkoen:
Welp, that's unfortunate for this proposed change and for #167

shaobo-he-aws (approved):
LGTM.

cdisselkoen:
Subsumed by #375.

PR #171, opened 2023-07-10
Entity storage from database or disk

prakol16:
## Description of changes

Currently, we support loading entities from an `entities.json` file. However, this is not scalable as the number of entities grows; it becomes wasteful to load all entities from disk when most requests only 'touch' a handful of entities.

In this PR, we will provide API to load entities from a database instead of a JSON file. Note that this is currently very much a rough draft.

In a future PR, we will introduce the ability to turn a residual expression that has type boolean into a corresponding SQL query that fetches the entities that make the expression true.

- A backwards-compatible change requiring a minor version bump to `cedar-policy` (e.g., addition of a new API).

- [ ] Does not require updates because my change does not impact the Cedar Dafny model or DRT infrastructure.

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/authorizer.rs
+++ b/cedar-policy-core/src/authorizer.rs
@@ -21,7 +21,7 @@
 //! the "authorization engine".
 
 use crate::ast::*;
-use crate::entities::Entities;
+use crate::entities::{Entities, EntityDatabase};
 use crate::evaluator::{EvaluationError, Evaluator};
 use crate::extensions::Extensions;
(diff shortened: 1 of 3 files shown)


andrewmwells-amazon:
Thanks for the PR. I suspect this change should go through our RFC process (https://github.com/cedar-policy/rfcs). For now I'll close the PR and you can open an RFC and link to it there. If you disagree that this merits an RFC, we can reopen this PR and discuss here.

PR #178, opened 2023-07-13
make public partial evaluation API experimental

andrewmwells-amazon:
make public partial evaluation API experimental

## Disclaimer


shaobo-he-aws (approved):
LGTM.

cedar-policy/src/api.rs
```diff
@@ -301,6 +301,7 @@ impl Authorizer {
         self.0.is_authorized(&r.0, &p.ast, &e.0).into()
     }
 
+    #[cfg(feature = "partial_eval")]
```
cdisselkoen, line 304: Typically this annotation would go below the `///` docs, not above them

cdisselkoen (changes requested):
I feel like we should do slightly more in this PR. 
- move `Entities::partial()` behind this flag
- move `PartialResponse` behind this flag
- move `ResidualResponse` behind this flag

andrewmwells-amazon:
I think we want to make sure the build with / without `--features "experimental"` has no warnings? That's what I've done in the latest commit. The downside is it requires `#[allow(unused)]` annotations we may then forget to remove if/when the feature is no longer experimental. I added comments naming the feature to each usage; maybe that's good enough.

cedar-policy/Cargo.toml
```diff
@@ -37,6 +37,10 @@ corpus-timing = []
 
 integration_testing = []
 
+# Experimental features
+experimental = ["partial-eval"]
```
cdisselkoen, line 41: Is the intent that `experimental` would turn on all experimental features?  If so can we have a comment documenting that?

cedar-policy/src/api.rs
```diff
@@ -1507,6 +1509,7 @@ impl Template {
     /// create the EST from the policy text or CST instead, as the conversion
     /// to AST is lossy. ESTs generated by this method will reflect the AST and
     /// not the original policy syntax.
+    #[allow(unused)] //Feature: partial-eval
```
cdisselkoen, line 1512: This could be `#[cfg_attr(feature = "partial-eval", allow(unused))]`

Likewise elsewhere

andrewmwells-amazon:
> * move `PartialResponse` behind this flag

Hmm, this change will be more invasive. Specifically it would require rewriting `is_authorized_core(`.

andrewmwells-amazon:
> I feel like we should do slightly more in this PR.
> 
> * move `Entities::partial()` behind this flag
> * move `ResidualResponse` behind this flag

Did these two.

andrewmwells-amazon:
> I think we want to make sure the build with / without `--features "experimental"` has no warnings? That's what I've done in the latest commit. The downside is it requires `#[allow(unused)]` annotations we may then forget to remove if/when the feature is no longer experimental. I added comments naming the feature to each usage; maybe that's good enough.

`#[cfg_attr(not(feature = "partial-eval"), allow(unused))]` works instead

shaobo-he-aws (approved):

cedar-policy/src/api.rs
```diff
@@ -2338,6 +2346,7 @@ pub fn eval_expression(
 }
 
 #[cfg(test)]
+#[cfg(feature = "partial-eval")]
 mod test {
```
cdisselkoen, line 2350: could we maybe rename the `test` module to be clear it's only tests related to partial evaluation?  Doesn't have to block this PR though

cdisselkoen (approved):

andrewmwells-amazon:
>Hmm, this change will be more invasive. Specifically it would require rewriting `is_authorized_core(`

Did the more granular change I think Craig intended instead of the more invasive change of hiding just the `Residual(ResidualResponse)` variant.

aaronjeline (approved):

PR #186, opened 2023-07-17
Add `AuthorizationError`

khieta:
## Description of changes

This PR adds the `AuthorizationError` type discussed in PR #123. I argue that this is a non-breaking change because errors returned by authorization were never truly exposed to users of `cedar-policy` -- they were (and still are) exposed only through a `String`. This PR slightly modifies the `String` that users will see for certain types of errors.

In particular, where before they may have seen a message like
```
while evaluating policy policy0, encountered the following error: type error: expected bool, got long
```
They will now see
```
error occurred while evaluating policy `policy0`: type error: expected bool, got long
```

The reason for the large diff is that we rely heavily on this type of error message in our integration tests.

I recommend that this PR be included in our next patch release.

## Issue #, if available

Related to #114

- [*] A change "invisible" to users (e.g., documentation, changes to "internal" crates like `cedar-policy-core`, `cedar-validator`, etc.)

- [*] Does not update the CHANGELOG because my change does not significantly impact released code.

- [*] Does not require updates because my change does not impact the Cedar Dafny model or DRT infrastructure.

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/ast/restricted_expr.rs
+++ b/cedar-policy-core/src/ast/restricted_expr.rs
@@ -315,7 +315,7 @@ impl<'a> Hash for RestrictedExprShapeOnly<'a> {
 }
 
 /// Errors generated in the restricted_expr module
-#[derive(Debug, Clone, PartialEq, Hash, Error)]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Hash, Error)]
 pub enum RestrictedExpressionError {
     /// A "restricted" expression contained a feature that is not allowed
     /// in "restricted" expressions. The `feature` is just a string description
--- a/cedar-policy-core/src/authorizer.rs
+++ b/cedar-policy-core/src/authorizer.rs
@@ -29,6 +29,9 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
 use std::iter::once;
 
+mod err;
+pub use err::AuthorizationError;
+
 /// Authorizer
 pub struct Authorizer {
     /// Cedar `Extension`s which will be used during requests to this `Authorizer`
@@ -99,11 +102,10 @@ impl Authorizer {
                 // `error_handling` is `SkipOnerror`, no forbids evaluated to a concrete response, and some permits evaluated to `true`
                 let mut errors = partial.diagnostics.errors;
                 errors.extend(partial.residuals.policies().map(|p| {
-                    format!(
-                        "while evaluating policy {}, encountered the following error: {}",
-                        p.id(),
-                        EvaluationError::NonValue(p.condition())
-                    )
+                    AuthorizationError::PolicyEvaluationError {
+                        id: p.id().clone(),
+                        error: EvaluationError::NonValue(p.condition()),
+                    }
                 }));
 
                 let idset = partial.residuals.policies().map(|p| p.id().clone());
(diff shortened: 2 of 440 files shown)


khieta:
Looks like I was wrong in my assertion that `cedar-spec` wouldn't need to be modified -- I'll look into the error tomorrow. Feel free to add comments to this PR in the meantime. I assume that I'll be making a separate PR to `cedar-spec` rather than editing the current PR to accommodate.

cedar-policy-core/src/ast/restricted_expr.rs
```diff
@@ -315,7 +315,7 @@ impl<'a> Hash for RestrictedExprShapeOnly<'a> {
 }
 
 /// Errors generated in the restricted_expr module
-#[derive(Debug, Clone, PartialEq, Hash, Error)]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Hash, Error)]
```
cdisselkoen, line 318: Why this change? Are other errors Ser/De?

cedar-policy-core/src/authorizer.rs
```diff
@@ -386,16 +375,29 @@ impl Default for Authorizer {
...
+        Self {
+            satisfied_permits: vec![],
+            satisfied_forbids: vec![],
+            global_deny_policies: HashSet::new(),
+            errors: vec![],
+            permit_residuals: vec![],
+            forbid_residuals: vec![],
+        }
+    }
+}
```
cdisselkoen, line 399: This is the same as what we would previously have gotten with the auto-derived `Default`; can we keep using that?

cedar-policy/src/api.rs
```diff
@@ -371,6 +359,19 @@ pub struct Diagnostics {
...
 }
 
+impl From<authorizer::Diagnostics> for Diagnostics {
+    fn from(diagnostics: authorizer::Diagnostics) -> Self {
+        Self {
+            reason: diagnostics.reason.into_iter().map(PolicyId).collect(),
+            errors: diagnostics
+                .errors
+                .into_iter()
+                .map(|e| e.to_string())
```
cdisselkoen, line 369: could be written `.map(ToString::to_string)`; not sure if that's clearer

cdisselkoen (approved):
Looks good

cedar-policy/src/api.rs
khieta: Fixed

cedar-policy-core/src/authorizer.rs
khieta: In some version of my edits, the automatic derivation of `Default` failed, which is why I implemented it manually. Forgot to go back and check whether it was still required (it's not).

cedar-policy-core/src/ast/restricted_expr.rs
khieta: `cedar-policy-core`'s `Response` type was originally Ser/De, so when I nested the error type inside it I needed to make that type Ser/De too (which involved adding `Serialize` and `Deserialize` in a bunch of places). Looking at the code again, I'm not sure that `Response` actually needs to be Ser/De. Possibly for `cedar-spec`? I'll play around with it and get back to you.

john-h-kastner-aws (approved):

cedar-policy-core/src/ast/restricted_expr.rs
khieta: Getting rid of Ser/De does indeed break DRT, but it looks pretty simple to fix. I haven't looked at FFI yet -- but it's probably a bad to break our serialized representations since they may have wider impact (e.g., on external contributors writing language bindings). Thanks for pointing that out.

It looks like I can use `#[serde_as(as = "DisplayFromStr")]` [[link](https://docs.rs/serde_with/latest/serde_with/struct.DisplayFromStr.html)] to use `Display` for serialization. It expects a `FromStr` for deserialization though. I don't think we want to write that for error messages. Thoughts?

cdisselkoen: Can we get away with having serialization but not deserialization?

Alternatively can we have deserialization use the structure but serialization do strings?  it might be weird that they don't roundtrip, but that might be an arrangement that makes sense in this case for backwards compatibility?  or maybe we don't want to go there

khieta: Actually, nvm. It looks like our json interface isn't trying to serialize `cedar-policy-core`'s `Response` type. It just exposes the `String` error messages exposed through `cedar-policy`. I think the best approach is to remove Ser/De for `Response`s and make the necessary fixes in DRT.

khieta:
Draft of a fix for DRT in `cedar-spec` [PR#77](https://github.com/cedar-policy/cedar-spec/pull/77).

cdisselkoen (approved):

PR #203, opened 2023-07-19
Fix panic is cedar_policy::PolicySet::link

john-h-kastner-aws:
Removes two `expects` in this function. One was known to panic when the function was called with a `template_id` corresponding to a static policy. The other could not panic, but small code change can elimiante it.

- [x] Does not require updates because my change does not impact the Cedar Dafny model or DRT
## Disclaimer

### DIFF
--- a/cedar-policy-core/src/ast/policy_set.rs
+++ b/cedar-policy-core/src/ast/policy_set.rs
@@ -187,7 +187,10 @@ impl PolicySet {
         }
     }
 
-    /// Attempt to create a new template linked policy and add it to the policy set
+    /// Attempt to create a new template linked policy and add it to the policy
+    /// set. Returns a references to the new template linked policy if
+    /// successful.
+    ///
     /// Errors for two reasons
     ///   1) The the passed SlotEnv either does not match the slots in the templates
     ///   2) The passed link Id conflicts with an Id already in the set
@@ -196,7 +199,7 @@ impl PolicySet {
         template_id: PolicyID,
         new_id: PolicyID,
         values: HashMap<SlotId, EntityUID>,
-    ) -> Result<(), LinkingError> {
+    ) -> Result<&Policy, LinkingError> {
         let t = self
             .get_template(&template_id)
             .ok_or_else(|| LinkingError::NoSuchTemplate {
@@ -209,10 +212,7 @@ impl PolicySet {
             self.links.entry(new_id.clone()),
             self.templates.entry(new_id),
         ) {
-            (Entry::Vacant(links_entry), Entry::Vacant(_)) => {
-                links_entry.insert(r);
-                Ok(())
-            }
+            (Entry::Vacant(links_entry), Entry::Vacant(_)) => Ok(links_entry.insert(r)),
             (Entry::Occupied(oentry), _) => Err(LinkingError::PolicyIdConflict {
                 id: oentry.key().clone(),
             }),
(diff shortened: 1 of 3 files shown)


cedar-policy-core/src/ast/policy_set.rs
```diff
@@ -182,7 +182,7 @@ impl PolicySet {
         template_id: PolicyID,
         new_id: PolicyID,
         values: HashMap<SlotId, EntityUID>,
-    ) -> Result<(), LinkingError> {
+    ) -> Result<&Policy, LinkingError> {
```
cdisselkoen, line 185: Since we're now returning a value, can we document what we're returning in the docstring?  (It's probably obvious that it returns the new linked policy, but good to say so explicitly anyhow)

cedar-policy/src/api.rs
```diff
@@ -1121,6 +1121,9 @@ pub enum PolicySetError {
     /// Expected an static policy, but a template-linked policy was provided.
     #[error("Expected static policy, but a template-linked policy was provided")]
     ExpectedStatic,
+    /// Expected a template, but a static policy was provided.
+    #[error("expected template, but a static policy was provided")]
```
cdisselkoen, line 1125:
```suggestion
    #[error("expected a template, but a static policy was provided")]
```

cedar-policy/src/api.rs
```diff
@@ -1280,9 +1283,12 @@ impl PolicySet {
 
     /// Attempt to link a template and add the new template-linked policy to the policy set.
     /// If link fails, the `PolicySet` is not modified.
-    /// Failure can happen for two reasons
+    /// Failure can happen for three reasons
     ///   1) The map passed in `vals` may not match the slots in the template
     ///   2) The `new_id` may conflict w/ a policy that already exists in the set
+    ///   3) `template_id` does not correspond to a template. Either the id is
+    ///   not in the policy set, or it is in the policy set but is a static
+    ///   policy rather than a template
```
cdisselkoen, line 1291: what if it's in the policy set but it's a linked policy, rather than a static policy or a template?

cedar-policy/src/api.rs
```diff
@@ -1294,24 +1300,24 @@ impl PolicySet {
...
             .map_err(PolicySetError::LinkingError)?;
-        let linked_ast = self
-            .ast
-            .get(&new_id.0)
-            .expect("ast.link() didn't fail above, so this shouldn't fail");
         let linked_lossless = self
             .templates
             .get(&template_id)
-            .expect("ast.link() didn't fail above, so this shouldn't fail")
+            .ok_or(PolicySetError::ExpectedTemplate)?
```
cdisselkoen, line 1314: can we document why this is the right error?  Looking at these couple lines in isolation, it looks like this could fire when the template simply does not exist, which is a different error.  The code is correct but a comment would be helpful for whoever looks at this next

cedar-policy/src/api.rs
```diff
@@ -1294,24 +1300,24 @@ impl PolicySet {
...
             .lossless
             .clone()
             .link(unwrapped_vals.iter().map(|(k, v)| (*k, v)))
+            // The only error case for `lossless.link` is a template with slots
+            // which are not filled by the provided values. `ast.link` will have
```
cdisselkoen, line 1319:
```suggestion
            // The only error case for `lossless.link()` is a template with slots
            // which are not filled by the provided values. `ast.link()` will have
```

shaobo-he-aws (approved):

cdisselkoen (approved):

shaobo-he-aws (approved):

andrewmwells-amazon (approved):

PR #222, opened 2023-07-21
error bug bash

andrewmwells-amazon:
## Description of changes
Improve error messages for parsing `ToAST`

## Issue #, if available
https://github.com/cedar-policy/cedar/issues/223 (partially completed)

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/est/expr.rs
+++ b/cedar-policy-core/src/est/expr.rs
@@ -18,7 +18,7 @@ use super::utils::unwrap_or_clone;
 use super::FromJsonError;
 use crate::ast;
 use crate::entities::{EscapeKind, JSONValue, JsonDeserializationError, TypeAndId};
-use crate::parser::cst;
+use crate::parser::cst::{self, Ident};
 use crate::parser::err::{ParseError, ParseErrors, ToASTError};
 use crate::parser::unescape;
 use crate::parser::ASTNode;
@@ -1106,7 +1106,10 @@ fn interpret_primary(p: cst::Primary) -> Result<Either<ast::Name, Expr>, ParseEr
                 ))),
                 (path, Some(id)) => {
                     let (l, r) = match (path.first(), path.last()) {
-                        (Some(l), Some(r)) => (l.info.range_start(), r.info.range_end()),
+                        (Some(l), Some(r)) => (
+                            l.info.range_start(),
+                            r.info.range_end() + ident_to_str_len(&id),
+                        ),
                         (_, _) => (0, 0),
                     };
                     Err(ParseError::ToAST(ToASTError::InvalidExpression(cst::Name {
@@ -1318,7 +1321,10 @@ impl TryFrom<cst::Name> for Expr {
             (&[], Some(cst::Ident::Context)) => Ok(Expr::var(ast::Var::Context)),
             (path, Some(id)) => {
                 let (l, r) = match (path.first(), path.last()) {
-                    (Some(l), Some(r)) => (l.info.range_start(), r.info.range_end()),
+                    (Some(l), Some(r)) => (
+                        l.info.range_start(),
+                        r.info.range_end() + ident_to_str_len(&id),
+                    ),
                     (_, _) => (0, 0),
                 };
                 Err(ParseError::ToAST(ToASTError::InvalidExpression(cst::Name {
@@ -1332,6 +1338,30 @@ impl TryFrom<cst::Name> for Expr {
     }
 }
 
+/// Get the string length of an `Ident`. Used to print the source location for error messages
+fn ident_to_str_len(i: &Ident) -> usize {
+    match i {
+        Ident::Principal => 9,
+        Ident::Action => 6,
+        Ident::Resource => 8,
+        Ident::Context => 7,
+        Ident::True => 4,
+        Ident::False => 5,
+        Ident::Permit => 6,
+        Ident::Forbid => 6,
+        Ident::When => 4,
+        Ident::Unless => 6,
+        Ident::In => 2,
+        Ident::Has => 3,
+        Ident::Like => 4,
+        Ident::If => 2,
+        Ident::Then => 4,
+        Ident::Else => 4,
+        Ident::Ident(s) => s.len(),
+        Ident::Invalid(s) => s.len(),
+    }
+}
+
 #[test]
 fn test_invalid_expr_from_cst_name() {
     let path = vec![ASTNode::new(
@@ -1339,7 +1369,7 @@ fn test_invalid_expr_from_cst_name() {
         0,
         12,
     )];
-    let name = ASTNode::new(Some(cst::Ident::Else), 12, 15);
+    let name = ASTNode::new(Some(cst::Ident::Else), 13, 16);
     let cst_name = cst::Name { path, name };
 
     match Expr::try_from(cst_name) {
@@ -1349,7 +1379,7 @@ fn test_invalid_expr_from_cst_name() {
             match &e[0] {
                 ParseError::ToAST(ToASTError::InvalidExpression(e)) => {
                     println!("{:?}", e);
-                    assert_eq!(e.name.info.range_end(), 12);
+                    assert_eq!(e.name.info.range_end(), 16);
                 }
                 _ => panic!("wrong error"),
             }
--- a/cedar-policy-core/src/est/expr.rs
+++ b/cedar-policy-core/src/est/expr.rs
@@ -1105,9 +1105,13 @@ fn interpret_primary(p: cst::Primary) -> Result<Either<ast::Name, Expr>, ParseEr
                         .collect::<Result<Vec<ast::Id>, _>>()?,
                 ))),
                 (path, Some(id)) => {
+                    let (l, r) = match (path.first(), path.last()) {
+                        (Some(l), Some(r)) => (l.info.range_start(), r.info.range_end()),
+                        (_, _) => (0, 0),
+                    };
                     Err(ParseError::ToAST(ToASTError::InvalidExpression(cst::Name {
                         path: path.to_vec(),
-                        name: ASTNode::new(Some(id), 0, 0),
+                        name: ASTNode::new(Some(id), l, r),
                     }))
                     .into())
                 }
@@ -1312,12 +1316,43 @@ impl TryFrom<cst::Name> for Expr {
             (&[], Some(cst::Ident::Action)) => Ok(Expr::var(ast::Var::Action)),
             (&[], Some(cst::Ident::Resource)) => Ok(Expr::var(ast::Var::Resource)),
             (&[], Some(cst::Ident::Context)) => Ok(Expr::var(ast::Var::Context)),
-            (path, Some(id)) => Err(ParseError::ToAST(ToASTError::InvalidExpression(cst::Name {
-                path: path.to_vec(),
-                name: ASTNode::new(Some(id), 0, 0),
-            }))
-            .into()),
+            (path, Some(id)) => {
+                let (l, r) = match (path.first(), path.last()) {
+                    (Some(l), Some(r)) => (l.info.range_start(), r.info.range_end()),
+                    (_, _) => (0, 0),
+                };
+                Err(ParseError::ToAST(ToASTError::InvalidExpression(cst::Name {
+                    path: path.to_vec(),
+                    name: ASTNode::new(Some(id), l, r),
+                }))
+                .into())
+            }
             (_, None) => Err(ParseError::ToAST(ToASTError::MissingNodeData).into()),
+        }
+    }
+}
+
+#[test]
+fn test_invalid_expr_from_cst_name() {
+    let path = vec![ASTNode::new(
+        Some(cst::Ident::Ident("some_long_str".into())),
+        0,
+        12,
+    )];
+    let name = ASTNode::new(Some(cst::Ident::Else), 12, 15);
+    let cst_name = cst::Name { path, name };
+
+    match Expr::try_from(cst_name) {
+        Ok(_) => panic!("wrong error"),
+        Err(e) => {
+            assert!(e.len() == 1);
+            match &e[0] {
+                ParseError::ToAST(ToASTError::InvalidExpression(e)) => {
+                    println!("{:?}", e);
+                    assert_eq!(e.name.info.range_end(), 12);
+                }
+                _ => panic!("wrong error"),
+            }
         }
     }
 }
--- a/cedar-policy-core/src/parser/err.rs
+++ b/cedar-policy-core/src/parser/err.rs
@@ -98,7 +98,7 @@ pub enum ToASTError {
     #[error("this policy is missing the {0} variable in the scope")]
     MissingScopeConstraint(Var),
     /// Returned when a policy has an extra scope clause. This is not valid syntax
-    #[error("this policy an extra head constraint in the scope; a policy must have exactly `principal`, `action`, and `resource` constraints: {0}")]
+    #[error("this policy has an extra head constraint in the scope; a policy must have exactly `principal`, `action`, and `resource` constraints: {0}")]
     ExtraHeadConstraints(cst::VariableDef),
     /// Returned when a policy uses a reserved keyword as an identifier.
     #[error("this identifier is reserved and cannot be used: {0}")]


cedar-policy-core/src/est/expr.rs
```diff
@@ -1076,7 +1076,7 @@ fn interpret_primary(p: cst::Primary) -> Result<Either<ast::Name, Expr>, ParseEr
                 }
             }
             Some(cst::Ref::Ref { .. }) => {
-                Err(ParseError::ToAST("entity-lookup syntax is not supported".to_string()).into())
+                Err(ParseError::ToAST("entity-lookup syntax is not supported: please use `Namespace::EntityType::EntityID` or similar".to_string()).into())
```
cdisselkoen, line 1079: should this be `Namespace::EntityType::"EntityID"` to remind people quotes are required?

cedar-policy-core/src/est/expr.rs
```diff
@@ -1183,8 +1183,8 @@ impl TryFrom<cst::Member> for Expr {
                             Either::Right(expr) => Either::Right(Expr::get_attr(expr, i)),
                         };
                     }
-                    Some(_i) => {
-                        return Err(ParseError::ToAST("Invalid Identifier".to_string()).into())
+                    Some(i) => {
+                        return Err(ParseError::ToAST(format!("invalid Identifier: {:?}", i)).into())
```
cdisselkoen, line 1187: probably lowercase `identifier` in this message?

cedar-policy-core/src/est/expr.rs
```diff
@@ -1302,12 +1302,14 @@ impl TryFrom<cst::Literal> for Expr {
...
                 Ok(Expr::lit(JSONValue::Long(n.try_into().map_err(|_| {
-                    ParseError::ToAST("numeric literal out of range".to_string())
+                    ParseError::ToAST(format!("numeric literal out of range: {n}"))
                 })?)))
             }
             cst::Literal::Str(ASTNode { node, .. }) => match node {
                 Some(cst::Str::String(s)) => Ok(Expr::lit(JSONValue::String(s))),
-                Some(_) => Err(ParseError::ToAST("invalid string".to_string()).into()),
+                Some(cst::Str::Invalid(s)) => {
+                    Err(ParseError::ToAST(format!("invalid string: {s}")).into())
```
cdisselkoen, line 1311: might be a good idea to use `{s:?}` instead of `{s}` in this message, since I'm imagining invalid strings might contain weird characters or whitespace or etc

cdisselkoen (approved):

cedar-policy-core/src/parser/err.rs
```diff
@@ -48,6 +48,76 @@ pub enum ParseError {
     #[error("poorly formed: {0}")]
     #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
     ToAST(String),
+
+    /// Error in the CST -> AST transform, found an illegal, empty node.
+    #[error("poorly formed: {0}")]
+    #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
+    NodeShouldNotBeEmpty(String),
```
khieta, line 55: I think this may be a type of `InternalInvariantViolation` since I imagine that the message "node should not be empty" reflects an error on our end.

khieta (commented):
I love the new errors types, but I wish they didn't all use string arguments. I think it'll be pretty straightforward to get rid of most of the strings. A quick sketch:
```
    #[error("too many -'s")]
    TooManyDashes(String),
    ....
    #[error("literal `{0}` is too large")]
    NumericLiteralOutOfRange(Int),
```

I also think we can drop "poorly formed:" since this is implied by it being a `ParseError`.

andrewmwells-amazon:
> I think it'll be pretty straightforward to get rid of most of the strings.

I'm not convinced this is a good idea. E.g., if we change to `NumericLiteralOutOfRange(i64)` then the error message will print out some `i64` when I think we should print out the string.

> I also think we can drop "poorly formed:" since this is implied by it being a ParseError.

Agreed and pushed.

cedar-policy-core/src/parser/err.rs
```diff
@@ -48,6 +48,71 @@ pub enum ParseError {
     #[error("poorly formed: {0}")]
     #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
     ToAST(String),
+
+    /// Error in the CST -> AST transform, internal invariant violation.
+    #[error("poorly formed: {0}")]
```
khieta, line 53: My preference would be `#[error("internal error: {0}")]`

cedar-policy-core/src/parser/err.rs
```diff
@@ -48,6 +48,71 @@ pub enum ParseError {
...
+
+    /// Error in the CST -> AST transform, too many `-`s.
+    #[error("too many -'s: {0}")]
+    #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
+    TooManyDashes(String),
+
+    /// Error in the CST -> AST transform, too many `!`s.
+    #[error("too many !'s: {0}")]
+    #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
+    TooManyBangs(String),
```
khieta, line 65: Remove the string argument from these. The current error message will be `"too many -'s: too many -'s"`

cedar-policy-core/src/parser/err.rs
```diff
@@ -48,6 +48,71 @@ pub enum ParseError {
...
+
+    /// Error in the CST -> AST transform, illegal entity lookup syntax.
+    #[error("unsupproted entity lookup syntax: {0}")]
```
khieta, line 68: unsupproted -> unsupported

cedar-policy-core/src/est/expr.rs
```diff
@@ -1300,15 +1375,18 @@ impl TryFrom<cst::Literal> for Expr {
         match lit {
             cst::Literal::True => Ok(Expr::lit(JSONValue::Bool(true))),
             cst::Literal::False => Ok(Expr::lit(JSONValue::Bool(false))),
-            cst::Literal::Num(n) => {
-                Ok(Expr::lit(JSONValue::Long(n.try_into().map_err(|_| {
-                    ParseError::ToAST("numeric literal out of range".to_string())
-                })?)))
-            }
+            cst::Literal::Num(n) => Ok(Expr::lit(JSONValue::Long(n.try_into().map_err(|_| {
+                ParseError::NumericLiteralOutOfRange(format!("numeric literal out of range: {n}"))
```
khieta, line 1379: Should just be `n.to_string()`

cedar-policy-core/src/est/expr.rs
```diff
@@ -1300,15 +1375,18 @@ impl TryFrom<cst::Literal> for Expr {
...
+            })?))),
             cst::Literal::Str(ASTNode { node, .. }) => match node {
                 Some(cst::Str::String(s)) => Ok(Expr::lit(JSONValue::String(s))),
-                Some(_) => Err(ParseError::ToAST("invalid string".to_string()).into()),
-                None => Err(ParseError::ToAST("node should not be empty".to_string()).into()),
+                Some(cst::Str::Invalid(s)) => {
+                    Err(ParseError::InvalidString(format!("invalid string: {s:?}")).into())
```
khieta, line 1384: Should just be `s`

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -668,7 +670,7 @@ impl ASTNode<Option<cst::Str>> {
             cst::Str::String(s) => Some(s),
             // at time of comment, all strings are valid
             cst::Str::Invalid(s) => {
-                errs.push(err::ParseError::ToAST(format!(
+                errs.push(err::ParseError::InvalidString(format!(
```
khieta, line 673: Should just be `s` + same comment for all other instances in this file

cedar-policy-core/src/parser/err.rs
```diff
@@ -48,6 +48,71 @@ pub enum ParseError {
...
+    #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
+    NumericLiteralOutOfRange(String),
+
+    /// Error in the CST -> AST transform, invalid string.
+    #[error("invalid string: {0}")]
+    #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
+    InvalidString(String),
+
+    /// Error in the CST -> AST transform, invalid expression.
+    #[error("invalid expr: {0}")]
```
khieta, line 83: expr -> expression, just because I'm being picky 😄

cedar-policy-core/src/est/expr.rs
```diff
@@ -1096,40 +1150,40 @@ fn interpret_primary(p: cst::Primary) -> Result<Either<ast::Name, Expr>, ParseEr
...
                                         "node should not be empty".to_string(),
                                     )])
                                 })
                                 .and_then(|id| id.to_string().parse().map_err(Into::into))
                         })
                         .collect::<Result<Vec<ast::Id>, _>>()?,
                 ))),
-                (path, Some(id)) => Err(ParseError::ToAST(format!(
+                (path, Some(id)) => Err(ParseError::InvalidExpr(format!(
                     "{} is not a valid expr",
```
khieta, line 1162: Drop the "is not a valid expr" + same comment for the other instances in this file

cedar-policy-core/src/est/expr.rs
```diff
@@ -1167,27 +1221,37 @@ impl TryFrom<cst::Member> for Expr {
...
                                 .into())
                             }
                             Either::Right(expr) => Either::Right(Expr::get_attr(expr, i)),
                         };
                     }
-                    Some(_i) => {
-                        return Err(ParseError::ToAST("Invalid Identifier".to_string()).into())
+                    Some(i) => {
+                        return Err(ParseError::InvalidIdentifier(format!(
+                            "invalid identifier: {:?}",
```
khieta, line 1245: Drop the "invalid identifier"

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -330,13 +330,13 @@ impl ASTNode<Option<cst::Ident>> {
             | cst::Ident::In
             | cst::Ident::Has
             | cst::Ident::Like => {
-                errs.push(err::ParseError::ToAST(format!(
+                errs.push(err::ParseError::InvalidIdentifier(format!(
                     "This identifier is reserved and cannot be used: {ident}"
```
khieta, line 334: Maybe reorder this one to something like "{ident}. This identifier is reserved and cannot be used" so that it concatenates nicely with "invalid identifier:"

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -756,19 +760,21 @@ impl ExprOrSpecial<'_> {
...
+                errs.push(err::ParseError::InvalidString(format!(
                     "not a string literal: {var}"
                 )));
                 None
             }
             Self::Name(name) => {
-                errs.push(err::ParseError::ToAST(format!(
+                errs.push(err::ParseError::InvalidString(format!(
                     "not a string literal: {name}"
                 )));
```
khieta, line 771: I wonder if the errors in this function are more `IllegalNonLiteral` errors that `InvalidString` errors? Also, looking at this code again, I think the `Var`, `Name`, `Expr` cases can be smooshed into one case. Same comments about the function below.

cedar-policy-core/src/est/expr.rs
```diff
@@ -1244,33 +1308,41 @@ impl TryFrom<cst::Member> for Expr {
                             }
                         }
                         _ => {
-                            return Err(
-                                ParseError::ToAST("malformed method call".to_string()).into()
+                            return Err(ParseError::MalformedMethodCall(
+                                "malformed method call".to_string(),
                             )
+                            .into())
```
khieta, line 1314: This error will print as "malformed method call: malformed method call"

khieta (commented):
I ended up leaving a bunch of comments 😁 Just address whichever ones seem reasonable.

cedar-policy-core/src/parser/err.rs
shaobo-he-aws: I found the naming kind of confusing to me. A user may wonder what invariants are for a parser.

cedar-policy-core/src/parser/err.rs
```diff
@@ -48,6 +48,71 @@ pub enum ParseError {
...
     #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
     ToAST(String),
+
+    /// Error in the CST -> AST transform, internal invariant violation.
+    #[error("poorly formed: {0}")]
+    #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
+    InternalInvariantViolation(String),
+
+    /// Error in the CST -> AST transform, too many `-`s.
+    #[error("too many -'s: {0}")]
```
shaobo-he-aws, line 58: I would prefer `integer negation operators` to simply `-`.

cedar-policy-core/src/parser/err.rs
shaobo-he-aws: Apart from Kesha's comment, please see my comment above.

cedar-policy-core/src/parser/err.rs
```diff
@@ -48,6 +48,71 @@ pub enum ParseError {
...
+
+    /// Error in the CST -> AST transform, too many `!`s.
+    #[error("too many !'s: {0}")]
+    #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
+    TooManyBangs(String),
+
+    /// Error in the CST -> AST transform, illegal entity lookup syntax.
+    #[error("unsupproted entity lookup syntax: {0}")]
+    #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
+    UnsupportedEntityLookupSyntax(String),
```
shaobo-he-aws, line 70: Maybe remove `Syntax`?

cedar-policy-core/src/parser/err.rs
```diff
@@ -48,6 +48,71 @@ pub enum ParseError {
...
+
+    /// Error in the CST -> AST transform, invalid identifier.
+    #[error("invalid identifier: {0}")]
+    #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
+    InvalidIdentifier(String),
+
+    /// Error in the CST -> AST transform, non literal where literal was expected.
+    #[error("expected a literal. Got: {0}")]
+    #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
+    IllegalNonLiteral(String),
```
shaobo-he-aws, line 95: Change `Illegal` to `Unexpected` for consistency?

cedar-policy-core/src/parser/err.rs
```diff
@@ -48,6 +48,71 @@ pub enum ParseError {
...
+
+    /// Error in the CST -> AST transform, illegal slot.
+    #[error("illegal slot: {0}")]
+    #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
+    IllegalSlot(String),
```
shaobo-he-aws, line 100: See above.

cedar-policy-core/src/parser/err.rs
```diff
@@ -48,6 +48,71 @@ pub enum ParseError {
...
+
+    /// Error in the CST -> AST transform, illegal operation (e.g., division or modulo).
+    #[error("illegal operation: {0}")]
+    #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
+    IllegalOperation(String),
```
shaobo-he-aws, line 105: See above.

cedar-policy-core/src/parser/err.rs
```diff
@@ -48,6 +48,71 @@ pub enum ParseError {
...
+
+    /// Error in the CST -> AST transform, malformed method call.
+    #[error("malformed method call: {0}")]
+    #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
+    MalformedMethodCall(String),
+
+    /// Error in the CST -> AST transform, illegal function args.
+    #[error("illegal function args: {0}")]
+    #[diagnostic(code(cedar_policy_core::parser::to_ast_err))]
+    IllegalFunctionArgs(String),
```
shaobo-he-aws, line 115: `Illegal` to `Invalid`?

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -1926,7 +1934,7 @@ impl ASTNode<Option<cst::Literal>> {
             cst::Literal::Num(n) => match i64::try_from(*n) {
                 Ok(i) => Some(ExprOrSpecial::Expr(construct_expr_num(i, src.clone()))),
                 Err(_) => {
-                    errs.push(ParseError::ToAST(format!("Literal {n} is too large")));
+                    errs.push(ParseError::NumericLiteralOutOfRange(n.to_string()));
```
shaobo-he-aws, line 1937: I wonder if we can let the variant `NumericLiteralOutOfRange` to carry the number instead of the error message? And use `#[error(...)]` to specify the message.

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -1728,7 +1736,7 @@ impl ASTNode<Option<cst::Primary>> {
         match prim {
             cst::Primary::Literal(l) => l.to_expr_or_special(errs)?.into_string_literal(errs),
             _ => {
-                errs.push(err::ParseError::ToAST(format!(
+                errs.push(err::ParseError::InvalidString(format!(
```
shaobo-he-aws, line 1739: See below.

cedar-policy-core/src/parser/err.rs
andrewmwells-amazon: I think it helps clarify that the syntax isn't supported. Otherwise it might sound like it's unsupported in certain circumstances but not others.

cedar-policy-core/src/parser/cst_to_ast.rs
andrewmwells-amazon: Personally I think `Invalid` makes more sense here.

cedar-policy-core/src/parser/cst_to_ast.rs
andrewmwells-amazon: Agree with `IllegalNonLiteral` part. I think it's reasonable to keep the separate errors for `Identifier` `Expr` and `String`. If we had more time it could be useful to try to explain why something isn't a valid `identifier` and if we do that we'd want to handle each case separately.

shaobo-he-aws (approved):

john-h-kastner-aws (approved):

cedar-policy-core/src/est/expr.rs
```diff
@@ -1243,36 +1295,34 @@ impl TryFrom<cst::Member> for Expr {
                                 }
                             }
                         }
-                        _ => {
-                            return Err(
-                                ParseError::ToAST("malformed method call".to_string()).into()
-                            )
-                        }
+                        _ => return Err(ParseError::MalformedMethodCall("".to_string()).into()),
```
cdisselkoen, line 1298: empty-string error message seems not ideal ... should we have a `String` in this error variant at all?

cedar-policy-core/src/est/expr.rs
```diff
@@ -1302,13 +1355,16 @@ impl TryFrom<cst::Literal> for Expr {
             cst::Literal::False => Ok(Expr::lit(JSONValue::Bool(false))),
             cst::Literal::Num(n) => {
                 Ok(Expr::lit(JSONValue::Long(n.try_into().map_err(|_| {
-                    ParseError::ToAST("numeric literal out of range".to_string())
+                    ParseError::NumericLiteralOutOfRange(n.to_string())
```
cdisselkoen, line 1358: instead of a `String`, it would probably be more natural to have the argument be a number directly

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -211,7 +211,7 @@ impl ASTNode<Option<cst::Policy>> {
 
         for e in conds.iter() {
             for slot in e.slots() {
-                errs.push(ParseError::ToAST(format!("Template slots are currently unsupported in policy condition clauses, found slot {slot}")));
+                errs.push(ParseError::UnexpectedSlot(format!("Template slots are currently unsupported in policy condition clauses, found slot {slot}")));
```
cdisselkoen, line 214: Can the explanation be in the `#[error]` annotation for this variant, and the argument just be the `{slot}`?

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -386,13 +384,15 @@ impl ASTNode<Option<cst::Ident>> {
             Some(cst::Ident::Action) => Some(ast::Var::Action),
             Some(cst::Ident::Resource) => Some(ast::Var::Resource),
             Some(ident) => {
-                errs.push(err::ParseError::ToAST(format!(
-                    "expected an identifier, got {ident}"
+                errs.push(err::ParseError::InvalidIdentifier(format!(
+                    "{ident}; expected an identifier"
```
cdisselkoen, line 388: The "expected an identifier" part is probably redundant

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -386,13 +384,15 @@ impl ASTNode<Option<cst::Ident>> {
...
                 )));
                 None
             }
             None => {
-                errs.push(err::ParseError::ToAST("expected an identifier".to_string()));
+                errs.push(err::ParseError::InvalidIdentifier(
+                    "expected an identifier".to_string(),
```
cdisselkoen, line 394: the argument here should be the actual text that failed to parse as an identifier. Or is this another empty-node error and shouldn't be `InvalidIdentifier` at all?

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -756,19 +754,21 @@ impl ExprOrSpecial<'_> {
                 }
             },
             Self::Var(var, _) => {
-                errs.push(err::ParseError::ToAST(format!(
+                errs.push(err::ParseError::UnexpectedNonLiteral(format!(
                     "not a string literal: {var}"
```
cdisselkoen, line 758: this error message also won't work with the `#[error]` template on `UnexpectedNonLiteral`.  And more occurrences below

cdisselkoen: Would love to see a few tests that explicitly compare the error message to some constant string, so that we can look at the constant strings and see they're reasonable, rather than piecing things together ourselves in our heads

cdisselkoen:
Marked request-changes mostly because of the error messages that don't work with the templating -- my other comments are optional, take or leave

cedar-policy-core/src/est/expr.rs
andrewmwells-amazon: What if the number doesn't fit in `n` bits?

cdisselkoen: Right.  String is good.  I was thinking that the `n` here was already a number type, but it's probably a string slice, isn't it

cedar-policy-core/src/est/expr.rs
andrewmwells-amazon: The other uses have messages: `expected method name, found {}` and `this item does not have methods` which I think are useful. I can print out the item here, but I'm not sure how legible that will be. If it's a typo that's probably useful. If it's some complex AST node probably not.

cedar-policy-core/src/parser.rs
```diff
@@ -267,9 +267,14 @@ pub(crate) fn parse_literal(val: &str) -> Result<ast::Literal, err::ParseErrors>
     let cst = text_to_cst::parse_primary(val)?;
     let Some(ast) = cst.to_expr(&mut errs) else { return Err(errs); };
     if errs.is_empty() {
-        match ast.into_expr_kind() {
+        let expr_kind = ast.into_expr_kind();
+        match expr_kind {
             ast::ExprKind::Lit(v) => Ok(v),
-            _ => Err(err::ParseError::ToAST("text is not a literal".to_string()).into()),
+            _ => Err(err::ParseError::UnexpectedNonLiteral(format!(
+                "text is not a literal: {:?}",
```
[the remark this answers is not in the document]
andrewmwells-amazon: Fixed but in the process I introduced clones `node_copy_for_error`, `pat_expr_copy_for_error` and `field_copy_for_error`. These codepaths probably aren't exercised a ton, but I'm not sure we want to clone Exprs just for error messages. Take a look and let me know what you think.

shaobo-he-aws (approved):

andrewmwells-amazon:
Pretty much entirely superseded by https://github.com/cedar-policy/cedar/pull/243

shaobo-he-aws (approved):

aaronjeline (approved):

aaronjeline (approved):

PR #230, opened 2023-07-26
`Entities` structure eager evaluation

aaronjeline:
## Description of changes

Adds a new method to the `Entities` structure to allow eager evaluation of entity attributes and re-use them across calls to `is_authorized`. Currently this takes the form of a new method `Entities::evaluate()`, but in a future major version we could potentially add this to the constructor of `Entities`. 

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/ast/extension.rs
+++ b/cedar-policy-core/src/ast/extension.rs
@@ -328,7 +328,7 @@ impl std::fmt::Debug for ExtensionFunction {
 /// Anything implementing this trait can be used as a first-class value in
 /// Cedar. For instance, the `ipaddr` extension uses this mechanism
 /// to implement IPAddr as a Cedar first-class value.
-pub trait ExtensionValue: Debug + Display {
+pub trait ExtensionValue: Debug + Display + Send + Sync {
     /// Get the name of the type of this value.
     ///
     /// Cedar has nominal typing, so two values have the same type iff they
@@ -355,7 +355,7 @@ pub struct ExtensionValueWithArgs {
 
 impl ExtensionValueWithArgs {
     /// Get the internal value
-    pub fn value(&self) -> &dyn InternalExtensionValue {
+    pub fn value(&self) -> &(dyn InternalExtensionValue) {
         self.value.as_ref()
     }
 
@@ -365,7 +365,11 @@ impl ExtensionValueWithArgs {
     }
 
     /// Constructor
-    pub fn new(value: Arc<dyn InternalExtensionValue>, args: Vec<Expr>, constructor: Name) -> Self {
+    pub fn new(
+        value: Arc<dyn InternalExtensionValue + Send + Sync>,
+        args: Vec<Expr>,
+        constructor: Name,
+    ) -> Self {
         Self {
             value,
             args,
@@ -436,7 +440,7 @@ pub trait InternalExtensionValue: ExtensionValue {
     fn cmp_extvalue(&self, other: &dyn InternalExtensionValue) -> std::cmp::Ordering;
 }
 
-impl<V: 'static + Eq + Ord + ExtensionValue> InternalExtensionValue for V {
+impl<V: 'static + Eq + Ord + ExtensionValue + Send + Sync> InternalExtensionValue for V {
     fn as_any(&self) -> &dyn Any {
         self
     }
--- a/cedar-policy-core/src/ast/extension.rs
+++ b/cedar-policy-core/src/ast/extension.rs
@@ -22,6 +22,7 @@ use crate::evaluator;
 use std::any::Any;
 use std::collections::HashMap;
 use std::fmt::{Debug, Display};
+use std::panic::{RefUnwindSafe, UnwindSafe};
 use std::sync::Arc;
 
 /// Cedar extension.
@@ -328,7 +329,7 @@ impl std::fmt::Debug for ExtensionFunction {
 /// Anything implementing this trait can be used as a first-class value in
 /// Cedar. For instance, the `ipaddr` extension uses this mechanism
 /// to implement IPAddr as a Cedar first-class value.
-pub trait ExtensionValue: Debug + Display + Send + Sync {
+pub trait ExtensionValue: Debug + Display + Send + Sync + UnwindSafe + RefUnwindSafe {
     /// Get the name of the type of this value.
     ///
     /// Cedar has nominal typing, so two values have the same type iff they
--- a/cedar-policy-core/src/entities.rs
+++ b/cedar-policy-core/src/entities.rs
@@ -270,9 +270,9 @@ pub struct EntityAttrValues<'a> {
     entities: &'a Entities,
 }
 
-fn build_evaluated_entities<'e>(
+fn build_evaluated_entities(
     entities: &Entities,
-    extensions: &'e Extensions<'e>,
+    extensions: &'_ Extensions<'_>,
 ) -> std::result::Result<EvaluatedEntities, EvaluationError> {
     let restricted_eval = RestrictedEvaluator::new(extensions);
     // Eagerly evaluate each attribute expression in the entities.
--- a/cedar-policy-core/src/entities.rs
+++ b/cedar-policy-core/src/entities.rs
@@ -20,10 +20,10 @@ use crate::ast::*;
 use crate::evaluator::{EvaluationError, RestrictedEvaluator};
 use crate::extensions::Extensions;
 use crate::transitive_closure::{compute_tc, enforce_tc_and_dag};
+use std::borrow::Cow;
 use std::collections::{hash_map, HashMap};
 use std::fmt::Write;
 
-use either::Either;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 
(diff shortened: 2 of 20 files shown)


cedar-policy-core/src/entities.rs
```diff
@@ -48,6 +52,9 @@ pub struct Entities {
     #[serde_as(as = "Vec<(_, _)>")]
     entities: HashMap<EntityUID, Entity>,
 
+    #[serde(skip)]
+    evaluated_entities: Option<EvaluatedEntities>,
```
cdisselkoen, line 56: With this change, when we do have the evaluated entities (ie when that field is `Some`), the `Entities` object carries both the evaluated and unevaluated entities.  Would it make more sense to have the `entities` field be an enum, either unevaluated entities (a variant that carries `HashMap<EntityUID, Entity>`) or evaluated entities (a variant that carries `EvaluatedEntities`, or perhaps `HashMap<EntityUID, HashMap<SmolStr, PartialValue>>` directly)?

In addition to being "cleaner" and maybe having the memory footprint of the `Entities` be smaller, this might also prevent us from shooting ourselves in the foot, in that no other code could accidentally look up the unevaluated definitions and evaluate them again, if the evaluated ones are already available.

cedar-policy-core/src/entities.rs
```diff
@@ -216,6 +226,109 @@ impl Entities {
         dot_str.write_str("}\n")?;
         Ok(dot_str)
     }
+
+    /// Attempt to eagerly compute the values of attributes for all entities in the slice.
+    /// This can fail if evaluation of the [`RestrictedExpr`] fails.
+    /// In a future major version, we will likely make this function automatically called via the constructor.
+    pub fn evaluate(self) -> std::result::Result<Self, EvaluationError> {
```
cdisselkoen, line 233: It might be useful for this to return a different type than `Self` (at least a newtype wrapper), so that the type system actually distinguishes between "unevaluated Entities" and "evaluated Entities" objects.

cedar-policy-core/src/entities.rs
```diff
@@ -216,6 +226,109 @@ impl Entities {
...
+            })
+            .collect::<std::result::Result<
+                HashMap<EntityUID, HashMap<SmolStr, PartialValue>>,
+                EvaluationError,
+            >>()?;
+    Ok(attrs)
+}
+
+impl<'a> EntityAttrValues<'a> {
+    /// Construct an [EntityAttrValues] with either an owned or borrowed set of evlauted attributes.
```
cdisselkoen, line 310:
```suggestion
    /// Construct an [`EntityAttrValues`] with either an owned or borrowed set of evaluated attributes.
```

cedar-policy-core/src/evaluator.rs
```diff
@@ -17,8 +17,9 @@
 //! This module contains the Cedar evaluator.
 
 use crate::ast::*;
-use crate::entities::{Dereference, Entities};
+use crate::entities::{Dereference, Entities, EntityAttrValues};
 use crate::extensions::Extensions;
+#[cfg(test)]
 use std::collections::HashMap;
```
cdisselkoen, line 23: instead of the `cfg(test)`, could we move this import into the test module(s) where it's needed?

cedar-policy/src/api.rs
```diff
@@ -170,6 +170,15 @@ impl Entities {
         Self(self.0.partial())
     }
 
+    /// Attempt to eagerly compute the values of attributes for all entities in the slice.
+    /// This can fail if evaluation of the [`RestrictedExpression`] fails.
+    /// In a future major version, we will likely make this function automatically called via the constructor.
```
cdisselkoen, line 175: why does that change have to wait for next major version?

cdisselkoen (commented):
Thanks for the work on this!

Can we also get a better PR title than "First draft" at some point? :)

cedar-policy/src/api.rs
aaronjeline: Because it introduces a new way creation of an entities object can fail, and the `EntityError` enum isn't `non_exhaustive`

cedar-policy-core/src/evaluator.rs
aaronjeline: It's used in a top level function that is `#[cfg(test)]`, so I'm not sure?

cedar-policy-core/src/entities.rs
aaronjeline: But we need `is_authorized to still accept both, no?

cedar-policy-core/src/entities.rs
aaronjeline: The problem is serialization. You can't throw out the restricted expressions or else you can't serialize anymore

cedar-policy-core/src/evaluator.rs
cdisselkoen: Ah got it

cedar-policy-core/src/entities.rs
cdisselkoen: I think we don't ever serialize these except for DRT.  Unfortunate that we have to accept performance/memory overhead in the hot path for the sake of DRT, but I'll allow it.  (Or am I wrong?)

cedar-policy-core/src/entities.rs
cdisselkoen: Ideally I'd love to live in a world where only the evaluated Entities are exposed, and thus all is_authorized callers always pass the evaluated Entities (because they're automatically evaluated during construction of `Entities` object).  But I understand we can't live in that world until next major.  So yes, I take your point

cedar-policy-core/src/entities.rs
```diff
@@ -216,6 +226,109 @@ impl Entities {
...
+pub struct EntityAttrValues<'a> {
+    /// The evaluated entity attributes. The attributes may either be owner or borrowed.
+    attrs: Either<&'a EvaluatedEntities, EvaluatedEntities>,
+    /// The original entity slice, which contains hierarchy information.
+    entities: &'a Entities,
+}
+
+fn build_evaluated_entities<'e>(
+    entities: &Entities,
+    extensions: &'e Extensions<'e>,
```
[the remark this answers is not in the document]
aaronjeline: Clippy did not warn me, but you are correct

cedar-policy-core/src/ast/extension.rs
```diff
@@ -355,7 +356,7 @@ pub struct ExtensionValueWithArgs {
 
 impl ExtensionValueWithArgs {
     /// Get the internal value
-    pub fn value(&self) -> &dyn InternalExtensionValue {
+    pub fn value(&self) -> &(dyn InternalExtensionValue) {
```
cdisselkoen, line 359: Curious what this change does / what this syntax means?

cdisselkoen (approved):

cedar-policy-core/src/ast/extension.rs
aaronjeline: Ah, it's no-op. It was needed as I working through it as my first draft was changing `&dyn ...` to `&(dyn ... + Send + Sync`. Ended up pushing the bounds on `Send`/`Sync` further down

cedar-policy-core/src/entities.rs
khieta: Same comment as @cdisselkoen - it would be cleaner to use an enum to represent entities that are either evaluated or not evaluated. Make sure to file an issue about this before you close the PR so we don't forget

khieta (approved):

cedar-policy-core/src/entities.rs
prakol16, line 56: Oh no my merge conflicts :(

We are solving essentially the same problem in pretty different ways. I've been trying to keep my fork up to date with cedar `main` so that merging is/will be less troublesome, but I think that this will force a split because it solves an existing problem in a different way from how I solved it.

I made `Entity` and `Entities` generic around a type `T` which is the type of attributes. Then, `T` can take the value of either `RestrictedExpr` or `PartialValue`  (with `RestrictedExpr` the default to prevent breaking changes). I then changed the evaluator to only work with `EntityDatabase` (which is implemented only for `Entities<PartialValue>`, not `Entities<RestrictedExpr>`, because the `get` method needs to return attributes of type `PartialValue`). Finally I modified `is_authorized` to convert an `Entities<RestrictedExpr>` to an `Entities<PartialValue>` and duplicated the API for `is_authorized` to take in `Entities<PartialValue>`  instead of`Entities<RestrictedExpr>` in case the user wants to do this parsing/caching themselves (the goal is eventually the duplicated API will be the default and the other/old API will be deprecated/removed).

Could you comment on the pros/cons of this approach? Besides minor performance benefits because you don't duplicate the `HashMap`, the biggest reason I need this is there is an explicit type for entities with evaluated attributes, which can implement `EntityDatabase`. The major drawback is it duplicates all existing API to take in the new type instead of just adding a method to `Entities` like yours does.

prakol16 (approved):

khieta (approved):

cedar-policy-core/src/entities.rs
cdisselkoen: Discussed in other comments (eg [this one](https://github.com/cedar-policy/cedar/pull/230#discussion_r1276666878)) -- I'd also love for "entities with evaluated attributes" to be an explicitly different type from "entities without evaluated attributes", but Aaron's solution in this PR avoids changing the `is_authorized` signature for back-compatibility.  Later, on next major, we could maybe make the type-level changes and change the signature of `is_authorized`.

PR #233, opened 2023-07-27
Abort parsing when invalid clause identifiers are seen

### LINKED ISSUE #232: Odd error message when parsing not valid policy condition

### Bug Category

Cedar Parser

### Describe the bug

Odd error message when trying to parse this policy

```
permit (principal, action, resource)
whenless {};
```
>  poorly formed: not a valid policy condition: whenless
>  poorly formed: whenless clause should not be empty

If `whenless` is not valid, then why say it should not be empty?

### Expected behavior

A single error message would be best (preferably with line/character offsets)

> poorly formed: not a valid policy condition: whenless

### Reproduction steps

Save the above policy into a file named `whenless.cedar` then use the `cedar` cli tool

```
cedar check-parse -p whenless.cedar 
failed to parse policy set from file whenless.cedar: 
  poorly formed: not a valid policy condition: whenless
  poorly formed: whenless clause should not be empty
```

### Code Snippet

_No response_

### Log output

_No response_

### Additional configuration

_No response_

### Operating  System

_No response_

_No response_

shaobo-he-aws:
Fixes #232

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/parser/cst_to_ast.rs
+++ b/cedar-policy-core/src/parser/cst_to_ast.rs
@@ -650,11 +650,13 @@ impl ASTNode<Option<cst::Cond>> {
             }
         };
 
-        match (maybe_is_when, maybe_expr) {
-            (true, Some(e)) => Some(e),
-            (false, Some(e)) => Some(construct_expr_not(e, src.clone())),
-            _ => None,
-        }
+        maybe_expr.map(|e| {
+            if maybe_is_when {
+                e
+            } else {
+                construct_expr_not(e, src.clone())
+            }
+        })
     }
 }
 
--- a/cedar-policy-core/src/parser/cst_to_ast.rs
+++ b/cedar-policy-core/src/parser/cst_to_ast.rs
@@ -637,7 +637,7 @@ impl ASTNode<Option<cst::Cond>> {
         // return right away if there's no data, parse provided error
         let cond = maybe_cond?;
 
-        let maybe_is_when = cond.cond.to_cond_is_when(errs);
+        let maybe_is_when = cond.cond.to_cond_is_when(errs)?;
 
         let maybe_expr = match &cond.expr {
             Some(expr) => expr.to_expr(errs),
@@ -651,8 +651,8 @@ impl ASTNode<Option<cst::Cond>> {
         };
 
         match (maybe_is_when, maybe_expr) {
-            (Some(true), Some(e)) => Some(e),
-            (Some(false), Some(e)) => Some(construct_expr_not(e, src.clone())),
+            (true, Some(e)) => Some(e),
+            (false, Some(e)) => Some(construct_expr_not(e, src.clone())),
             _ => None,
         }
     }


cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -650,11 +650,13 @@ impl ASTNode<Option<cst::Cond>> {
...
-            (Some(false), Some(e)) => Some(construct_expr_not(e, src.clone())),
-            _ => None,
-        }
+        maybe_expr.map(|e| {
+            if maybe_is_when {
+                e
+            } else {
+                construct_expr_not(e, src.clone())
+            }
+        })
```
khieta, line 659: I think this change results in a different program logic -- what happens when `maybe_expr` is `None`? The quick fix would be to add a question mark on line 651.

OTOH, given that the CI passed, perhaps this logic wasn't important. Or we're just not doing a good job of testing parse errors :)

john-h-kastner-aws:
The error `poorly formed: whenless clause should not be empty` is odd, but still somewhat useful, so I'm not convinced that eliminating it entirely for a typoed keyword is best. Could we instead check if the keyword not `when` or `unless`, and then generate a different, more generic, error message like `poorly formed: policy body clause should not be empty`?

cdisselkoen:
I second John's suggestion, but "policy body clause" might not be the clearest wording, maybe "policy clause"?  What do our docs call these?

cedar-policy-core/src/parser/cst_to_ast.rs
cdisselkoen: On my reading, the behavior when `maybe_expr` is `None` is unchanged -- `Option::map` returns `None` if the receiver is `None`

khieta: Haha good point. I missed the `map`.

khieta:
`when {...}` and `unless {...}` are "conditions", so I think the `{...}` would be the "condition body".

But I feel that the pair of errors
```
  poorly formed: not a valid policy condition: whenless
  poorly formed: condition body should not be empty
```
are still confusing together. I'd still prefer just the first (possibly modified to point out that the valid options are "when" and "unless"). Then after the user fixes the first error, they would get the second.

cdisselkoen:
That makes sense to me too

cdisselkoen:
As a meta point, I don't think that "poorly formed" adds anything to these messages (or other messages) -- should we remove it from all `ToAST` parse errors?

shaobo-he-aws:
> As a meta point, I don't think that "poorly formed" adds anything to these messages (or other messages) -- should we remove it from all `ToAST` parse errors?

Andrew is working on ToAST errors. We can bring the suggestion in his PR.

shaobo-he-aws:
> The error `poorly formed: whenless clause should not be empty` is odd, but still somewhat useful, so I'm not convinced that eliminating it entirely for a typoed keyword is best. Could we instead check if the keyword not `when` or `unless`, and then generate a different, more generic, error message like `poorly formed: policy body clause should not be empty`?

I think this PR won't remove the error message `... clause should not be empty`. Like @khieta pointed out, it would show up after the user corrects the identifier.

john-h-kastner-aws:
> it would show up after the user corrects the identifier.

I don't agree that this is better than showing up at the same time. We know that's the body shouldn't be empty, regardless of what type of body it's meant to be.

khieta (approved):
I'm happy with the proposed change. I prefer the current implementation, but would also be ok with @john-h-kastner-aws's alternative.

cdisselkoen (approved):
After consideration, I'm on the same page as Kesha's comment immediately above.

Issue #182, opened 2023-07-14
`miette` integration for all errors


cdisselkoen:
### Category

User level API changes

### Describe the feature you'd like to request

In #148, we refactored some of our errors to use `miette`.  We should do this for all errors in Core and Validator.

### Describe the solution you'd like

As above.

### Describe alternatives you've considered

N/A

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [X] ⚠️ This feature might incur a breaking change

cdisselkoen:
To be explicit: This involves threading source location through many more places.  That could be tracked as a separate issue or as part of this one.

Issue #177, opened 2023-07-13
Improved error message when trying to use `in` for set membership


cdisselkoen:
### Category

Other

### Describe the feature you'd like to request

As pointed out by @khieta in [this comment on RFC 18](https://github.com/cedar-policy/rfcs/pull/18#discussion_r1262583601).  For errors like
```
context.solution_id in principal.solution_ids
```
instead of the not-very-helpful error message
```
Errors: while evaluating policy policy0, encountered the following error: type error: expected (entity of type any_entity_type), got string
```
we should have something more helpful like this
```
type error: in expects an entity on the LHS. Use .contains() for set membership
```

### Describe the solution you'd like

As above.

### Describe alternatives you've considered

RFC 18 discusses alternatives, but this is something we can do now to help while RFC 18 continues to be discussed and/or implemented.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #253, opened 2023-08-01
Add `advice` to `EvaluationError`s

khieta:
## Description of changes

* Refactored `EvaluationError` into the following form, which matches our current `ValidationError` structure. The `advice` field contains an optional string with additional help information. If we were using [miette](https://docs.rs/miette/latest/miette/), this field would be labeled as `#[help]`. But for this PR, I'm considering updating to miette out of scope (see issue #182). This new structure should make supporting location information simpler in the future.
```
pub enum EvaluationError {
  advice: Option<String>,
  error_kind: EvaluationErrorKind.
}
```
* Added advice in 4 error cases that seem likely:
  * Using `in` instead of `.contains()`
     `3 in [34, -2, 7]` --> "type error: expected (entity of type any_entity_type), got long. `in` is for checking the entity hierarchy, use `.contains()` to test set membership"
  * Using `in` instead of `has`
    `"foo" in { "foo": 2, "bar": true }` --> "type error: expected (entity of type any_entity_type), got string. `in` is for checking the entity hierarchy, use `has` to test if a record has a key"
  * Using an IP-related function on a string instead of an IP value
    `"127.0.0.1".isIpv4()` --> "type error: expected ipaddr, got string. Maybe you forgot to apply the `ip` constructor?"
  * Using a decimal-related function on a string instead of a decimal value
    `"1.23".lessThan("2.34")` --> "type error: expected decimal, got string. Maybe you forgot to apply the `decimal` constructor?"

## Issue #, if available

#177

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/ast/extension.rs
+++ b/cedar-policy-core/src/ast/extension.rs
@@ -158,11 +158,11 @@ impl ExtensionFunction {
                 if args.is_empty() {
                     func()
                 } else {
-                    Err(evaluator::EvaluationError::WrongNumArguments {
-                        function_name: name.clone(),
-                        expected: 0,
-                        actual: args.len(),
-                    })
+                    Err(evaluator::EvaluationError::wrong_num_arguments(
+                        name.clone(),
+                        0,
+                        args.len(),
+                    ))
                 }
             }),
             Some(return_type),
@@ -182,11 +182,11 @@ impl ExtensionFunction {
             style,
             Box::new(move |args: &[Value]| match args.get(0) {
                 Some(arg) => func(arg.clone()),
-                None => Err(evaluator::EvaluationError::WrongNumArguments {
-                    function_name: name.clone(),
-                    expected: 1,
-                    actual: args.len(),
-                }),
+                None => Err(evaluator::EvaluationError::wrong_num_arguments(
+                    name.clone(),
+                    1,
+                    args.len(),
+                )),
             }),
             None,
             vec![arg_type],
@@ -206,11 +206,11 @@ impl ExtensionFunction {
             style,
             Box::new(move |args: &[Value]| match &args {
                 &[arg] => func(arg.clone()),
-                _ => Err(evaluator::EvaluationError::WrongNumArguments {
-                    function_name: name.clone(),
-                    expected: 1,
-                    actual: args.len(),
-                }),
+                _ => Err(evaluator::EvaluationError::wrong_num_arguments(
+                    name.clone(),
+                    1,
+                    args.len(),
+                )),
             }),
             Some(return_type),
             vec![arg_type],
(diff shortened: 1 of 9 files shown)


cdisselkoen (approved):
To your questions in the description: 

Yes, I think it's a good idea to make that structural change to `EvaluationError`.

No, I don't think it has to be in scope for this PR (but if it isn't, let's make a separate issue for the backlog?)

Something I do think we could do in this PR is make the same change for `decimal` that we did for `ip` -- i.e., a hint for the case `"23.57".greaterThan(...)`

khieta:
Finished the first pass of refactoring! The diff is larger now, but the only important changes are in `evaluator/err.rs`, `extensions/ipaddr.rs`, `extensions/decimal.rs`, and `evaluator.rs` L398-411.

cedar-policy-core/src/ast/extension.rs
```diff
@@ -158,11 +158,11 @@ impl ExtensionFunction {
...
-                    Err(evaluator::EvaluationError::WrongNumArguments {
-                        function_name: name.clone(),
-                        expected: 0,
-                        actual: args.len(),
-                    })
+                    Err(evaluator::EvaluationError::wrong_num_arguments(
+                        name.clone(),
+                        0,
+                        args.len(),
+                    ))
```
cdisselkoen, line 165: side note that it's a shame this has to become less readable like this, I wish Rust had named arguments to functions

cedar-policy-core/src/evaluator.rs
```diff
@@ -3777,6 +3648,16 @@ pub mod test {
         );
     }
 
+    fn assert_restricted_expression_error(v: Result<PartialValue>) {
+        match v {
+            Err(e) => assert!(matches!(
```
cdisselkoen, line 3653: `assert_matches!` gives nicer errors than `assert!(matches!(` which just will say "hey this was false and expected true"

cedar-policy-core/src/extensions/decimal.rs
```diff
@@ -290,17 +301,20 @@ mod tests {
...
+                    println!("{msg}");
+                    assert_eq!(
+                        *extension_name,
+                        Name::parse_unqualified_name("decimal")
+                            .expect("should be a valid identifier")
+                    )
+                }
+                _ => panic!("Expected an decimal ExtensionErr, got {:?}", e),
+            },
             Ok(_) => panic!("Expected an decimal ExtensionErr, got Ok"),
```
cdisselkoen, line 318:
```suggestion
                _ => panic!("Expected a decimal ExtensionErr, got {:?}", e),
            },
            Ok(_) => panic!("Expected a decimal ExtensionErr, got Ok"),
```

cdisselkoen (approved):

cedar-policy-core/src/evaluator.rs
khieta: `assert_matches` is a nightly feature, so to enable it we'd have to add `#![feature(assert_matches)]` to the crate -- are we ok with adding this crate-level attribute?

john-h-kastner-aws (approved):
Looks great.

PR #266, opened 2023-08-08
Add a check-schema command to the CLI

john-h-kastner-aws:
## Description of changes

I added `check-schema` to the CLI so that I could play with schema error messages conveniently. Might be useful to merge.

## Disclaimer

### DIFF
--- a/cedar-policy-cli/CHANGELOG.md
+++ b/cedar-policy-cli/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Added
+- A `check-schema` command to check that a schema file parses.
+
 ### Changed
 - Input policies for `check-parse` command can be read from standard input.
 - Duplicate policy ids in `@id` annotations cause the CLI to exit gracefully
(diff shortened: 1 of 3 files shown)


shaobo-he-aws:
QQ: can we add an enum to the `check-parse` subcommand so that it can check wether a policy, schema, or entity store parses?

john-h-kastner-aws:
> QQ: can we add an enum to the `check-parse` subcommand so that it can check wether a policy, schema, or entity store parses?

This might be nicer.

andrewmwells-amazon (approved):

hakanson:
How might this behave if [RFC 24](https://github.com/cedar-policy/rfcs/blob/schema-syntax/text/0024-schema-syntax.md) is adopted?  Would you add a `--format` parameter with `--format=json` the future default and `--format=rfc24` a future option, or try and auto-detect the schema syntax (either by pre-parsing or trying both)?

This issue would also affect `cedar validate --schema XXX ...` or `cedar evaluate --schema XXX ...`, so maybe a comment I should add to RFC 24 instead?

Note, I have used ` cedar evaluate --schema cedarschema.json true` before to get schema parsing errors.

john-h-kastner-aws:
Good comments about the new format. Definitely needs considering, so you may as well drop a comment on the RFC.

I've left this PR idle for a while without making the modification shaobo suggested, so I'll just close this PR for now.

PR #274, opened 2023-08-14
Typecheck template based on possible instantiation types

john-h-kastner-aws:
Rather than typechecking a template once using `AnyEntity` for the type of the slot, we can typecheck it once for each possible entity type that could instantiate that slot. 

This does not affect current permissive validation. It is instead intended to support strict typechecking templates after the coming strict validation changes. It will also support more precise template typechecking if template slots are allowed in policy bodies in the future.

- [x] Does not require updates because my change does not impact the Cedar Dafny model or DRT 

## Disclaimer

### DIFF
--- a/cedar-policy-validator/src/schema.rs
+++ b/cedar-policy-validator/src/schema.rs
@@ -1422,6 +1422,18 @@ impl ValidatorEntityType {
     pub fn attributes(&self) -> impl Iterator<Item = (&SmolStr, &AttributeType)> {
         self.attributes.iter()
     }
+
+    /// Return `true` if this entity type has an `EntityType` declared as a
+    /// possible descendant in the schema. This takes an `EntityType` rather
+    /// than a `Name`, It's not possible to declare the unspecified entity type
+    /// is a descendant of an entity type in the schema, so we can return false
+    /// in the unspecified case.
+    pub fn has_descendant_entity_type(&self, ety: &EntityType) -> bool {
+        match ety {
+            EntityType::Concrete(ety) => self.descendants.contains(ety),
+            EntityType::Unspecified => false,
+        }
+    }
 }
 
 impl TCNode<Name> for ValidatorEntityType {
--- a/cedar-policy-validator/src/typecheck.rs
+++ b/cedar-policy-validator/src/typecheck.rs
@@ -510,8 +510,8 @@ impl<'a> Typechecker<'a> {
 
     /// Get the entity types which could instantiate the slot given in this
     /// template based on the policy scope constraints. We use this function to
-    /// avoid typechecking with for slot bindings that will always be false
-    /// based only on the scope constraints.
+    /// avoid typechecking with slot bindings that will always be false based
+    /// only on the scope constraints.
     fn possible_slot_instantiations(
         &self,
         t: &Template,
--- a/cedar-policy-validator/src/typecheck.rs
+++ b/cedar-policy-validator/src/typecheck.rs
@@ -50,9 +50,10 @@ use crate::{
 use super::type_error::TypeError;
 
 use cedar_policy_core::ast::{
-    BinaryOp, EntityType, EntityUID, Expr, ExprBuilder, ExprKind, Literal, Name, Template, UnaryOp,
-    Var,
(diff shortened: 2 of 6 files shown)


cedar-policy-validator/src/typecheck.rs
```diff
@@ -441,24 +457,100 @@ impl<'a> Typechecker<'a> {
...
+                context: env.context,
+                principal_slot: p_slot.clone(),
+                resource_slot: r_slot.clone(),
+            })
+        })
+    }
+
+    /// Get the entity types which could instantiate the slot given in this
+    /// template based on the policy scope constraints. We use this function to
+    /// avoid typechecking with for slot bindings that will always be false
```
andrewmwells-amazon, line 513: [nit] with for

andrewmwells-amazon (approved):

andrewmwells-amazon (approved):

cedar-policy-validator/src/schema.rs
```diff
@@ -1422,6 +1422,18 @@ impl ValidatorEntityType {
     pub fn attributes(&self) -> impl Iterator<Item = (&SmolStr, &AttributeType)> {
         self.attributes.iter()
     }
+
+    /// Return `true` if this entity type has an `EntityType` declared as a
+    /// possible descendant in the schema. This takes an `EntityType` rather
+    /// than a `Name`, It's not possible to declare the unspecified entity type
```
khieta, line 1428:
```suggestion
    /// than a `Name`. It's not possible to declare the unspecified entity type
```

cedar-policy-validator/src/typecheck.rs
```diff
@@ -441,24 +457,100 @@ impl<'a> Typechecker<'a> {
...
+                // appear in head constraints, but if we ever see this, then the
+                // only correct way to proceed is by returning all entity types
+                // as possible instantiations.
+                PrincipalOrResourceConstraint::Any => Box::new(
+                    all_entity_types.map(|(name, _)| Some(EntityType::Concrete(name.clone()))),
+                ),
+            }
+        } else {
+            // If the template does not contain this slot, then we don't need to
+            // consider its instantiations..
```
khieta, line 549:
```suggestion
            // consider its instantiations.
```

cedar-policy-validator/src/typecheck.rs
```diff
@@ -1087,9 +1179,23 @@ impl<'a> Typechecker<'a> {
...
+                        .map(|ety| Type::possibly_unspecified_entity_reference(ety))
+                        .unwrap_or(Type::any_entity_reference())
+                } else if slotid.is_resource() {
+                    request_env
+                        .resource_slot
+                        .clone()
+                        .map(|ety| Type::possibly_unspecified_entity_reference(ety))
+                        .unwrap_or(Type::any_entity_reference())
+                } else {
+                    Type::any_entity_reference()
```
khieta, line 1195: Should these `unwrap` and `else` cases be errors? Why is it ok to return `any_entity_reference`?

cedar-policy-validator/src/typecheck.rs
```diff
@@ -441,24 +457,100 @@ impl<'a> Typechecker<'a> {
...
+                resource_slot: r_slot.clone(),
+            })
+        })
+    }
+
+    /// Get the entity types which could instantiate the slot given in this
+    /// template based on the policy scope constraints. We use this function to
+    /// avoid typechecking with slot bindings that will always be false based
+    /// only on the scope constraints.
+    fn possible_slot_instantiations(
```
khieta, line 515: Would be more consistent to stick with the "link" terminology rather than switching to "instantiation" here.

khieta (approved):
Lgtm. Added a couple nit picky comments, but they're not blocking.

cedar-policy-validator/src/typecheck.rs
john-h-kastner-aws: If I see some other slot that I didn't account for in the query env, it's still safe to call it `AnyEntity`. This could be a panic, because it shouldn't happen currently, but this behavior gives a sound fall-back if it does occur.

Issue #272, opened 2023-08-14
provide a way to build `Entities` "by hand"


verrchu:
### Category

User level API changes

### Describe the feature you'd like to request

I have this situation where I obtain "initial" `Entities" from json (via `Entities::from_json_value`) and then I have to add several hand-crafted `Entity` instances on top of that but current API doesn't allow it.

### Describe alternatives you've considered

an alternative for my particular use case would be to provide a way to obtain `Enitty` (not `Entities`) form json. this way I could have alist of "initial" `Entity` instances and then build `Entities` from them combined with the hand-crafted ones

### Additional context

as suggested on Slack a workaround for this issue would be to obtain an iterator of `Entity` from `Entities` via `Entities::iter()` and go from there :)

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

emina:
A third alternative is to provide an API that adds a collection of entity instances to `Entities`.  This may trigger re-computation of transitive closure, depending on the implementation, which is O(n^2).  A bulk API lets us recompute only once and avoid copying if multiple entities need to be added.

aaronjeline:
I'm going to make a second issue for your alternative @verrchu, because I think we want both.

cdisselkoen:
Closing as partially completed in #276, with the remaining work separated out as #275.  Thanks for the report!

PR #276, opened 2023-08-15
Adds APis for adding entities to `Entities`

aaronjeline:
## Description of changes
Adds APIs to the `Entities` structure that allow you to add a collection of `Entity` objects to an existing `Entities` structure.
## Issue #, if available
#272 

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/entities.rs
+++ b/cedar-policy-core/src/entities.rs
@@ -104,6 +104,36 @@ impl Entities {
         self.entities.values()
     }
 
+    /// Adds the `[crate::ast::Entity]` to this `[Entities]`.
+    /// Fails if the passed iterator contains any duplicate entities with this structure,
+    /// or if any error is encountered in the transitive closure computation.
+    ///
+    /// If you pass `TCComputation::AssumeAlreadyComputed`, then the caller is
+    /// responsible for ensuring that TC and DAG hold before calling this method.
+    pub fn add_entities(
+        mut self,
+        collection: impl IntoIterator<Item = Entity>,
+        mode: TCComputation,
+    ) -> Result<Self> {
+        for entity in collection.into_iter() {
+            match self.entities.entry(entity.uid()) {
+                hash_map::Entry::Occupied(_) => return Err(EntitiesError::Duplicate(entity.uid())),
+                hash_map::Entry::Vacant(vacant_entry) => {
+                    vacant_entry.insert(entity);
+                }
+            }
+        }
+        match mode {
+            TCComputation::AssumeAlreadyComputed => (),
+            TCComputation::EnforceAlreadyComputed => {
+                enforce_tc_and_dag(&self.entities).map_err(Box::new)?
+            }
+            TCComputation::ComputeNow => compute_tc(&mut self.entities, true).map_err(Box::new)?,
+        };
+        self.evaluated_entities = None;
+        Ok(self)
+    }
+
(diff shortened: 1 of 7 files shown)


cedar-policy-core/src/entities.rs
```diff
@@ -104,6 +104,36 @@ impl Entities {
         self.entities.values()
     }
 
+    /// Adds the [`crate::ast::Entity`] to this [`Entities`].
```
cdisselkoen, line 107: might be clearer to point out this could add more than one `crate::ast::Entity`

cedar-policy-core/src/entities.rs
```diff
@@ -104,6 +104,36 @@ impl Entities {
...
+            }
+        }
+        match mode {
+            TCComputation::AssumeAlreadyComputed => (),
+            TCComputation::EnforceAlreadyComputed => {
+                enforce_tc_and_dag(&self.entities).map_err(Box::new)?
+            }
+            TCComputation::ComputeNow => compute_tc(&mut self.entities, true).map_err(Box::new)?,
+        };
+        self.evaluated_entities = None;
```
cdisselkoen, line 133: should instead, we look to see if `evaluated_entities` is Some, and if it is, evaluate these new entities as well so that `evaluated_entities` remains Some?

Relatedly, what happens today if someone calls `.evaluate()` twice?

cedar-policy-core/src/entities.rs
```diff
@@ -433,6 +463,129 @@ mod json_parsing_tests {
     use super::*;
     use crate::extensions::Extensions;
 
+    #[test]
+    fn adds_extends_tc() {
```
cdisselkoen, line 467: would be great to have another test involving `EnforceAlreadyComputed` and ensure that it (a) works when TC is in fact respected, and (b) fails properly when TC is not respected

cedar-policy-core/src/entities/json/entities.rs
```diff
@@ -96,28 +96,64 @@ impl<'e, S: Schema> EntityJsonParser<'e, S> {
...
-    /// Parse an entities JSON file (in `std::io::Read` form) into an `Entities` object
+    /// Parse an entities JSON file (in [`std::io::Read`] form) into an [`Entities`] object
     pub fn from_json_file(&self, json: impl std::io::Read) -> Result<Entities, EntitiesError> {
         let ejsons: Vec<EntityJSON> =
             serde_json::from_reader(json).map_err(JsonDeserializationError::from)?;
         self.parse_ejsons(ejsons)
     }
 
-    /// internal function that creates an `Entities` from a stream of `EntityJSON`
+    /// Parse an entities JSON file (in [`&str`] form) into an iterator over [Entity]s
```
cdisselkoen, line 120:
```suggestion
    /// Parse an entities JSON file (in [`&str`] form) into an iterator over [`Entity`]s
```

cdisselkoen (approved):

cedar-policy-core/src/entities.rs
aaronjeline: The problem w/ doing that is that it introduces a new error mode, which means this method will have different error semantics from the normal constructor

cdisselkoen: Makes sense, carry on

shaobo-he-aws (approved):

cdisselkoen (approved):

PR #311, opened 2023-09-18
Formatting of error messages

max2me:
## Description of changes
This PR formats error messages by following sentence case, wrapping tokens/keywords in ' and refining some of the strings.

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/authorizer/err.rs
+++ b/cedar-policy-core/src/authorizer/err.rs
@@ -22,11 +22,11 @@ use thiserror::Error;
 #[derive(Debug, PartialEq, Eq, Clone, Error)]
 pub enum AuthorizationError {
     /// Failed to eagerly evaluate entity attributes when initializing the `Evaluator`.
-    #[error("error occurred while evaluating entity attributes: {0}")]
+    #[error("Error occurred while evaluating entity attributes: {0}.")]
     AttributeEvaluationError(EvaluationError),
 
     /// An error occurred when evaluating a policy.
-    #[error("error occurred while evaluating policy `{}`: {}", &.id, &.error)]
+    #[error("Error occurred while evaluating policy '{}': {}.", &.id, &.error)]
     PolicyEvaluationError {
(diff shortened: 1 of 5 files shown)


khieta:
The build is currently failing for two reasons: 1) we require code to be formatted using Rust's formatter, which can be done by running `cargo fmt`; 2) some of our tests check the content of error messages, so the tests need to be updated when error strings change. These are both simple to fix & I can do that once we've resolved any other outstanding comments on this PR.

cedar-policy-core/src/authorizer/err.rs
```diff
@@ -22,11 +22,11 @@ use thiserror::Error;
 #[derive(Debug, PartialEq, Eq, Clone, Error)]
 pub enum AuthorizationError {
     /// Failed to eagerly evaluate entity attributes when initializing the `Evaluator`.
-    #[error("error occurred while evaluating entity attributes: {0}")]
+    #[error("Error occurred while evaluating entity attributes: {0}.")]
     AttributeEvaluationError(EvaluationError),
 
     /// An error occurred when evaluating a policy.
-    #[error("error occurred while evaluating policy `{}`: {}", &.id, &.error)]
+    #[error("Error occurred while evaluating policy '{}': {}.", &.id, &.error)]
```
khieta, line 29: I recall now that the reason we decided not to use proper sentence capitalization/punctuation was to support "combining" error messages using colons. Because `&.error` is also an error message, the current change will result in messages like the following: 

```
Error occurred while evaluating policy 'policy2': Cannot access attribute of unspecified entity 'jobLevel'..
```

(Note the double period.) There are a couple ways to solve this: 1) continue using lowercase & no punctuation, 2) combine error messages using periods instead of colons.

I don't have strong feelings either way, but it seems like the precedent in Rust is the former (see, e.g. <https://rustc-dev-guide.rust-lang.org/diagnostics.html#diagnostic-structure>). Any other opinions?

PR #330, opened 2023-09-28
Add a new command to the CLI that creates dummy Cedar project

shaobo-he-aws:
## Description of changes
Now users should be able to create a Cedar project named `hello-world` by issuing `cedar new hello-world`. The dummy project contains a schema, a policy, and an entity store.

## Disclaimer

### DIFF
--- a/cedar-policy-cli/src/lib.rs
+++ b/cedar-policy-cli/src/lib.rs
@@ -22,7 +22,7 @@
 mod err;
 
 use clap::{Args, Parser, Subcommand, ValueEnum};
-use miette::{miette, NamedSource, Report, Result, WrapErr};
+use miette::{miette, IntoDiagnostic, NamedSource, Report, Result, WrapErr};
 use serde::{Deserialize, Serialize};
 use std::{
     collections::HashMap,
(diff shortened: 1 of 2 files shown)


cedar-policy-cli/src/lib.rs
```diff
@@ -506,6 +513,104 @@ pub fn format_policies(args: &FormatArgs) -> CedarExitCode {
     }
 }
 
+fn generate_schema(path: &Path) -> Result<()> {
+    std::fs::write(
+        path,
+        serde_json::to_string_pretty(&serde_json::json!(
+        {
+            "": {
```
khieta, line 521: Might be good to use a nonempty namespace here since the `"": {` might look strange to first time users.

khieta (approved):
LGTM. Make sure to update the instructions in the CLI README before merging.

andrewmwells-amazon (approved):

cedar-policy-cli/src/lib.rs
shaobo-he-aws: Good point. I think adding a namespace will make entity literals more complicated. We can revisit it later.

Issue #328, opened 2023-09-27
Provide remove methods for PolicySet


0x00A5:
### Category

User level API changes

### Describe the feature you'd like to request

The `PolicySet` struct provides `add`, `add_template`, `link` methods to insert static policy, policy template and dynamic policy to a policy set. It would be useful to also provide the counterpart methods to remove these policies from a policy set.

```rust
/// Remove a static policy from the policy set
fn remove(&mut self, policy_id: PolicyId) -> Result<(), PolicySetError>;

/// Remove a policy template and all dynamic policies linked to it
fn remove_template(&mut self, template_id: PolicyId) -> Result<(), PolicySetError>;

/// Unlink a dynamic policy from the policy set
fn unlink(&mut self, policy_id: PolicyId) -> Result<(), PolicySetError>;
```

### Describe alternatives you've considered

Alternatively, these methods could be replaced by a simple `remove` method that takes a policy id and according to the policy type, removes the policy/template or unlink the dynamic policy.

### Is this something that you'd be interested in working on?

- [x] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

andrewmwells-amazon:
Closed with https://github.com/cedar-policy/cedar/pull/337

PR #337, opened 2023-10-04
Add apis to remove policies

andrewmwells-amazon:
## Description of changes
Add APIs for removing policies / policy templates and for unlinking templates.

## Issue #, if available
https://github.com/cedar-policy/cedar/issues/328

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/ast/policy_set.rs
+++ b/cedar-policy-core/src/ast/policy_set.rs
@@ -67,10 +67,10 @@ impl TryFrom<LiteralPolicySet> for PolicySet {
         for template in &templates {
             template_to_links_map.insert(template.0.clone(), HashSet::new());
         }
-        for linked_policy in &links {
-            let template = linked_policy.1.template().id();
+        for (link_id, link) in &links {
+            let template = link.template().id();
             match template_to_links_map.entry(template.clone()) {
-                Entry::Occupied(t) => t.into_mut().insert(linked_policy.0.clone()),
+                Entry::Occupied(t) => t.into_mut().insert(link_id.clone()),
                 Entry::Vacant(_) => return Err(ReificationError::NoSuchTemplate(template.clone())),
             };
         }
--- a/cedar-policy-core/src/ast/policy_set.rs
+++ b/cedar-policy-core/src/ast/policy_set.rs
@@ -339,7 +339,7 @@ impl PolicySet {
     /// Get the list of policies linked to `template_id`.
     /// Returns all p in `links` s.t. `p.template().id() == template_id`
     pub fn get_linked_policies(
-        &mut self,
+        &self,
         template_id: &PolicyID,
     ) -> Result<impl Iterator<Item = &PolicyID>, PolicySetGetLinksError> {
         match self.template_to_links_map.get(template_id) {
--- a/cedar-policy-core/src/ast/policy_set.rs
+++ b/cedar-policy-core/src/ast/policy_set.rs
@@ -138,9 +138,7 @@ pub enum PolicySetUnlinkError {
 #[derive(Error, Debug)]
 pub enum PolicySetTemplateRemovalError {
     /// There was no [`PolicyID`] template in the list of templates.
-    #[error(
-        "unable to remove static policy id `{0}` from template list because it does not exist"
-    )]
+    #[error("unable to remove template id `{0}` from template list because it does not exist")]
     RemovePolicyNoTemplateError(PolicyID),
     /// There are still active links to template [`PolicyID`].
     #[error(
@@ -343,9 +341,9 @@ impl PolicySet {
     pub fn get_linked_policies(
         &mut self,
         template_id: &PolicyID,
-    ) -> Result<Vec<&PolicyID>, PolicySetGetLinksError> {
+    ) -> Result<impl Iterator<Item = &PolicyID>, PolicySetGetLinksError> {
         match self.template_to_links_map.get(template_id) {
-            Some(s) => Ok(s.iter().collect_vec()),
+            Some(s) => Ok(s.iter()),
             None => Err(PolicySetGetLinksError::MissingTemplate(template_id.clone())),
         }
     }
--- a/cedar-policy-core/src/ast/policy_set.rs
+++ b/cedar-policy-core/src/ast/policy_set.rs
@@ -40,7 +40,7 @@ pub struct PolicySet {
     /// A `Template` may have zero or many links
     links: HashMap<PolicyID, Policy>,
 
-    /// Map from a template `PolicyID`` to the set of `PolicyID`s in `links` that are linked to that template.
+    /// Map from a template `PolicyID` to the set of `PolicyID`s in `links` that are linked to that template.
     /// There is a key `t` iff `templates` contains the key `t`. The value of `t` will be a (possibly empty)
     /// set of every `p` in `links` s.t. `p.template().id() == t`.
     template_to_links_map: HashMap<PolicyID, HashSet<PolicyID>>,
--- a/cedar-policy-core/src/ast/policy_set.rs
+++ b/cedar-policy-core/src/ast/policy_set.rs
@@ -20,7 +20,7 @@ use super::{
 };
 use itertools::Itertools;
 use serde::{Deserialize, Serialize};
-use std::collections::{hash_map::Entry, HashMap};
+use std::collections::{hash_map::Entry, HashMap, HashSet};
 use std::{borrow::Borrow, sync::Arc};
 use thiserror::Error;
 
@@ -39,6 +39,11 @@ pub struct PolicySet {
     ///   (this is managed by `PolicySet::add)
     /// A `Template` may have zero or many links
     links: HashMap<PolicyID, Policy>,
+
+    /// Map from a template `PolicyID`` to the set of `PolicyID`s in `links` that are linked to that template.
+    /// There is a key `t` iff `templates` contains the key `t`. The value of `t` will be a (possibly empty)
+    /// set of every `p` in `links` s.t. `p.template().id() == t`.
+    template_to_links_map: HashMap<PolicyID, HashSet<PolicyID>>,
 }
 
 /// Converts a LiteralPolicySet into a PolicySet, ensuring the invariants are met
@@ -57,7 +62,24 @@ impl TryFrom<LiteralPolicySet> for PolicySet {
             .into_iter()
             .map(|(id, literal)| literal.reify(&templates).map(|linked| (id, linked)))
             .collect::<Result<HashMap<PolicyID, Policy>, ReificationError>>()?;
-        Ok(Self { templates, links })
+
+        let mut template_to_links_map = HashMap::new();
+        for template in &templates {
+            template_to_links_map.insert(template.0.clone(), HashSet::new());
+        }
+        for linked_policy in &links {
+            let template = linked_policy.1.template().id();
+            match template_to_links_map.entry(template.clone()) {
+                Entry::Occupied(t) => t.into_mut().insert(linked_policy.0.clone()),
+                Entry::Vacant(_) => return Err(ReificationError::NoSuchTemplate(template.clone())),
+            };
+        }
+
+        Ok(Self {
+            templates,
+            links,
+            template_to_links_map,
+        })
     }
 }
 
@@ -94,6 +116,14 @@ pub enum PolicySetError {
         /// [`PolicyID`] that was duplicate
         id: PolicyID,
     },
+}
+
+/// Potential errors when working with `PolicySet`s.
+#[derive(Error, Debug)]
+pub enum PolicySetGetLinksError {
+    /// There was no [`PolicyID`] in the set of templates.
+    #[error("No template `{0}`")]
+    MissingTemplate(PolicyID),
 }
 
 /// Potential errors when unlinking from a `PolicySet`.
@@ -140,6 +170,7 @@ impl PolicySet {
         Self {
             templates: HashMap::new(),
             links: HashMap::new(),
+            template_to_links_map: HashMap::new(),
         }
     }
 
@@ -175,7 +206,19 @@ impl PolicySet {
         // if we get here, there will be no errors.  So actually do the
         // insertions.
         if let Some(ventry) = template_ventry {
+            self.template_to_links_map.insert(
+                t.id().clone(),
+                vec![policy.id().clone()]
+                    .into_iter()
+                    .collect::<HashSet<PolicyID>>(),
+            );
             ventry.insert(t);
+        } else {
+            //`template_ventry` is None, so `templates` has `t` and we never use the `HashSet::new()`
+            self.template_to_links_map
+                .entry(t.id().clone())
+                .or_insert_with(|| HashSet::new())
+                .insert(policy.id().clone());
         }
         if let Some(ventry) = link_ventry {
             ventry.insert(policy);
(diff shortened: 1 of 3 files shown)


andrewmwells-amazon:
We don't have `2.5.x` so I targeted this PR against `2.4.x` If we don't save it for 3.0, it should go into a minor.

andrewmwells-amazon:
Targeted at `3.0`

shaobo-he-aws (approved):

cedar-policy/src/api.rs
```diff
@@ -1709,6 +1726,14 @@ impl PolicySet {
         Ok(())
     }
 
+    /// Remove a `Template` from the `PolicySet`
+    pub fn remove_template(&mut self, template_id: PolicyId) -> Result<(), PolicySetError> {
```
aaronjeline, line 1730: We should be explicit in the documentation about what happens if you remove a template that has active links?

aaronjeline (changes requested):

cedar-policy/src/api.rs
```diff
@@ -1701,6 +1710,14 @@ impl PolicySet {
         }
     }
 
+    /// Remove a static `Policy` from the `PolicySet`
+    pub fn remove(&mut self, policy_id: PolicyId) -> Result<(), PolicySetError> {
```
john-h-kastner-aws, line 1714: Should these functions also return the `Policy` or `Template` they removed?

cedar-policy/src/api.rs
```diff
@@ -1595,6 +1595,15 @@ pub enum PolicySetError {
...
     ExpectedTemplate,
+    /// Error when removing a policy
+    #[error("unable to remove static policy: {0}")]
+    RemovePolicyError(PolicyId),
+    /// Error when removing a template
+    #[error("unable to remove policy template: {0}")]
+    RemoveTemplateError(PolicyId),
+    /// Error when unlinking a template
+    #[error("unable to unlink policy template: {0}")]
+    UnlinkingError(PolicyId),
```
cdisselkoen, line 1606: do we want more detail in these errors about _why_ the operation failed? like, "because it does not exist"?

cedar-policy/src/api.rs
```diff
@@ -1701,6 +1710,14 @@ impl PolicySet {
         }
     }
 
+    /// Remove a static `Policy` from the `PolicySet`
+    pub fn remove(&mut self, policy_id: PolicyId) -> Result<(), PolicySetError> {
+        match self.policies.remove(&policy_id) {
+            Some(_) => Ok(()),
+            None => Err(PolicySetError::RemovePolicyError(policy_id)),
+        }
```
cdisselkoen, line 1718: I think we need to also remove from `self.ast`.  (Also, do we have a test that would catch this? i.e., that confirms via an authz query that the policy was removed?)

cedar-policy/src/api.rs
```diff
@@ -1709,6 +1726,14 @@ impl PolicySet {
         Ok(())
     }
 
+    /// Remove a `Template` from the `PolicySet`
+    pub fn remove_template(&mut self, template_id: PolicyId) -> Result<(), PolicySetError> {
+        match self.templates.remove(&template_id) {
+            Some(_) => Ok(()),
+            None => Err(PolicySetError::RemoveTemplateError(template_id)),
+        }
```
cdisselkoen, line 1734: same here, I think we need to also remove from `self.ast` and add testing that authz queries are affected

cedar-policy/src/api.rs
```diff
@@ -1812,6 +1837,14 @@ impl PolicySet {
         Ok(())
     }
 
+    /// Unlink a dynamic policy from the policy set
+    pub fn unlink(&mut self, policy_id: PolicyId) -> Result<(), PolicySetError> {
+        match self.policies.remove(&policy_id) {
+            Some(_) => Ok(()),
+            None => Err(PolicySetError::UnlinkingError(policy_id)),
+        }
```
cdisselkoen, line 1845: same here, I think we need to also remove from `self.ast` and add testing that authz queries are affected

cdisselkoen (changes requested):

cedar-policy/src/api.rs
andrewmwells-amazon: I decided that should be an error.

cedar-policy-core/src/ast/policy_set.rs
```diff
@@ -187,6 +228,31 @@ impl PolicySet {
         }
     }
 
+    /// Remove a template from the policy set.
+    /// This involves an expensive scan over all linked policies to see if they are linked to the template
```
andrewmwells-amazon, line 232: Not sure if this is what we should finalize on, but it seems reasonable.

cedar-policy/src/api.rs
```diff
@@ -1701,6 +1710,26 @@ impl PolicySet {
...
+            None => return Err(PolicySetError::RemovePolicyError(policy_id)),
+        };
+        match self
+            .ast
+            .remove(&ast::PolicyID::from_string(policy_id.to_string()))
+        {
+            Ok(_) => Ok(policy),
+            Err(_) => {
+                //This should be unreachable.
+                debug_assert!(false);
```
andrewmwells-amazon, line 1726: I'll remove this if we all agree it's unreachable.

cedar-policy-core/src/ast/policy_set.rs
andrewmwells-amazon: @aaronjeline Not sure if this is what we want.

Pros: If don't allow you to remove templates with active links, people can't be confused about what will happen.

Cons: If `T1` has 0 active links and `T2` has 10,000 and you want to remove `T1`, you still need to look at 10,000 links.

cedar-policy-core/src/ast/policy_set.rs
```diff
@@ -96,6 +96,22 @@ pub enum PolicySetError {
     },
 }
 
+/// Potential errors when working with `PolicySet`s.
```
andrewmwells-amazon, line 99:
```suggestion
/// Potential errors when removing things from `PolicySet`s.
```

cedar-policy-core/src/ast/policy_set.rs
```diff
@@ -148,6 +164,31 @@ impl PolicySet {
         Ok(())
     }
 
+    /// Remove a `Policy`` from the `PolicySet`.
+    pub fn remove(&mut self, policy_id: &PolicyID) -> Result<Policy, PolicySetRemovalError> {
+        // Invariant: if `policy_id` is a key in both `self.links` and `self.templates`,
```
andrewmwells-amazon, line 169: TODO: double-check this

cedar-policy-core/src/ast/policy_set.rs
```diff
@@ -187,6 +228,31 @@ impl PolicySet {
...
+    pub fn remove_template(
+        &mut self,
+        policy_id: &PolicyID,
+    ) -> Result<Template, PolicySetRemovalError> {
+        if self
+            .links
+            .iter()
+            .any(|(_, p)| p.template().id().eq(policy_id))
+        {
+            return Err(PolicySetRemovalError::RemovePolicyNoTemplateError(
```
andrewmwells-amazon, line 243: Should have different error type

cedar-policy/src/api.rs
```diff
@@ -1595,6 +1595,15 @@ pub enum PolicySetError {
     /// Expected a template, but a static policy was provided.
     #[error("expected a template, but a static policy was provided")]
     ExpectedTemplate,
+    /// Error when removing a policy
+    #[error("unable to remove static policy: {0} because it does not exist")]
+    RemovePolicyError(PolicyId),
+    /// Error when removing a template
+    #[error("unable to remove policy template: {0} because it does not exist")]
```
andrewmwells-amazon, line 1602: Need new variant for templates with active links

cedar-policy/src/api.rs
```diff
@@ -3684,6 +3752,155 @@ mod policy_set_tests {
...
+        let response = authorizer.is_authorized(&request, &pset, &entities);
+        assert_matches!(
+            response,
+            Response {
+                decision: Decision::Allow,
+                diagnostics: _
+            }
+        );
+
+        assert_matches!(
```
andrewmwells-amazon, line 3839:
```suggestion
        //Can't call `remove` on a template
        assert_matches!(
```

cedar-policy-core/src/ast/policy_set.rs
aaronjeline: Yeah, what we really want is a two way map. :(

aaronjeline (approved):

cedar-policy-core/src/ast/policy_set.rs
```diff
@@ -39,6 +39,11 @@ pub struct PolicySet {
     ///   (this is managed by `PolicySet::add)
     /// A `Template` may have zero or many links
     links: HashMap<PolicyID, Policy>,
+
+    /// Map from a template `PolicyID`` to the set of `PolicyID`s in `links` that are linked to that template.
```
cdisselkoen, line 43:
```suggestion
    /// Map from a template `PolicyID` to the set of `PolicyID`s in `links` that are linked to that template.
```

cedar-policy-core/src/ast/policy_set.rs
```diff
@@ -57,7 +62,24 @@ impl TryFrom<LiteralPolicySet> for PolicySet {
...
+        for template in &templates {
+            template_to_links_map.insert(template.0.clone(), HashSet::new());
+        }
+        for linked_policy in &links {
+            let template = linked_policy.1.template().id();
+            match template_to_links_map.entry(template.clone()) {
+                Entry::Occupied(t) => t.into_mut().insert(linked_policy.0.clone()),
+                Entry::Vacant(_) => return Err(ReificationError::NoSuchTemplate(template.clone())),
+            };
+        }
```
cdisselkoen, line 76: readability nit:
```suggestion
        for (link_id, link) in &links {
            let template = link.template().id();
            match template_to_links_map.entry(template.clone()) {
                Entry::Occupied(t) => t.into_mut().insert(link_id.clone()),
                Entry::Vacant(_) => return Err(ReificationError::NoSuchTemplate(template.clone())),
            };
        }
```

cedar-policy-core/src/ast/policy_set.rs
```diff
@@ -96,6 +118,50 @@ pub enum PolicySetError {
...
+}
+
+/// Potential errors when removing templates from a `PolicySet`.
+#[derive(Error, Debug)]
+pub enum PolicySetTemplateRemovalError {
+    /// There was no [`PolicyID`] template in the list of templates.
+    #[error(
+        "unable to remove static policy id `{0}` from template list because it does not exist"
+    )]
+    RemovePolicyNoTemplateError(PolicyID),
```
cdisselkoen, line 144: unable to remove a static policy, or unable to remove a template? the error message and error variant name don't seem to match

cedar-policy-core/src/ast/policy_set.rs
```diff
@@ -181,12 +298,58 @@ impl PolicySet {
...
+            None => panic!("Found in template_to_links_map but not in templates"),
+        }
+    }
+
+    /// Get the list of policies linked to `template_id`.
+    /// Returns all p in `links` s.t. `p.template().id() == template_id`
+    pub fn get_linked_policies(
+        &mut self,
+        template_id: &PolicyID,
+    ) -> Result<Vec<&PolicyID>, PolicySetGetLinksError> {
```
cdisselkoen, line 346: instead of `Vec<&PolicyID>`, we could slightly more flexibly/performantly return `impl Iterator<Item = &PolicyID>`; this means we wouldn't have to `collect()` (and thus allocate) in the `Ok` case

cedar-policy/src/api.rs
```diff
@@ -1606,6 +1606,18 @@ pub enum PolicySetError {
     /// Expected a template, but a static policy was provided.
     #[error("expected a template, but a static policy was provided")]
     ExpectedTemplate,
+    /// Error when removing a static policy
+    #[error("unable to remove static policy: {0} because it does not exist")]
```
cdisselkoen, line 1610:
```suggestion
    #[error("unable to remove static policy `{0}` because it does not exist")]
```

cedar-policy/src/api.rs
```diff
@@ -1606,6 +1606,18 @@ pub enum PolicySetError {
...
+    PolicyNonexistentError(PolicyId),
+    /// Error when removing a template that doesn't exist
+    #[error("unable to remove policy template: {0} because it does not exist")]
```
cdisselkoen, line 1613:
```suggestion
    #[error("unable to remove policy template `{0}` because it does not exist")]
```

cedar-policy/src/api.rs
```diff
@@ -1606,6 +1606,18 @@ pub enum PolicySetError {
...
+    TemplateNonexistentError(PolicyId),
+    /// Error when removing a template with active links
+    #[error("unable to remove policy template: {0} because it has active links")]
```
cdisselkoen, line 1616:
```suggestion
    #[error("unable to remove policy template `{0}` because it has active links")]
```

cedar-policy/src/api.rs
```diff
@@ -1606,6 +1606,18 @@ pub enum PolicySetError {
...
+    RemoveTemplateWithActiveLinksError(PolicyId),
+    /// Error when unlinking a template
+    #[error("unable to unlink policy template: {0} because it does not exist")]
```
cdisselkoen, line 1619:
```suggestion
    #[error("unable to unlink policy template `{0}` because it does not exist")]
```

cedar-policy/src/api.rs
```diff
@@ -1720,6 +1751,47 @@ impl PolicySet {
...
+            }
+        }
+    }
+
+    /// Get policies linked to a `Template` in the `PolicySet`.
+    /// If any policy is linked to the template, this will error
+    pub fn get_linked_policies(
+        &mut self,
+        template_id: PolicyId,
+    ) -> Result<Vec<PolicyId>, PolicySetError> {
```
cdisselkoen, line 1785: same here we could return `Iterator<Item = PolicyId>` to avoid the `collect()`

cedar-policy/src/api.rs
```diff
@@ -1851,6 +1923,25 @@ impl PolicySet {
         entity_uids
     }
 
+    /// Unlink a dynamic policy from the policy set
```
cdisselkoen, line 1926:
```suggestion
    /// Unlink a template link from the policy set.
    /// Returns the policy that was unlinked.
```

cdisselkoen (approved):

cedar-policy/src/api.rs
andrewmwells-amazon: Doing this above is fine, but here we convert from `PolicyID` to `PolicyId` so I think we'd need to introduce a new type `PolicyIdRef(&PolicyID)` to handle this, which doesn't seem worth it to avoid the `.collect()` to me.

cdisselkoen: That is fine

cdisselkoen: Or, (not exactly seeing your code) does it work to use `.ref_cast()` to turn `&PolicyID` into `&PolicyId`?

cdisselkoen: (as we do for other types many other places in api.rs)

andrewmwells-amazon: Worked like a charm

aaronjeline (approved):

PR #340, opened 2023-10-05
Improving our panic audit

aaronjeline:
## Description of changes
Adds a new flag to the list of clippy flags, and the panic audit script.
## Issue #, if available
N/A

## Disclaimer

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
       - run: cargo test --verbose --features "experimental"
       - run: cargo test --verbose
       - run: cargo doc --all-features
-      - run: cargo clippy
+      - run: cargo clippy --all-features
       - run: ./panic_safety.sh
       - run: cargo test --verbose -- --ignored
       - run: cargo bench --no-run
--- a/cedar-policy-validator/src/schema_file_format.rs
+++ b/cedar-policy-validator/src/schema_file_format.rs
@@ -526,6 +526,8 @@ impl SchemaType {
 }
 
 #[cfg(feature = "arbitrary")]
+// PANIC SAFETY property testing code
+#[allow(clippy::panic)]
 impl<'a> arbitrary::Arbitrary<'a> for SchemaType {
     fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<SchemaType> {
         use cedar_policy_core::ast::Name;


cedar-policy-core/src/ast/request.rs
```diff
@@ -238,3 +246,14 @@ impl std::fmt::Display for Context {
...
+
+#[cfg(test)]
+mod test {
+
+    #[test]
+    fn test_json_from_str_non_record() {
+        let src = "1";
+        let context = super::Context::from_json_str(src);
+        let _vec: Vec<_> = context.iter().collect();
+    }
```
cdisselkoen, line 258: why doesn't this test fail? should it?

cedar-policy-core/src/entities/json/context.rs
```diff
@@ -80,6 +80,8 @@ impl<'e, 's, S: ContextSchema> ContextJsonParser<'e, 's, S> {
         let rexpr = vparser.val_into_rexpr(json, expected_ty.as_ref(), || {
             JsonDeserializationErrorContext::Context
         })?;
+        // INVARIANT `Context:from_exprs` requires that `rexpr` is a `Record`.
```
cdisselkoen, line 83:
```suggestion
        // INVARIANT `Context::from_exprs` requires that `rexpr` is a `Record`.
```

cedar-policy-validator/src/extensions/decimal.rs
```diff
@@ -21,9 +21,13 @@ use cedar_policy_core::evaluator::RestrictedEvaluator;
 use cedar_policy_core::extensions::{decimal, Extensions};
 use std::str::FromStr;
 
+/// Note on panic safety
+/// This module depends on the cedar parser only constructings ASTs with valid extension calls
```
cdisselkoen, line 25:
```suggestion
/// This module depends on the cedar parser only constructing ASTs with valid extension calls
```

cedar-policy-validator/src/extensions/ipaddr.rs
```diff
@@ -21,9 +21,13 @@ use cedar_policy_core::evaluator::RestrictedEvaluator;
 use cedar_policy_core::extensions::{ipaddr, Extensions};
 use std::str::FromStr;
 
+/// Note on panic safety
+/// This module depends on the cedar parser only constructings ASTs with valid extension calls
 /// If any of the panics in this file are triggered, that means that this file has become
 /// out-of-date with the ipaddr extension definition in CedarCore.
```
cdisselkoen, line 27: as-is, the `///` attaches this as a doc comment to `get_argument_types()`.  Maybe that's fine, in which case we don't need the extra backreference in the docs on `get_argument_types()`.  Or, we could use `//!` here to make a doc comment on the entire module; or, we could use `//` to make this block not a doc comment at all

cdisselkoen, line 27:
```suggestion
/// If any of the panics in this file are triggered, that means that this file has become
/// out-of-date with the ipaddr extension definition in `cedar-policy-core`.
```

cedar-policy-validator/src/extensions/partial_evaluation.rs
```diff
@@ -18,9 +18,13 @@ use crate::extension_schema::{ExtensionFunctionType, ExtensionSchema};
 use crate::types::{self, Type};
 use cedar_policy_core::extensions::partial_evaluation;
 
+/// Note on panic safety
+/// This module depends on the cedar parser only constructings ASTs with valid extension calls
 /// If any of the panics in this file are triggered, that means that this file has become
 /// out-of-date with the decimal extension definition in CedarCore.
```
cdisselkoen, line 24:
```suggestion
/// If any of the panics in this file are triggered, that means that this file has become
/// out-of-date with the decimal extension definition in `cedar-policy-core`.
```

cedar-policy-validator/src/extensions/partial_evaluation.rs
```diff
@@ -18,9 +18,13 @@ use crate::extension_schema::{ExtensionFunctionType, ExtensionSchema};
 use crate::types::{self, Type};
 use cedar_policy_core::extensions::partial_evaluation;
 
+/// Note on panic safety
```
cdisselkoen, line 21: same note as above about `///`

cedar-policy-validator/src/typecheck.rs
```diff
@@ -2135,13 +2148,16 @@ impl<'a> Typechecker<'a> {
 
     /// Utility called by the main typecheck method to handle extension function
     /// application.
+    /// INVARIANT `ext_expr` must be a `ExtensionFucntionApp`
```
cdisselkoen, line 2151:
```suggestion
    /// INVARIANT `ext_expr` must be a `ExtensionFunctionApp`
```

cedar-policy/src/api.rs
```diff
@@ -2205,6 +2216,7 @@ impl Policy {
     /// Get the head constraint on this policy's action
     pub fn action_constraint(&self) -> ActionConstraint {
         // Clone the data from Core to be consistant with the other constraints
+        // Invariant: all of the EntityUids come from a policy, which must have Concrete EntityUids
```
andrewmwells-amazon, line 2219: Invariant -> INVARIANT

cedar-policy-validator/src/extensions/ipaddr.rs
aaronjeline: I think module comment is the way to go, and is what I assumed it was doing when I edited to match the style of existing comment.

shaobo-he-aws (approved):

cedar-policy-validator/src/extensions/decimal.rs
```diff
@@ -13,6 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+//! Note on panic safety
+//! This module depends on the cedar parser only constructing ASTs with valid extension calls
+//! If any of the panics in this file are triggered, that means that this file has become
+//! out-of-date with the decimal extension definition in CedarCore.
```
john-h-kastner-aws, line 19: I don't think this depends on the AST at all.

The function `extension_schema` is the only place here the `get_argument_types`, `get_return_type` and `get_argument_check` functions are called where they are called on elements of `decimal::extension().funcs()`. So, the panic safety condition is that the we've accounted for every extension function in the match.

If we add a test that just executes `extension_schema()` we can guarantee that these panics don't occur at runtime. This is probably covered by tests elsewhere that construct the extension data structure, but we can retest here.

aaronjeline: Ok sounds good. I'll make that test

cedar-policy/benches/cedar_benchmarks.rs
```diff
@@ -23,6 +23,8 @@ use cedar_policy::{
 
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
 
+// PANIC SAFETY unit tests
+#[allow(clippy::unwrap_used)]
```
john-h-kastner-aws, line 27: Do you know why this annotation wasn't required before?

john-h-kastner-aws (approved):

cedar-policy/benches/cedar_benchmarks.rs
aaronjeline: `cargo clippy` from the cmdline runs without the `#[cfg(test)]` enabled. In VSCodee it runs with `#[cfg(test)]` enabled. Raises the point of making sure we test this in CI with all features, double checking on that now

john-h-kastner-aws (approved):

cedar-policy-core/src/ast/request.rs
```diff
@@ -164,13 +165,15 @@ impl Context {
...
+    /// INVARIANT: `from_expr` must only be called with an `expr` that is a `Record`
     pub fn from_expr(expr: RestrictedExpr) -> Self {
         debug_assert!(matches!(expr.expr_kind(), ExprKind::Record { .. }));
         Self { context: expr }
     }
 
     /// Create a `Context` from a map of key to `RestrictedExpr`, or a Vec of
     /// `(key, RestrictedExpr)` pairs, or any other iterator of `(key, RestrictedExpr)` pairs
+    // INVARIANT: always constructs a record
     pub fn from_pairs(pairs: impl IntoIterator<Item = (SmolStr, RestrictedExpr)>) -> Self {
```
cdisselkoen, line 177: in one of these, the `INVARIANT` is in the doc string, in the other it is not

cdisselkoen (approved):

cedar-policy-core/src/ast/request.rs
aaronjeline: I put it in a docstring when it's something the caller needs to consider

john-h-kastner-aws: I think the second isn't an invariant on the functions, it's just stating that the function maintains the datatype invariant.

PR #343, opened 2023-10-05
fix clippy

andrewmwells-amazon:
## Description of changes
Fix clippy errs for not using `cmp` in `partial_cmp` if the type also implements `Ord`. Also fix some warnings.

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/ast/value.rs
+++ b/cedar-policy-core/src/ast/value.rs
@@ -296,9 +296,7 @@ impl Eq for Set {}
 // HashSet doesn't implement PartialOrd
 impl PartialOrd<Set> for Set {
     fn partial_cmp(&self, other: &Set) -> Option<std::cmp::Ordering> {
-        self.authoritative
-            .as_ref()
-            .partial_cmp(other.authoritative.as_ref())
+        Some(self.cmp(other))
     }
 }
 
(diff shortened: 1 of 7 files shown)


cedar-policy-core/src/ast/value.rs
```diff
@@ -296,9 +296,7 @@ impl Eq for Set {}
 // HashSet doesn't implement PartialOrd
 impl PartialOrd<Set> for Set {
     fn partial_cmp(&self, other: &Set) -> Option<std::cmp::Ordering> {
-        self.authoritative
-            .as_ref()
-            .partial_cmp(other.authoritative.as_ref())
+        Some(self.cmp(other))
```
cdisselkoen, line 299: This implementation is here precisely because we want to compare on `authoritative`, not on all of `self` and `other`.  Why this change?

cdisselkoen (approved):

cedar-policy-core/src/ast/value.rs
andrewmwells-amazon: Clippy errors without this (See the build from Kesha's PR: https://github.com/cedar-policy/cedar/actions/runs/6421688661/job/17436607880?pr=341)

khieta (approved):

cedar-policy-core/src/ast/value.rs
shaobo-he-aws: I think Clippy just tries to ensure the `PartialOrd/Ord` rule: https://doc.rust-lang.org/std/cmp/trait.Ord.html. Calling `cmp` on `self` and `other` doesn't mean a structural comparison.

shaobo-he-aws (approved):

PR #345, opened 2023-10-06
Return `Err` from `Context::from_expr` on non-record expr

john-h-kastner-aws:
## Description of changes

A small change to make an invariant more robust.

The `Context` struct requires as an invariant that it may only contain a record expression, but `Context::from_expr` could construct contexts violating the invariant (although this fail a debug assertion).

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/ast/request.rs
+++ b/cedar-policy-core/src/ast/request.rs
@@ -164,11 +164,16 @@ impl Context {
         Self::from_pairs([])
     }
 
-    /// Create a `Context` from a `RestrictedExpr`, which must be a `Record`
-    /// INVARIANT: `from_expr` must only be called with an `expr` that is a `Record`
-    pub fn from_expr(expr: RestrictedExpr) -> Self {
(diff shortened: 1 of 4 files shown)


aaronjeline:
Two questions:
1) How does this manage to change the public API?
2) Instead of having `Context` hold a `RestrictedExpr`, can we have it hold a the pairs directly? That way we strip out all of the panics.

john-h-kastner-aws:
> Two questions:
> 
>   1. How does this manage to change the public API?

It doesn't.

>   2. Instead of having `Context` hold a `RestrictedExpr`, can we have it hold a the pairs directly? That way we strip out all of the panics.

This would be nice, but I didn't make that change because `Context` derives `Serialize` and `Deserialize`, and I think any change to the serialization would change the JSON accepted by `cedar_policy::Context::from_json_value` in the public API. If we're confident the change doesn't change the serialization then we could do it.

aaronjeline (approved):

shaobo-he-aws (approved):

cdisselkoen (approved):

PR #347, opened 2023-10-06
RFC 24 Prototype

shaobo-he-aws:
## Description of changes
This PR implements a parser and a pretty printer of the custom schema syntax proposed in https://github.com/cedar-policy/rfcs/blob/main/text/0024-schema-syntax.md.

## Disclaimer


andrewmwells-amazon (approved):

aaronjeline (approved):

john-h-kastner-aws:
Does this parser provide access to source ranges in a way that they could be stored on the parsed schema and used for error reporting later? This is lacking from the JSON schema format, and we have said before that we want to include this type of error reporting when we introduce the custom format.

shaobo-he-aws:
> Does this parser provide access to source ranges in a way that they could be stored on the parsed schema and used for error reporting later? This is lacking from the JSON schema format, and we have said before that we want to include this type of error reporting when we introduce the custom format.

It doesn't do it now but shouldn't be hard to implement. I'll make sure we have this feature for the final version.

cedar-policy-validator/src/schema_syntax.rs
```diff
@@ -0,0 +1,801 @@
...
+            optional(choice((
+                (accept(Token::Comma), AppParser()).map(|(_, cdr)| cdr),
+                accept(Token::Comma).map(|_| ApplySpec {
+                    resource_types: None,
+                    principal_types: None,
+                    context: AttributesOrContext::default(),
+                }),
+            ))),
+        )
+            .map(|(singleton, cdr)| {
```
john-h-kastner-aws, line 323: This logic is a bit confusing.

john-h-kastner-aws (commented):
Looks nice. The approach or parsing into the `SchemaFragment` and `NamespaceDefinition` is good. This should let us share the rest of the schema construction code.

The parser combinators library is nice, but I do think it's better to use lalrpop for the policy and schema grammar like we said earlier. Not sure what the performance implication are, but i wouldn't want to end up using two parser libraries.

Issue #358, opened 2023-10-11
Duplicate entries silently ignored in JSON `frontend` module functions


john-h-kastner-aws:
- [x] Duplicate policy IDs in policies
```rust
    #[test]
    fn test_authorized_fails_on_duplicate_policy_ids() {
        let call = r#"{
            "principal" : "User::\"alice\"",
            "action" : "Photo::\"view\"",
            "resource" : "Photo::\"door\"",
            "context" : {},
            "slice" : {
                "policies" : {
                  "ID0": "permit(principal, action, resource);",
                  "ID0": "permit(principal, action, resource);"
                },
                "entities" : [],
                "templates" : {},
                "template_instantiations" : [ ]
            }
        }"#;
        assert_is_failure(
            &json_is_authorized(call),
            true,
            "",
        );
    }
```

the same problem exists for `json_validate`

```rust
        #[test]
        fn test_validate_fails_on_duplicate_policy_id() {
            let call_json = r#"{
                "schema": { "": { "entityTypes": {}, "actions": {} } },
                "policySet": {
                  "ID0": "permit(principal, action, resource);",
                  "ID0": "permit(principal, action, resource);"
                }
            }"#
            .to_string();
            let result = json_validate(&call_json);
            assert_is_failure(
                &result,
                false,
                "error parsing call: invalid entry: found duplicate key",
            );
        }
```

- [x] Duplicate templates IDs in templates
```rust
    #[test]
    fn test_authorized_fails_on_duplicate_template_ids() {
        let call = r#"{
            "principal" : "User::\"alice\"",
            "action" : "Photo::\"view\"",
            "resource" : "Photo::\"door\"",
            "context" : {},
            "slice" : {
                "policies" : {},
                "entities" : [],
                "templates" : {
                    "ID0": "permit(principal == ?principal, action, resource);",
                    "ID0": "permit(principal == ?principal, action, resource);"
                },
                "template_instantiations" : [ ]
            }
        }"#;
        assert_is_failure(
            &json_is_authorized(call),
            true,
            "",
        );
    }
```

- [x] Duplicate slot instantiations (#352)
```rust
    #[test]
    fn test_authorized_fails_on_duplicate_slot_instantiation() {
        let call = r#"{
            "principal" : "User::\"alice\"",
            "action" : "Photo::\"view\"",
            "resource" : "Photo::\"door\"",
            "context" : {},
            "slice" : {
                "policies" : {},
                "entities" : [],
[...]

Issue #352, opened 2023-10-09
JSON `is_authorized` API silently ignored duplicate template slot values


jamiewinder:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Cedar Parser

### Describe the bug

I'm using Cedar through the JSON API (`json_is_authorized`). I'm not sure if the normal API is affected.

This is an example request:
```json
{
        "principal":"User::\"1\"",
        "action":"Action::\"View\"",
        "resource":"Document::\"123\"",
        "context":{},
        "schema":null,
        "slice":{
           "entities":[],
           "policies": {},
           "templates":{
              "test_template":"permit(\r\n  principal in ?principal, \r\n  action, \r\n  resource\r\n);"
           },
           "template_instantiations":[
              {
                 "template_id":"test_template",
                 "result_policy_id":"test_template_result",
                 "instantiations":[
                    {
                       "slot":"?principal",
                       "value":{
                          "ty":"User",
                          "eid":"2"
                       }
                    },
                    {
                       "slot":"?principal",
                       "value":{
                          "ty":"User",
                          "eid":"1"
                       }
                    }
                 ]
              }
           ]
        }
     }
```
The result of this is Allow. Note there are two `template_instantiations`. If these are swapped around, the result becomes Deny.

It appears the last template instantiation overwrites any previous ones?

### Expected behavior

Multiple template instantiations should not overwrite each other.

### Reproduction steps

Include multiple template instantiations.

### Log output

```
// Put your output below this line

```

### Operating  System

Windows

andrewmwells-amazon:
Thanks for the issue report. I think the current behavior is intentional but confusing; As you noted, we take the last instantiation. This matches how we handle duplicate record keys (https://github.com/cedar-policy/rfcs/blob/main/text/0020-unique-record-keys.md). We'll look into changing this because the current behavior is confusing and it's unclear when it would be desired.

jamiewinder:
Thanks for your reply. I'd assume this behaviour isn't found in the non-JSON API? 

[...]

PR #365, opened 2023-10-16
Bugfix/aaronjeline/358

aaronjeline:
## Description of changes
Moves our parsers to reject duplicate keys, instead of having last-write-wins semantics.
Adds a new dependency on `serde-with`
## Issue #, if available
#358 and #352 

## Disclaimer


cedar-policy-cli/src/lib.rs
```diff
@@ -935,25 +928,6 @@ fn read_schema_file(filename: impl AsRef<Path> + std::marker::Copy) -> Result<Sc
         })
 }
 
-fn load_actions_from_schema(entities: Entities, schema: &Option<Schema>) -> Result<Entities> {
```
aaronjeline, line 938: Removing this function has it will be incorporated into the main cedar library, and we want a very specific error semantics for it

cedar-policy/tests/example_use_cases_doc.rs
```diff
@@ -68,9 +68,7 @@ fn scenario_4a() {
 // note: 4b currently omitted because it requires date/timestamp functionality
 
 /// currently failing, as the validator does not support action attributes
-#[should_panic(
-    expected = "error occurred while evaluating policy `policy0`: entity `Action::\\\"view\\\"` does not exist"
-)]
+#[should_panic(expected = "Action::\\\"view\\\"` does not have the attribute `readOnly`")]
```
aaronjeline, line 71: Actually having the entity specifies changes the error message.

cedar-policy-cli/src/lib.rs
cdisselkoen: This overlaps with my #360

cedar-policy/tests/example_use_cases_doc.rs
cdisselkoen: This overlaps with my #360

aaronjeline: Not surprising. We'll want to make sure yours wins.

cdisselkoen: but I think it's the same change, so we should be able to accept both in either order

cedar-policy-cli/src/lib.rs
cdisselkoen: but I think it's the same change, so we should be able to accept both in either order

cedar-policy-core/src/entities/json/entities.rs
```diff
@@ -417,3 +417,41 @@ impl EntityJSON {
...
+                },
+                "attrs" : {},
+                "parents": []
+            }
+        ]);
+        let eparser: EntityJsonParser<'_, NoEntitiesSchema> =
+            EntityJsonParser::new(None, Extensions::all_available(), TCComputation::ComputeNow);
+        let e = eparser.from_json_value(json).err().unwrap();
+        let bad_euid: EntityUID = r#"User::"alice""#.parse().unwrap();
+        match e {
```
john-h-kastner-aws, line 450: You can do something nice like
```
assert_matches(e, Err(EntitiesError::Duplicate(euid)) => {
  assert_eq!(bad_euid, euid, r#"Returned euid should be User::"alice""#);
});
```

for this sort of assertion

cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -175,7 +181,9 @@ impl AuthorizationCall {
             None => None,
         };
 
-        let context = Context::from_json_value(self.context, schema.as_ref().map(|s| (s, &action)))
+        let context = serde_json::to_value(self.context)
+            .map_err(|e| [format!("Error encoding the context as JSON: {e}")])?;
```
john-h-kastner-aws, line 185: This change should be made to the `Context` constructor.

cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -228,62 +277,62 @@ struct RecvdSlice {
     template_instantiations: Option<Vec<TemplateLink>>,
 }
 
-impl RecvdSlice {
```
john-h-kastner-aws, line 231: Is all this whitespace change to split up the function for the clippy lint?

cedar-policy/src/frontend/utils.rs
```diff
@@ -20,11 +20,13 @@ use std::collections::HashMap;
 
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(untagged)]
+#[serde(expecting = "policy map must have no duplicate IDs,")]
```
john-h-kastner-aws, line 23: I expect this will give a bad error message if the policy specification happens to be an int or null or something else. Maybe best to spell it out "policies as a concatenated string or multiple policies as a hashmap where the policy Id is the key with no duplicate IDs". Although maybe that's a bit wordy.

cedar-policy/src/frontend/is_authorized.rs
aaronjeline: To make it more readable, it was a bunch of nested functions.

cedar-policy/CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 - The `Response::new()` constructor now expects a `Vec<AuthorizationError>` as its third argument.
 - Implements RFC #19, making validation slightly more strict, but more explainable.
 - Improved formatting for error messages.
+- Standardized on duplicates being errors instead of last-write-wins for parsers.
```
cdisselkoen, line 25: I think we should be slightly more specific -- maybe list the affected APIs?

cedar-policy/src/frontend/is_authorized.rs
cdisselkoen: are we serializing here or deserializing here?

cdisselkoen (approved):

cedar-policy/src/frontend/is_authorized.rs
aaronjeline: De-serialized to a `map` so that we can use an annotation to fail on duplicates. Before we pass back to Context constructor we have to serialize back to a json-value. If we want to forbid duplicates in the context contructor itself, than this api can't just take a json value.

aaronjeline:
Blocked on: https://github.com/cedar-policy/cedar-java/pull/53

andrewmwells-amazon (approved):

PR #372, opened 2023-10-19
Improvement/aaronjeline/infallible parsing

aaronjeline:
## Description of changes
Parsing eids always suceeds, so it's `Error` type should be `Infallible`, not `ParseErrors`

## Disclaimer

### DIFF
--- a/cedar-policy/CHANGELOG.md
+++ b/cedar-policy/CHANGELOG.md
@@ -32,6 +32,7 @@
 	+ Template instantiation records
 	+ Entity slice JSONs
 	+ Context JSONs
+- `EntityId::FromStr::Error` is now `Infallible` instead of `ParseErrors`
 
 ## 2.4.1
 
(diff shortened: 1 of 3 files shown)


cedar-policy/CHANGELOG.md
```diff
@@ -32,6 +32,7 @@
 	+ Template instantiation records
 	+ Entity slice JSONs
 	+ Context JSONs
+- `EntityId::FromStr::Error` is now `Infallible` instead of `ParseErrors`
```
cdisselkoen, line 35:
```suggestion
- `<EntityId as FromStr>::Error` is now `Infallible` instead of `ParseErrors`
```

cdisselkoen (commented):
do we want to commit to this always being infallible? do we envision a future where it might fail, e.g., we prohibit unmatched BIDI characters, or apply confusable_string_checker, or impose a length limit?

aaronjeline:
> do we want to commit to this always being infallible? do we envision a future where it might fail, e.g., we prohibit >unmatched BIDI characters, or apply confusable_string_checker, or impose a length limit?

That would be a breaking change, so we could change the type back in the same version. In that case people who were depending on the current behavior would get a compile time notice that it the behavior changed. Right now you can very reasonable call `.unwrap()` on parsing eids, and you'd get no warning if we changed the behavior.

cdisselkoen (approved):

john-h-kastner-aws (approved):

PR #375, opened 2023-10-19
Implement RFC 20

cdisselkoen:
## Description of changes

Disallows duplicate keys in record values (including record literals in policies, request `context`, and records in entity attributes).

Includes the changes in #169 (and thus subsumes #169) -- it is now both possible and desirable to store records as maps in the AST, and internally expresses the no-duplicate-keys invariant.  (The EST already stored records as maps, which was a discrepancy between EST and AST semantics noted in #167.)

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -416,6 +416,12 @@ impl Expr {
     }
 
     /// Create an `Expr` which evaluates to a Record with the given key-value mapping.
+    ///
+    /// If you have an iterator of pairs, generally prefer calling
+    /// `Expr::record()` instead of `.collect()`-ing yourself and calling this,
+    /// potentially for efficiency reasons but also because `Expr::record()`
+    /// will properly handle duplicate keys but your own `.collect()` will not
+    /// (by default).
     pub fn record_arc(map: Arc<BTreeMap<SmolStr, Expr>>) -> Self {
         ExprBuilder::new().record_arc(map)
     }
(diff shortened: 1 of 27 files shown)


cedar-policy-core/src/est/expr.rs
```diff
@@ -536,59 +542,64 @@ impl TryFrom<Expr> for ast::Expr {
...
-            )),
-            Expr::ExprNoExt(ExprNoExt::Record(map)) => Ok(ast::Expr::record(
-                map.into_iter()
-                    .map(|(k, v)| Ok((k, v.try_into()?)))
-                    .collect::<Result<HashMap<SmolStr, _>, Self::Error>>()?,
+                    .map(|el| el.try_into_ast(id.clone()))
+                    .collect::<Result<Vec<_>, FromJsonError>>()?,
             )),
+            Expr::ExprNoExt(ExprNoExt::Record(map)) => {
+                // PANIC SAFETY: can't have duplicate keys here because the input was already a HashMap
```
john-h-kastner-aws, line 591: Most of the `expect`s here are asserting that `.iter().map()` (and similar operation on `unzip()` or on `keys()` and `values()`) affecting only values without touching keys cannot introduce duplicates. Maybe this could be abstracted into a utility function.

cedar-policy-core/src/ast/expr.rs
```diff
@@ -1052,10 +1049,29 @@ impl<T> ExprBuilder<T> {
...
+                btree_map::Entry::Vacant(ventry) => {
+                    ventry.insert(v);
+                }
+            }
+        }
+        Ok(self.with_expr_kind(ExprKind::Record(Arc::new(map))))
+    }
+
+    /// Create an `Expr` which evalutes to a Record with the given key-value mapping.
+    pub fn record_arc(self, map: Arc<BTreeMap<SmolStr, Expr<T>>>) -> Expr<T> {
```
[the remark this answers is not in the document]
cdisselkoen: Added note here (and in the other place I added `record_arc()`)

john-h-kastner-aws (approved):

cedar-policy-core/src/est/expr.rs
cdisselkoen: I think adding abstraction is just as likely to obscure potential bugs.  With this design, although we have more `expect`s, each `expect` can be verified with local reasoning about the ~10 lines of code surrounding it

andrewmwells-amazon (approved):

PR #381, opened 2023-10-24
Fix `cedar_policy_core::PrincipalConstraint::is_in_slot()`

john-h-kastner-aws:
## Description of changes

Fixes a bug in `cedar_policy_core::PrincipalConstraint::is_in_slot()` where the function was typo-ed to call `PrincpalOrResourceContraint::is_eq_slot()` instead of `is_in_slot()`. The function is only called by our test code, so the bug doesn't effect anything else. 

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/ast/policy.rs
+++ b/cedar-policy-core/src/ast/policy.rs
@@ -1011,7 +1011,7 @@ impl PrincipalConstraint {
     /// Hierarchical constraint to Slot
     pub fn is_in_slot() -> Self {
         Self {
-            constraint: PrincipalOrResourceConstraint::is_eq_slot(),
+            constraint: PrincipalOrResourceConstraint::is_in_slot(),
         }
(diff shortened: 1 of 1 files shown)


shaobo-he-aws (approved):

cdisselkoen (approved):

cdisselkoen:
If I'm seeing correctly, it was only called in the test `ir_binding()` in `ast/policy.rs`.  Nonetheless, why didn't that test fail because of this?  do we know?  should it have failed?

john-h-kastner-aws:
> If I'm seeing correctly, it was only called in the test `ir_binding()` in `ast/policy.rs`. Nonetheless, why didn't that test fail because of this? do we know? should it have failed?

It doesn't seem like it was intended to test this function. It just needs to construct an arbitrary policy with a template slot (which this function did do).

PR #386, opened 2023-10-26
Use cedar-policy `README.md` as crate top-level documentation

### LINKED ISSUE #362: docs.rs documentation improvement
### Category

Documentation and code comments

On the Rust crate documentation (e.g., https://docs.rs/cedar-policy/latest/cedar_policy/) there are no usage examples on the landing pages, just snippet examples for the particular API calls. When you get to the landing page, you are looking for help, but instead you get a long list of types and API calls with no explanation.

I like the worked example we have on the [crates.io page for cedar-policy](https://crates.io/crates/cedar-policy); could we something like that for our various packages, too? I see this is done for other Rust packages, e.g., https://docs.rs/graph/latest/graph/ and https://docs.rs/bigdecimal/latest/bigdecimal/ as two random cases.

Not really. Doing nothing = status quo.

### Additional context

_No response_

john-h-kastner-aws:
Edit: updated to make it top level docs.

---

It's a little bit hacky, but this is the simplest way I found to run the actual code that's in the readme. ([See the test execute here](https://github.com/cedar-policy/cedar/actions/runs/6655106480/job/18084606177?pr=386#step:7:1502))

A nicer option that avoids adding the dummy function could be running `rustdoc --test README.md` in CI, but `rustdoc` doesn't resolve the `cedar_policy::*` import, and `cargo doc --test` doesn't exist.

Another alternative could be using `#![doc = include_str!("../README.md")]` as the top level crate documentation. This would resolve the docs issue #362 and test the example at the same time, if were happy to use the exact same text in the README and crate docs.

## Disclaimer

### DIFF
--- a/cedar-policy/src/lib.rs
+++ b/cedar-policy/src/lib.rs
@@ -39,3 +39,10 @@ pub mod frontend;
 
 #[cfg(feature = "integration_testing")]
 pub mod integration_testing;
+
+/// Includes README.md as the documentation for this unused empty function so
+/// that the example usage in the README is executed as a doctest,
+/// protecting against accidental breakage of the example.
+#[doc = include_str!("../README.md")]
+#[allow(dead_code)]
+fn readme_as_doc_test() {}
--- a/cedar-policy/src/lib.rs
+++ b/cedar-policy/src/lib.rs
@@ -20,7 +20,6 @@
 // to add a separate test specifically for README examples by introducing a
 // private, empty, and unused function with `#[doc = include_str!("../README.md")]`.
 #![doc = include_str!("../README.md")]
-
 #![forbid(unsafe_code)]
 #![warn(rust_2018_idioms, clippy::pedantic, clippy::nursery)]
 #![deny(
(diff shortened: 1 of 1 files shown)


khieta:
Thanks @john-h-kastner-aws! We should definitely be making sure that our README example compiles.

I like your `#![doc = include_str!("../README.md")]` alternative. What are the downsides of that?

john-h-kastner-aws:
> What are the downsides of that?

Mainly that some part of the readme might not be useful to some one looking at the crates docs like the Documentation, Building, and Contributing sections. There could be further drift over time, but if we're happy using the current readme in both places, then we could address any other issues later.

khieta:
I think we should just go for it. Even if the README has some info that is not relevant on crates.io, at least it will be maintained and the two sources will stay in sync.

cdisselkoen:
Yeah it's definitely an improvement on the status quo for very little effort

john-h-kastner-aws:
Looks nice!

image

khieta (approved):

PR #389, opened 2023-10-26
split `cedar_policy_validator::schema` into some submodules

cdisselkoen:
## Description of changes

`schema.rs` is quite large, and it can get hard to find what you're looking for, or for relative newcomers to the codebase to understand what's related to each other.  This PR splits some related groups of types/definitions into submodules.  Feel free to critique my choice of what goes together into a submodule.

Organization this PR proposes:

* split `ValidatorNamespaceDef`, and the types that make up components of it, into their own submodule, as these are somewhat separate from the `ValidatorSchema` types in the same file
* For each of the two main components of a `ValidatorSchema` -- `ValidatorEntityType` and `ValidatorActionId` -- split them into their own submodules, along with the types that make up components of those
* `ValidatorSchema` itself, and `ValidatorSchemaFragment`, remain in `schema.rs` directly
* wasn't sure where to put `HeadVar` and `ActionHeadVar`, but since they are used in methods on `ValidatorSchema`, I left them in `schema.rs` as well

## Disclaimer


aaronjeline (approved):

john-h-kastner-aws:
Looks fine. There are probably some other things that would make sense in the new files, but that might require some changes to the code.

john-h-kastner-aws (approved):

PR #390, opened 2023-10-26
Bug fix for ipaddr

shaobo-he-aws:
## Description of changes
The parser used to leverage `u8` parser to parse prefixes, which allows numbers like `+1` and `001`. And implementation of `isLoopback` and `isMulticast` is not correct.

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/extensions/ipaddr.rs
+++ b/cedar-policy-core/src/extensions/ipaddr.rs
@@ -100,8 +100,8 @@ impl IPAddr {
     /// Return true if this is a loopback address
     fn is_loopback(&self) -> bool {
         // Loopback addresses are "127.0.0.0/8" for IpV4 and "::1" for IpV6
-        // If `addr` is a loopback address, its prefix is `127` or `1`
-        // We need to just make sure the prefix is greater than or equal to `8` or `128`
+        // If `addr` is a loopback address, its prefix is `0x7f` or `0x00000000000000000000000000000001`
+        // We need to just make sure the prefix length (i.e., `prefix`) is greater than or equal to `8` or `128`
         self.addr.is_loopback() && self.prefix >= if self.is_ipv4() { 8 } else { PREFIX_MAX_LEN_V6 }
     }
 
@@ -756,6 +756,13 @@ mod tests {
             eval.interpret_inline_policy(&Expr::call_extension_fn(
                 Name::parse_unqualified_name("isLoopback").expect("should be a valid identifier"),
                 vec![ip("::2")]
+            )),
+            Ok(Value::from(false))
+        );
+        assert_eq!(
+            eval.interpret_inline_policy(&Expr::call_extension_fn(
+                Name::parse_unqualified_name("isLoopback").expect("should be a valid identifier"),
+                vec![ip("127.255.200.200/0")]
             )),
             Ok(Value::from(false))
         );
--- a/cedar-policy-core/src/extensions/ipaddr.rs
+++ b/cedar-policy-core/src/extensions/ipaddr.rs
@@ -100,20 +100,15 @@ impl IPAddr {
     /// Return true if this is a loopback address
     fn is_loopback(&self) -> bool {
         // Loopback addresses are "127.0.0.0/8" for IpV4 and "::1" for IpV6
-        // Unlike the implementation of `is_multicast`, we don't need to test prefix
-        // The reason for IpV6 is obvious: There's only one loopback address
-        // The reason for IpV4 is that provided the truncated ip address is a loopback address, its prefix cannot be less than 8 because otherwise its more significant byte cannot be 127
-        self.addr.is_loopback()
+        // If `addr` is a loopback address, its prefix is `127` or `1`
+        // We need to just make sure the prefix is greater than or equal to `8` or `128`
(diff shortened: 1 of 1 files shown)


aaronjeline (approved):

cedar-policy-core/src/extensions/ipaddr.rs
```diff
@@ -100,75 +114,112 @@ impl IPAddr {
...
         match (&self.addr, &other.addr) {
             (std::net::IpAddr::V4(self_v4), std::net::IpAddr::V4(other_v4)) => {
-                let netmask = |subnet: u8| u32::MAX.checked_shl(32 - subnet as u32).unwrap_or(0);
-                let hostmask = |subnet: u8| u32::MAX.checked_shr(subnet as u32).unwrap_or(0);
+                let netmask = |prefix: u8| {
+                    u32::MAX
+                        .checked_shl((PREFIX_MAX_LEN_V4 - prefix).into())
+                        .unwrap_or(0)
+                };
+                let hostmask = |prefix: u8| u32::MAX.checked_shr(prefix as u32).unwrap_or(0);
```
cdisselkoen, line 129: why do we use `.into()` in one case, but `as u32` in the other case, to convert to `u32`?  I feel like being consistent would make this easier to read

cedar-policy-core/src/extensions/ipaddr.rs
```diff
@@ -100,75 +114,112 @@ impl IPAddr {
...
             }
             (std::net::IpAddr::V6(self_v6), std::net::IpAddr::V6(other_v6)) => {
-                let netmask = |subnet: u8| u128::MAX.checked_shl(128 - subnet as u32).unwrap_or(0);
-                let hostmask = |subnet: u8| u128::MAX.checked_shr(subnet as u32).unwrap_or(0);
+                let netmask = |prefix: u8| {
+                    u128::MAX
+                        .checked_shl((PREFIX_MAX_LEN_V6 - prefix).into())
+                        .unwrap_or(0)
+                };
+                let hostmask = |prefix: u8| u128::MAX.checked_shr(prefix as u32).unwrap_or(0);
```
cdisselkoen, line 143: same comment here

cedar-policy-core/src/extensions/ipaddr.rs
```diff
@@ -100,75 +114,112 @@ impl IPAddr {
...
+        return Err(format!("error parsing prefix: encountered non-digit"));
+    }
+    if s.starts_with('0') && s != "0" {
+        return Err(format!("error parsing prefix: leading zero(s)"));
+    }
+    let res: u8 = s
+        .parse()
+        .map_err(|err| format!("error parsing prefix from the string `{s}`: {err}"))?;
+    if res > max {
+        return Err(format!("error parsing prefix: too large"));
```
cdisselkoen, line 170:
```suggestion
        return Err(format!("error parsing prefix: {res} is larger than the limit {max}"));
```

cdisselkoen (approved):

cedar-policy-core/src/extensions/ipaddr.rs
```diff
@@ -86,89 +100,123 @@ impl IPAddr {
     /// Return true if this is a loopback address
     fn is_loopback(&self) -> bool {
         // Loopback addresses are "127.0.0.0/8" for IpV4 and "::1" for IpV6
-        // Unlike the implementation of `is_multicast`, we don't need to test prefix
-        // The reason for IpV6 is obvious: There's only one loopback address
-        // The reason for IpV4 is that provided the truncated ip address is a loopback address, its prefix cannot be less than 8 because otherwise its more significant byte cannot be 127
-        self.addr.is_loopback()
+        // If `addr` is a loopback address, its prefix is `127` or `1`
+        // We need to just make sure the prefix is greater than or equal to `8` or `128`
```
cdisselkoen, line 104: I think this comment mixes two different meanings of the word "prefix"?

cedar-policy-core/src/extensions/ipaddr.rs
```diff
@@ -86,89 +100,123 @@ impl IPAddr {
...
+        self.addr.is_loopback() && self.prefix >= if self.is_ipv4() { 8 } else { PREFIX_MAX_LEN_V6 }
```
cdisselkoen, line 105: why `8` in the V4 case but `PREFIX_MAX_LEN_V6` in the V6 case?

cdisselkoen (changes requested):
Can we add a couple tests for `isLoopback` that capture the problem we're fixing?

cedar-policy-core/src/extensions/ipaddr.rs
shaobo-he-aws: Yes. I'll fix it.

cedar-policy-core/src/extensions/ipaddr.rs
shaobo-he-aws: Because the prefix length of loopback addresses for V4 is 8 and there's only one V6 loopback address.

cdisselkoen: Makes sense.

cdisselkoen (approved):

PR #393, opened 2023-10-30
Optionally validate requests

### LINKED ISSUE #191: New API for validating requests
### Category

User level API changes

The content of this issue was migrated from [RFC#11](https://github.com/cedar-policy/rfcs/pull/11).

### Overview

Given a schema, _validation_ checks that policies conform to the schema, and _schema-based parsing_ checks that entities and context conform to the schema. This issue proposes to add support for checking that a request conforms to a schema.

### Example

Say the schema specifies that an action `readFile` should only be used with principals of type `User` and resources of type `File`. This RFC proposes to add a new constructor for the `Request` type, which takes a schema argument that is used to validate fields when constructing a `Request`.

Examples (using informal syntax for EUIDs):

```rust
Request::new(Some(principal), Some(action), Some(resource), context); // original API; unchanged by this RFC

Request::new_with_validation(Some(User::"alice"), Some(Action::"readFile"), Some(File::"secret_file.txt"), context, schema); // returns a Request

Request::new_with_validation(Some(User::"alice"), Some(Action::"readFile"), Some(Folder::"some_folder"), context, schema); // returns an error (invalid resource)
```

See [RFC#11](https://github.com/cedar-policy/rfcs/pull/11) for additional motivation and examples.

Add a new constructor for the `Request` type, which takes a schema argument that is used to validate fields when constructing a `Request`. More concretely, the new API will have the following signature:

```rust
pub fn new_with_validation(
        principal: Option<EntityUid>,
        action: Option<EntityUid>,
        resource: Option<EntityUid>,
        context: Context,
        schema: &Schema
    ) -> Result<Request> {
        ...
    }
```

(This is the same as the signature for `Request::new`, aside from the schema argument and return type.)

If the action is `None` (indicating that it is "unspecified"), then the function performs no additional checks. If the action is `Some`, then the function checks that the specified action is present in the schema, and that the principal and resource are consistent with the action's `appliesTo` lists in the schema.

See discussion on [RFC#11](https://github.com/cedar-policy/rfcs/pull/11).

### Additional context

_No response_

cdisselkoen:
## Description of changes

Adds `Request::new_with_validation()` to the public API.

## Issue #, if available

Fixes #191

## Disclaimer


khieta (commented):
Overall changes look good to me, but I'd like to relitigate the decision to add a new `Request::new_with_validation` API vs. adding an optional schema argument to `Request::new`. I feel that the latter is more consistent with our choice for the `Entities` constructors (#360), although I do concede that it is a more invasive change for users.

khieta:
Another comment I forgot to add: might be useful to add test cases for the new `validate_request` function.

cdisselkoen:
Added a bunch of tests (and did some reorg while I was at it).  In the process I fixed a (bug?) in entity validation that exists on the 2.4.x release branch.

> I'd like to relitigate the decision to add a new `Request::new_with_validation` API vs. adding an optional schema argument to `Request::new`.

This PR follows #191, which resulted from discussion in [RFC 11](https://github.com/cedar-policy/rfcs/pull/11).  I'm happy to relitigate this, but we should make sure to include the points discussed on 191 and on RFC 11, and to make sure everyone relevant is included in the discussion.

cdisselkoen:
Recapping the RFC 11 discussion as I see it, along with the above point:

Let's call the solution described in #191, with a separate `Request::new_with_validation()` constructor, the "two-constructor solution", and the alternative, where we add an optional schema parameter to `Request::new()`, the "one-constructor-solution". 

* The two-constructor solution is back-compatible, while the one-constructor solution breaks existing users. 
* @mwhicks1 argues for the two-constructor solution "because I don't want to break existing users, and because I expect that the non-validation path is the common one. People will largely use the validation path for debugging."
* @khieta argues that the one-constructor solution does break existing users, but the benefit is that this will force them to be aware of the new feature and start validating their requests.
* As precedent for the two-constructor solution, we already have `Request::new_with_unknowns()` for partial evaluation.  We did not break `Request::new()` when adding partial evaluation.  (Note that partial evaluation APIs are still experimental, and we could still change them)
* As precedent for the one-constructor solution, we decided to add an optional schema argument to some `Entities` methods in #360, breaking existing users of those methods.
* @john-h-kastner-aws said at the time "somewhat in favor of making the breaking change"
* I said at the time "Same. It seems like a worthwhile change to force people to adapt to when updating to Cedar 3.x."
* @ChunghaSung's comment on #191 supports request-validation-by-default, finding the current default (no request validation) to be unintuitive.  I interpret this as support of the one-constructor solution.

mwhicks1:
I could be persuaded that one constructor is Ok, on the basis that old code will not typecheck, and the fix to get it typecheck (include `None` as the second argument) is easy.

I wonder: Is it easy/possible to build tools that might carry out such a change automatically? We have other APIs we might like to change in ways that are not back-compatible, but are easy to migrate.

cdisselkoen:
Another point is that, as you can see in the current version of this PR, the two-constructor solution really becomes a four-constructor solution when you consider partial evaluation: we have `Request::new()`, `Request::new_with_unknowns()`, `Request::new_with_validation()`, and `Request::new_with_unknowns_and_validation()`.  This is a little unwieldy IMO.  And it definitely doesn't scale to any future features/options we might want to add to `Request` construction.

andrewmwells-amazon:
I would vote for the one-constructor solution. 

> Is it easy/possible to build tools that might carry out such a change automatically?

In Jetbrains products you can do "Change Method Signature" (tested with RustRover). It looks like RustAnalyzer has an issue for this (https://github.com/rust-lang/rust-analyzer/issues/6499) but it isn't done yet.

cedar-policy-validator/src/schema.rs
```diff
@@ -522,34 +519,22 @@ impl ValidatorSchema {
...
-            ContextSchema(crate::types::Type::record_with_attributes(
-                action_id
-                    .context
-                    .iter()
-                    .map(|(k, v)| (k.clone(), v.clone())),
-                OpenTag::ClosedAttributes,
-            ))
-        })
+    pub fn context_type(&self, action: &EntityUID) -> Option<Type> {
+        // INVARIANT: `ValidatorActionId::context_type` always returns a closed
```
andrewmwells-amazon, line 527: nit: I don't think we need to repeat this since it's in the docstring above.

cedar-policy-validator/src/types.rs
```diff
@@ -556,9 +556,9 @@ impl Type {
                                 }
                                 None => {
                                     // attrs has the attribute, self_attrs does not.
-                                    // if required in attrs, incompatible.
-                                    // otherwise fine
-                                    !v.is_required()
+                                    // we require all attributes are declared in
```
andrewmwells-amazon, line 559:
```suggestion
                                    // we require that all attributes are declared in
```

andrewmwells-amazon (approved):

cedar-policy-validator/src/schema.rs
cdisselkoen: The docstring above is to document `ValidatorSchema::context_type`'s return invariant.  The comment here is explaining how we meet that invariant, namely by discharging it to `ValidatorActionId::context_type`.

cedar-policy-validator/src/types.rs
```diff
@@ -556,9 +556,9 @@ impl Type {
                                 }
                                 None => {
                                     // attrs has the attribute, self_attrs does not.
-                                    // if required in attrs, incompatible.
-                                    // otherwise fine
-                                    !v.is_required()
+                                    // we require that all attributes are declared in
```
john-h-kastner-aws, line 559: I think the comment at the top of this method isn't accurate after this change.

john-h-kastner-aws (approved):
Still in favor of the one-constructor option.

cedar-policy-validator/src/coreschema.rs
```diff
@@ -0,0 +1,806 @@
...
+                }
+                if let Some(context) = request.context() {
+                    let actual_context_ty = entities::type_of_restricted_expr(
+                        context.as_ref().as_borrowed(),
+                        extensions,
+                    )?;
+                    println!("actual context ty is {actual_context_ty}");
+                    let expected_context_ty = validator_action_id.context_type();
+                    println!("expected context ty is {expected_context_ty}");
+                    if !expected_context_ty.is_consistent_with(&actual_context_ty) {
```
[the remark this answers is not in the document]
cdisselkoen: I agree, and wrote the function to directly check a RestrictedExpr has a given (validator) type.  And I reverted the change to `is_consistent_with`.

cedar-policy-validator/src/types.rs
```diff
@@ -614,6 +625,109 @@ impl Type {
...
+                    for (k, attr_ty) in attrs.iter() {
+                        if attr_ty.is_required && !record.contains_key(k) {
+                            return Ok(false);
+                        }
+                    }
+                    Ok(true)
+                }
+                None => Ok(false),
+            },
+            Type::ExtensionType { name } => match restricted_expr.as_extn_fn_call() {
```
john-h-kastner-aws, line 720: Does this not need to check the arguments? I suppose that doesn't effect the type of the expression, assuming it is well typed, but it's probably good to reject a request with a bad extension function call.

john-h-kastner-aws (approved):

cedar-policy-validator/src/types.rs
cdisselkoen: Fixed in 8908ee5

khieta (approved):

PR #397, opened 2023-11-01
Avoid duplicating build_and_test workflow

john-h-kastner-aws:
Extract `build_and_test` into a [re-usable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows)run for both PRs and nightly build. This also re-orders the steps a bit to run faster steps before some of the slower steps.

## Disclaimer

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -6,6 +6,4 @@ rustflags = [
     "-Dclippy::unreachable",
     "-Dclippy::indexing_slicing",
     "-Dclippy::panic",
-    "-Dwarnings",
-    "-Funsafe-code"
 ]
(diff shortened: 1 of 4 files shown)


.cargo/config.toml
```diff
@@ -6,4 +6,6 @@ rustflags = [
     "-Dclippy::unreachable",
     "-Dclippy::indexing_slicing",
     "-Dclippy::panic",
+    "-Dwarnings",
```
cdisselkoen, line 9: does this deny all clippy warnings?  I think we probably don't want that

cdisselkoen (approved):

andrewmwells-amazon (approved):

PR #411, opened 2023-11-08
Improve error message for template slot mismatch

svanderbleek:
## Description of changes

Improve error message when there is a template slot mismatch such as `principal = ?resource` instead of `principal = ?principal`.

The change in this PR is:

I confirm that this PR:

## Disclaimer


cdisselkoen (changes requested):
Looks great but can we add a unit test for the failing case, that checks that the error message is appopriate?

cdisselkoen (approved):

shaobo-he-aws (approved):

Issue #382, opened 2023-10-24
Should `type_of_restricted_expr()` mark all record attributes as optional?


cdisselkoen:
### Category

Internal refactors/changes

### Describe the feature you'd like to request

This issue comes out of [this discussion](https://github.com/cedar-policy/cedar/pull/360#discussion_r1370281297) on #360.  Currently, the function `ValueParser::type_of_rexpr()` (before #360) or `type_of_restricted_expr()` (after #360) returns `SchemaType`s in which all record attributes are marked optional.  This has some consequences when used in Core's `SchemaType::is_consistent_with()`.  (note that all of the code and types just referenced are in Core, not referring to the validator's type system.)  @khieta and @john-h-kastner-aws are confused why this is the case, so it merits investigation at a minimum, and at least some clarifying comments, if not a behavior change or a reorganization of Core's `SchemaType` or an adjustment to how Core gets schema information from the validator crate.

### Describe alternatives you've considered

It's possible that the current behavior is correct, in which case we should just add comments clarifying why it is correct for future folks that might happen across this code.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

andrewmwells-amazon:
Marking attributes as required affects which sets are considered heterogeneous, which would be a breaking change. I think we can close this because that's not viable.

cdisselkoen:
Should we update the comments in the code?

PR #430, opened 2023-11-14
Implementation of RFC 34

cdisselkoen:
## Description of changes

See [RFC 34](https://github.com/cedar-policy/rfcs/pull/34) for high-level description.

Some internal notes: 
* Core's `Entity`, Validator's `ValidatorNamespaceDef`, and Validator's `ValidatorActionId` were all updated to store attributes as partial value (action attributes, in the validator case); but the existing serialization behavior for these types was preserved -- they still serialize as `RestrictedExpr`.
* RFC 34 changes where a bunch of errors are thrown, which means a bunch of internal APIs throw more or less errors than they used to.  Some error types had variants added or removed, and some (non-public-facing) error types were renamed for clarity.  Some tests have to do more `.unwrap()`, some less.
* We do get to remove not only `Entities::evaluate()` but quite a bit of code that was supporting it.

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/ast/entity.rs
+++ b/cedar-policy-core/src/ast/entity.rs
@@ -446,7 +446,7 @@ impl std::fmt::Display for PartialValueSerializedAsExpr {
 /// Contains some extra contextual information and the underlying
 /// `EvaluationError`.
 #[derive(Debug, Error)]
-#[error("failed to evaluate attribute `{attr}` of entity `{uid}`: {err}")]
+#[error("failed to evaluate attribute `{attr}` of `{uid}`: {err}")]
 pub struct EntityAttrEvaluationError {
     /// UID of the entity where the error was encountered
     pub uid: EntityUID,
--- a/cedar-policy-core/src/ast/entity.rs
+++ b/cedar-policy-core/src/ast/entity.rs
@@ -15,13 +15,17 @@
  */
 
 use crate::ast::*;
+use crate::evaluator::{EvaluationError, RestrictedEvaluator};
+use crate::extensions::Extensions;
 use crate::parser::err::ParseErrors;
 use crate::transitive_closure::TCNode;
 use crate::FromNormalizedStr;
 use itertools::Itertools;
 use serde::{Deserialize, Serialize};
+use serde_with::{serde_as, TryFromInto};
 use smol_str::SmolStr;
 use std::collections::{HashMap, HashSet};
+use thiserror::Error;
 
 /// We support two types of entities. The first is a nominal type (e.g., User, Action)
 /// and the second is an unspecified type, which is used (internally) to represent cases
@@ -216,8 +220,8 @@ pub struct Entity {
     /// Internal HashMap of attributes.
     ///
     /// In the serialized form of `Entity`, attribute values appear as
-    /// `RestrictedExpr`s.
-    attrs: HashMap<SmolStr, RestrictedExpr>,
+    /// `RestrictedExpr`s, for mostly historical reasons.
+    attrs: HashMap<SmolStr, PartialValueSerializedAsExpr>,
 
     /// Set of ancestors of this `Entity` (i.e., all direct and transitive
     /// parents), as UIDs
@@ -229,6 +233,53 @@ impl Entity {
     pub fn new(
         uid: EntityUID,
         attrs: HashMap<SmolStr, RestrictedExpr>,
+        ancestors: HashSet<EntityUID>,
+        extensions: &Extensions<'_>,
+    ) -> Result<Self, EntityAttrEvaluationError> {
+        let evaluator = RestrictedEvaluator::new(extensions);
+        let evaluated_attrs = attrs
+            .into_iter()
+            .map(|(k, v)| {
+                let attr_val = evaluator
+                    .partial_interpret(v.as_borrowed())
+                    .map_err(|err| EntityAttrEvaluationError {
+                        uid: uid.clone(),
+                        attr: k.clone(),
+                        err,
+                    })?;
+                Ok((k, attr_val.into()))
+            })
+            .collect::<Result<_, EntityAttrEvaluationError>>()?;
+        Ok(Entity {
+            uid,
+            attrs: evaluated_attrs,
+            ancestors,
+        })
+    }
+
+    /// Create a new `Entity` with this UID, attributes, and ancestors.
+    ///
+    /// Unlike in `Entity::new()`, in this constructor, attributes are expressed
+    /// as `PartialValue`.
+    pub fn new_with_attr_partialvalues(
+        uid: EntityUID,
+        attrs: HashMap<SmolStr, PartialValue>,
+        ancestors: HashSet<EntityUID>,
+    ) -> Self {
+        Entity {
+            uid,
+            attrs: attrs.into_iter().map(|(k, v)| (k, v.into())).collect(), // TODO: can we do this without disassembling and reassembling the HashMap
+            ancestors,
+        }
+    }
+
+    /// Create a new `Entity` with this UID, attributes, and ancestors.
+    ///
+    /// Unlike in `Entity::new()`, in this constructor, attributes are expressed
+    /// as `PartialValueSerializedAsExpr`.
+    pub fn new_with_attr_partialvalueserializedasexpr(
+        uid: EntityUID,
+        attrs: HashMap<SmolStr, PartialValueSerializedAsExpr>,
         ancestors: HashSet<EntityUID>,
     ) -> Self {
         Entity {
@@ -244,8 +295,8 @@ impl Entity {
     }
 
     /// Get the value for the given attribute, or `None` if not present
-    pub fn get(&self, attr: &str) -> Option<&RestrictedExpr> {
-        self.attrs.get(attr)
+    pub fn get(&self, attr: &str) -> Option<&PartialValue> {
+        self.attrs.get(attr).map(|v| v.as_ref())
     }
 
     /// Is this `Entity` a descendant of `e` in the entity hierarchy?
@@ -259,10 +310,8 @@ impl Entity {
     }
 
     /// Iterate over this entity's attributes
-    pub fn attrs(&self) -> impl Iterator<Item = (&str, BorrowedRestrictedExpr<'_>)> {
-        self.attrs
-            .iter()
-            .map(|(k, v)| (k.as_str(), v.as_borrowed()))
+    pub fn attrs(&self) -> impl Iterator<Item = (&SmolStr, &PartialValue)> {
+        self.attrs.iter().map(|(k, v)| (k, v.as_ref()))
     }
 
     /// Create an `Entity` with the given UID, no attributes, and no parents.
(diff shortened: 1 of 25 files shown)


cedar-policy-core/src/ast/entity.rs
```diff
@@ -230,6 +234,53 @@ impl Entity {
...
+            attrs: attrs.into_iter().map(|(k, v)| (k, v.into())).collect(), // TODO: can we do this without disassembling and reassembling the HashMap
+            ancestors,
+        }
+    }
+
+    /// Create a new `Entity` with this UID, attributes, and ancestors.
+    ///
+    /// Unlike in `Entity::new()`, in this constructor, attributes are expressed
+    /// as `PartialValueSerializedAsExpr`.
+    pub fn new_with_attr_partialvalueserializedasexpr(
```
aaronjeline, line 280: this name is kinda gross, can we add `_`s?

cedar-policy-core/src/ast/entity.rs
```diff
@@ -354,6 +404,58 @@ impl std::fmt::Display for Entity {
     }
 }
 
+/// `PartialValue`, but serialized as a RestrictedExpr, for historical reasons
```
aaronjeline, line 407: Might add in the comment the fact that we can't serialize all values (specifically extension values) so we have to convert back to exprs

cedar-policy-core/src/entities/json/value.rs
```diff
@@ -283,53 +283,113 @@ impl CedarValueJson {
...
+                        .map(|(k, v)| Ok((k.clone(), Self::from_value(v.clone())?)))
+                        .collect::<Result<JsonRecord, JsonSerializationError>>()?,
+                ))
+            }
+            Value::ExtensionValue(ev) => {
+                let ext_fn: &Name = &ev.constructor;
+                Ok(Self::ExtnEscape {
+                    __extn: FnAndArg {
+                        ext_fn: ext_fn.to_string().into(),
+                        arg: match ev.args.as_slice() {
```
aaronjeline, line 342: Does this really only work for extensions that only call one argument?

cedar-policy-core/src/entities/json/value.rs
```diff
@@ -283,53 +283,113 @@ impl CedarValueJson {
...
+/// helper function to check if the given keys contain any reserved keys,
+/// throwing an appropriate `JsonSerializationError` if so
+fn check_for_reserved_keys<'a>(
+    mut keys: impl Iterator<Item = &'a SmolStr>,
+) -> Result<(), JsonSerializationError> {
+    // We could be a little more permissive here, but to be
+    // conservative, we throw an error for any record that contains
+    // any key with a reserved name, not just single-key records
+    // with the reserved names.
+    let reserved_keys: HashSet<&str> = HashSet::from_iter(["__entity", "__extn", "__expr"]);
```
aaronjeline, line 383: Can we move this constant somewhere else instead of having it buried in a function. If we add a new reserved key we will need to update this, and that should be obvious.

aaronjeline (changes requested):

cedar-policy-core/src/entities/json/value.rs
cdisselkoen: Yes.  Our JSON formats only support extension functions that have exactly one argument.  This is currently fine because we do not have any extension functions with zero or 2+ arguments that are legal in restricted expressions.  Put another way, all of our extension _constructors_ have exactly one argument.

cedar-policy-core/src/ast/entity.rs
cdisselkoen: Fixed

cedar-policy-core/src/ast/entity.rs
cdisselkoen: Fixed

cedar-policy-core/src/entities/json/value.rs
aaronjeline: Fair enough

aaronjeline (approved):

khieta (approved):
Overall looks good. Left some minor comments.

cedar-policy-core/src/ast/entity.rs
```diff
@@ -290,8 +333,15 @@ impl Entity {
...
     // Only used for convenience in some tests and when fuzzing
     #[cfg(any(test, fuzzing))]
-    pub fn set_attr(&mut self, attr: SmolStr, val: RestrictedExpr) {
-        self.attrs.insert(attr, val);
+    pub fn set_attr(
+        &mut self,
+        attr: SmolStr,
+        val: RestrictedExpr,
+        extensions: &Extensions<'_>,
+    ) -> Result<(), EvaluationError> {
```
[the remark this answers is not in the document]
cdisselkoen: The existing setter isn't even full public, we try not to mutate `Entity` at all.  (It's marked `cfg(any(test, fuzzing))`.)  I'd rather we not add more setters unless/until they are needed

cedar-policy/CHANGELOG.md
```diff
@@ -37,7 +38,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Rename `cedar_policy_core::entities::JsonDeserializationError::ExtensionsError`
   to `cedar_policy_core::entities::JsonDeserializationError::ExtensionFunctionLookup`.
   (#360)
-- Rename variants in `cedar_policy::SchemaError`. (#231)
+- Rename variants in `SchemaError`. (#231)
+- `SchemaError` has a new variant corresponding to errors evaluating action
+  attributes. (#372)
```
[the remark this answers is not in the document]
cdisselkoen: In fact, must be a merge conflict, that changelog bullet is for this PR

cedar-policy-validator/src/types.rs
```diff
@@ -627,7 +630,163 @@ impl Type {
...
+    /// but does not typecheck for other permissible substitutions), this is
+    /// reported as an error.
+    ///
+    /// TODO(#437): Handling of `Unknown`s is not yet complete and doesn't
+    /// properly behave according to the above description, as of this writing.
+    pub(crate) fn typecheck_partial_value(
+        &self,
+        value: &PartialValue,
+        extensions: Extensions<'_>,
+    ) -> Result<bool, GetSchemaTypeError> {
```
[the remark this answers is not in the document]
cdisselkoen: The conversion to `RestrictedExpr` is fallible, because of the case of nontrivial residuals (residuals which are more than just a single Unknown).  Eventually we will need to handle the nontrivial-residuals case in a better way.

cedar-policy-core/src/authorizer/err.rs
```diff
@@ -21,16 +21,12 @@ use thiserror::Error;
 /// Errors that can occur during authorization
 #[derive(Debug, PartialEq, Eq, Clone, Error)]
 pub enum AuthorizationError {
```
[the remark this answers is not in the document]
cdisselkoen: I don't see a huge amount of value in doing this refactor, which we might just end up reverting in the future, as I could imagine us adding more ways authorization could fail other than policy evaluation failing?

Single-variant enums for errors are common enough, at least in our codebase, and in fact this PR introduces new single-variant error enums `PartialValueToRestrictedExprError` and `PartialValueToValueError` -- although there's currently only one way the operations could fail, the single-variant enum provides a nice name for the failure case (the variant name) and is future-compatible with additional failure modes we might want to add in the future.

cedar-policy-core/src/entities/conformance.rs
```diff
@@ -225,15 +223,66 @@ impl<'a, S: Schema> EntitySchemaConformanceChecker<'a, S> {
...
+/// If the typecheck passes, return `Ok(())`.
+/// If the typecheck fails, return an appropriate `Err`.
+pub fn typecheck_value_against_schematype(
+    value: &PartialValue,
+    expected_ty: &SchemaType,
+    extensions: Extensions<'_>,
+) -> Result<(), TypecheckError> {
+    // TODO: instead of computing the `SchemaType` of `value` and then checking
+    // whether the schematypes are "consistent", wouldn't it be less confusing,
+    // more efficient, and maybe even more precise to just typecheck directly?
```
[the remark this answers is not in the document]
cdisselkoen: I'm somewhat concerned that the schematype-is-"consistent" check is not the same as the direct result of typechecking the `value` in all edge cases; #382 is an example of how it's even unclear to us exactly what the schematype-is-"consistent" check should be/mean.  So I'd like to keep the comment (somewhere).

However, I do agree that we could reuse `typecheck_restricted_expr_against_schematype` here.  Converting `PartialValue` to `RestrictedExpr` is fallible in general, but in this case the failure (nontrivial residuals) cannot happen, and we already handle that case in this method by returning `Ok(())`, which we can easily still do with the code-reuse approach.

cedar-policy/CHANGELOG.md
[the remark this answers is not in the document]
cdisselkoen: Fixed

cedar-policy-core/src/authorizer.rs
```diff
@@ -165,16 +165,7 @@ impl Authorizer {
...
-            Ok(eval) => eval,
-            Err(e) => {
-                return ResponseKind::FullyEvaluated(Response::new(
-                    Decision::Deny,
-                    HashSet::new(),
-                    vec![AuthorizationError::AttributeEvaluationError(e)],
-                ));
-            }
-        };
+        let eval = Evaluator::new(q.clone(), entities, &self.extensions);
```
john-h-kastner-aws, line 168: `Request` is probably cheap to clone, but would it still be better to take an owned `Request` here?

khieta (approved):

cedar-policy-validator/src/types.rs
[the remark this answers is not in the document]
cdisselkoen: Looking at this again, I do think we can reduce the code duplication some though.  I think the appropriate place to delegate is to have `typecheck_value()` convert the `Value` to `RestrictedExpr` and then reuse `typecheck_restricted_expr()`.  That conversion is infallible, and while not free (requires cloning records, among other things), it's probably worth it to avoid the maintenance burden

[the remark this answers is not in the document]
cdisselkoen: Refactored

cedar-policy-validator/src/types.rs
```diff
@@ -627,7 +630,163 @@ impl Type {
...
+                primitive_type: Primitive::Bool,
+            } => Ok(matches!(value, Value::Lit(Literal::Bool(_)))),
+            Type::Primitive {
+                primitive_type: Primitive::Long,
+            } => Ok(matches!(value, Value::Lit(Literal::Long(_)))),
+            Type::Primitive {
+                primitive_type: Primitive::String,
+            } => Ok(matches!(value, Value::Lit(Literal::String(_)))),
+            Type::True => match value {
+                Value::Lit(Literal::Bool(b)) => Ok(*b),
```
john-h-kastner-aws: I'd find this easier to read as `Value::Lit(Literal::Bool(true)) => Ok(true)`

khieta (approved):

john-h-kastner-aws (approved):

cedar-policy-validator/src/types.rs
cdisselkoen: Actually removed this code due to thread above with Kesha

cedar-policy-core/src/authorizer.rs
cdisselkoen: That's fair, for Core here.  I don't think it's worth changing (breaking) the public API to take owned `Request`, especially because maybe in the future we find a way to avoid this clone.

Before this PR, the `Request` was effectively cloned in `Evaluator::new()`.  I've just made it more explicit by having `Evaluator::new()` take owned `Request` itself.  The end result in terms of performance should be the same before and after this PR (in this respect)

cdisselkoen: Refactored in Core, left the clone in api.rs

PR #446, opened 2023-11-16
Move tests out of api.rs

andrewmwells-amazon:
## Description of changes
Move tests out of api.rs. IMO this is cleaner than having almost 3k lines of tests in the public API.

## Disclaimer


aaronjeline (approved):

shaobo-he-aws (approved):

cdisselkoen (approved):

shaobo-he-aws:
LGTM. That being said, is there a convention where we want to put unit tests (e.g., a dedicated folder)?

cdisselkoen:
stupid question, did you do the simple check that the tests are still running?  (like, just break one and check.)  I'm a little surprised we didn't need `mod tests;` anywhere

andrewmwells-amazon:
>  LGTM. That being said, is there a convention where we want to put unit tests (e.g., a dedicated folder)?

I just followed e.g., https://github.com/rust-lang/rust/blob/master/compiler/rustc_graphviz/src/tests.rs here. I don't know if a one-size-fits-all strategy makes the most sense.

andrewmwells-amazon:
> stupid question, did you do the simple check that the tests are still running? (like, just break one and check.) I'm a little surprised we didn't need mod tests; anywhere

We did in fact need a `mod tests;`

cedar-policy/src/api.rs
```diff
@@ -2746,7 +2746,7 @@ impl Policy {
     /// conversion to AST is lossy. ESTs for policies generated by this method
     /// will reflect the AST and not the original policy syntax.
     #[cfg_attr(not(feature = "partial-eval"), allow(unused))]
-    fn from_ast(ast: ast::Policy) -> Self {
+    pub fn from_ast(ast: ast::Policy) -> Self {
```
andrewmwells-amazon, line 2749: Highlighting this change.

cedar-policy/src/tests.rs
```diff
@@ -0,0 +1,2737 @@
...
+            ast::Effect::Permit,
+            ast::Expr::unknown(ast::Unknown::new_with_type(
+                "test_entity_type::\"unknown\"",
+                ast::Type::Entity {
+                    ty: ast::EntityType::Concrete("test_entity_type".parse().unwrap()),
+                },
+            )),
+            ast::PolicyID::from_smolstr("static".into()),
+        );
+        let static_policy = Policy::from_ast(ast);
```
andrewmwells-amazon, line 841: Needed for this.

It's unclear to me if we want people to be able to directly construct policies with unknown entities (i.e., do we want `Policy::from_ast` to be `pub` with `partial-eval` enabled). OTOH, I don't know why we enforce this property in a unit test if we don't intend to allow people to construct policies with unknown entities from our public API. If we don't intend that, I think we should test that the output of some PE has the expected unknown entities.

andrewmwells-amazon:
Fails correctly with broken unit test: https://github.com/cedar-policy/cedar/actions/runs/6897658266/job/18766238211?pr=446

cedar-policy/src/api.rs
cdisselkoen: I don't think we should do this change. `ast::Policy` is a type we don't want to expose.  Is `pub(crate)` sufficient?

cedar-policy/src/tests.rs
cdisselkoen: Not familiar with this test, but if the test requires `ast::Policy`, then perhaps it shouldn't be in the `cedar-policy` crate?

cedar-policy/src/lib.rs
```diff
@@ -42,5 +42,8 @@ pub use api::*;
 /// Frontend utilities, see comments in the module itself
 pub mod frontend;
 
+#[cfg(test)] //redundant, but helps rust_analyzer not incorrectly flag `use` statements
```
john-h-kastner-aws, line 45: I think this could also be `#![cfg(test)]` at the top of the test file. A little cleaner IMO.

cedar-policy/src/tests.rs
```diff
@@ -0,0 +1,2737 @@
...
+pub use cedar_policy_validator::{
+    TypeErrorKind, UnsupportedFeature, ValidationErrorKind, ValidationWarningKind,
+};
+use std::collections::{HashMap, HashSet};
+use std::str::FromStr;
+
+pub use super::api::Response;
+
+// PANIC SAFETY unit tests
+#[allow(clippy::panic)]
```
john-h-kastner-aws, line 21: With this all as one module we can `#![allow(clippy::panic)]`.

cedar-policy/src/api.rs
john-h-kastner-aws: `pub(crate)` should work since these are unit test instead of integration.

john-h-kastner-aws (approved):
I like it.

john-h-kastner-aws (approved):

PR #448, opened 2023-11-17
disallow duplicate keys in FFI, integration testing

cdisselkoen:
## Description of changes

using the already existing `JsonValueWithNoDuplicateKeys` created for Core

## Disclaimer

### DIFF
--- a/cedar-policy/src/frontend/is_authorized.rs
+++ b/cedar-policy/src/frontend/is_authorized.rs
@@ -377,23 +377,25 @@ impl RecvdSlice {
 
         let mut errs = Vec::new();
 
-        let (mut policies, entities) =
-            match (Entities::from_json_value(entities.into(), schema), policy_set) {
-                (Ok(entities), Ok(policies)) => (policies, entities),
-                (Ok(_), Err(policy_parse_errors)) => {
-                    errs.extend(policy_parse_errors);
-                    (PolicySet::new(), Entities::empty())
-                }
-                (Err(e), Ok(_)) => {
-                    errs.push(e.to_string());
-                    (PolicySet::new(), Entities::empty())
-                }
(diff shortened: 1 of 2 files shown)


cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -143,17 +144,17 @@ enum AuthorizationAnswer {
...
+    resource: Option<JsonValueWithNoDuplicateKeys>,
     #[serde_as(as = "MapPreventDuplicates<_, _>")]
-    context: HashMap<String, serde_json::Value>,
+    context: HashMap<String, JsonValueWithNoDuplicateKeys>,
     /// Optional schema in JSON format.
     /// If present, this will inform the parsing: for instance, it will allow
     /// `__entity` and `__extn` escapes to be implicit, and it will error if
     /// attributes have the wrong types (e.g., string instead of integer).
     #[serde(rename = "schema")]
-    schema: Option<serde_json::Value>,
```
john-h-kastner-aws, line 156: Schema parsing _should_ already be forbidding duplicate keys everywhere, so this should just move the error message around a bit. I don't see anything wrong with that, and consistently using the no-duplicates variant is nice, but maybe think about if the new error is substantially better/worse than before.

john-h-kastner-aws (approved):

cedar-policy/src/frontend/is_authorized.rs
cdisselkoen: I don't think that's the case.  `serde_json::Value` cannot contain duplicate keys, so any schema parsing that accepts `serde_json::Value` won't ever see the duplicate keys -- they would already be silently dropped (last-value-wins) during whatever process created the `serde_json::Value`.

andrewmwells-amazon (approved):

Issue #455, opened 2023-11-21
remove_template and remove_static_policy for a linked policy panics


andrewmwells-amazon:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Other

### Describe the bug

Performing the following operations on a PolicySet
```
add_template("policy0")
link(template="policy0", id="policy3")
add_template("policy3")
remove_static("policy3")
remove_template("policy3")
```
results in a panic.

### Expected behavior

Cedar shouldn't panic.

### Reproduction steps

Perform the following operations on a PolicySet:

```
add_template("policy0")
link(template="policy0", id="policy3")
add_template("policy3")
remove_static("policy3")
remove_template("policy3")
```

### Code Snippet

```rust
// Put your code below this line.

    #[test]
    fn pset_removal_prop_test_1() {
        let template = Template::parse(
            Some("policy0".into()),
            "permit(principal == ?principal, action, resource);",
        )
        .expect("Template Parse Failure");
        // let static_policy = Policy::parse(
        //     Some("policy4".into()),
        //     "permit(principal, action, resource);",
        // )
        // .expect("Static parse failure");
        let mut pset = PolicySet::new();
        // pset.add(static_policy).unwrap();
        pset.add_template(template).unwrap();
        let env: HashMap<SlotId, EntityUid> =
        std::iter::once((SlotId::principal(), EntityUid::from_strs("Test", "test"))).collect();
        pset.link(PolicyId::from_str("policy0").unwrap(), PolicyId::from_str("policy3").unwrap(), env);
        let template = Template::parse(
            Some("policy3".into()),
            "permit(principal == ?principal, action, resource);",
        )
        .expect("Template Parse Failure");
        pset.add_template(template).unwrap();
        pset.remove_static(PolicyId::from_str("policy3").unwrap());
        pset.remove_template(PolicyId::from_str("policy3").unwrap());
        //Should not panic
    }
```

### Log output

```
// Put your output below this line

panic: `Found template policy in self.templates but not in self.ast`
```

andrewmwells-amazon:
Closed with https://github.com/cedar-policy/cedar/pull/456

PR #456, opened 2023-11-21
Fix bugs in policy set remove operations on wrong types. Fix bug in a…

andrewmwells-amazon:
…dd_template with existing link of same name.

## Description of changes
Fix bugs in policy set remove operations on wrong types. Fix bug in add_template with existing link of same name.

## Issue #, if available
https://github.com/cedar-policy/cedar/issues/455

## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [ ] A breaking change requiring a major version bump to `cedar-policy` (e.g., changes to the signature of an existing API).
- [ ] A backwards-compatible change requiring a minor version bump to `cedar-policy` (e.g., addition of a new API).
- [ ] A bug fix or other functionality change requiring a patch to `cedar-policy`.
- [ ] A change "invisible" to users (e.g., documentation, changes to "internal" crates like `cedar-policy-core`, `cedar-validator`, etc.)
- [ ] A change (breaking or otherwise) that only impacts unreleased or experimental code.

I confirm that this PR (choose one, and delete the other options):

- [ ] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).
- [ ] Does not update the CHANGELOG because my change does not significantly impact released code.

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [ ] Does not require updates because my change does not impact the Cedar Dafny model or DRT infrastructure.
- [ ] Requires updates, and I have made / will make these updates myself. (Please include in your description a timeline or link to the relevant PR in `cedar-spec`, and how you have tested that your updates are correct.)
- [ ] Requires updates, but I do not plan to make them in the near future. (Make sure that your changes are hidden behind a feature flag to mark them as experimental.)
- [ ] I'm not sure how my change impacts `cedar-spec`. (Post your PR anyways, and we'll discuss in the comments.)

## Disclaimer

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #457, opened 2023-11-21
Fix bug in add_template with existing link of same name.

andrewmwells-amazon:
## Description of changes
Fix bug in add_template with existing link of same name.

## Issue #, if available
Related to https://github.com/cedar-policy/cedar/issues/455

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/ast/policy_set.rs
+++ b/cedar-policy-core/src/ast/policy_set.rs
@@ -161,7 +161,11 @@ impl PolicySet {
     }
 
     /// Add a template to the policy set.
+    /// If a link with the same name already exists, this will error.
     pub fn add_template(&mut self, t: Template) -> Result<(), PolicySetError> {
+        if self.links.contains_key(t.id()) {
+            return Err(PolicySetError::Occupied);
+        }
         // TODO: Use `try_insert` when stabilized.
         // https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.try_insert
         match self.templates.entry(t.id().clone()) {
--- a/cedar-policy-core/src/ast/policy_set.rs
+++ b/cedar-policy-core/src/ast/policy_set.rs
@@ -161,7 +161,7 @@ impl PolicySet {
     }
 
     /// Add a template to the policy set.
-    /// If a link with the same name already exists, this will error.
+    /// If a link or template with the same name already exists, this will error.
     pub fn add_template(&mut self, t: Template) -> Result<(), PolicySetError> {
         if self.links.contains_key(t.id()) {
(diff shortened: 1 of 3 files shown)


cedar-policy-core/src/ast/policy_set.rs
```diff
@@ -161,7 +161,11 @@ impl PolicySet {
     }
 
     /// Add a template to the policy set.
+    /// If a link with the same name already exists, this will error.
```
cdisselkoen, line 164: what about if a static policy with the same name already exists?  or if a template with the same name already exists?  I'm sure these are both errors, but we can/should call that out in the doc comment

cedar-policy/src/api.rs
```diff
@@ -3662,6 +3662,36 @@ mod policy_set_tests {
...
+            Some("policy3".into()),
+            "permit(principal == ?principal, action, resource);",
+        )
+        .expect("Template Parse Failure");
+
+        assert_matches!(
+            pset.add_template(template),
+            Err(PolicySetError::AlreadyDefined)
+        );
+        //Should not panic
```
cdisselkoen, line 3693: do we already have tests that add a template with the same name as an existing static policy, or as an existing template, throws this error?

cdisselkoen (approved):

cedar-policy-core/src/ast/policy_set.rs
andrewmwells-amazon: Fair point.

cedar-policy/src/api.rs
andrewmwells-amazon: I didn't add tests for it but from reading the code I'm fairly confident (famous last words)

cdisselkoen: Adding tests for that is probably reasonable just to ensure we don't break it in the future

cdisselkoen:
Oh another comment -- are we confident this change isn't breaking?  in the sense that, does Cedar 2.4.2 allow a template and a link to have the same name, and now we're disallowing (breaking) that?

andrewmwells-amazon:
> Oh another comment -- are we confident this change isn't breaking? in the sense that, does Cedar 2.4.2 allow a template and a link to have the same name, and now we're disallowing (breaking) that?

Definitely breaking but I think the old behavior is buggy. I don't see why `link; add_template` would be allowed but `add_template; link` would be an error. The order in which you add policies shouldn't change behavior.

andrewmwells-amazon:
Since the 3.0 release is coming soon, I merged into `main` instead (https://github.com/cedar-policy/cedar/pull/456)

PR #491, opened 2023-12-12
Improve parse errors around invalid `is` expressions

### LINKED ISSUE #409: Improve parser errors around `is`
### Category

Other

There is some room to provide more specific parse errors for malformed `is` expressions.

To make this issue concrete: `is` currently parsers expecting a `Name` for the entity type. If the following text doesn't parse as a `Name` the user sees a cryptic "expected token" error message (see #176). We could improve these errors message by instead expecting an `Expr`, and then raising an specific error in the CST to AST conversion if the `Expr` isn't a `Name`.

Other parse error improvements improvements are covered by #176, #391, #407

.

### Additional context

_No response_

john-h-kastner-aws:
## Description of changes

Fix #409 by parsing an `Add` rather than an `Name` in `is` expression. This allows for parsing expression in this position so that `principal is User::"alice"` can parse, allowing for a more useful error in when building the ast. The `Add` production needs to be used instead of `Expr` to avoid ambiguity with the `.. is .. in ..` form.

## Checklist for requesting a review.

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/parser/grammar.lalrpop
+++ b/cedar-policy-core/src/parser/grammar.lalrpop
@@ -94,7 +94,7 @@ pub Policy: Node<Option<cst::Policy>> = {
     <l:@L> <err:!> <r:@R> => { errors.push(err); Node::new(None,l,r) },
 }
 
-// VariableDef := Variable [':' Name] [('in' | '==') Expr]
+// VariableDef := Variable [':' Name] ['is' Add] [('in' | '==') Expr]
 VariableDef: Node<Option<cst::VariableDef>> = {
     <l:@L> <variable: AnyIdent> <unused_type_name: (":" <Name>)?> <entity_type: (IS <Add>)?>
         <ineq: (RelOp Expr)?> <r:@R>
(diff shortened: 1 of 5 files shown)


cedar-policy-core/src/parser/grammar.lalrpop
```diff
@@ -94,9 +94,9 @@ pub Policy: Node<Option<cst::Policy>> = {
     <l:@L> <err:!> <r:@R> => { errors.push(err); Node::new(None,l,r) },
 }
 
-// VariableDef := Variable [':' Name] [('in' | '==') Expr]
+// VariableDef := Variable [':' Name] ['is' Add] [('in' | '==') Expr]
```
shaobo-he-aws, line 97: Nit: maybe we should add a comment here saying `Add` shows up here rather than `Name` because we want better error message. Otherwise it could be confusing here because the grammar specifies that a `Name` comes after `is`.

shaobo-he-aws (approved):

andrewmwells-amazon (approved):

cdisselkoen (approved):

shaobo-he-aws (approved):

Issue #461, opened 2023-11-27
CLI command to convert between Human and JSON policy formats


max2me:
### Category

CLI features/changes

### Describe the feature you'd like to request

CLI should let users convert policy to/from JSON.

### Describe alternatives you've considered

Doing it in Rust, as it's done today.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

cdisselkoen:
^ related to the name change above -- the CLI already accepts policies in JSON format everywhere it accepts policies in human format (except for the `format` command, because, obviously). so all that remains to "support JSON policy format in CLI" is to add a command to convert between human and JSON formats.

memark:
Could you give me some pointers on how to do this?

For `SchemaFragment` there is `as_json_string()` and `as_natural()`.

For `PolicySet` there is `from_str()` and `from_json_str()`. But I'm not sure how to convert it TO json or human form?

cdisselkoen:
To json: On `main` (not in any release as of this writing), there is a `PolicySet::to_json()`.
To human: The `Display` trait for `PolicySet` is probably what should be used I think

john-h-kastner-aws:
I looked at this a couple weeks ago and found that we don't have a public API for converting json policies to human. `Display` prints the lossless repr, which for a JSON policy is the JSON.

That extends the scope of this issue to adding a public `to_human` (maybe not the best name,  see https://github.com/cedar-policy/cedar/issues/842) function to policies and policy sets before using that function to implement the CLI feature.

john-h-kastner-aws:
> Would it make sense to implement only the translation from human to json for now?

That would be enough to be useful for seeding the corpus in  cedar-policy/cedar-spec#343, so yeah, I think it would make sense

john-h-kastner-aws:
With #987 we can convert to JSON. As discussed above, we need to implement json-to-human before we can add it to the CLI.

memark:
>That would be enough to be useful for seeding the corpus in https://github.com/cedar-policy/cedar-spec/pull/343, so yeah, I think it would make sense

Given that the translation capability now exists, I'd love to take a look at using it for seeding the corpus, as mentioned above. Could you give me some pointers on what needs to be done in that repo?

john-h-kastner-aws:
[...]

PR #493, opened 2023-12-12
[CLI] accept policies in JSON format and/or from stdin in more targets

cdisselkoen:
## Description of changes

Allows the CLI to accept policies in JSON (EST) format, using the flag `--policy-format=json`.  The default remains `--policy-format=human`, so this is a backwards compatible change.

Factors out the code for accepting policy sets, so that all targets (except `format`, which is special because it doesn't use the full parser) accept the same options under the same names and share code for parsing policies.  In particular, all of these targets now allow omitting `-p`/`--policies`, and thus reading policies from stdin.  (Previously this was supported on some targets and not others.)

## Issue #, if available

Related to #461, and probably a building block for #461, but doesn't resolve it.  #461 wants a CLI command for converting policies from human to json and back.

## Disclaimer

### DIFF
--- a/cedar-policy-cli/CHANGELOG.md
+++ b/cedar-policy-cli/CHANGELOG.md
@@ -9,6 +9,10 @@
 - Requests are now validated by default if a schema is provided. This can be
   disabled with `--request-validation=false`.
 - The `-s` short form can now be used for `--schema` across all subcommands.
+- The `-p`/`--policies` flag can now be omitted across all subcommands where it
+  is present. If the flag is omitted, policies will be read from `stdin`.
+- `--policy-format` flag to many subcommands, allowing you to pass policies in
+  JSON format. The default remains `human` format.
 
 ### Changed
 
--- a/cedar-policy-cli/src/lib.rs
+++ b/cedar-policy-cli/src/lib.rs
@@ -103,9 +103,9 @@ pub struct ValidateArgs {
     /// File containing the schema
     #[arg(short, long = "schema", value_name = "FILE")]
     pub schema_file: String,
-    /// File containing the policy set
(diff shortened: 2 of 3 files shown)


cedar-policy-cli/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
 - Requests are now validated by default if a schema is provided. This can be
   disabled with `--request-validation=false`.
 - The `-s` short form can now be used for `--schema` across all subcommands.
+- The `-p`/`--policies` flag can now be omitted across all subcommands where it
+  is present. If the flag is omitted, policies will be read from `stdin`.
+- `--policy-format` flag to many subcommands, allowing you to pass policies in
```
andrewmwells-amazon, line 14:
```suggestion
- Add `--policy-format` flag to many subcommands, allowing you to pass policies in
```

andrewmwells-amazon (approved):

cedar-policy-cli/CHANGELOG.md
cdisselkoen: This is in the `Added` section, so I thought we don't do this?  Although it seems we're more careful about that in the `cedar-policy` changelog than this CLI one

khieta (approved):
Lgtm. You may consider adding examples of using json input to the various readmes.

cdisselkoen:
The READMEs in `sample-data` are kinda presented as a pseudo-tutorial.  I think the audience for a pseudo-tutorial will be most interested in the human input format, and the JSON format is mostly for advanced programmatic users who generally know what they're doing and don't need CLI examples; pointing them to the CLI help is sufficient.  So I'll leave the READMEs alone for now

Issue #494, opened 2023-12-12
Generalize Integer Type in Cedar


andrewmwells-amazon:
### Category

Internal refactors/changes

### Describe the feature you'd like to request

Based on this request from slack (https://cedar-policy.slack.com/archives/C0547KH7R19/p1701977134218759):
```
I’m working on an authorization system around blockchain transactions; I believe current data types can’t support bigint operations, any idea how that could be supported easily? Any reason we couldn’t add an extension type like decimal to add Big Integer type/operations?
```
We don't want to change the default behavior of Cedar, but we can generalize our code and make it easy for people to play with different Integer types.

### Describe alternatives you've considered

We could keep the status quo. IMO, generalizing the code in this way is not much work and would provide more flexibility to people who want it.

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

andrewmwells-amazon:
Closed with https://github.com/cedar-policy/cedar/pull/495

PR #495, opened 2023-12-12
Generalize Integer type (so it can be changed to e.g., i128)

andrewmwells-amazon:
## Description of changes
Generalize Integer type (so it can be changed to e.g., i128)
(To change to i128, some tests for over/underflow behavior also need to be changed)

## Issue #, if available
https://github.com/cedar-policy/cedar/issues/494

## Disclaimer


cedar-policy-core/src/ast/integer.rs
```diff
@@ -0,0 +1,9 @@
+/// The integer type we use internally.
```
andrewmwells-amazon, line 1:
```suggestion
/// The integer types we use (both internally and for parsing).
```

aaronjeline (approved):
Seems very reasonable. Only wish we could change the text `Long` to `Integer`

shaobo-he-aws (approved):
LGTM.

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -3842,6 +3842,7 @@ mod tests {
                 "-9223372036854775809",
                 ToASTErrorKind::IntegerLiteralTooLarge(9223372036854775809),
             ),
+            // This test doesn't fail with an internal representation of i128:
             // Contrary to Rust, this expression is not valid because the
             // parser treats it as a negation operation whereas the operand
             // (9223372036854775808) is too large.
```
cdisselkoen, line 3848: can we reverse the order here -- first explain why the test has the behavior it does by default (which is the "Contrary to Rust" part) -- then we can have an additional comment about the behavior with i128 (and actually, why doesn't it fail with i128?)

cdisselkoen (commented):
I expected to see a change in the lalrpop definition file to use the `InputInteger` type, was that not necessary?

PR #498, opened 2023-12-14
fix clippy warnings

andrewmwells-amazon:
## Description of changes
fix some clippy warnings

## Disclaimer

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -19,6 +19,7 @@ jobs:
       - run: RUSTFLAGS="-D warnings -F unsafe-code" cargo build --verbose
       - run: cargo doc --all-features --no-deps
       - run: cargo clippy --all-features
+      - run: cargo clippy -- -D warnings
       - run: cargo test --verbose --features "experimental"
       - run: cargo test --verbose
       - run: cargo test --verbose --no-default-features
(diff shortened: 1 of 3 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -19,6 +19,7 @@ jobs:
       - run: RUSTFLAGS="-D warnings -F unsafe-code" cargo build --verbose
       - run: cargo doc --all-features --no-deps
       - run: cargo clippy --all-features
```
andrewmwells-amazon, line 21: We want to add `-D` to this line, but clippy wants us to add `#[must_use]` to a bunch of functions in api.rs, so I'll do than in another PR after we discuss whether than needs a semver bump.

john-h-kastner-aws (commented):
I think (a while ago) we explicitly decided not to fail CI on clippy. 

Since then, we have been keeping the code more-or-less free of clippy lints with regular PRs to fix them, so enforcing this in CI is probably fine, but we should get consensus on this from the whole team first.

andrewmwells-amazon:
AFAIK, the closest thing to "warn (but not block) if clippy fails" is to just override the failure and merge anyways.

andrewmwells-amazon:
Maybe the compromise could be to run clippy nightly and fail that build, but not run for each PR?

andrewmwells-amazon:
Discussed with the team and decided there's no need to run clippy in CI.

john-h-kastner-aws (approved):

khieta (approved):

PR #508, opened 2023-12-15
remove `Node::as_inner_pair`

cdisselkoen:
## Description of changes

Another refactor relating to the internal type `Node`.

I don't think `Node::as_inner_pair()` was helping.  Replaced it with calls to `as_inner()` and accessing `self.loc` directly, which I think is clearer.  (In some places, where there are more than one `SourceSpan`s involved, it makes it super clear which span is being used.)

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/parser/cst_to_ast.rs
+++ b/cedar-policy-core/src/parser/cst_to_ast.rs
@@ -216,7 +216,7 @@ impl Node<Option<cst::Policy>> {
         id: ast::PolicyID,
         errs: &mut ParseErrors,
     ) -> Option<ast::Template> {
-        let (maybe_policy, src) = self.as_inner_pair();
+        let maybe_policy = self.as_inner();
         // return right away if there's no data, parse provided error
         let policy = maybe_policy?;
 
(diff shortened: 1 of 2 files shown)


cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -216,7 +216,7 @@ impl Node<Option<cst::Policy>> {
         id: ast::PolicyID,
         errs: &mut ParseErrors,
     ) -> Option<ast::Template> {
-        let (maybe_policy, src) = self.as_inner_pair();
+        let maybe_policy = self.as_inner();
```
john-h-kastner-aws, line 219: I might go a step further and inline this with `let policy = ...` and do the same for similar uses in other functions.

john-h-kastner-aws (approved):

cedar-policy-core/src/parser/cst_to_ast.rs
cdisselkoen: Planning on that as a followup

khieta (approved):

PR #512, opened 2023-12-15
Integrate validation errors with miette

john-h-kastner-aws:
## Description of changes

Validation errors were not being printed using the pretty miette style. This PR makes the whole `ValidationResult` a `Diagnostic` implemented in the same style as a `ParseErrors` and uses this to print the result in the CLI. Source locations still aren't printed because the `Report` struct needs to be given the source code string, but this change allows for print help messages.

Pretty printing the `ValidationResult` with a miette `Report` did not work initially because `Report::new` wants arguments with static lifetimes, but `ValidationResult` is parametrized by a lifetime `'a` inherited from the policy set that was validated. Removing the lifetime on the struct would be an breaking change to the api, so I have worked around the issue by using the lifetime only on a `PhantomData` field in the struct, allowing the lifetime parameter to be set to `'static`. I think this avoids any breaking api changes, but review on that point is welcome.

## Disclaimer

### DIFF
--- a/cedar-policy-cli/src/lib.rs
+++ b/cedar-policy-cli/src/lib.rs
@@ -505,19 +505,21 @@ pub fn validate(args: &ValidateArgs) -> CedarExitCode {
         CedarExitCode::Success
     };
 
-    let mut errors = result.validation_errors().peekable();
+    let (errors, warnings) = result.into_errors_and_warnings();
+    let mut errors = errors.peekable();
+    let mut warnings = warnings.peekable();
+
     if errors.peek().is_some() {
         println!("Validation Errors:");
         for note in errors {
-            println!("{}", note);
+            println!("{:?}", Report::new(note.into_owned()));
         }
     }
 
-    let mut warnings = result.validation_warnings().peekable();
     if warnings.peek().is_some() {
         println!("Validation Warnings:");
         for note in warnings {
-            println!("{}", note);
+            println!("{:?}", Report::new(note.into_owned()));
         }
     }
 
--- a/cedar-policy-cli/src/lib.rs
+++ b/cedar-policy-cli/src/lib.rs
@@ -495,35 +495,21 @@ pub fn validate(args: &ValidateArgs) -> CedarExitCode {
     let validator = Validator::new(schema);
     let result = validator.validate(&pset, mode);
 
-    let exit_code = if !result.validation_passed()
+    if !result.validation_passed()
         || (args.deny_warnings && !result.validation_passed_without_warnings())
     {
-        println!("Validation Failed");
+        println!(
+            "{:?}",
+            Report::new(result).wrap_err("policy set validation failed")
+        );
         CedarExitCode::ValidationFailure
     } else {
-        println!("Validation Passed");
+        println!(
+            "{:?}",
+            Report::new(result).wrap_err("policy set validation passed")
+        );
         CedarExitCode::Success
-    };
-
-    let (errors, warnings) = result.into_errors_and_warnings();
-    let mut errors = errors.peekable();
-    let mut warnings = warnings.peekable();
-
-    if errors.peek().is_some() {
-        println!("Validation Errors:");
-        for note in errors {
-            println!("{:?}", Report::new(note.into_owned()));
-        }
     }
-
-    if warnings.peek().is_some() {
-        println!("Validation Warnings:");
-        for note in warnings {
-            println!("{:?}", Report::new(note.into_owned()));
-        }
-    }
-
-    exit_code
 }
 
 pub fn evaluate(args: &EvaluateArgs) -> (CedarExitCode, EvalResult) {
--- a/cedar-policy/CHANGELOG.md
+++ b/cedar-policy/CHANGELOG.md
@@ -35,6 +35,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- The `ValidationResult` returned from `Validator::validate` now has a static
+  lifetime, allowing it to be used in more contexts. The lifetime parameter
+  will be removed in a future major version. 
 - Improve parse error around invalid `is` expressions.
 - Improve parser error message when a policy includes an invalid template slot.
   The error now identifies that the policy used an invalid slot and suggests using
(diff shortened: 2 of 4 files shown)


cedar-policy-cli/src/lib.rs
```diff
@@ -495,33 +495,21 @@ pub fn validate(args: &ValidateArgs) -> CedarExitCode {
...
+        println!(
+            "{:?}",
+            Report::new(result).wrap_err("policy set validation failed")
+        );
         CedarExitCode::ValidationFailure
     } else {
-        println!("Validation Passed");
+        println!(
+            "{:?}",
+            Report::new(result).wrap_err("policy set validation passed")
```
cdisselkoen, line 509: why is this using `Report`? will this show as an error with miette's ❌?

cedar-policy-cli/src/lib.rs
```diff
@@ -495,33 +495,21 @@ pub fn validate(args: &ValidateArgs) -> CedarExitCode {
     let validator = Validator::new(schema);
     let result = validator.validate(&pset, mode);
 
-    let exit_code = if !result.validation_passed()
+    if !result.validation_passed()
```
cdisselkoen, line 498: usually I prefer switching the `then` and `else` branches in order to not have `if !cond`

cedar-policy/CHANGELOG.md
```diff
@@ -35,6 +35,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- The `ValidationResult` returned from `Validator::validate` now has a static
+  lifetime, allowing it to be used in more contexts. The lifetime parameter
+  will be removed in a future major version.
```
cdisselkoen, line 40: If we're going to say this, we should make an issue to remember to do it

shaobo-he-aws (approved):

cedar-policy/src/api.rs
```diff
@@ -1385,35 +1387,143 @@ impl<'a> ValidationResult<'a> {
...
 
     /// Get an iterator over the warnings found by the validator.
-    pub fn validation_warnings(&self) -> impl Iterator<Item = &ValidationWarning<'a>> {
+    pub fn validation_warnings(&self) -> impl Iterator<Item = &ValidationWarning<'static>> {
         self.validation_warnings.iter()
     }
+
+    fn first_error_or_warning(
+        &self,
+    ) -> Option<Either<&ValidationError<'a>, &ValidationWarning<'a>>> {
```
cdisselkoen, line 1401: would it simplify things to return `Option<&dyn Diagnostic>` instead of `Option<Either<...>>`?  do any callers need to discriminate between error vs warning?

cedar-policy/src/api.rs
```diff
@@ -1385,35 +1387,143 @@ impl<'a> ValidationResult<'a> {
...
+    }
 }
 
 /// An error generated by the validator when it finds a potential problem in a
 /// policy. The error contains a enumeration that specifies the kind of problem,
 /// and provides details specific to that kind of problem. The error also records
 /// where the problem was encountered.
 #[derive(Debug, Error)]
-#[error("validation error on `{}`: {}", self.location, self.error_kind())]
+#[error("validation error on {}: {}", self.location, self.error_kind())]
```
cdisselkoen, line 1522:
```suggestion
#[error("validation error on {location}: {}", self.error_kind())]
```

cedar-policy/src/api.rs
```diff
@@ -1519,13 +1631,14 @@ impl<'a> std::fmt::Display for SourceLocation<'a> {
     }
 }
 
-impl<'a> From<cedar_policy_validator::SourceLocation<'a>> for SourceLocation<'a> {
-    fn from(loc: cedar_policy_validator::SourceLocation<'a>) -> SourceLocation<'a> {
-        let policy_id: &'a PolicyId = PolicyId::ref_cast(loc.policy_id());
+impl<'a> From<cedar_policy_validator::SourceLocation<'a>> for SourceLocation<'static> {
+    fn from(loc: cedar_policy_validator::SourceLocation<'a>) -> SourceLocation<'static> {
+        let policy_id: PolicyId = PolicyId(loc.policy_id().clone());
```
cdisselkoen, line 1636:
```suggestion
        let policy_id = PolicyId(loc.policy_id().clone());
```

cedar-policy/src/api.rs
```diff
@@ -1535,18 +1648,19 @@ impl<'a> From<cedar_policy_validator::SourceLocation<'a>> for SourceLocation<'a>
...
-) -> impl Iterator<Item = ValidationWarning<'a>> {
+    templates: impl Iterator<Item = &'a Template> + 'a,
+) -> impl Iterator<Item = ValidationWarning<'static>> + 'a {
     cedar_policy_validator::confusable_string_checks(templates.map(|t| &t.ast))
         .map(std::convert::Into::into)
 }
 
 #[derive(Debug, Error)]
-#[error("validation warning on `{}`: {}", .location, .kind)]
+#[error("validation warning on {}: {}", .location, .kind)]
```
cdisselkoen, line 1658:
```suggestion
#[error("validation warning on {location}: {kind}")]
```

cdisselkoen (approved):

cedar-policy-cli/src/lib.rs
andrewmwells-amazon: I don't understand what you're suggesting here? Negate the whole:
`!result.validation_passed() || (args.deny_warnings && !result.validation_passed_without_warnings())` and swap the `then` and `else` branches? IMO `result.validation_passed() && (!args.deny_warnings || result.validation_passed_without_warnings())` isn't any nicer.

andrewmwells-amazon (approved):

cedar-policy-cli/src/lib.rs
john-h-kastner-aws: I set the severity to advice, so we get
image

or, if there's a warning:

image

cedar-policy/src/api.rs
john-h-kastner-aws: I think that will be easier

cedar-policy/CHANGELOG.md
john-h-kastner-aws: sure

cedar-policy-cli/src/lib.rs
cdisselkoen: Beautiful

cedar-policy-cli/src/lib.rs
cdisselkoen: Missed that the condition was more complicated.  Ignore me, sorry

cdisselkoen (approved):

PR #513, opened 2023-12-15
Add `help` to some type errors

john-h-kastner-aws:
## Description of changes

Add help messages on to type errors where we can suggest a useful fix. For example, an error from using `in` with a non-entity on the left will suggest using `contains`, `containsAll` or `containsAny` instead.

## Disclaimer

### DIFF
--- a/cedar-policy-validator/src/type_error.rs
+++ b/cedar-policy-validator/src/type_error.rs
@@ -112,6 +112,7 @@ impl TypeError {
         on_expr: Expr,
         expected: impl IntoIterator<Item = Type>,
         actual: Type,
+        help: Option<UnexpectedTypeHelp>,
     ) -> Self {
         Self {
             on_expr: Some(on_expr),
@@ -119,6 +120,7 @@ impl TypeError {
             kind: TypeErrorKind::UnexpectedType(UnexpectedType {
                 expected: expected.into_iter().collect::<BTreeSet<_>>(),
                 actual,
+                help,
             }),
         }
     }
@@ -243,13 +245,8 @@ impl TypeError {
 pub enum TypeErrorKind {
     /// The typechecker expected to see a subtype of one of the types in
     /// `expected`, but saw `actual`.
-    #[error("unexpected type: expected {} but saw {}",
-        match .0.expected.iter().next() {
-            Some(single) if .0.expected.len() == 1 => format!("{}", single),
-            _ => .0.expected.iter().join(", or ")
-        },
-        .0.actual
-    )]
+    #[error(transparent)]
+    #[diagnostic(transparent)]
     UnexpectedType(UnexpectedType),
     /// The typechecker could not compute a least upper bound for `types`.
     #[error("unable to find upper bound for types: [{}]", .0.types.iter().join(","))]
@@ -262,6 +259,7 @@ pub enum TypeErrorKind {
     /// The typechecker could not conclude that an access to an optional
     /// attribute was safe.
     #[error("unable to guarantee safety of access to optional attribute {}", .0.attribute_access)]
+    #[diagnostic(help("try testing for the attribute with `e has a && ..`"))]
     UnsafeOptionalAttributeAccess(UnsafeOptionalAttributeAccess),
     /// The typechecker found that a policy condition will always evaluate to false.
     #[error(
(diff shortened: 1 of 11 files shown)


shaobo-he-aws (approved):

cedar-policy-validator/src/type_error.rs
```diff
@@ -262,6 +259,7 @@ pub enum TypeErrorKind {
     /// The typechecker could not conclude that an access to an optional
     /// attribute was safe.
     #[error("unable to guarantee safety of access to optional attribute {}", .0.attribute_access)]
+    #[diagnostic(help("try testing for the attribute with `e has a && ..`"))]
```
cdisselkoen, line 262: we could perhaps make this suggestion better by using the real attribute name instead of `a` (or even the real LHS instead of `e`)

cedar-policy-validator/src/typecheck.rs
```diff
@@ -960,6 +969,13 @@ impl<'a> Typechecker<'a> {
                     expr,
                     &[Type::any_entity_reference(), Type::any_record()],
                     type_errors,
+                    |actual| match actual {
+                        Type::Set { .. } => Some(UnexpectedTypeHelp::TryUsingContains),
```
cdisselkoen, line 973: what's the rationale for this suggestion?  if I'm understanding correctly, something like `[1, 2, 4].foo` or `[1, 2, 4]["foo"]` will suggest `.contains()`?

cedar-policy-validator/src/typecheck.rs
```diff
@@ -1149,6 +1180,7 @@ impl<'a> Typechecker<'a> {
                     expr,
                     Type::any_entity_reference(),
                     type_errors,
+                    |_| Some(UnexpectedTypeHelp::TypeTestNotSupported),
```
cdisselkoen, line 1183: this is the case where we have something like `principal.manager is User::"alice"`?  should we suggest using `==` instead of `is` to compare two entity references?

cedar-policy-validator/src/typecheck.rs
```diff
@@ -1388,12 +1422,22 @@ impl<'a> Typechecker<'a> {
             }
 
             BinaryOp::Add | BinaryOp::Sub => {
+                let help_builder = |actual: &Type| match (op, actual) {
+                    (
+                        BinaryOp::Add,
+                        Type::Primitive {
+                            primitive_type: Primitive::String,
+                        },
+                    ) => Some(UnexpectedTypeHelp::ConcatenationNotSupported),
```
cdisselkoen, line 1431: I would suggest also throwing `ConcatenationNotSupported` in the case of `+` on sets, maybe also records

andrewmwells-amazon (approved):

cedar-policy-validator/src/type_error.rs
john-h-kastner-aws: The attribute names are already available on the `AttributeAccess` struct, so I can do something with those.

cedar-policy-validator/src/typecheck.rs
john-h-kastner-aws: That's already suggested by the parser when the cst to ast conversion fails

cdisselkoen: Hmm I see.  When do we actually hit this case then?

cedar-policy-validator/src/typecheck.rs
john-h-kastner-aws: Also makes sense to say some on `-` for sets

cedar-policy-validator/src/typecheck.rs
john-h-kastner-aws: I was thinking using `[1, 2][0]`, but that doesn't actually parse, so it'd have to be `[1, 2]["0"]`. I'll just remove these.

cedar-policy-validator/src/typecheck.rs
john-h-kastner-aws: Something like `1 is Long`, if `Long` were a valid entity type, so not often

john-h-kastner-aws: The error also makes sense if some did `1 is User` expect it to be `false` because `1` isn't a user.

cdisselkoen: Makes sense.

cedar-policy-validator/src/type_error.rs
```diff
@@ -465,27 +469,56 @@ impl AttributeAccess {
...
+            // shouldn't be possible. If it does happen, just use a placeholder
+            // attribute name `f` since we'd rather avoid panicking.
+            None => (&[] as &[SmolStr], "f".into()),
+        };
+
+        let full_expr = std::iter::once(&base_expr)
+            .chain(safe_attrs.iter().rev())
+            .join(".");
+        format!("{full_expr} has {err_attr}")
+    }
```
cdisselkoen, line 503: This is complicated enough that it might make sense to have a test checking that the generated help message is sensible. maybe in particular for a case like `context.foo.bar has spam`.

cdisselkoen (approved):

shaobo-he-aws (approved):

PR #516, opened 2023-12-18
carry source code in `Loc` struct

cdisselkoen:
## Description of changes

Currently (before this PR), source code snippets are only available in the CLI because of the CLI's use of `.with_source_code()` on particular errors.  This doesn't help other consumers of `cedar-policy` that don't call `.with_source_code()` on their own errors.

This PR works towards rectifying this problem, by ensuring that the errors we currently have location info for, carry not only the `SourceSpan` but also a reference to the original source code -- everything that a downstream needs in order to print a pretty error message with source snippet.  This is done by encapsulating source info into a new struct `Loc` which is carried around everywhere instead of carrying around `SourceSpan` directly.

The actual source references, which now live in eg every CST and AST node, are made `Arc<str>`.  Not `&'static str` because source code is (often) not static, but not `&'a str` partly because that would add lifetime parameters to `Node` and `Expr`, and partly because we need source code references to outlive `Node` and `Expr` anyway when they are returned in errors.  I modified the parser to expect a new parameter with the `Arc<str>` so that it can just clone the `Arc` into every CST node, not create a new `Arc` for every CST node which would be very wasteful of memory.

I believe the changes here are back-compatible.  Some struct fields changed types from `SourceSpan` to `Loc`, but only private fields, without affecting those struct's public APIs.

## Issue #, if available

N/A

## Disclaimer

### DIFF
--- a/cedar-policy-cli/src/lib.rs
+++ b/cedar-policy-cli/src/lib.rs
@@ -956,9 +956,7 @@ fn read_from_file(filename: impl AsRef<Path>, context: &str) -> Result<String> {
 
 /// Read a policy set, in Cedar human syntax, from the file given in `filename`,
 /// or from stdin if `filename` is `None`.
-fn read_policy_set(
-    filename: Option<impl AsRef<Path> + std::marker::Copy>,
-) -> miette::Result<PolicySet> {
+fn read_policy_set(filename: Option<impl AsRef<Path> + std::marker::Copy>) -> Result<PolicySet> {
     let context = "policy set";
     let ps_str = read_from_file_or_stdin(filename, context)?;
     let ps = PolicySet::from_str(&ps_str)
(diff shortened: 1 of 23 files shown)


andrewmwells-amazon (approved):

john-h-kastner-aws (approved):
Looks good.

Is this change enough to remove the explicit call to `Report::with_source_code` in the CLI?

cdisselkoen:
I looked at removing the explicit call to `Report::with_source_code` in the CLI.  However, that call does add the filename, which otherwise would be lost.  Leaving the call does no harm.  I added separate tests that we have source info on the appropriate errors even without the extra (external) `Report::with_source_code` call.

john-h-kastner-aws (approved):

Issue #485, opened 2023-12-08
Better source locations for evaluation errors and entity-parsing errors


cdisselkoen:
### Category

Internal refactors/changes

### Describe the feature you'd like to request

Similar to #174 for schema parsing errors, we can make similar improvements to reporting source locations in evaluation errors, entity-parsing errors, and elsewhere.  Especially now that #477 is merged.

### Describe alternatives you've considered

We've basically committed to doing this in #182 / #477 

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

cdisselkoen:
Splitting this issue into two: 

This issue (#485) will be only for source locations in evaluation errors.  That work was completed in #582, and as such, this issue will be closed.

A new issue (#599) covers source locations in "entity-parsing errors and elsewhere" as this issue's description puts it.

PR #523, opened 2023-12-19
propagate source locations to Values

cdisselkoen:
## Description of changes

Just like `Expr`s, now `Value`s also have (optional) attached source `Loc`s.

For now, most `Value` constructors don't attach a `Loc`.  The main way `Value`s get `Loc`s attached is that when the evaluator evaluates an `Expr` into a `Value`, it propagates that `Expr`'s `Loc` to the `Value`.

Some tests that this is done correctly, but we'll get more thorough tests when we start actually propagating these `Loc`s to evaluator errors and testing that the errors have the proper `Loc`s.

## Issue #, if available

A stepping stone towards #485 -- sometimes we need the `Loc` associated with a `Value`, not just an `Expr`.

## Disclaimer


cedar-policy-core/src/ast/value.rs
```diff
@@ -26,58 +27,255 @@ use thiserror::Error;
 
 /// This describes all the values which could be the dynamic result of evaluating an `Expr`.
 /// Cloning is O(1).
-#[derive(Debug, Clone, PartialOrd, Ord, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(into = "Expr")]
 #[serde(try_from = "Expr")]
 pub enum Value {
```
shaobo-he-aws, line 33: Does it make sense to have a struct `Value`, which has a `ValueKind` and a `Loc`?

shaobo-he-aws (approved):

cedar-policy-core/src/ast/value.rs
cdisselkoen: We could do it that way.  I felt that in this case, the benefit of being able to `match value { Value::Lit { lit, .. } }` is better ergonomics for ignoring the `Loc` than something like `match value.value_kind() { ValueKind::Lit(lit) }` would be, but I'm open to the `ValueKind` approach too.

A more involved change, which I'd also be wiling to consider, would be renaming `Node<T>` to `Located<T>` or something, since it's really just a general wrapper around a `T` that adds a `Loc`.  Then we could leave `Value` alone but use `Located<Value>` in places where we expect a `Value` with optional `Loc` attached.  

Thoughts on either of these?

john-h-kastner-aws: For our _large_ enums I definitely prefer the `*Kind` pattern. The particular benefits I see are that it avoids redundancy and facilitates attaching extra meta-data later, but I agree that attaching it inline is more ergonomic when there number of variants is small. The `Value` enum is small enough that I'd have probably used the inline approach as well.

There is some value in standardizing on one pattern, but IMO it's not critical, and it may be better to just use whichever is nicer in a given situation. 

The `Located` option is an improvement over re-implementing the `*Kind` pattern everywhere, but it might not work out so cleanly given that `Node` and `ToAstError` currently have non-optional `Loc`s, and the `Expr` struct also stores the `data` entry used to type annotation.

cdisselkoen: Discussed offline with the Cedar team and we decided to prefer the `*Kind` pattern in general.  I will adapt this PR after the holidays.

cedar-policy-core/src/parser.rs
```diff
@@ -469,52 +471,51 @@ mod eval_tests {
         let entities = eval::test::basic_entities();
         let exts = Extensions::none();
         let evaluator = eval::Evaluator::new(request, &entities, &exts);
+        // The below tests check not only that we get the expected `Value`, but
+        // that it has the expected source location.
+        // We have to check that separately because the `PartialEq` and `Eq`
+        // impls for `Value` do not compare source locations.
```
andrewmwells-amazon, line 477: Should they? At least personally I'd expect this.

andrewmwells-amazon (approved):

cedar-policy-core/src/parser.rs
cdisselkoen: I feel that the evaluator should not consider source locations when comparing values for equality.  Source locations are not part of the Cedar semantics.  So I think we definitely need two senses of equality for Values -- one that ignores source locations, and is used by the evaluator; and another that considers source locations, and is used in tests / for debugging / maybe other purposes.

The question here is then, which sense should get the actual `PartialEq` / `Eq` implementation, and which one should be a different method on `Value`.  I sense that you're proposing we have `PartialEq` / `Eq` be the sense of equality that considers source locations.  I'm inclined to agree.  Did I interpret correctly?

andrewmwells-amazon: Yeah, definitely need both and my inclination was to have `PartialEq` / `Eq` consider location. Looking at the changes needed for that I'm not as inclined now. My reasoning was that since we need both we should just use the default `derive` where we can, but I think I actually prefer the current version.

cdisselkoen: All right, I'll leave it as is then.

cedar-policy-core/src/ast/value.rs
cdisselkoen: Completed this change.

shaobo-he-aws (approved):

cedar-policy-core/src/ast/value.rs
```diff
@@ -30,35 +30,34 @@ use thiserror::Error;
...
-    ExtensionValue {
-        /// the value
-        ev: Arc<ExtensionValueWithArgs>,
-        /// Source location associated with the value, if any
-        loc: Option<Loc>,
-    },
+    ExtensionValue(Arc<ExtensionValueWithArgs>),
+}
+
+// Custom impl of `Ord`: ignore the `Loc`s
```
andrewmwells-amazon, line 56:
```suggestion
// Custom impl of `Ord`, ignoring the `Loc`s
```

cedar-policy-core/src/ast/value.rs
```diff
@@ -299,6 +287,44 @@ pub struct Set {
...
+            .collect::<Option<HashSet<Literal>>>()
+            .map(Arc::new);
+        Self {
+            authoritative: Arc::new(authoritative),
+            fast,
+        }
+    }
+
+    /// Create a set with the given `Literal`s as elements
+    pub fn of_lits(lits: impl IntoIterator<Item = Literal>) -> Self {
```
andrewmwells-amazon, line 313: [nit] Isn't `from_lits` more idiomatic?

andrewmwells-amazon (approved):

cedar-policy-core/src/ast/value.rs
cdisselkoen: You are correct.  I think I got sidetracked by wanting to write `Set::of_lits` which reads well in English

cdisselkoen: Updated

john-h-kastner-aws (approved):

PR #529, opened 2023-12-20
Removes some panics

aaronjeline:
## Description of changes
Removes some panics
1) Removes a panic by refactoring code to use iterators instead of length check -> then access
2) Removes a panic by refactoring `TypeError` to enforce it's invariant statically
## Issue #, if available
N/A

## Disclaimer


aaronjeline:
Ensure if this change is Major, do we expose the `TypeError` constructor?

cdisselkoen:
It appears we do expose `EvaluationErrorKind` and thus all its variants?  probably since #260?  `EvaluationErrorKind` is documented in the 3.0.0 docs for `cedar-policy` [on docs.rs](https://docs.rs/cedar-policy/latest/cedar_policy/enum.EvaluationErrorKind.html)

cedar-policy-core/src/evaluator/err.rs
```diff
@@ -115,25 +116,46 @@ impl EvaluationError {
     }
 
     /// Construct a [`TypeError`] error
-    pub(crate) fn type_error(expected: Vec<Type>, actual: Type) -> Self {
+    pub(crate) fn type_error(expected_first: Type, expected_rest: Vec<Type>, actual: Type) -> Self {
```
cdisselkoen, line 119: it's worth a comment on this function specifying what `expected_first` and `expected_rest` are.

I think we should consider using https://crates.io/crates/nonempty instead of essentially rolling it ourselves, if only to make code like this (and below) clearer and more ergonomic.

cedar-policy-core/src/evaluator/err.rs
```diff
@@ -115,25 +116,46 @@ impl EvaluationError {
...
+            error_kind: EvaluationErrorKind::TypeError {
+                expected_first,
+                expected_rest,
+                actual,
+            },
             advice: None,
         }
     }
 
+    pub(crate) fn type_error_single(expected_first: Type, actual: Type) -> Self {
```
cdisselkoen, line 130: also a comment here on what `expected_first` is (maybe rename to just `expected`?) and what the difference is between this function and `type_error()`.  (It's clear in this PR, but might not be immediately clear to future people coming to this code for the first time)

cedar-policy-core/src/evaluator/err.rs
```diff
@@ -115,25 +116,46 @@ impl EvaluationError {
...
+        Self::type_error(expected_first, vec![], actual)
+    }
+
     /// Construct a [`TypeError`] error with the advice field set
     pub(crate) fn type_error_with_advice(
-        expected: Vec<Type>,
+        expected_first: Type,
+        expected_rest: Vec<Type>,
```
cdisselkoen, line 137: same here

cedar-policy-core/src/evaluator/err.rs
```diff
@@ -115,25 +116,46 @@ impl EvaluationError {
...
+                actual,
+            },
             advice: Some(advice),
         }
     }
 
+    pub(crate) fn type_error_with_advice_single(
+        expected_first: Type,
+        actual: Type,
+        advice: String,
```
cdisselkoen, line 154: and here

cedar-policy-core/src/evaluator/err.rs
```diff
@@ -300,22 +323,14 @@ pub enum EvaluationErrorKind {
...
-        #[allow(clippy::indexing_slicing)]
-        1 => format!("type error: expected {}, got {}", expected[0], actual),
-        _ => {
-            use itertools::Itertools;
-            format!(
-                "type error: expected one of [{}], got {actual}",
-                expected.iter().join(", ")
-            )
-        }
+fn pretty_type_error(first_expected: &Type, rest_expected: &[Type], actual: &Type) -> String {
```
cdisselkoen, line 326: why `first_expected` and `rest_expected` here, but `expected_first` and `expected_rest` everywhere else?

cedar-policy-core/src/evaluator/err.rs
```diff
@@ -300,22 +323,14 @@ pub enum EvaluationErrorKind {
...
+    if rest_expected.is_empty() {
+        format!("type error: expected {}, got {}", first_expected, actual)
+    } else {
+        format!(
+            "type error: expected one of [{}], got {actual}",
+            rest_expected.iter().join(", ")
```
cdisselkoen, line 332: case in point about how https://crates.io/crates/nonempty might help us avoid mistakes.  I think this message silently drops the `first_expected`.  It should probably be something like 
```suggestion
            std::iter::once(first_expected).chain(rest_expected.iter()).join(", ")
```

cedar-policy-core/src/fuzzy.rs
```diff
@@ -0,0 +1,158 @@
+use smol_str::SmolStr;
+
+/// Fuzzy string matching using the Levenshtein distance algorithm
+pub fn fuzzy_search(key: &str, lst: &[impl AsRef<str>]) -> Option<SmolStr> {
```
cdisselkoen, line 4: where did this code come from? I feel like we already had Levenshtein algorithm implemented somewhere in Cedar, did you just move it here?  if so, can we delete it from wherever it used to be?

cedar-policy/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Changed `TypeError` to enforce it's invariant statically
```
cdisselkoen, line 18:
```suggestion
- Changed `TypeError` to enforce its invariant statically
```

cdisselkoen (changes requested):
Marking request-changes only because of the suspected bug in `pretty_type_error()`.  the rest doesn't have to be blocking

cedar-policy-core/src/evaluator/err.rs
aaronjeline: Yup, fair enough! Nonempty does seem better.

cedar-policy-core/src/fuzzy.rs
aaronjeline: Shoot, accidental commit. Was planning to move it to this crate it the process of resolving another issue, but not part of this PR

shaobo-he-aws (approved):

cedar-policy-core/Cargo.toml
```diff
@@ -28,6 +28,7 @@ miette = { version = "5.9.0", features = ["serde"] }
 
 # decimal extension requires regex
 regex = { version = "1.8", features = ["unicode"], optional = true }
+nonempty = "0.9.0"
```
cdisselkoen, line 31: let's put this in the main section of deps, not under decimal extension

cedar-policy-core/src/evaluator/err.rs
```diff
@@ -266,12 +258,10 @@ pub enum EvaluationErrorKind {
...
     /// Tried to evaluate an operation on values with incorrect types for that
     /// operation
-    #[error("{}", pretty_type_error(expected_first, expected_rest, actual))]
+    #[error("{}", pretty_type_error(expected, actual))]
     TypeError {
-        /// Expected this type
-        expected_first: Type,
-        /// or one of these types
-        expected_rest: Vec<Type>,
+        /// Expected one of these type
```
cdisselkoen, line 263:
```suggestion
        /// Expected one of these types
```

cdisselkoen (approved):

shaobo-he-aws (approved):

PR #531, opened 2023-12-21
Support source location in JSON schema internal representations

shaobo-he-aws:

## Disclaimer


john-h-kastner-aws (approved):
This approach to adding source information looks good.

We might need to added it onto more parts of the schema struct `ActionEntityType::member_of` stands out, and it also look like we don't have source information on the top level `NamespaceDefinition`s in a `SchemaFragment`.

cdisselkoen (approved):

PR #550, opened 2023-12-26
Delete unused integration test

khieta:
## Description of changes

* Deleted the "4c" integration test, which was expected to fail in our testing harness.
* Removed comments referencing other tests that we ignore (which have already been deleted).
* Minor edit in the integration testing harness to ignore evaluation errors when a "custom implementation" is used. (This will enable a `cedar-spec` edit I plan to make once this PR is merged.)

## Disclaimer


john-h-kastner-aws (approved):

khieta:
@john-h-kastner-aws asked (offline) if it would make sense to update the "4c" test to pass instead of deleting it entirely.

After looking into it, although we could update the json test file to say that every request should fail, I think it makes more sense to just delete it because it's testing a feature that we don't currently support (action attributes). Users might reasonably expect the policies in the `example_use_cases_doc` folder to be well-formed.

shaobo-he-aws (approved):

cdisselkoen:
> a feature that we don't currently support (action attributes)

We do support action attributes, but only in non-schema mode.  Would it make sense instead to keep the test, but not require validation and not provide a schema for it?  (I realize I'm late to this thread)

PR #555, opened 2023-12-27
Re-enable clippy

john-h-kastner-aws:
## Description of changes

When discussing #498 we forgot that clippy was responsible for denying `unwrap()` expressions. Re-enable clippy in CI so that it will continue to detect these. 

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -19,7 +19,7 @@ jobs:
       - run: RUSTFLAGS="-D warnings -F unsafe-code" cargo build --verbose
       - run: cargo doc --all-features --no-deps
       # Clippy is configured by `.cargo/config.toml` to deny on lints like
-      # `unwrap_used`. The aren't detected by `panic_safety.sh` which only
+      # `unwrap_used`. They aren't detected by `panic_safety.sh` which only
       # looks for comments where we've added an `allow` directive for clippy.
       - run: cargo clippy --all-features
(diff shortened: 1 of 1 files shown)


shaobo-he-aws (approved):

.github/workflows/build_and_test.yml
```diff
@@ -18,6 +18,10 @@ jobs:
       - run: RUSTFLAGS="-D warnings -F unsafe-code" cargo build --verbose --features "experimental"
       - run: RUSTFLAGS="-D warnings -F unsafe-code" cargo build --verbose
       - run: cargo doc --all-features --no-deps
+      # Clippy is configured by `.cargo/config.toml` to deny on lints like
+      # `unwrap_used`. The aren't detected by `panic_safety.sh` which only
```
khieta, line 22:
```suggestion
      # `unwrap_used`. They aren't detected by `panic_safety.sh` which only
```

khieta (commented):
To clarify: clippy is set to deny on `unwrap`, but warn on other issues, right? And the warnings will just be printed to the CI log? (Just making sure I understand the impact of reverting #498.)

john-h-kastner-aws:
Yes. Clippy lints will be warnings, printed into the CI log and ignored by us for the most part. 

Anything listed in https://github.com/cedar-policy/cedar/blob/main/.cargo/config.toml will be an error, causing the build to fail unless we add an `#[allow()]`.

khieta (approved):

PR #557, opened 2023-12-28
RFC 24 implementation

shaobo-he-aws:
## Description of changes

Implement [RFC 24](https://github.com/cedar-policy/rfcs/blob/main/text/0024-schema-syntax.md)

## Disclaimer


cedar-policy-cli/sample-data/sandbox_a/schema.cedarschema
```diff
@@ -0,0 +1,17 @@
...
+entity Video in [Account, Album];
+entity User in [UserGroup];
+entity UserGroup;
+entity Administrator;
+entity Photo in [Account, Album];
+entity Album in [Account];
+entity Account;
+
+// Common type declarations
+
```
mwhicks1, line 11: Is it possible to drop this comment when there are no common types?

cedar-policy-cli/src/lib.rs
```diff
@@ -94,10 +94,28 @@ pub enum Commands {
     Link(LinkArgs),
     /// Format a policy set
     Format(FormatArgs),
+    /// Translate JSON schema to custom schema syntax and vice versa
```
mwhicks1, line 97: > and vice versa

except for comments?

cedar-policy-cli/src/lib.rs
```diff
@@ -114,6 +132,9 @@ pub struct ValidateArgs {
     /// experimental `partial-validate` feature enabled.
     #[arg(long = "partial-validate")]
     pub partial_validate: bool,
+    /// Schema format (JSON or Custom)
+    #[arg(long, value_enum, default_value_t = SchemaFormat::Json)]
```
mwhicks1, line 136: We should change the default to be non-JSON, and then explicitly use it in our existing codebase. We want users to think of the custom format as the preferred one.

cedar-policy-validator/src/custom_schema/to_json_schema.rs
```diff
@@ -0,0 +1,372 @@
...
+            NamedType::Primitive(PrimitiveType::Bool) => Self::Type(SchemaTypeVariant::Boolean),
+            NamedType::Primitive(PrimitiveType::Long) => Self::Type(SchemaTypeVariant::Long),
+            NamedType::Primitive(PrimitiveType::String) => Self::Type(SchemaTypeVariant::String),
+        }
+    }
+}
+
+impl Context {
+    pub fn lookup(&self, node: &Path) -> Option<NamedType> {
+        let name = <Path as Into<Name>>::into(node.clone());
```
john-h-kastner-aws, line 58: Could this just be `Into<Name>::into`?

cedar-policy-validator/src/custom_schema/to_json_schema.rs
```diff
@@ -0,0 +1,372 @@
...
+
+impl TryFrom<NonEmpty<Node<AppDecl>>> for ApplySpec {
+    type Error = ToValidatorSchemaError;
+    fn try_from(value: NonEmpty<Node<AppDecl>>, context: &Context) -> Result<Self, Self::Error> {
+        let mut resource_types = Vec::new();
+        let mut principal_types = Vec::new();
+        let mut attr_context: Option<AttributesOrContext> = None;
+        for decl in value {
+            match decl.node {
+                AppDecl::PR(PRAppDecl { ty, entity_tys }) => match ty.node {
```
john-h-kastner-aws, line 218: Is it ever possible that either arm of this match is hit more than once? 

If not, the way this is written in a loop is a bit confusing. Might be nicer if you could have a function `value.pr_app_decl` and `value.context`.

cedar-policy-cli/src/lib.rs
```diff
@@ -94,10 +94,28 @@ pub enum Commands {
...
+    #[arg(short, long, value_name = "FILE")]
+    pub input_file: Option<String>,
+}
+
+#[derive(Debug, Clone, ValueEnum)]
+pub enum SchemaFormat {
+    /// Default JSON format
+    Json,
+    /// Custom format
+    Custom,
```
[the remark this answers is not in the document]
cdisselkoen: I suggest `Human`

cedar-policy-validator/src/custom_schema/to_json_schema.rs
cdisselkoen: Even better, `Name::from`?

cedar-policy-cli/src/lib.rs
aaronjeline: Is that a breaking change?

jszmajda:
BTW - one feature request here - please sort the keys of the output json, I'm having to pass it through jq to stabilize the output json (and make diffs nice)

cedar-policy-cli/Cargo.toml
```diff
@@ -13,6 +13,7 @@ repository = "https://github.com/cedar-policy/cedar"
 [dependencies]
 cedar-policy = { version = "=3.0.0", path = "../cedar-policy" }
 cedar-policy-formatter = { version = "=3.0.0", path = "../cedar-policy-formatter" }
+cedar-policy-validator = { version = "=3.0.0", path = "../cedar-policy-validator" }
```
cdisselkoen, line 16: why does CLI need a direct dependency on Validator?  should we just expose the relevant Validator functionality in `cedar-policy` instead?

cedar-policy-cli/sample-data/sandbox_a/schema.cedarschema
```diff
@@ -0,0 +1,15 @@
...
+entity Photo in [Account, Album];
+entity Album in [Account];
+entity Account;
+
+
+// Action declarations
+action listPhotos
+  appliesTo { principal: [User], resource: [Album, Photo, Video] };
+action view, delete, edit
+  appliesTo { principal: [User], resource: [Photo, Video, Album] };
```
cdisselkoen, line 15: Do we want/need the comments "Entity type declarations" and "Action declarations"?  (here and in the other schemas)

cedar-policy-cli/src/lib.rs
```diff
@@ -94,10 +97,34 @@ pub enum Commands {
...
     /// Create a Cedar project
     New(NewArgs),
 }
 
+#[derive(Args, Debug)]
+pub struct TranslateSchemaArgs {
+    #[arg(long)]
+    pub from_json: bool,
+    #[arg(short, long = "schema", value_name = "FILE")]
+    pub input_file: Option<String>,
```
cdisselkoen, line 111: Doc comments here would be helpful, would appear in the CLI `help`.

I wonder if we want an enum `"json-to-human", "human-to-json"` instead of a `bool` flag.  Or even, two separate CLI commands, one for each direction. so `cedar schema-json-to-human` and `cedar schema-human-to-json` or something

cedar-policy-cli/src/lib.rs
cdisselkoen: That's a breaking change we should make for Cedar 4.  But we shouldn't make that breaking change in this PR, because we intend this PR for Cedar 3.1

cedar-policy-cli/src/lib.rs
```diff
@@ -1010,14 +1095,42 @@ fn read_json_policy(filename: Option<impl AsRef<Path> + std::marker::Copy>) -> R
...
+                            "failed to parse schema from file {}",
+                            filename.as_ref().display()
+                        )
+                    })?;
+            let (json_schema, _) = custom_schema_to_json_schema(custom_schema)
+                .map_err(|err| miette!("failed to translate custom schema: {err:?}"))?;
+            Schema::from_str(
+                &serde_json::to_string(&json_schema)
+                    .map_err(|err| miette!("failed to serialize schema: {err}"))?,
+            )
```
cdisselkoen, line 1130: why is CLI code converting from `custom_schema` to `json_schema` and then calling `Schema::from_str()`?  The fact that custom schemas are parsed by first converting to JSON seems like it should be an implementation detail of the `validator` crate, and not something that external users (like CLI) should need to know or be aware of

cedar-policy-cli/tests/sample.rs
```diff
@@ -465,6 +469,7 @@ fn run_validate_test(policies_file: &str, schema_file: &str, exit_code: CedarExi
         },
         deny_warnings: false,
         partial_validate: false,
+        schema_format: cedar_policy_cli::SchemaFormat::Json,
```
cdisselkoen, line 472:
```suggestion
        schema_format: SchemaFormat::Json,
```

cedar-policy-core/src/test_utils.rs
```diff
@@ -122,6 +122,8 @@ impl<'a> std::fmt::Display for OriginalInput<'a> {
 /// `src` is the original input text, just for better assertion-failure messages.
 /// This function accepts any `impl Into<OriginalInput>` for `src`,
 /// including `&str` and `&serde_json::Value`.
+/// PANIC SAFETY: testing code
+#[allow(clippy::panic)]
```
cdisselkoen, line 126: just put this on the whole module, not the individual function?

cedar-policy-validator/src/custom_schema/ast.rs
```diff
@@ -0,0 +1,187 @@
...
+use nonempty::NonEmpty;
+use smol_str::SmolStr;
+
+use super::to_json_schema::PrimitiveType;
+
+const IPADDR_EXTENSION: &str = "ipaddr";
+const DECIMAL_EXTENSION: &str = "decimal";
+pub(super) const CEDAR_NAMESPACE: &str = "__cedar";
+
+pub type Ident = Node<Id>;
```
cdisselkoen, line 16: now that we're using `Node` for these other purposes, `Node` isn't really a great name.  Should we change the name?  It's not a public type, so that wouldn't be a breaking change.  `WithLoc`?

cedar-policy-validator/src/custom_schema/ast.rs
```diff
@@ -0,0 +1,187 @@
...
+pub type Str = Node<SmolStr>;
+pub type Name = Either<Ident, Str>;
+pub type Schema = Vec<Node<Namespace>>;
+
+#[derive(Debug, Clone)]
+pub struct Path {
+    pub base: Ident,
+    pub prefix: Vec<Ident>,
```
cdisselkoen, line 24: can we reuse Core's `Name` here, or not?

cedar-policy-validator/src/custom_schema/ast.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub(super) kind: PathKind,
+}
+
+#[derive(Debug, Clone)]
+pub(super) enum PathKind {
+    CedarBuiltin(Option<PrimOrExtension>),
+    Unqualified(Option<PrimOrExtension>),
+    Other,
```
cdisselkoen, line 32: what does `None` mean in these cases?

cedar-policy-validator/src/custom_schema/ast.rs
```diff
@@ -0,0 +1,187 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub(super) enum PrimOrExtension {
```
cdisselkoen, line 36: doc string

cedar-policy-validator/src/custom_schema/ast.rs
```diff
@@ -0,0 +1,187 @@
...
+
+#[derive(Debug, Clone)]
+pub struct Path {
+    pub base: Ident,
+    pub prefix: Vec<Ident>,
+    pub(super) kind: PathKind,
+}
+
+#[derive(Debug, Clone)]
+pub(super) enum PathKind {
```
cdisselkoen, line 29: doc string

cedar-policy-validator/src/custom_schema/ast.rs
```diff
@@ -0,0 +1,187 @@
...
+        }
+    }
+
+    fn parse_prim_or_ext(base: &Ident) -> Option<PrimOrExtension> {
+        match base.node.as_ref() {
+            IPADDR_EXTENSION => Some(PrimOrExtension::IpAddr),
+            DECIMAL_EXTENSION => Some(PrimOrExtension::Decimal),
+            _ => Self::parse_prim_type(base).map(PrimOrExtension::Prim),
+        }
+    }
```
cdisselkoen, line 73: not actually related to this code, but this made me think of it: We should have thorough tests for "Rules 1-5" in the RFC text.  For instance, that primitive vs entity name collisions are resolved in favor of entity, etc.  Perhaps we already have this, I'm just making a note here as I continue my review

cedar-policy-validator/src/custom_schema/ast.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    fn parse_prim_type(base: &Ident) -> Option<PrimitiveType> {
+        match base.node.as_ref() {
+            "Bool" => Some(PrimitiveType::Bool),
+            "String" => Some(PrimitiveType::String),
+            "Long" => Some(PrimitiveType::Long),
+            _ => None,
+        }
+    }
```
cdisselkoen, line 82: should this be a `FromStr for PrimitiveType` implementation instead?

cedar-policy-validator/src/custom_schema/ast.rs
```diff
@@ -0,0 +1,187 @@
...
+pub enum Type {
+    Set(Box<Node<Type>>),
+    Ident(Path),
+    Record(Vec<Node<AttrDecl>>),
+}
+
+#[derive(Debug, Clone)]
+pub struct AttrDecl {
+    pub name: Name,
+    pub required: Option<()>,
```
cdisselkoen, line 154: `bool` instead of `Option<()>`?

cedar-policy-validator/src/custom_schema/ast.rs
```diff
@@ -0,0 +1,187 @@
...
+pub enum AppDecl {
+    PR(PRAppDecl),
+    Context(Vec<Node<AttrDecl>>),
+}
+
+#[derive(Debug, Clone)]
+pub struct Ref {
+    pub ty: Option<Path>,
+    pub id: Name,
+}
```
cdisselkoen, line 180: is `id: Name` correct here? can `id` be namespaced?

should this `Ref` just be `EntityUID` instead?

cedar-policy-validator/src/custom_schema/err.rs
```diff
@@ -0,0 +1,322 @@
...
+use smol_str::SmolStr;
+use thiserror::Error;
+
+use super::ast::Str;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum UserError {
+    EmptyList,
+    StringEscape(NonEmpty<UnescapeError>),
+}
```
cdisselkoen, line 21: if this is an error type, should it derive `Error` according to our style in Core/Validator crates?

cedar-policy-validator/src/custom_schema/err.rs
```diff
@@ -0,0 +1,322 @@
...
+        ("NAMESPACE", "`namespace`"),
+        ("TYPE", "`type`"),
+    ]);
+}
+
+/// For errors during parsing
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ParseError {
+    /// Error generated by lalrpop
+    ToAST(OwnedRawParseError),
```
cdisselkoen, line 52: Rust style is not to all-caps acronyms in PascalCase identifiers.  For instance, `HttpError` and `JsonError`, not `HTTPError` and `JSONError`.  We aren't super consistent with this in our existing code (except in our public API where we comply), but we might as well follow it in new code.  So this would be `ToAst`

cedar-policy-validator/src/custom_schema/err.rs
```diff
@@ -0,0 +1,322 @@
...
+}
+
+impl ParseError {
+    pub(crate) fn from_raw_parse_err(err: RawParseError<'_>) -> Self {
+        Self::ToAST(err.map_token(|token| token.to_string()))
+    }
```
cdisselkoen, line 58: should this be a `From` impl?

cedar-policy-validator/src/custom_schema/err.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    pub(crate) fn from_raw_err_recovery(recovery: RawErrorRecovery<'_>) -> Self {
+        Self::from_raw_parse_err(recovery.error)
+    }
```
cdisselkoen, line 62: should this be a `From` impl?

cedar-policy-validator/src/custom_schema/err.rs
```diff
@@ -0,0 +1,322 @@
...
+            } => SourceSpan::from(*token_start..*token_end),
+            OwnedRawParseError::ExtraToken {
+                token: (token_start, _, token_end),
+            } => SourceSpan::from(*token_start..*token_end),
+            OwnedRawParseError::User { error } => error.loc.span,
+        }
+    }
+}
+
+impl Display for ParseError {
```
cdisselkoen, line 82: can we define this using `thiserror`'s `#[error]` macros, like we do for other error types in our codebase

cedar-policy-validator/src/custom_schema/err.rs
```diff
@@ -0,0 +1,322 @@
...
+            OwnedRawParseError::ExtraToken { .. } => LabeledSpan::underline(primary_source_span),
+            OwnedRawParseError::User { .. } => LabeledSpan::underline(primary_source_span),
+        };
+        Some(Box::new(std::iter::once(labeled_span)))
+    }
+}
+
+/// Multiple parse errors.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct ParseErrors(pub Vec<ParseError>);
```
cdisselkoen, line 131: `NonEmpty` instead of `Vec`?

cedar-policy-validator/src/custom_schema/err.rs
```diff
@@ -0,0 +1,322 @@
...
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        self.first().and_then(std::error::Error::source)
+    }
+}
+
+// Except for `.related()`, everything else is forwarded to the first error, if it is present.
+// This ensures that users who only use `Display`, `.code()`, `.labels()` etc, still get rich
+// information for the first error, even if they don't realize there are multiple errors here.
+// See cedar-policy/cedar#326.
+impl Diagnostic for ParseErrors {
```
cdisselkoen, line 244: this whole file has a lot of duplication with the Core `ParseErrors` code.  I wonder if we can factor some of it out

cedar-policy-validator/src/custom_schema/err.rs
```diff
@@ -0,0 +1,322 @@
...
+        self.first().and_then(Diagnostic::diagnostic_source)
+    }
+}
+
+/// For errors during schema format conversion
+#[derive(Clone, Debug, Error, PartialEq, Eq)]
+pub enum ToJsonSchemaError {
+    /// Error raised when there are duplicate keys
+    #[error("Duplicate keys: `{0}`")]
+    DuplicateKeys(SmolStr, (Loc, Loc)),
```
cdisselkoen, line 288: named fields would probably be better than `(SmolStr, (Loc, Loc))` just for clarity.  also in other variants below

cedar-policy-validator/src/custom_schema/fmt.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+impl Display for SchemaType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            SchemaType::Type(ty) => match ty {
+                SchemaTypeVariant::Boolean => write!(f, "__cedar::Bool"),
+                SchemaTypeVariant::Entity { name } => write!(f, "{name}"),
+                SchemaTypeVariant::Extension { name } => write!(f, "__cedar::{name}"),
+                SchemaTypeVariant::Long => write!(f, "__cedar::Long"),
```
cdisselkoen, line 47: it would be great if we didn't always add `__cedar`, and only add it when necessary.  99.9% of customers will not need `__cedar` and it is ugly.

cedar-policy-validator/src/custom_schema/fmt.rs
```diff
@@ -0,0 +1,198 @@
...
+                write!(f, "[")?;
+                for et in head {
+                    write!(f, "{et}, ")?
+                }
+                write!(f, "{tail}")?;
+                write!(f, "]")?;
+                Ok(())
+            }
+        }
+        // PANIC SAFETY: input should be non-empty a slice
```
cdisselkoen, line 88:
```suggestion
        // PANIC SAFETY: input should be a non-empty slice
```

cedar-policy-validator/src/custom_schema/fmt.rs
```diff
@@ -0,0 +1,198 @@
...
+        #[allow(clippy::unreachable)]
+        None => unreachable!("input list must not be empty"),
```
cdisselkoen, line 90: This seems like an invariant that should be documented in this function's doc string, at a minimum.

cedar-policy-validator/src/custom_schema/fmt.rs
```diff
@@ -0,0 +1,198 @@
...
+                    (None, None) => {
+                        write!(f, "appliesTo {{")?;
+                        write!(f, "  context: {}", &spec.context.0)?;
+                        write!(f, "\n}}")?;
+                    }
+                }
+            }
+            // applies to unspecified principals and resources
+            None => {
+                write!(f, " appliesTo {{ context: {{}}}}")?;
```
cdisselkoen, line 158: Rust's raw-string-literal syntax might help you out here

cedar-policy-validator/src/custom_schema/test.rs
```diff
@@ -0,0 +1,597 @@
...
+            entity Public in [DocumentShare];
+            entity Drive;
+        }"#,
+        );
+        assert!(res.is_ok(), "{res:?}");
+    }
+}
+
+#[cfg(test)]
+mod parser_tests {
```
cdisselkoen, line 61: high-level comment: most of these tests only check that parse was successful.  I think our bar should be a lot higher -- we should check that it parsed into the "correct" structure, or at least check some basic properties of the parse's output.

cedar-policy-validator/src/custom_schema/to_json_schema.rs
```diff
@@ -0,0 +1,542 @@
...
+use super::{
+    ast::{
+        ActionDecl, AppDecl, AttrDecl, Declaration, EntityDecl, Namespace, PRAppDecl, Path,
+        PathKind, PrimOrExtension, Ref, Schema, Str, Type, CEDAR_NAMESPACE, PR,
+    },
+    err::ToJsonSchemaError,
+};
+
+#[derive(Debug, Clone)]
+pub(super) struct Resolver {
```
cdisselkoen, line 25: doc string?

cedar-policy-validator/src/custom_schema/to_json_schema.rs
```diff
@@ -0,0 +1,542 @@
...
+    global_common_types: HashSet<Node<Name>>,
+    global_entity_types: HashSet<Node<Name>>,
+}
+
+pub(super) type LookupFn<'a> = Box<dyn Fn(&Path) -> Option<NamedType> + 'a>;
```
cdisselkoen, line 30: doc string?

cedar-policy-validator/src/custom_schema/to_json_schema.rs
```diff
@@ -0,0 +1,542 @@
...
+                }
+            }
+        }
+        Resolver {
+            global_common_types: common_tys,
+            global_entity_types: ets,
+        }
+    }
+
+    pub(super) fn get_lookup_fn(&self, namespace: &Namespace) -> (LookupFn, TypeNameCollisions) {
```
cdisselkoen, line 65: doc string?

cedar-policy-validator/src/custom_schema/to_json_schema.rs
```diff
@@ -0,0 +1,542 @@
...
+            Some(NamedType::Common(node.clone().into()))
+        } else if entity_types.contains(name) || self.global_entity_types.contains(name) {
+            Some(NamedType::Entity(node.clone().into()))
+        } else {
+            None
+        }
+    }
+}
+
+pub type TypeNameCollisions = Vec<TypeNameCollision>;
```
cdisselkoen, line 166: `NonEmpty<TypeNameCollision>`?

cedar-policy-validator/src/custom_schema/to_json_schema.rs
```diff
@@ -0,0 +1,542 @@
...
+
+#[derive(Debug, Clone)]
+pub enum TypeNameCollision {
+    CommonTypeAndEntityType(Str, Str),
+    Builtin(Str),
```
cdisselkoen, line 171: doc strings?  In particular, what's the `Builtin` -- a `Str` colliding with itself?  (I assume a `Str` colliding with an implied builtin, but a doc string would clarify)

cedar-policy-validator/src/custom_schema/to_json_schema.rs
```diff
@@ -0,0 +1,542 @@
...
+        Either::Right(s) => s,
+    }
+}
+
+fn ref_to_action_euid(name: Ref) -> ActionEntityUID {
+    let Ref { ty, id } = name;
+    ActionEntityUID {
+        id: name_to_str(id).node,
+        ty: ty.map(|p| p.into()),
+    }
```
cdisselkoen, line 216:
```suggestion
    ActionEntityUID {
        id: name_to_str(name.id).node,
        ty: name.ty.map(|p| p.into()),
    }
```

cedar-policy-validator/src/custom_schema/to_json_schema.rs
```diff
@@ -0,0 +1,542 @@
...
+                Some(principal_types.into())
+            } else {
+                None
+            },
+            context: attr_lookup_func.map_or(AttributesOrContext::default(), |attrs| attrs),
+        })
+    }
+}
+
+/// If there is exactly one PRDecl here, map it to it's list of entity types
```
cdisselkoen, line 356:
```suggestion
/// If there is exactly one PRDecl here, map it to its list of entity types
```

cedar-policy-validator/src/custom_schema/to_json_schema.rs
```diff
@@ -0,0 +1,542 @@
...
+    name: &'static str,
+) -> Result<Option<NonEmpty<SmolStr>>, ToJsonSchemaError> {
+    Ok(exactly_once(input, name)?.map(|decl| decl.entity_tys.map(|et| et.node.into())))
+}
+
+/// Fold over the list, producing the following:
+///
+/// * Err(_) -> if there is more than one element in the array
+/// * Ok(None) -> if there are no elements in the array
+/// * Ok(Some(_)) -> if there is exactly one element in the array
```
cdisselkoen, line 368: this is `Itertools::exactly_one()` (https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.exactly_one)

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -78,6 +78,19 @@ impl NamespaceDefinition {
             actions: actions.into_iter().collect(),
         }
     }
+
+    pub fn merge_actions_by_type(&self) -> Vec<(&ActionType, HashSet<SmolStr>)> {
```
cdisselkoen, line 82: doc string? also why return a `Vec` of pairs instead of a `HashMap`?

cdisselkoen (changes requested):
Not all of these comments need to be blocking for merging this PR, but I think some of them are important to address before merging

cedar-policy-validator/src/custom_schema/test.rs
cdisselkoen: maybe the correct answer is roundtrip tests JSON->Human->JSON and Human->JSON->Human, similar to the tests we have for EST in Core

aaronjeline:
> BTW - one feature request here - please sort the keys of the output json, I'm having to pass it through jq to stabilize the output json (and make diffs nice)

Can try to do this for cosmetic reasons but we can't guarantee order won't change in the future or by whatever JSON parser you are using.

cedar-policy-validator/src/custom_schema/ast.rs
aaronjeline: No because these fields have `Loc`s attached

cdisselkoen: Got it

cedar-policy-validator/src/custom_schema/ast.rs
cdisselkoen: `ty` is a confusing name for a namespace.

So I'm to understand that if `ty` is present, it is concatenated on the left of `id`, but if `ty` is not present, we concatenate the implicit schema namespace on the left of `id` instead?

cdisselkoen (commented):
Another comment, needs changelog entries in both the `cedar-policy` and `cedar-policy-cli` CHANGELOGs

cedar-policy-validator/src/schema_file_format.rs
shaobo-he-aws: I don't think this function is used anywhere now.

cedar-policy-cli/src/lib.rs
```diff
@@ -94,10 +94,48 @@ pub enum Commands {
     Link(LinkArgs),
     /// Format a policy set
     Format(FormatArgs),
+    /// Translate JSON schema to custom schema syntax and vice versa (except comments)
```
cdisselkoen, line 97:
```suggestion
    /// Translate JSON schema to natural schema syntax and vice versa (except comments)
```

cedar-policy-cli/src/lib.rs
```diff
@@ -94,10 +94,48 @@ pub enum Commands {
...
+    /// JSON -> Natural schema syntax
+    JsonToNatural,
+    /// Natural schema syntax -> JSON
+    NaturalToJson,
+}
+
+#[derive(Debug, Clone, Copy, ValueEnum)]
+pub enum SchemaFormat {
+    /// Default human-readable format
+    Human,
```
cdisselkoen, line 128: here we use "Human" but in other help text we use "Natural".  We should be consistent, at least in things visible to CLI users in the CLI commands and help

cedar-policy-cli/src/lib.rs
```diff
@@ -588,6 +639,41 @@ pub fn format_policies(args: &FormatArgs) -> CedarExitCode {
...
+fn translate_to_natural(json_src: impl AsRef<str>) -> Result<String> {
+    let fragment = SchemaFragment::from_str(json_src.as_ref())?;
+    let output = fragment.as_natural()?;
+    Ok(output)
+}
+
+fn translate_to_json(natural_src: impl AsRef<str>) -> Result<String> {
+    let (fragment, warnings) = SchemaFragment::from_str_natural(natural_src.as_ref())?;
+    for warning in warnings {
+        eprintln!("{warning}");
```
cdisselkoen, line 651: if these are `miette::Diagnostic` then we can make a `miette::Report` containing all the warnings, and print that (with Debug, because that's how you get the pretty miette output; the Display is still always just the main error message)

cedar-policy-cli/src/lib.rs
```diff
@@ -588,6 +639,41 @@ pub fn format_policies(args: &FormatArgs) -> CedarExitCode {
...
+    let output = fragment.as_json_string()?;
+    Ok(output)
+}
+
+fn translate_schema_inner(args: &TranslateSchemaArgs) -> Result<String> {
+    let translate = match args.direction {
+        TranslationDirection::JsonToNatural => translate_to_natural,
+        TranslationDirection::NaturalToJson => translate_to_json,
+    };
+    read_from_file_or_stdin(args.input_file.clone(), "translate schema").and_then(translate)
```
cdisselkoen, line 662:
```suggestion
    read_from_file_or_stdin(args.input_file.clone(), "schema").and_then(translate)
```

cedar-policy-cli/src/lib.rs
```diff
@@ -588,6 +639,41 @@ pub fn format_policies(args: &FormatArgs) -> CedarExitCode {
...
+}
+pub fn translate_schema(args: &TranslateSchemaArgs) -> CedarExitCode {
+    match translate_schema_inner(args) {
+        Ok(sf) => {
+            println!("{sf}");
+            CedarExitCode::Success
+        }
+        Err(err) => {
+            eprintln!("Error: {err:?}");
```
cdisselkoen, line 671: Assuming it's `miette`, we don't want/need the `Error: ` prefix

```suggestion
            eprintln!("{err:?}");
```

cedar-policy-cli/src/lib.rs
```diff
@@ -94,10 +94,48 @@ pub enum Commands {
     Link(LinkArgs),
     /// Format a policy set
     Format(FormatArgs),
+    /// Translate JSON schema to custom schema syntax and vice versa (except comments)
+    TranslateSchema(TranslateSchemaArgs),
```
cdisselkoen, line 98: Should we also say in the doc comment that the translated schema will be printed on stdout?  or is that implied?

cedar-policy-cli/src/lib.rs
```diff
@@ -1010,14 +1097,26 @@ fn read_json_policy(filename: Option<impl AsRef<Path> + std::marker::Copy>) -> R
...
+            format!(
+                "failed to parse schema from file {}",
+                filename.as_ref().display()
+            )
+        }),
+        SchemaFormat::Human => {
+            let (schema, warnings) = Schema::from_str_natural(&schema_src)?;
+            for warning in warnings {
+                eprintln!("{warning}");
+            }
```
cdisselkoen, line 1116: same comment about `miette::Report`

cedar-policy-core/src/parser/node.rs
```diff
@@ -72,6 +72,11 @@ impl<T> Node<T> {
     pub fn to_ast_err(&self, error_kind: impl Into<ToASTErrorKind>) -> ToASTError {
         ToASTError::new(error_kind.into(), self.loc.clone())
     }
+
+    pub fn swap<A>(self, a: A) -> (T, Node<A>) {
+        let (t, loc) = self.into_inner();
+        (t, Node::with_source_loc(a, loc))
+    }
```
cdisselkoen, line 79: doc comment

cedar-policy-validator/src/custom_schema/ast.rs
```diff
@@ -0,0 +1,299 @@
...
+pub struct Path(pub Node<NonEmpty<Id>>);
+
+impl Path {
+    /// Create a [`Path`] with a single entry
+    pub fn single(node: Id, loc: Loc) -> Self {
+        Self(Node::with_source_loc(NonEmpty::new(node), loc))
+    }
+
+    /// Create [`Path`] with a head and an iterator
+    pub fn new(head: Id, rest: impl IntoIterator<Item = Id>, loc: Loc) -> Self {
```
cdisselkoen, line 36: for the doc comment: is `head` the first component of the path, or the last component (the actual typename)?

cedar-policy-validator/src/custom_schema/ast.rs
```diff
@@ -0,0 +1,299 @@
...
+        &self.0.loc
+    }
+
+    /// Consume the [`Path`] and get an owned iterator over the elements
+    pub fn into_iter(self) -> impl Iterator<Item = Node<Id>> {
+        let loc = self.0.loc;
+        self.0
+            .node
+            .into_iter()
+            .map(move |x| Node::with_source_loc(x, loc.clone()))
```
cdisselkoen, line 57: This gives the `loc` of the entire Path as the `loc` of each component.  Is that intended?

cedar-policy-validator/src/custom_schema/ast.rs
```diff
@@ -0,0 +1,299 @@
...
+        let s = self.0.node.iter().map(|id| id.as_ref()).join("::");
+        write!(f, "{s}")
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct QualName {
+    pub path: Option<Path>,
+    pub id: SmolStr,
+}
```
cdisselkoen, line 86: two comments

1) I thought `Path` already was an optional number of namespaces, plus a final `id`
2) `id` is a confusing name for a field of type `SmolStr`.  is it an identifier or an arbitrary string?

basically I'm unsure at a glance whether this represents a (potentially qualified) typename, or an EntityUid

cedar-policy-validator/src/custom_schema/ast.rs
```diff
@@ -0,0 +1,299 @@
...
+pub fn is_extension_name(s: impl AsRef<str>) -> bool {
+    EXTENSIONS.contains(&s.as_ref())
+}
+
+/// A [`Namespace`] has a name and a collection declaration
+/// A schema is made up of a series of fragments
+/// A fragment is a series of namespaces
+#[derive(Debug, Clone)]
+pub struct Namespace {
+    /// The name of this name. If [`None`], then this is the unqualified namespace
```
cdisselkoen, line 111:
```suggestion
    /// The name of this namespace. If [`None`], then this is the unqualified namespace
```

cedar-policy-validator/src/custom_schema/ast.rs
```diff
@@ -0,0 +1,299 @@
...
+    /// Is this [`Namespace`] the builtin `__cedar` namespace?
+    pub fn is_cedar_namespace(&self) -> bool {
+        match &self.name {
+            Some(path) => path.iter().exactly_one().map(is_cedar).unwrap_or(false),
+            None => false,
+        }
+    }
+
+    /// Does this [`Namespace`] start with `__`, the reserved pattern
+    pub fn is_reserved_namespaces(&self) -> bool {
```
cdisselkoen, line 132:
```suggestion
    pub fn is_reserved_namespace(&self) -> bool {
```

cedar-policy-validator/src/err.rs
```diff
@@ -128,6 +130,12 @@ pub enum SchemaError {
     #[error("the `__expr` escape is no longer supported")]
     #[diagnostic(help("to create an entity reference, use `__entity`; to create an extension value, use `__extn`; and for all other values, use JSON directly"))]
     ExprEscapeUsed,
+    #[error("{0}")]
+    NaturalSyntaxError(#[from] NaturalSyntaxParseErrors),
+    #[error("{0}")]
+    IOError(#[from] std::io::Error),
+    #[error("{0}")]
+    ToNaturalSyntaxError(#[from] custom_schema::ToCustomSchemaStrError),
```
cdisselkoen, line 138: We should consider adding `#[diagnostic(transparent)]` to these, if any of these contained types impl `Diagnostic`

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -38,16 +41,37 @@ pub struct SchemaFragment(
...
         serde_json::from_value(json).map_err(Into::into)
     }
 
-    /// Create a `SchemaFragment` directly from a file.
+    /// Create a [`SchemaFragment`] directly from a file containing a JSON object.
     pub fn from_file(file: impl std::io::Read) -> Result<Self> {
         serde_json::from_reader(file).map_err(Into::into)
     }
+
+    /// Parse the schema format from a string
```
cdisselkoen, line 55:
```suggestion
    /// Parse the schema (in natural schema syntax) from a string
```

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -38,16 +41,37 @@ pub struct SchemaFragment(
...
+    pub fn from_str_natural(src: &str) -> Result<(Self, impl Iterator<Item = SchemaWarning>)> {
+        let tup = parse_natural_schema_fragment(src)?;
+        Ok(tup)
+    }
+
+    /// Parse the schema format from a reader
```
cdisselkoen, line 61:
```suggestion
    /// Parse the schema (in natural schema syntax) from a reader
```

cedar-policy/src/api.rs
```diff
@@ -1237,6 +1314,52 @@ pub enum SchemaError {
...
     ExprEscapeUsed,
+    /// Errors parsing the natural schema syntax.
+    #[error("{0}")]
+    NaturalSyntaxError(#[from] NaturalSyntaxError),
+    /// IO errors when calling `from_reader`
+    #[error("I/O error while reading schmea: {0}")]
+    IOError(#[from] std::io::Error),
+    /// Error converting a into natural syntax
+    #[error("{0}")]
+    ToNaturalSyntax(#[from] ToNaturalSyntaxError),
```
cdisselkoen, line 1325: same here, we should consider `#[diagnostic(transparent)]`

cedar-policy/src/api.rs
```diff
@@ -1237,6 +1314,52 @@ pub enum SchemaError {
...
+}
+
+/// Errors serializing Schemas to the natural syntax
+#[derive(Debug, Error, Diagnostic)]
+pub enum ToNaturalSyntaxError {
+    /// Duplicate names were found in the schema
+    #[error("There are type name collisions: [{}]", .0.iter().join(","))]
```
cdisselkoen, line 1332:
```suggestion
    #[error("There are type name collisions: [{}]", .0.iter().join(", "))]
```

cedar-policy/src/api.rs
```diff
@@ -1237,6 +1314,52 @@ pub enum SchemaError {
...
+                collisions,
+            ) => Self::NameCollisions(collisions),
+        }
+    }
+}
+
+/// Errors when parsing natural schema syntax
+#[derive(Debug, Diagnostic, Error)]
+pub enum NaturalSyntaxError {
+    /// Error Parsing a Schema in natural syntax
```
cdisselkoen, line 1349:
```suggestion
    /// Error parsing a schema in natural syntax
```

cedar-policy/src/api.rs
```diff
@@ -1237,6 +1314,52 @@ pub enum SchemaError {
...
+
+impl From<cedar_policy_validator::custom_schema::parser::NaturalSyntaxParseErrors>
+    for NaturalSyntaxError
+{
+    fn from(
+        value: cedar_policy_validator::custom_schema::parser::NaturalSyntaxParseErrors,
+    ) -> Self {
+        Self::ParseError(value)
+    }
 }
```
cdisselkoen, line 1363: instead of this I think we can just use `#[from]` in the declaration of `NaturalSyntaxError::ParseError`

cdisselkoen (commented):
looking great, this is coming together

cedar-policy-cli/sample-data/sandbox_a/schema.cedarschema
cdisselkoen: Fixed

cedar-policy/src/api.rs
cdisselkoen: Fixed

cedar-policy/src/api.rs
cdisselkoen: Fixed

cedar-policy-validator/src/err.rs
cdisselkoen: Fixed

cedar-policy-core/src/parser/node.rs
cdisselkoen: Resolving this, looks like this function was just removed entirely

cedar-policy-validator/src/schema_file_format.rs
cdisselkoen: Removed it

cedar-policy-cli/src/lib.rs
aaronjeline: I think that's obvious

cedar-policy-validator/src/custom_schema/ast.rs
aaronjeline: Yeah, by this point we don't have the individual locs, this way at least the path will get updated

cedar-policy-validator/src/custom_schema/ast.rs
aaronjeline: EntityUId. A path is a nonempty set of namespaces, but they all have to be `id`s. An actions entity uid can be any string

aaronjeline:
One todo, the warnings in cli currently present as errors

cedar-policy-cli/src/lib.rs
```diff
@@ -94,10 +94,48 @@ pub enum Commands {
...
     New(NewArgs),
 }
 
+#[derive(Args, Debug)]
+pub struct TranslateSchemaArgs {
+    /// The direction of translation,
+    ///   JSON -> Human
+    ///   Human -> JSON
+    #[arg(long)]
+    pub direction: TranslationDirection,
```
cdisselkoen, line 109: does this doc comment render correctly in `cedar help translate-schema`?

cedar-policy-cli/src/lib.rs
```diff
@@ -94,10 +94,48 @@ pub enum Commands {
...
+    HumanToJson,
+}
+
+#[derive(Debug, Clone, Copy, ValueEnum)]
+pub enum SchemaFormat {
+    /// Human-readable format
+    Human,
+    /// JSON format
+    Json,
+}
```
cdisselkoen, line 131: did we decide Human vs Natural in the CLI interface?

cedar-policy-cli/src/lib.rs
```diff
@@ -114,6 +152,9 @@ pub struct ValidateArgs {
     /// experimental `partial-validate` feature enabled.
     #[arg(long = "partial-validate")]
     pub partial_validate: bool,
+    /// Schema format (Human-readable or yes)
```
cdisselkoen, line 155:
```suggestion
    /// Schema format (Human-readable or json)
```

cedar-policy-cli/src/lib.rs
```diff
@@ -588,6 +639,42 @@ pub fn format_policies(args: &FormatArgs) -> CedarExitCode {
...
+    read_from_file_or_stdin(args.input_file.clone(), "schema").and_then(translate)
+}
+pub fn translate_schema(args: &TranslateSchemaArgs) -> CedarExitCode {
+    match translate_schema_inner(args) {
+        Ok(sf) => {
+            println!("{sf}");
+            CedarExitCode::Success
+        }
+        Err(err) => {
+            eprintln!("{err:?}");
```
cdisselkoen, line 672: what is the type of `err` here? ideally we'd get `miette::Report` for prettiness.  I know we install some `miette` hooks at the toplevel which maybe take care of some coercing for us?

cedar-policy-validator/src/human_schema/ast.rs
```diff
@@ -0,0 +1,441 @@
...
+        Self(Node::with_source_loc(
+            PathInternal {
+                basename,
+                namespace: vec![],
+            },
+            loc,
+        ))
+    }
+
+    /// Create [`Path`] with a head and an iterator. Most significant name first.
```
cdisselkoen, line 37: I read "Most significant" as being leftmost, but then this function sticks it in `basename` which I interpret as rightmost?  but maybe I just interpret one of those terms weirdly / different from everyone else

cedar-policy-validator/src/human_schema/ast.rs
```diff
@@ -0,0 +1,441 @@
...
+            .into_iter()
+            .map(move |x| Node::with_source_loc(x, loc.clone()))
+    }
+
+    /// Get the base type name as well as the (potentially empty) prefix
+    pub fn split_last(self) -> (Vec<Id>, Id) {
+        (self.0.node.namespace, self.0.node.basename)
+    }
+
+    /// Is this referring to a name _in__ the __cedar namespace: ex: __cedar::Bool or the unqualified namespace
```
cdisselkoen, line 73:
```suggestion
    /// Is this referring to a name in the `__cedar` namespace (eg: `__cedar::Bool`) or the unqualified namespace
```

cedar-policy-validator/src/human_schema/ast.rs
```diff
@@ -0,0 +1,441 @@
...
+    pub fn is_in_unqualified_or_cedar(&self) -> bool {
+        self.0.node.is_in_unqualified_or_cedar()
+    }
+
+    /// Is this referring to a name _in__ the __cedar namespace: ex: __cedar::Bool
```
cdisselkoen, line 78:
```suggestion
    /// Is this referring to a name in the `__cedar` namespace (eg: `__cedar::Bool`)
```

cedar-policy-validator/src/human_schema/ast.rs
cdisselkoen: oh this means most significant _inside the iterator_ first.  the parameter name `basename` is sufficient documentation for what that represents

cedar-policy-validator/src/human_schema/ast.rs
```diff
@@ -0,0 +1,441 @@
...
+}
+
+impl Namespace {
+    /// Is this [`Namespace`] unqualfiied?
+    pub fn is_unqualified(&self) -> bool {
+        self.name.is_none()
+    }
+
+    /// Does this [`Namespace`] start with `__`, the reserved pattern
+    pub fn is_reserved_namespace(&self) -> bool {
```
cdisselkoen, line 179: [RFC 52](https://github.com/cedar-policy/rfcs/pull/52) isn't accepted yet, are we still ok here?

cedar-policy/src/api.rs
```diff
@@ -1237,6 +1314,44 @@ pub enum SchemaError {
...
+    #[diagnostic(transparent)]
+    NaturalSyntaxError(#[from] NaturalSyntaxError),
+    /// IO errors when calling `from_reader`
+    #[error(transparent)]
+    IOError(#[from] std::io::Error),
+    /// Error converting a schema into natural syntax
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    ToNaturalSyntax(#[from] ToNaturalSyntaxError),
+}
```
cdisselkoen, line 1328: `SchemaError` isn't marked `non_exhaustive`, so adding variants is technically a breaking change :(

cdisselkoen (commented):
Getting close.  I'm concerned about the `SchemaError` change in `cedar-policy`, as noted there

cedar-policy-cli/src/lib.rs
aaronjeline: `Human` as we already called it `human` for policies

cedar-policy-validator/src/human_schema/ast.rs
aaronjeline: I can disable this for now yeah

cedar-policy-cli/src/lib.rs
aaronjeline: Yup, it's a `Report`

cdisselkoen (approved):

cedar-policy-validator/build.rs
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright 2022-2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andrewmwells-amazon, line 2: I believe Legal signed off on changing this to `Copyright Cedar Contributors` and removing the dates. If we don't want to do that yet, we should use 2024.

andrewmwells-amazon (approved):

PR #259, opened 2023-08-04
Add api docstring examples

anwarmamat:
## Description of changes
Adds API docstring examples

## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [ ] A breaking change requiring a major version bump to `cedar-policy` (e.g., changes to the signature of an existing API).
- [ ] A backwards-compatible change requiring a minor version bump to `cedar-policy` (e.g., addition of a new API).
- [ ] A bug fix or other functionality change requiring a patch to `cedar-policy`.
- [X] A change "invisible" to users (e.g., documentation, changes to "internal" crates like `cedar-policy-core`, `cedar-validator`, etc.)
- [ ] A change (breaking or otherwise) that only impacts unreleased or experimental code.

I confirm that this PR (choose one, and delete the other options):

- [ ] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).
- [X] Does not update the CHANGELOG because my change does not significantly impact released code.

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [X] Does not require updates because my change does not impact the Cedar Dafny model or DRT infrastructure.
- [ ] Requires updates, and I have made / will make these updates myself. (Please include in your description a timeline or link to the relevant PR in `cedar-spec`, and how you have tested that your updates are correct.)
- [ ] Requires updates, but I do not plan to make them in the near future. (Make sure that your changes are hidden behind a feature flag to mark them as experimental.)
- [ ] I'm not sure how my change impacts `cedar-spec`. (Post your PR anyways, and we'll discuss in the comments.)

## Disclaimer

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #64, opened 2023-05-24
More ergonomic ways of writing Ids


aaronjeline:
### Category

User level API changes

### Describe the feature you'd like to request

It would be nice to have more ergonomic ways of writing down policy and entity ids. Currently you have write them as a string, and then call `.parse()` on them. For fully static data, this is cumbersome and delays error detection until runtime. 

### Describe the solution you'd like

One solution here would be to have macros for writing ids. You could image something like:
```rust
entityuid!(#r"User::"Alice""#)
```
Parse rules would be checked at compile time, allowing you skip calling `.unwrap` or handling the error, and making it compile time error if the ID was invalid. 

### Describe alternatives you've considered

The current situation is workable.

### Additional context

This is somewhat inspired by `serde_json`s `json!` macro.

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

cdisselkoen:
In addition to policy and entity ids, we can also do this for `Name`s in general.

shaobo-he-aws:
I wrote a prototype in #744, which proposes a different syntax. Instead of having an argument which is the string representation of entity uids, it asks for two argument: an entity type and an eid. The eid is simply a string. This design avoids the nuance around string escaping and provides users a more flexible way to specify eids. For instance, they can use raw string syntax when they need literal backslashes like `euid!(Foo, r#"\"#).

PR #561, opened 2024-01-02
Simplify Doc Tests

john-h-kastner-aws:
## Description of changes

This PR tries to make our doc tests easier to understand at a glance by removing unnecessary detail. 

* Hide `use` statements everywhere. It's implied that things will need to be imported from our library, and an IDE should be able to generate these automatically.
* Hide asserts where not relevant to the specific function. On a case-by-case basis, some asserts aren't necessary to demonstrate a function. Asserts are left as hidden because they are useful as tests. 
* ~~Construct entities using `EntityUid::from_str`. This occupies less space on the page, and IMO is easier to mentally parse than constructing an `EntityId` and `EntityTypeName` first.~~

## Disclaimer

### DIFF
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -3515,16 +3515,9 @@ impl Context {
     ///   ("age".to_string(), RestrictedExpression::from_str("18").unwrap()),
     /// ]).unwrap();
     /// # // create a request
-    /// # let p_eid = EntityId::from_str("alice").unwrap();
-    /// # let p_name: EntityTypeName = EntityTypeName::from_str("User").unwrap();
-    /// # let p = EntityUid::from_type_name_and_id(p_name, p_eid);
-    /// #
-    /// # let a_eid = EntityId::from_str("view").unwrap();
-    /// # let a_name: EntityTypeName = EntityTypeName::from_str("Action").unwrap();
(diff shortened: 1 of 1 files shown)


khieta (approved):
Lgtm, but I seem to recall that we switched to using `EntityUid::from_type_name_and_id` because we preferred use of that constructor over `EntityUid::from_str` to avoid confusion over handling of escaped chars in the eid. Can another reviewer confirm that we're ok with switching back to the old constructor?

(Personally, I'm fine with it because it makes the examples more readable.)

john-h-kastner-aws:
> we switched to using `EntityUid::from_type_name_and_id` because we preferred use of that constructor over `EntityUid::from_str`.

Make sense. We should generally encourage using `from_type_name_and_id`. 

That said, I think their value as examples is helped more by improved clarity than it is hurt by not using best-practices to construct these example entities (but the real solution here is #64 if someone's got time for it). I looked through #259 and didn't see anyone commenting on this precise issue, but anyone else who commented on that should take a look. 

For reference for future commenters, here is the example for `Entity::new`.

Before: 

image

After:

image

john-h-kastner-aws:
Update to remove change to using `EntityUid::from_str` except in some hidden code, since I like it better there.

andrewmwells-amazon (approved):

Issue #570, opened 2024-01-05
Partial evaluation is not exposed to FFIs


ygrignon-sfdc:
### Category

User level API changes

### Describe the feature you'd like to request

There should be an FFI method `cedar::frontend::is_authorized::json_is_authorized_partial` providing the partial evaluation equivalent to `cedar::frontend::is_authorized::json_is_authorized` just like there is `Authorizer::is_authorized_partial` providing the partial evaluation equivalent to `Authorizer::is_authorized_partial`

### Describe alternatives you've considered

Supporting both regular and partial evaluation in `cedar::frontend::is_authorized::json_is_authorized` by adding a boolean to the AuthorizationCall structure. This would introduce a breaking change forcing integrators to support different response formats.

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #571, opened 2024-01-05
FFI support for partial evaluation (for discussion)

ygrignon-sfdc:
## Description of changes
Adding support for partial evaluation for FFIs via the new json_is_authorized_partial function

## Issue #, if available
#570

- A backwards-compatible change requiring a minor version bump to `cedar-policy` (e.g., addition of a new API).

- Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).

- Does not require updates because my change does not impact the Cedar Dafny model or DRT infrastructure.

## Disclaimer


cedar-policy-core/src/authorizer.rs
```diff
@@ -892,4 +892,8 @@ pub enum Decision {
     /// that no decision could be safely reached; for example, errors parsing
     /// the policies.
     Deny,
+    /// The `Authorizer` determined that some information required to reach
+    /// a decision is missing.
+    #[cfg(feature = "partial-eval")]
+    NoDecision,
```
john-h-kastner-aws, line 898: One initial comment from a cursory look at the code: I'm a bit concerned with adding `NoDecision` to this struct. We eventually want to stabilize partial evaluation which would mean removing the feature flag (or enabling it by default), at which point non-partial-eval users would need to know to ignore this variant.

john-h-kastner-aws (commented):
Hi @ygrignon-sfdc, thanks for the contribution! We'd all be excited to see partial evaluation added to the json interface (and presumably language bindings later on). We'll take a closer look and get back to you with any comments on the design/code.

cedar-policy-core/src/authorizer.rs
ygrignon-sfdc: One way to avoid this would be to have a completely different interface response format for residual decisions that wouldn't include a decision member at all. Interestingly, [com.cedarpolicy.model.AuthorizationResponse.Decision](https://github.com/cedar-policy/cedar-java/blob/main/CedarJava/src/main/java/com/cedarpolicy/model/AuthorizationResponse.java#L48) in cedar-java/CedarJava already has a NoDecision enum value.

cdisselkoen: That's true, but there is already an open issue to remove the `NoDecision` enum value: https://github.com/cedar-policy/cedar-java/issues/37.  (Obviously this could be reconsidered)

ygrignon-sfdc: I removed the NoDecision and reworked the logic. I think this solution is cleaner. The JSON object for a concrete response is exactly the same as what would be produced by json_is_authorized(). The JSON object produced when residuals policies are returned only contains residuals and diagnostics keys.

cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -135,11 +188,54 @@ impl InterfaceDiagnostics {
...
+}
+
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(untagged)]
 enum AuthorizationAnswer {
     ParseFailed { errors: Vec<String> },
-    Success { response: InterfaceResponse },
+    Concrete { response: InterfaceResponse },
+    #[cfg(feature = "partial-eval")]
+    Residuals { response: InterfaceResidualResponse },
```
andrewmwells-amazon, line 238: I think we may want to have a separate struct `PartialAuthorizationAnswer` and keep this unmodified. If we turn the `parial-eval` flag on by default at some point, I think it will make more sense to have them separate.

andrewmwells-amazon:
Looks good overall.

cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -62,7 +64,44 @@ pub fn json_is_authorized(input: &str) -> InterfaceResult {
...
+            )
+        },
+        Err(errors) => AuthorizationAnswer::ParseFailed { errors },
+    }
+}
+
+/// public string-based JSON interfaced to be invoked by FFIs. In the policies portion of
+/// the `RecvdSlice`, you can either pass a `Map<String, String>` where the values are all single policies,
+/// or a single String which is a concatenation of multiple policies. If you choose the latter,
+/// policy id's will be auto-generated for you in the format `policyX` where X is a Whole Number (zero or a positive int)
```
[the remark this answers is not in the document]
ygrignon-sfdc: This was a cut'n'paste from the comment on is_authorized but I can change both np

cedar-policy/src/frontend/is_authorized.rs
ygrignon-sfdc: This enum is private to the module and only used to drive JSON serialization

cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -1398,4 +1559,154 @@ mod test {
         }"#;
         assert_is_failure(&json_is_authorized(call), true, "found duplicate key");
     }
+
+    #[cfg(feature = "partial-eval")]
```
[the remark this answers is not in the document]
ygrignon-sfdc: It looks like in Rust individual files and directories are implicit modules however it doesn't seem like the `#[cfg(feature = "...")]` construct can be used at the file level. These's an explicit `mod` keyword that can presumably be used to define sub-modules with a single file but then we would end up with `cedar_policy::frontend::is_authorized::partial::json_is_authorized_partial` as the public method signature.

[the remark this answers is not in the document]
john-h-kastner-aws: I think Andrew is suggesting that the tests could be a module like

```
#[cfg(feature = "partial-eval")]
#[cfg(test)]
mod test_partial {
...
}
```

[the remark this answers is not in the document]
ygrignon-sfdc: Ah, that makes sense. I thought that applied to all diffs behind the feature flag.

cedar-policy/src/frontend/is_authorized.rs
john-h-kastner-aws: IMO this is probably fine for the JSON format.  When we document the interface we should document it as the JSON objects for each are  distinct separate types, but unifying them in the implementation is fine.

cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -212,6 +308,45 @@ impl AuthorizationCall {
...
+                EntityUid::from_json(r.into())
+                    .map_err(|e| ["Failed to parse resource".into(), e.to_string()])?,
+            ),
+            None => None,
+        };
+
+        let context = serde_json::to_value(self.context)
+            .map_err(|e| [format!("Error encoding the context as JSON: {e}")])?;
+        let context = Context::from_json_value(context, schema.as_ref().map(|s| (s, &action)))
+            .map_err(|e| [e.to_string()])?;
```
john-h-kastner-aws, line 339: The this logic be easily shared with `get_components`?

john-h-kastner-aws (approved):
Looks good to me. 

I'll start the CI run (but a few targets are currently expected to fail due an issue in the scripts)

cedar-policy/src/frontend/is_authorized.rs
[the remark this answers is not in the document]
cdisselkoen: > it doesn't seem like the `#[cfg(feature = "...")]` construct can be used at the file level

It can, you need to add `!`, like so: `#![cfg(feature = "...")]`

cedar-policy/src/frontend/is_authorized.rs
ygrignon-sfdc: I just changed it as per @andrewmwells-amazon 's recommendation. At the end of the day, it's easy to change and private to the module so you do whatever you want with it ;)

andrewmwells-amazon:
Looking into the build failures. For one, you can run `cargo fmt` to fix. Alternatively, if you're happy with the current state, I can take over the PR to get CI passing.

andrewmwells-amazon:
The other failures should hopefully be fixed by rebasing to pick up https://github.com/cedar-policy/cedar/pull/574

cedar-policy/src/frontend/is_authorized.rs
ygrignon-sfdc: Not with my level of proficiency in Rust. I think something like this would work with similar code for get_component_partial(). But I'm getting move issues and I'm not sure to get over them.
```
impl AuthorizationCall {
    fn parse_schema(self) -> Result<Option<Schema>, Vec<String>> {
        self.schema
            .map(|v| Schema::from_json_value(v.into()))
            .transpose()
            .map_err(|e| vec!(e.to_string()))
    }
    ...
    fn get_components(self) -> Result<(Request, PolicySet, Entities), Vec<String>> {
        let schema = self.parse_schema()?;
        let principal = self.parse_principal()?;
        let action = self.parse_action()?;
        let resource = self.parse_resource()?;
        let context = self.parse_context(schema, &action)?;
        ...
    }
}
```

cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -212,6 +330,49 @@ impl AuthorizationCall {
...
+        let mut b = Request::builder().action(Some(action)).context(context);
+        if principal.is_some() {
+            b = b.principal(principal)
+        }
+        if resource.is_some() {
+            b = b.resource(resource)
+        }
+        if self.enable_request_validation && schema.is_some() {
+            b = b.schema(schema.as_ref().unwrap());
```
john-h-kastner-aws, line 370: regarding one of the clippy errors in CI: This can `match` on `schema` to avoid the `unwrap()`.

cedar-policy/src/frontend/is_authorized.rs
ygrignon-sfdc: @john-h-kastner-aws I pushed another commit that does some factoring out. Let me know what you think.

ygrignon-sfdc:
@andrewmwells-amazon I rebased my fork and addressed the fmt issues.
@john-h-kastner-aws got rid of the unwrap() you pointed to

cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -169,34 +288,43 @@ fn constant_true() -> bool {
...
-            None => None,
-        };
-
-        let context = serde_json::to_value(self.context)
-            .map_err(|e| [format!("Error encoding the context as JSON: {e}")])?;
-        let context = Context::from_json_value(context, schema.as_ref().map(|s| (s, &action)))
-            .map_err(|e| [e.to_string()])?;
+        let schema = parse_schema(self.schema)?;
+        let principal = parse_entity_uid(self.principal, "principal")?;
+        let action = parse_entity_uid(Some(self.action), "action")?.unwrap();
```
john-h-kastner-aws, line 325: One last thing. We audit everywhere in our code where rust could potentially panic, so `clippy` is configured by default to reject `unwrap` and `panic!`. To pass the last CI check you'll need to add `#[allow()]` and a comment explaining why it won't fail. e.g., 

```suggestion
        // PANIC SAFETY `parse_entity_uid` always returns `Some` given a `Some` argument
        #[allow(clippy::unwrap_used)]
        let action = parse_entity_uid(Some(self.action), "action")?.unwrap();
```

ygrignon-sfdc: I made changes to avoid the warnings rather than allowing them

john-h-kastner-aws (approved):
Looks good to me. Maybe @aaronjeline wants to look at this also

aaronjeline (approved):

PR #576, opened 2024-01-09
clippy fixes

andrewmwells-amazon:
## Description of changes
Automatically applied clippy fixes

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/est.rs
+++ b/cedar-policy-core/src/est.rs
@@ -356,7 +356,7 @@ mod test {
         let ast = est
             .try_into_ast_policy(None)
             .expect("Failed to convert to AST");
-        ast.try_into().expect("Failed to convert to EST")
+        ast.into()
     }
 
     /// helper function to take EST-->AST-->EST for templates.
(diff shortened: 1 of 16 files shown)


john-h-kastner-aws (approved):

cedar-policy-core/src/est.rs
```diff
@@ -356,7 +356,7 @@ mod test {
         let ast = est
             .try_into_ast_policy(None)
             .expect("Failed to convert to AST");
-        ast.try_into().expect("Failed to convert to EST")
```
andrewmwells-amazon, line 359: Not sure if we actually want this change.

cdisselkoen: I think this is fine.  Probably the AST->EST transform was originally fallible but now is not (as we can see from there now being an `Into` impl)?

cdisselkoen (approved):

cedar-policy-core/src/est.rs
john-h-kastner-aws: Seems legit. We have `impl From<ast::Policy> for Policy`, so the `TryFrom` should be automatic from that

andrewmwells-amazon:
It seems the CedarJava failure is caused by this change: https://github.com/cedar-policy/cedar/actions/runs/7465002067/job/20313236263?pr=577

andrewmwells-amazon:
Nvm, the issue is a property test that will fail non-deterministically (as Victor noticed here: https://github.com/cedar-policy/cedar-java/pull/72#discussion_r1437156141). Unrelated to this PR.

john-h-kastner-aws (approved):

Issue #285, opened 2023-08-24
Nested action groups don't work properly in schema-based parsing


cdisselkoen:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Policy Evaluation

### Describe the bug

When you include actions both in the schema and in the entities passed to schema-based parsing, Cedar ensures the definition of each action agrees across the two sources.  However, this check is done _after_ computing TC for the schema actions but _before_ computing TC for the entities' actions, meaning that in order to pass this check, the caller has to precompute TC manually for their actions, even if they pass `TCComputation::ComputeNow`.  Instead, the check for agreement of action definitions should be done after the TC computation in both cases.

### Expected behavior

When you pass the same actions to the schema-based entity JSON parser as appear in the schema, this should work properly, even in the presence of nested action groups, without having to manually compute TC in either place.

### Reproduction steps

1. Create a schema with nested action groups (e.g. action 1 -> action 2 -> action 3)
2. Try to construct an `Entities` using schema-based JSON parsing, which includes the actions in the same non-TC form as they appeared in the schema
3. Observe the error `definition of action [X] does not match its schema declaration`

khieta:
Resolved by #581

Issue #304, opened 2023-09-11
Schema Validation for Storing Actions in Entities


warmuthc:
### Category

Cedar validation features

### Describe the feature you'd like to request

Our application has a User entity and a Role entity. The Role entity contains a list of actions that users with that role are allowed to take. For instance, here is what the role entity looks like:
```
{
    "uid": {
        "type": "Role",
        "id": "admin"
    },
    "attrs": {
        "actions": [
             Action::"listSessions",
             Action::"deleteSession"
        ]
    }
}
```

and the Users have a role attribute like so:
```
{
    "uid": {
        "type": "User",
        "id": "user1"
    },
    "parents": [],
    "attrs": {
         "role": {
              "type": "Role",
              "id": "admin"
          }
     }
}
```

Our policies check if the action is allowed in the role like so:
```
permit (
    principal,
    action,
    resource
) when {
    action in principal.role.actions
};
```

The problem is that it isn't possible to validate the presence of the action list inside the role. Writing a Schema like
```
Role": {
    "shape": {
     "type": "Record",
      "attributes": {
        "actions": {
          "type": "Set",
          "element": {
            "type": "Entity",
            "name": "Action"
          }
        }
      }
    }
  },
```
produces an error `Undeclared entity types: {"Action"}`

It would be great if we could get an extension for the validator to allow us to list actions inside entities.

Thank you!

### Describe alternatives you've considered

We originally had the permissions for each user listed as Cedar policies, but having the permissions inside the Role entity allows us to quickly return what permissions a role has, as well as have a configuration file that defines every permission for each role.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #581, opened 2024-01-11
Fix for #285

khieta:
## Description of changes

Reordered schema-based parsing code so that action entities are check _after_ TC computation instead of before. See code comments for why (I think) this is ok.

## Issue #, if available

#285 

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/entities.rs
+++ b/cedar-policy-core/src/entities.rs
@@ -160,21 +160,18 @@ impl Entities {
     ) -> Result<Self> {
         let mut entity_map = create_entity_map(entities.into_iter())?;
         if let Some(schema) = schema {
-            // validate entities against schema.
-            // we do this before adding the actions, because we trust the
+            // Validate non-action entities against schema.
+            // We do this before adding the actions, because we trust the
             // actions were already validated as part of constructing the
             // `Schema`
             let checker = EntitySchemaConformanceChecker::new(schema, extensions);
             for entity in entity_map.values() {
-                checker.validate_entity(entity)?;
+                let uid = entity.uid();
+                let etype = uid.entity_type();
+                if !etype.is_action() {
+                    checker.validate_entity(entity)?;
+                }
             }
-            // now add the action entities from the schema
-            entity_map.extend(
-                schema
-                    .action_entities()
-                    .into_iter()
-                    .map(|e| (e.uid(), unwrap_or_clone(e))),
-            );
         }
         match tc_computation {
             TCComputation::AssumeAlreadyComputed => {}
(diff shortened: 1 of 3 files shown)


cedar-policy-core/src/entities.rs
```diff
@@ -160,21 +160,18 @@ impl Entities {
...
+            // Validate non-action entities against schema.
+            // We do this before adding the actions, because we trust the
             // actions were already validated as part of constructing the
             // `Schema`
             let checker = EntitySchemaConformanceChecker::new(schema, extensions);
             for entity in entity_map.values() {
-                checker.validate_entity(entity)?;
+                let uid = entity.uid();
+                let etype = uid.entity_type();
+                if !etype.is_action() {
```
cdisselkoen, line 171:
```suggestion
                if !entity.uid().is_action() {
```

cedar-policy-core/src/entities.rs
```diff
@@ -185,6 +182,28 @@ impl Entities {
                 compute_tc(&mut entity_map, true).map_err(Box::new)?;
             }
         }
+        // Now that TC has been enforced, we can check action entities for
+        // conformance with the schema and add action entities to the store.
+        // This is fine to do after TC because the action hierarchy in the
+        // schema already satisfies TC, and action and non-action entities
+        // will never be in the same hierarchy.
```
cdisselkoen, line 189: nit: although actions are not allowed to have non-action parents, I believe that (due to oversight) non-actions are allowed to have action parents.  If so, the comment is inaccurate, but the code may still be correct

cedar-policy-core/src/entities.rs
```diff
@@ -185,6 +182,28 @@ impl Entities {
...
+        if let Some(schema) = schema {
+            let checker = EntitySchemaConformanceChecker::new(schema, extensions);
+            for entity in entity_map.values() {
+                let uid = entity.uid();
+                let etype = uid.entity_type();
+                if etype.is_action() {
```
cdisselkoen, line 195:
```suggestion
                if entity.uid().is_action() {
```

cdisselkoen (approved):

cedar-policy-core/src/entities.rs
khieta: Schemas do not allow specifying "Action" in the "memberOfTypes" list (a by-product of #304), so by this point in the code we've ruled out any cases of non-actions with action parents.

khieta: (Updated comment to clarify)

cdisselkoen (approved):

john-h-kastner-aws (approved):

PR #582, opened 2024-01-12
add source locations to evaluation errors

cdisselkoen:
## Description of changes

Source locations for evaluation errors.

I _think_ the required changes to `EvaluationError` manage to be non-breaking -- `EvaluationError` is a public type, but its fields are private, this PR doesn't change the `EvaluationErrorKind` enum, and the constructors I changed are all marked `pub(crate)`.  But someone should check me on this.

(I did remove two `From` impls, which can be added back if necessary. But they should be deprecated in favor of the new constructors which also take a source location.)

## Issue #, if available

Towards resolving #485.  (Just evaluation errors, not the entity-parsing errors or other errors also mentioned in #485.)

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/evaluator.rs
+++ b/cedar-policy-core/src/evaluator.rs
@@ -19,6 +19,7 @@
 use crate::ast::*;
 use crate::entities::{Dereference, Entities};
 use crate::extensions::Extensions;
+use crate::parser::Loc;
 #[cfg(test)]
 use std::collections::HashMap;
 use std::sync::Arc;
@@ -91,10 +92,39 @@ impl<'e> RestrictedEvaluator<'e> {
     /// Interpret a `RestrictedExpr` into a `Value` in this evaluation environment.
     ///
     /// May return an error, for instance if an extension function returns an error
-    pub fn partial_interpret(&self, e: BorrowedRestrictedExpr<'_>) -> Result<PartialValue> {
+    pub fn partial_interpret(&self, expr: BorrowedRestrictedExpr<'_>) -> Result<PartialValue> {
         stack_size_check()?;
 
-        match e.as_ref().expr_kind() {
+        let res = self.partial_interpret_internal(&expr);
+
+        // set the returned value's source location to the same source location
+        // as the input expression had.
+        // we do this here so that we don't have to set/propagate the source
+        // location in every arm of the big `match` in `partial_interpret_internal()`.
+        // also, if there is an error, set its source location to the source
+        // location of the input expression as well, unless it already had a
+        // more specific location
+        res.map(|pval| pval.with_maybe_source_loc(expr.source_loc().cloned()))
+            .map_err(|err| match err.source_loc() {
+                None => err.with_maybe_source_loc(expr.source_loc().cloned()),
+                Some(_) => err,
+            })
+    }
+
+    /// Internal function to interpret a `RestrictedExpr`. (External callers,
+    /// use `interpret()` or `partial_interpret()`.)
+    ///
+    /// Part of the reason this exists, instead of inlining this into
+    /// `partial_interpret()`, is so that we can use `?` inside this function
+    /// without immediately shortcircuiting into a return from
+    /// `partial_interpret()` -- ie, so we can make sure the source locations of
+    /// all errors are set properly before returning them from
+    /// `partial_interpret()`.
+    fn partial_interpret_internal(
+        &self,
+        expr: &BorrowedRestrictedExpr<'_>,
+    ) -> Result<PartialValue> {
+        match expr.as_ref().expr_kind() {
             ExprKind::Lit(lit) => Ok(lit.clone().into()),
             ExprKind::Set(items) => {
                 let vals = items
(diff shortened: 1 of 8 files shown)


cedar-policy-core/src/evaluator/err.rs
```diff
@@ -79,137 +92,188 @@ impl EvaluationError {
...
+        self.advice.as_deref()
+    }
+
     /// Set the advice field of an error
     pub fn set_advice(&mut self, advice: String) {
         self.advice = Some(advice);
     }
 
+    /// Return the `EvaluationError`, but with the new `source_loc` (or `None`).
+    pub fn with_maybe_source_loc(self, source_loc: Option<Loc>) -> Self {
```
john-h-kastner-aws, line 111: Should this be `pub(crate)`?

john-h-kastner-aws (approved):
Looks good. I don't see anything that looks like a breaking change

cedar-policy-core/src/evaluator/err.rs
```diff
@@ -44,6 +48,23 @@ impl Diagnostic for EvaluationError {
...
     }
 
+    fn source_code(&self) -> Option<&dyn miette::SourceCode> {
+        self.source_loc
+            .as_ref()
+            .map(|loc| &loc.src as &dyn miette::SourceCode)
+            .or_else(|| self.error_kind.source_code())
+    }
+
+    fn labels(&self) -> Option<Box<dyn Iterator<Item = miette::LabeledSpan> + '_>> {
```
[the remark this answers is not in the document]
cdisselkoen: We could, but I'm worried that sometimes we have the same source location both in the `source_loc` and in the `error_kind`.  Would miette handle that correctly?  or would it underline the span twice or something?  I can experiment

cedar-policy-core/src/evaluator.rs
```diff
@@ -91,18 +92,47 @@ impl<'e> RestrictedEvaluator<'e> {
...
+        // set the returned value's source location to the same source location
+        // as the input expression had.
+        // we do this here so that we don't have to set/propagate the source
+        // location in every arm of the big `match` in `partial_interpret_internal()`.
+        // also, if there is an error, set its source location to the source
+        // location of the input expression as well, unless it already had a
+        // more specific location
+        res.map(|pval| pval.with_maybe_source_loc(expr.source_loc().cloned()))
+            .map_err(|err| match err.source_loc() {
+                None => err.with_maybe_source_loc(expr.source_loc().cloned()),
```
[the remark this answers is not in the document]
cdisselkoen: It's possible no errors do.  Even if that's the case, I'm inclined to keep the code this way, for symmetry with the code in `Evaluator::partial_interpret()`, and in case new errors get added to `RestrictedEvaluator::partial_interpret_internal()` in the future?  What do you think?

cedar-policy-core/src/evaluator/err.rs
cdisselkoen: Yes, good catch

cedar-policy-core/src/evaluator/err.rs
```diff
@@ -30,10 +31,13 @@ pub struct EvaluationError {
     error_kind: EvaluationErrorKind,
     /// Optional advice on how to fix the error
     advice: Option<String>,
+    /// Source location of the error. (This overrides other sources if present,
```
andrewmwells-amazon, line 34: This seems error-prone to me. Maybe create an issue to fix it?

andrewmwells-amazon (approved):

cedar-policy-core/src/evaluator/err.rs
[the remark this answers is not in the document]
cdisselkoen: Tested this and it does underline the span twice (kinda) if you put the same span in the iterator twice.  I think we should avoid doing that.

It makes sense to me to provide deference to the explicit `source_loc` in the structure if it's `Some`, with documentation that that's what's happening.

The vast majority of the `EvaluationErrorKind`s do not have a source loc field -- the only cases that do have a source loc are `InvalidRestrictedExpression` and `NonValue`, and those just because they store an `Expr` as part of the error.

cedar-policy-core/src/evaluator/err.rs
cdisselkoen: Happy to create an issue, curious what you find to be error-prone about this?

Are you concerned that the source location in this field might disagree with the source location embedded in `error_kind`?  If it makes you feel better, the vast majority of the `EvaluationErrorKind`s do not have a source loc field -- the only cases that do have a source loc are `InvalidRestrictedExpression` and `NonValue`, and those just because they store an `Expr` as part of the error.  We could make an issue to remove these somehow, so that `error_kind` never contains a source loc; is that what you're suggesting?

Going the other way -- having all the source locs in `error_kind` so that we never need this field -- is backwards; we established recently that we prefer the convention to use the `*Kind` pattern to have a single place the source loc field is defined, rather than adding a source loc field in N places, once for each enum variant.  The struct here conforms to that practice.

andrewmwells-amazon: Yeah, it's just a general concern about "code smell," not anything concrete. Even if we always remember to keep them in sync, I'd worry about that complexity hiding other bugs. E.g., I think the bug we had where it was possible to add a link and a template with colliding names wouldn't have happened if we didn't have duplicate information in our lists of policies+templates and links+templates. Even though it wasn't caused by forgetting to keep data in sync, the added complexity (e.g., rolling back changes to one representation if modifying the other failed) distracts code reviewers from the obvious "you forgot to check that no link with this ID exists."

Issue #584, opened 2024-01-16
Fix warnings in our docs


aaronjeline:
### Describe the improvement you'd like to request

Our `cargo doc` has a couple of warnings we should fix up.

### Additional context

```
warning: unresolved link to `ToASTError::WrongNode`
   --> cedar-policy-core/src/parser/err.rs:397:31
    |
397 |     /// Constructor for the [`ToASTError::WrongNode`] error
    |                               ^^^^^^^^^^^^^^^^^^^^^ the struct `ToASTError` has no field or associated item named `WrongNode`
    |
    = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default

warning: unresolved link to `ToASTError::WrongArity`
   --> cedar-policy-core/src/parser/err.rs:410:31
    |
410 |     /// Constructor for the [`ToASTError::WrongArity`] error
    |                               ^^^^^^^^^^^^^^^^^^^^^^ the struct `ToASTError` has no field or associated item named `WrongArity`

warning: `cedar-policy-core` (lib doc) generated 2 warnings
    Checking cedar-policy v3.0.0 (/local/home/aeline/cedar-github/cedar/cedar-policy)
 Documenting cedar-policy v3.0.0 (/local/home/aeline/cedar-github/cedar/cedar-policy)
    Checking cedar-policy-cli v3.0.0 (/local/home/aeline/cedar-github/cedar/cedar-policy-cli)
 Documenting cedar-policy-cli v3.0.0 (/local/home/aeline/cedar-github/cedar/cedar-policy-cli)
warning: this URL is not a hyperlink
   --> cedar-policy-cli/src/lib.rs:319:72
    |
319 |     /// The standard human-readable Cedar policy format, documented at https://docs.cedarpolicy.com/policies/syntax-policy.html
    |                                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use an automatic link instead: `<https://docs.cedarpolicy.com/policies/syntax-policy.html>`
    |
    = note: bare URLs are not automatically turned into clickable links
    = note: `#[warn(rustdoc::bare_urls)]` on by default

warning: this URL is not a hyperlink
   --> cedar-policy-cli/src/lib.rs:322:51
    |
322 |     /// Cedar's JSON policy format, documented at https://docs.cedarpolicy.com/policies/json-format.html
[...]

PR #585, opened 2024-01-16
Fixing some docs issues

aaronjeline:
## Description of changes
Just fixing some links in the cargo doc
## Issue #, if available
#584 

## Disclaimer

### DIFF
--- a/cedar-policy-cli/src/lib.rs
+++ b/cedar-policy-cli/src/lib.rs
@@ -316,10 +316,10 @@ pub struct AuthorizeArgs {
 
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq, ValueEnum)]
 pub enum PolicyFormat {
-    /// The standard human-readable Cedar policy format, documented at https://docs.cedarpolicy.com/policies/syntax-policy.html
+    /// The standard human-readable Cedar policy format, documented at <https://docs.cedarpolicy.com/policies/syntax-policy.html>
     #[default]
     Human,
-    /// Cedar's JSON policy format, documented at https://docs.cedarpolicy.com/policies/json-format.html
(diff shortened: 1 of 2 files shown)


cedar-policy-cli/src/lib.rs
```diff
@@ -316,10 +316,10 @@ pub struct AuthorizeArgs {
 
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq, ValueEnum)]
 pub enum PolicyFormat {
-    /// The standard human-readable Cedar policy format, documented at https://docs.cedarpolicy.com/policies/syntax-policy.html
+    /// The standard human-readable Cedar policy format, documented at <https://docs.cedarpolicy.com/policies/syntax-policy.html>
     #[default]
     Human,
-    /// Cedar's JSON policy format, documented at https://docs.cedarpolicy.com/policies/json-format.html
+    /// Cedar's JSON policy format, documented at <https://docs.cedarpolicy.com/policies/json-format.html>
```
john-h-kastner-aws, line 322: These strings appear verbatim in the CLI `--help`. I don't see any harm in including the angle brackets there, but thought it was worth pointing out.

john-h-kastner-aws (approved):

cedar-policy-cli/src/lib.rs
aaronjeline: If we don't add the angle-brackets it won't be a clickable link in the docs

john-h-kastner-aws: This is in CLI, so I don't expect too many people to be looking at them

aaronjeline: Fair enough, I can role that back

cdisselkoen (approved):

Issue #586, opened 2024-01-16
Use error info during integration testing


khieta:
### Describe the improvement you'd like to request

Our integration testing framework supports providing a "custom implementation" of the Cedar validator and authorizer to allow testing new implementations. If a custom implementation is provided, then we ignore the "errors" field of the integration tests, which records which policies are expected to error during an authorization request. 

We'll eventually want to update this to enforce that error behavior is consistent between different implementations of Cedar.

 _Note:_ We can’t do this until we update the implementations in [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) to return authorization error info.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this internal improvement
- [ ] ⚠️ This feature might incur a breaking change

khieta:
Closing this issue, since it was addressed during the last round of integration test revisions. I believe this particular issue was addressed in #707.

PR #589, opened 2024-01-18
Update integration tests to store id of erroring policies

khieta:
## Description of changes

This PR updates the “errors” field of the integration tests to store only the id of erroring policies. This will prevent us from having to change the corpus tests whenever an error message changes (which is so far the only reason we’ve had to update the tests), and going forward (#586) will allow for easier comparison with other implementations of Cedar that may produce different messages, but should agree on which policies error.

Specifically, this PR:
* Adds `id` method for `AuthorizationError` (requires a minor version bump)
* Updates integration test format
    * Exposes `JsonTest` and `JsonRequest` fields for reuse in `cedar-spec` and `cedar-policy-cli`
    * Changes errors field to list policy ids rather than error messages
    * Updates corpus tests to match new test format. You can skip this part of the diff.

- [X] Updates the "Unreleased" section of the CHANGELOG with a description of my change.

- [X] equires updates, and I have made / will make these updates myself.

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/authorizer/err.rs
+++ b/cedar-policy-core/src/authorizer/err.rs
@@ -32,3 +32,12 @@ pub enum AuthorizationError {
         error: EvaluationError,
     },
 }
+
+impl AuthorizationError {
+    /// Get the id of the erroring policy
+    pub fn id(&self) -> &PolicyID {
+        match self {
+            Self::PolicyEvaluationError { id, error: _ } => id,
+        }
+    }
+}
--- a/cedar-policy-core/src/authorizer/err.rs
+++ b/cedar-policy-core/src/authorizer/err.rs
@@ -32,12 +32,3 @@ pub enum AuthorizationError {
         error: EvaluationError,
     },
 }
-
-impl AuthorizationError {
-    /// Get the id of the erroring policy
-    pub fn id(&self) -> &PolicyID {
-        match self {
-            Self::PolicyEvaluationError { id, error: _ } => id,
-        }
-    }
-}
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -906,8 +906,10 @@ impl Diagnostics {
     }
 
     /// Get the `PolicyId`s of the policies where an error occurred during authorization.
-    pub fn error_ids(&self) -> impl Iterator<Item = &PolicyId> {
-        self.errors.iter().map(|e| PolicyId::ref_cast(e.id()))
+    pub fn error_policy_ids(&self) -> impl Iterator<Item = &PolicyId> {
+        self.errors.iter().map(|e| match e {
+            AuthorizationError::PolicyEvaluationError { id, error: _ } => PolicyId::ref_cast(id),
+        })
     }
 }
 
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -904,6 +904,11 @@ impl Diagnostics {
     pub fn errors(&self) -> impl Iterator<Item = &AuthorizationError> + '_ {
         self.errors.iter()
     }
+
+    /// Get the `PolicyId`s of the policies where an error occurred during authorization.
+    pub fn error_ids(&self) -> impl Iterator<Item = &PolicyId> {
+        self.errors.iter().map(|e| PolicyId::ref_cast(e.id()))
+    }
 }
 
 impl Response {
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -27,7 +27,6 @@ use cedar_policy_core::ast::{
     ContextCreationError, ExprConstructionError, Integer, RestrictedExprParseError,
 }; // `ContextCreationError` is unsuitable for `pub use` because it contains internal types like `RestrictedExpr`
 use cedar_policy_core::authorizer;
-pub use cedar_policy_core::authorizer::AuthorizationError;
 use cedar_policy_core::entities::{
     self, ContextJsonDeserializationError, ContextSchema, Dereference, JsonDeserializationError,
     JsonDeserializationErrorContext,
@@ -735,6 +734,40 @@ impl Authorizer {
     }
 }
 
+/// Errors that can occur during authorization
+#[derive(Debug, Diagnostic, PartialEq, Eq, Error, Clone)]
+pub enum AuthorizationError {
+    /// An error occurred when evaluating a policy.
+    #[error("while evaluating policy `{id}`: {error}")]
+    PolicyEvaluationError {
+        /// Id of the policy with an error
+        id: ast::PolicyID,
+        /// Underlying evaluation error
+        #[diagnostic(transparent)]
+        error: EvaluationError,
+    },
+}
+
+impl AuthorizationError {
+    /// Get the id of the erroring policy
+    pub fn id(&self) -> &PolicyId {
+        match self {
+            Self::PolicyEvaluationError { id, error: _ } => PolicyId::ref_cast(id),
+        }
+    }
+}
+
+#[doc(hidden)]
+impl From<authorizer::AuthorizationError> for AuthorizationError {
+    fn from(value: authorizer::AuthorizationError) -> Self {
+        match value {
+            authorizer::AuthorizationError::PolicyEvaluationError { id, error } => {
+                Self::PolicyEvaluationError { id, error }
+            }
+        }
+    }
+}
+
 /// Authorization response returned from the `Authorizer`
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct Response {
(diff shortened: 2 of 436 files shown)


khieta:
Apologies for the absurdly large diff. The important part of the diff (i.e., the code outside `cedar-integration-tests/corpus_tests`) is small!

khieta:
Here are a few things I noticed, but didn’t change because they would require a major version bump. Lmk if you think these should be fixed in a followup PR, or if i'm just being picky!

1. The “errors” field of the `InterfaceDiagnostics` type should by `HashSet<PolicyId>` instead of `HashSet<String>`.
2. In the `Response` structure, the field is called `reason` (singular), whereas in the integration tests it is called `reasons` (plural). I didn’t change these to be consistent because I’m conflicted about the right option. I think “reasons” makes more sense since there may be multiple policy ids (and it’s consistent with “errors”), but changing the `Response` type is a breaking change.
3. Rather than adding an `error_ids` method to `Diagnostics`, it would make more sense to add an `id` method to the publicly exposed `AuthorizationError` type that returns a `PolicyId`. I don’t think this is possible without a breaking change.

khieta:
Looks like there's a downstream failure in CedarJava. This should be easy to fix, but does beg the question of whether this is PR is truly backwards compatible. I maintain that it is because our integration test format should not be considered part of the stable API... but open to second opinions.

cedar-policy-core/src/authorizer/err.rs
```diff
@@ -32,3 +32,12 @@ pub enum AuthorizationError {
         error: EvaluationError,
     },
 }
+
+impl AuthorizationError {
+    /// Get the id of the erroring policy
+    pub fn id(&self) -> &PolicyID {
```
john-h-kastner-aws, line 38: Should this not  be `pub` since `AuthorizationError` is `pub use`'ed but this return the internal `PolicyId` type?

The enum already exposes this to anyone who matches on it, but returning it from a function could be more confusing IMO. Better to just have the function on `Diagnostics` that returns the wrapped `PolicyId` type.

cedar-policy/src/api.rs
```diff
@@ -904,6 +904,11 @@ impl Diagnostics {
     pub fn errors(&self) -> impl Iterator<Item = &AuthorizationError> + '_ {
         self.errors.iter()
     }
+
+    /// Get the `PolicyId`s of the policies where an error occurred during authorization.
+    pub fn error_ids(&self) -> impl Iterator<Item = &PolicyId> {
```
john-h-kastner-aws, line 909: Maybe a more explicit name like `error_policy_ids`. `error_ids` makes it sound like they might be error code.

cedar-policy-core/src/authorizer/err.rs
john-h-kastner-aws: Responding to your earlier comment: 

> Rather than adding an error_ids method to Diagnostics, it would make more sense to add an id method to the publicly exposed AuthorizationError type that returns a PolicyId. I don’t think this is possible without a breaking change.

Agreed. I'm not sure how we would structure it, but the ideal situation would be a public `AuthorizationError` type with a function returning the public `PolicyId` type, but I also don't think that could be done without a breaking change. For now I'd say we drop this function.

khieta: Good catch. No, we don't want to expose a function that returns a `PolicyID`. I'll make a separate PR to change the `AuthorizationError` definition in the next major release.

cedar-policy/src/api.rs
khieta: Fair point. Changed in the latest commit.

john-h-kastner-aws (approved):

shaobo-he-aws (approved):

andrewmwells-amazon (approved):

andrewmwells-amazon:
> Looks like there's a downstream failure in CedarJava. This should be easy to fix, but does beg the question of whether this is PR is truly backwards compatible. I maintain that it is because our integration test format should not be considered part of the stable API... but open to second opinions.

I think you can safely override this. I still consider it backwards compatible.

andrewmwells-amazon:
> The “errors” field of the InterfaceDiagnostics type should by HashSet<PolicyId> instead of HashSet<String>.
Agree we should probably change but don't think it's urgent.

> In the Response structure, the field is called reason (singular), whereas in the integration tests it is called reasons (plural). I didn’t change these to be consistent because I’m conflicted about the right option. I think “reasons” makes more sense since there may be multiple policy ids (and it’s consistent with “errors”), but changing the Response type is a breaking change.
I could see saying "The reason is this (non-empty) list." or "The reasons are the items on this list." I prefer the later, but I'm not convinced it's worth a breaking change.

> Rather than adding an error_ids method to Diagnostics, it would make more sense to add an id method to the publicly exposed AuthorizationError type that returns a PolicyId. I don’t think this is possible without a breaking change.
Confused by this. Adding new methods shouldn't be a breaking change?

john-h-kastner-aws:
> Confused by this. Adding new methods shouldn't be a breaking change?

As I understand it, the issue is that `AuthorizationError` is in `cedar-policy-core`, so a function on it can't return `cedar-policy::PolicyId` without moving the enum into `cedar-policy`.

khieta:
Thanks for the reviews! I believe I've addressed all comments.

> 2. In the `Response` structure, the field is called `reason` (singular), whereas in the integration tests it is called `reasons` (plural).

I changed the name in the corpus test format because the inconsistency was bothering me. This is not a breaking change.

> 3. Rather than adding an `error_ids` method to `Diagnostics`, it would make more sense to add an `id` method to the publicly exposed `AuthorizationError` type that returns a `PolicyId`. I don’t think this is possible without a breaking change.

Like @john-h-kastner-aws said, the reason this would be breaking is that we'd have to (re)define `AuthorizationError` in `cedar-policy`.

cdisselkoen (approved):

cdisselkoen:
> Like @john-h-kastner-aws said, the reason this would be breaking is that we'd have to (re)define `AuthorizationError` in `cedar-policy`.

If `AuthorizationError` is already a publically exported type from `cedar-policy`, why does changing where it's defined constitute a breaking change?

khieta:
Fix for Java CI failure: https://github.com/cedar-policy/cedar-java/pull/78

khieta:
> If `AuthorizationError` is already a publicly exported type from `cedar-policy`, why does changing where it's defined constitute a breaking change?

`AuthorizationError` is currently defined in `cedar-policy-core` (and needs to stay there because it has other dependencies). There is no method to pull out the `id` field, but you can pattern match to get it. The ideal change (IMO) is to define a new `AuthorizationError` in `cedar-policy` that is a wrapper around the internal version, and not publicly expose the internal version. This means that you could no longer pattern match on the internal error type, which is a change in behavior.

cdisselkoen:
I suppose instead of wrapping the internal version, we could mirror the internal version, similar to how `SchemaError` works.  Then users could still pattern match on the internal error type

khieta:
> I suppose instead of wrapping the internal version, we could mirror the internal version, similar to how SchemaError works. Then users could still pattern match on the internal error type

I hadn't thought about that. The `error_policy_ids` method felt a little clunky, so I switched to your suggestion in the latest (non-merge) commit. Unfortunately, we have to reference the internal `PolicyID` type in the struct to avoid a breaking change, but at least the `id()` method can return a `PolicyId`. Lmk what you think!

andrewmwells-amazon (approved):

cedar-policy/src/api.rs
```diff
@@ -735,6 +734,40 @@ impl Authorizer {
...
 }
 
+/// Errors that can occur during authorization
+#[derive(Debug, Diagnostic, PartialEq, Eq, Error, Clone)]
+pub enum AuthorizationError {
+    /// An error occurred when evaluating a policy.
+    #[error("while evaluating policy `{id}`: {error}")]
+    PolicyEvaluationError {
+        /// Id of the policy with an error
+        id: ast::PolicyID,
```
john-h-kastner-aws, line 744: We should `#[doc(hidden)]` this to avoid confusion with the actual getter function. That should be fair game for a minor release since you can still access it if really want to.

john-h-kastner-aws (approved):
Seems legit. I like this better and AFAIK it shouldn't be a breaking change.

cedar-policy/src/api.rs
khieta: Good call. Done in the last commit.

khieta:
Pushed one more edit -- I just realized that `cedar-policy-cli` was redefining `JsonTest` and `JsonRequest` when it could be importing them from `cedar_policy` instead.

I'm done now I swear!

john-h-kastner-aws (approved):

Issue #209, opened 2023-07-19
Separate partial evaluation from concrete evaluation


shaobo-he-aws:
### Category

Internal refactors/changes

### Describe the feature you'd like to request

Currently, the concrete evaluator simply calls the partial evaluator. It's better if they are separate functions (i.e., tree walks with calls to shared functions).

### Describe alternatives you've considered

N/A

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

emina:
I'm strongly in favor of this change.

Separating interpretation and PE will give us code that is easier maintain and a cleaner high-level design.

At the code level, we'll be able to change PE independently of the concrete evaluator.  Right now, we have to change the evaluator code to fix PE bugs, add more rewrite rules, tweak PE-specific error messages, etc.  The addition of PE code also makes the authorizer and evaluator code harder to read and understand. Finally, the PE code adds runtime overhead, however small, to every authorization call.

At the design level, PE isn't a part of the concrete semantics.  It's an additional tool just like the validator, which is implemented as a separate AST traversal.  So, implementing PE as a separate traversal lets us keep the various "liftings" of the concrete semantics separate from its definitive implementation (the concrete evaluator).

PR #590, opened 2024-01-18
draft to Separate partial evaluation from concrete evaluation 

shaobo-he-aws:
## Issue #, if available

#209

## Disclaimer


cedar-policy-core/src/evaluator.rs
```diff
@@ -252,9 +252,383 @@ impl<'e> Evaluator<'e> {
...
-        match self.partial_interpret(e, slots)? {
-            PartialValue::Value(v) => Ok(v),
-            PartialValue::Residual(r) => Err(EvaluationError::non_value(r)),
+        stack_size_check()?;
+        let res = self.interpret_internal(e, slots);
+
+        // set the returned value's source location to the same source location
+        // as the input expression had.
+        // we do this here so that we don't have to set/propagate the source
+        // location in every arm of the big `match` in `partial_interpret_internal()`.
```
cdisselkoen, line 261:
```suggestion
        // location in every arm of the big `match` in `interpret_internal()`.
```

cedar-policy-core/src/evaluator.rs
```diff
@@ -252,9 +252,383 @@ impl<'e> Evaluator<'e> {
...
+        // more specific location
+        res.map(|pval| pval.with_maybe_source_loc(e.source_loc().cloned()))
+            .map_err(|err| match err.source_loc() {
+                None => err.with_maybe_source_loc(e.source_loc().cloned()),
+                Some(_) => err,
+            })
+    }
+
+    /// Internal function to interpret an `Expr`. (External callers, use
+    /// `interpret()` or `partial_interpret()`.)
```
cdisselkoen, line 273:
```suggestion
    /// `interpret()`.)
```

cedar-policy-core/src/evaluator.rs
```diff
@@ -252,9 +252,383 @@ impl<'e> Evaluator<'e> {
...
+    ///
+    /// Part of the reason this exists, instead of inlining this into
+    /// `partial_interpret()`, is so that we can use `?` inside this function
```
cdisselkoen, line 276:
```suggestion
    /// `interpret()`, is so that we can use `?` inside this function
```

cedar-policy-core/src/evaluator.rs
```diff
@@ -252,9 +252,383 @@ impl<'e> Evaluator<'e> {
...
+    /// without immediately shortcircuiting into a return from
+    /// `partial_interpret()` -- ie, so we can make sure the source locations of
```
cdisselkoen, line 278:
```suggestion
    /// `interpret()` -- ie, so we can make sure the source locations of
```

cedar-policy-core/src/evaluator.rs
```diff
@@ -252,9 +252,383 @@ impl<'e> Evaluator<'e> {
...
+    /// all errors are set properly before returning them from
+    /// `partial_interpret()`.
```
cdisselkoen, line 280:
```suggestion
    /// `interpret()`.
```

cedar-policy-core/src/evaluator.rs
```diff
@@ -252,9 +252,383 @@ impl<'e> Evaluator<'e> {
...
+                            EvaluationError::integer_overflow(
+                                IntegerOverflowError::UnaryOp { op: *op, arg },
+                                loc.cloned(),
+                            )
+                        }),
+                }
+            }
+            ExprKind::BinaryApp { op, arg1, arg2 } => {
+                let (arg1, arg2) = (self.interpret(arg1, slots)?, self.interpret(arg2, slots)?);
+                // Borrow from cedar-spec
```
cdisselkoen, line 341: I don't know what this comment refers to

cedar-policy-core/src/evaluator.rs
```diff
@@ -252,9 +252,383 @@ impl<'e> Evaluator<'e> {
...
+                    (_, _, BinaryOp::In) => Err(EvaluationError::type_error_with_advice(
+                        nonempty![Type::entity_type(names::ANY_ENTITY_TYPE.clone())],
+                        &arg1,
+                        "the LHS of `in` should be an entity".into(),
+                    )),
+                    // contains, which works on Sets
+                    (_, _, BinaryOp::Contains) => match arg1.value {
+                        ValueKind::Set(Set { fast: Some(h), .. }) => Ok(arg2
+                            .try_as_lit()
+                            .map_or(false, |lit| h.contains(lit))
```
cdisselkoen, line 481: maybe would be good to copy the comment from the partial evaluator, that the `map_or(false)` works because we know the set can't contain a non-literal

cedar-policy-core/src/evaluator.rs
```diff
@@ -252,9 +252,383 @@ impl<'e> Evaluator<'e> {
...
+                        Value {
+                            value:
+                                ValueKind::Set(Set {
+                                    authoritative: _,
+                                    fast: Some(arg2_set),
+                                }),
+                            ..
+                        },
+                        BinaryOp::ContainsAll,
+                    ) => Ok((arg2_set.is_subset(&arg1_set)).into()),
```
cdisselkoen, line 508: maybe would be good to copy the comment from the partial evaluator, that in this case both sets only contain literals and thus we have this fast hashset-based implementation

cedar-policy-core/src/evaluator.rs
```diff
@@ -252,9 +252,383 @@ impl<'e> Evaluator<'e> {
...
+                        Value {
+                            value:
+                                ValueKind::Set(Set {
+                                    authoritative: _,
+                                    fast: Some(arg2_set),
+                                }),
+                            ..
+                        },
+                        BinaryOp::ContainsAny,
+                    ) => Ok((arg2_set.is_disjoint(&arg1_set)).into()),
```
cdisselkoen, line 527: I think we need a `!` here.  Unit tests should catch this

cedar-policy-core/src/evaluator.rs
```diff
@@ -252,9 +252,383 @@ impl<'e> Evaluator<'e> {
...
+                            value: ValueKind::Set(arg1_set),
+                            ..
+                        },
+                        Value {
+                            value: ValueKind::Set(arg2_set),
+                            ..
+                        },
+                        BinaryOp::ContainsAll,
+                    ) => Ok(arg2_set
+                        .authoritative
```
cdisselkoen, line 539: maybe would be good to copy the comment from the partial evaluator, that in this case one or both sets are in slow form, ie, contain a non-literal, which is why we need this slow implementation

cedar-policy-core/src/evaluator.rs
```diff
@@ -252,9 +252,383 @@ impl<'e> Evaluator<'e> {
...
+                            value: ValueKind::Set(arg1_set),
+                            ..
+                        },
+                        Value {
+                            value: ValueKind::Set(arg2_set),
+                            ..
+                        },
+                        BinaryOp::ContainsAny,
+                    ) => Ok(arg1_set
+                        .authoritative
```
cdisselkoen, line 554: same here

cedar-policy-core/src/evaluator.rs
```diff
@@ -252,9 +252,383 @@ impl<'e> Evaluator<'e> {
...
+                            },
+                            loc.cloned(),
+                        )
+                    })
+            }
+            ExprKind::ExtensionFunctionApp { fn_name, args } => {
+                let efunc = self
+                    .extensions
+                    .func(&fn_name)
+                    .map_err(|err| EvaluationError::extension_function_lookup(err, loc.cloned()))?;
```
cdisselkoen, line 591: I think this will be a behavior change vs the current partial evaluator.  The current partial evaluator evaluates all the arguments first, before checking if the efunc lookup fails.  So if an argument has an error _and_ the efunc lookup is invalid, the current partial evaluator will prefer the argument error, while this code will prefer the efunc lookup error.  I don't think we have a strong reason to prefer one over the other, and I don't think we mind that a different error is thrown now (I think?), but just flagging this

cedar-policy-core/src/evaluator.rs
```diff
@@ -252,9 +252,383 @@ impl<'e> Evaluator<'e> {
...
+            }
+            ExprKind::GetAttr { expr, attr } => {
+                self.get_attr(self.interpret(expr, slots)?, attr, loc)
+            }
+            ExprKind::HasAttr { expr, attr } => {
+                self.has_attr(self.interpret(expr, slots)?, attr, loc)
+            }
+            ExprKind::Like { expr, pattern } => {
+                let v = self.interpret(expr, slots)?.get_as_string()?.clone();
+                Ok((pattern.wildcard_match(&v)).into())
```
cdisselkoen, line 606: do we need the `.clone()` followed by `&v`?

cedar-policy-core/src/evaluator.rs
```diff
@@ -252,9 +252,383 @@ impl<'e> Evaluator<'e> {
...
+                Ok(Value::set(vals, loc.cloned()))
+            }
+            ExprKind::Record(map) => {
+                let map = map
+                    .iter()
+                    .map(|(k, v)| Ok((k.clone(), self.interpret(v, slots)?)))
+                    .collect::<Result<Vec<_>>>()?;
+                let (names, evalled): (Vec<SmolStr>, Vec<Value>) = map.into_iter().unzip();
+
+                Ok(Value::record(names.into_iter().zip(evalled), loc.cloned()))
```
cdisselkoen, line 630:
```suggestion
                Ok(Value::record(names.into_iter().zip(evalled), loc.cloned()))
```

cedar-policy-core/src/evaluator.rs
```diff
@@ -711,10 +1064,90 @@ impl<'e> Evaluator<'e> {
...
+            )),
+        }
+    }
+
+    fn get_attr(&self, val: Value, attr: &SmolStr, source_loc: Option<&Loc>) -> Result<Value> {
+        match val {
+            Value {
+                value: ValueKind::Record(record),
+                ..
+            } => record.as_ref().get(attr).map(|v| v.clone()).ok_or_else(|| {
```
cdisselkoen, line 1098:
```suggestion
            } => record.as_ref().get(attr).cloned().ok_or_else(|| {
```

cdisselkoen (commented):
High-level comments:

- With this PR, the concrete evaluator is definitely a lot simpler than the partial evaluator, which was one of the goals.  This is good.
- However, I'm concerned by the amount of code duplication between the concrete evaluator and the partial evaluator.  There's a lot of potential for these to drift out of sync in the future.  Would it be possible to share more code?  I like how the code is shraed for `GetAttr`
- Could we add a property-based test that the two evaluators always give the same result if all the inputs are concrete?

PR #598, opened 2024-01-23
Delete dead code in `cst_to_ast.rs`

john-h-kastner-aws:
## Description of changes

Check me on this, but this code can't be reached. 

## Disclaimer


cdisselkoen (approved):
I think it can be reached but will always match the first branch, resulting in no-op.  (Note the `?` on line 546.)

CI passing is also a nice check that this code wasn't necessary.

john-h-kastner-aws:
Yeah, the interesting code is unreachable.

john-h-kastner-aws:
The one way this could change anything would be if `to_var` added an error to `errs` without returning `None`. Then that error would be added once instead of twice.

shaobo-he-aws (approved):

Issue #596, opened 2024-01-23
Parseable policy cannot be converted to json


khieta:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Cedar Parser

### Describe the bug

The following policy can be successfully parsed:
```
permit(principal, action, resource) when { -9223372036854775808 };
```
But when you try to convert the parsed policy to json, you will get a `IntegerLiteralTooLarge` error.

(The policy in my example is ill-typed, but similar well-typed policies will have the same issue.)

### Expected behavior

Ideally, all policies that can be parsed by `Policy::from_str` should be able to be converted to json using `Policy::to_json()`.

### Code Snippet

```rust
let policy_str = r#"permit(principal, action, resource) when { -9223372036854775808 };"#;
let p = Policy::from_str(policy_str).expect("should parse successfully");
// next line will panic
p.to_json().expect("should convert to JSON successfully");
```

### Log output

```
// Put your output below this line

```

khieta:
Followup based on conversation with the team: the underlying issue here is that JSON doesn't support true 64-bit integers, while Cedar does. This is a case of a Cedar policy that contains an integer literal that cannot be expressed exactly using JSON. The same issue will exist for our other json formats, like schemas and entities.

As such, I'll change the title to be more descriptive.

aaronjeline:
Follow up to the follow up: I think it actually has to do with `-9223372036854775808` being a valid 64 bit integer but `9223372036854775808` not being one, and the ambiguity around whether `-1` is a negative constant or the negation operation applied to a positive constant.

PR #601, opened 2024-01-24
Fixes an issue where certain cedar policies couldn't be converted to JSON

aaronjeline:
## Description of changes
Previously, the Cedar EST code would fail to generate an EST for any policy containing the constant `-9223372036854775808` (`i64::MIN`). This happened because it was turned into the CST representing the unary op `-` applied to the constant `9223372036854775808`, which outside the range of `i64`. This adds a tiny pass to ensure that unary op is constant folded.
## Issue #, if available
#596 

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/est/expr.rs
+++ b/cedar-policy-core/src/est/expr.rs
@@ -1052,6 +1052,7 @@ impl TryFrom<&Node<Option<cst::Unary>>> for Expr {
                             .map(|y| y == InputInteger::MAX as u64)
                             .unwrap_or(false)) =>
                     {
+                        num_dashes -= 1;
                         Expr::ExprNoExt(ExprNoExt::Value(CedarValueJson::Long(InputInteger::MIN)))
                     }
                     _ => (&u_node.item).try_into()?,
@@ -1074,15 +1075,11 @@ impl TryFrom<&Node<Option<cst::Unary>>> for Expr {
                         // not safe to collapse `--` to nothing
                         Ok(Expr::neg(Expr::neg(inner)))
                     }
-                    n => {
-                        if n % 2 == 0 {
-                            // safe to collapse to `--` but not to nothing
-                            Ok(Expr::neg(Expr::neg(inner)))
-                        } else {
-                            // safe to collapse to -
-                            Ok(Expr::neg(inner))
-                        }
-                    }
+                    3 => Ok(Expr::neg(Expr::neg(Expr::neg(inner)))),
+                    4 => Ok(Expr::neg(Expr::neg(Expr::neg(Expr::neg(inner))))),
+                    _ => Err(u
+                        .to_ast_err(ToASTErrorKind::UnaryOpLimit(ast::UnaryOp::Neg))
+                        .into()),
                 }
             }
             Some(cst::NegOp::OverBang) => Err(u
--- a/cedar-policy-core/src/est/expr.rs
+++ b/cedar-policy-core/src/est/expr.rs
@@ -1022,17 +1022,12 @@ impl TryFrom<&Node<Option<cst::Unary>>> for Expr {
                     3 => Ok(Expr::not(Expr::not(Expr::not(inner)))),
                     4 => Ok(Expr::not(Expr::not(Expr::not(Expr::not(inner))))),
                     _ => Err(u
-                        .to_ast_err(ToASTErrorKind::UnaryOpLimit(ast::UnaryOp::Neg))
+                        .to_ast_err(ToASTErrorKind::UnaryOpLimit(ast::UnaryOp::Not))
                         .into()),
                 }
             }
             Some(cst::NegOp::Dash(0)) => Ok((&u_node.item).try_into()?),
             Some(cst::NegOp::Dash(mut num_dashes)) => {
-                //
-                // && (x
-                //     .checked_sub(1)
-                //     .map(|y| y == InputInteger::MAX as u64)
-                //     .unwrap_or(false)) =>
                 let inner = match &u_node.item.to_lit() {
                     Some(cst::Literal::Num(num))
                         if num
--- a/cedar-policy-core/src/est/expr.rs
+++ b/cedar-policy-core/src/est/expr.rs
@@ -1011,25 +1011,51 @@ impl TryFrom<&Node<Option<cst::Unary>>> for Expr {
     type Error = ParseErrors;
     fn try_from(u: &Node<Option<cst::Unary>>) -> Result<Expr, ParseErrors> {
         let u_node = u.ok_or_missing()?;
-        let inner = (&u_node.item).try_into()?;
+
         match u_node.op {
-            Some(cst::NegOp::Bang(0)) => Ok(inner),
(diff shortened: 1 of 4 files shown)


cedar-policy-core/src/est/expr.rs
```diff
@@ -1011,25 +1011,51 @@ impl TryFrom<&Node<Option<cst::Unary>>> for Expr {
...
+                            }),
+                        ..
+                    } if access.is_empty()
+                        && (x
+                            .checked_sub(1)
+                            .map(|y| y == InputInteger::MAX as u64)
+                            .unwrap_or(false)) =>
+                    {
+                        Expr::ExprNoExt(ExprNoExt::Value(CedarValueJson::Long(InputInteger::MIN)))
+                    }
```
cdisselkoen, line 1056: don't we need to subtract one from `num_dashes` in this case?  (and do we have a test for this code, that makes sure the value is correct, or preserved across various roundtrips?)

cdisselkoen (commented):
Two more comments:

1) This approach transforms only INT_MIN.  Should we, more generally, represent (-X) as a negative integer literal instead of a negation operation of a positive integer literal, for all X?
2) It looks to me that only the CST->EST code was adjusted; is this sufficient?  Is there AST->EST code that needs to be adjusted?  I assume that EST->AST is OK

cedar-policy/src/api.rs
```diff
@@ -3924,3 +3924,91 @@ mod partial_eval_test {
...
+    }
+
+    fn test_single_int_with_dashes(x: i64, dashes: usize) {
+        let dashes = vec!['-'; dashes].into_iter().collect::<String>();
+        let src = format!(r#"permit(principal, action, resource) when {{ {dashes}{x} }};"#);
+        let p: Policy = src.parse().unwrap();
+        let json = p.to_json().unwrap();
+        let round_trip = Policy::from_json(None, json).unwrap();
+        let pretty_print = format!("{round_trip}");
+        assert!(pretty_print.contains(&x.to_string()));
```
khieta, line 3954: Might as well check for the correct number of dashes here too.

cedar-policy/src/api.rs
```diff
@@ -3924,3 +3924,91 @@ mod partial_eval_test {
...
+    }
+
+    // Serializing a valid 64-bit int that can't be represented in double precision float
+    #[test]
```
khieta, line 3958: A little confused about this comment: why is converting this int to json ok?

aaronjeline: From the JSON spec
> JSON is agnostic about the semantics of numbers. In any programming language, there can be a variety of
number types of various capacities and complements, fixed or floating, binary or decimal. That can make
interchange between different programming languages difficult. JSON instead offers only the representation of
numbers that humans use: a sequence of digits. All programming languages know how to make sense of digit
sequences even if they disagree on internal representations. That is enough to allow interchange

Serde in particular is happy to either ser/de JSON numbers as floats or ints, depending on what datastructure you're ser/deing to/from. This does mean that applications interfacing from Cedar whose JSON libraries take a different approach could lead to loss of precision.

aaronjeline:
>     1. This approach transforms only INT_MIN.  Should we, more generally, represent (-X) as a negative integer literal instead of a negation operation of a positive integer literal, for all X?
perhaps

>     2. It looks to me that only the CST->EST code was adjusted; is this sufficient?  Is there AST->EST code that needs to be adjusted?  I assume that EST->AST is OK
Yeah, all of those already have the integer literals as `i64`, whereas the CST has integers as `u64`s for the purpose of error messages

aaronjeline:
Do we want to remove this code:
```
                match num_dashes {
                    0 => Ok(inner),
                    1 => Ok(Expr::neg(inner)),
                    2 => {
                        // not safe to collapse `--` to nothing
                        Ok(Expr::neg(Expr::neg(inner)))
                    }
                    n => {
                        if n % 2 == 0 {
                            // safe to collapse to `--` but not to nothing
                            Ok(Expr::neg(Expr::neg(inner)))
                        } else {
                            // safe to collapse to -
                            Ok(Expr::neg(inner))
                        }
                    }
                }
```
If the intent of the EST is to lossless preserve the policy?

cedar-policy-core/src/est/expr.rs
```diff
@@ -1048,15 +1075,11 @@ impl TryFrom<&Node<Option<cst::Unary>>> for Expr {
...
-                        } else {
-                            // safe to collapse to -
-                            Ok(Expr::neg(inner))
-                        }
-                    }
+                    3 => Ok(Expr::neg(Expr::neg(Expr::neg(inner)))),
+                    4 => Ok(Expr::neg(Expr::neg(Expr::neg(Expr::neg(inner))))),
+                    _ => Err(u
+                        .to_ast_err(ToASTErrorKind::UnaryOpLimit(ast::UnaryOp::Neg))
+                        .into()),
```
khieta, line 1082: Looks like we're not enforcing the unary op limit in the bang case above, should we be enforcing it here?

aaronjeline: It's implicit in that it's assumed that `num_bangs` is going to be less than 4. In my opinion I'd like to change to it from a `usize` to an enum, so that we can ensure the domain is correct.

aaronjeline: But that's probably too big of a change for this PR

cedar-policy-core/src/est/expr.rs
```diff
@@ -1011,25 +1011,52 @@ impl TryFrom<&Node<Option<cst::Unary>>> for Expr {
...
+                                        ..
+                                    },
+                                access,
+                            }),
+                        ..
+                    } if access.is_empty()
+                        && (x
+                            .checked_sub(1)
+                            .map(|y| y == InputInteger::MAX as u64)
+                            .unwrap_or(false)) =>
```
khieta, line 1053: Looking at where [we handle this in the CST -> AST code](https://github.com/cedar-policy/cedar/blob/1a961126a7ef94d64c0c4a997e71928a819313f9/cedar-policy-core/src/parser/cst_to_ast.rs#L1521), possibly you can replace some of this code using `.to_lit()`?

aaronjeline: Oh yeah that's almost certainly going to be nicer :)

cedar-policy-core/src/est/expr.rs
```diff
@@ -1013,44 +1013,32 @@ impl TryFrom<&Node<Option<cst::Unary>>> for Expr {
...
-                                access,
-                            }),
-                        ..
-                    } if access.is_empty()
-                        && (x
+                //
+                // && (x
+                //     .checked_sub(1)
+                //     .map(|y| y == InputInteger::MAX as u64)
+                //     .unwrap_or(false)) =>
```
khieta, line 1035: Supposed to be deleted?

khieta (approved):

cdisselkoen (approved):
Happy to approve as-is, possible followup per my comment above (transforming all negative literals and not just INT_MIN)

PR #603, opened 2024-01-25
increase validator precision for some `Unspecified` cases

cdisselkoen:
## Description of changes

Increases validator precision (typing `False` instead of `Bool`) in some edge cases involving `Unspecified`.  Also, change from `TypecheckFail` to `TypecheckSuccess` in one of these edge cases in particular.

This is a backwards-incompatible change unless/until `ImpossiblePolicy` is moved from an error to a warning.  We should probably not merge this until we've decided whether that will happen.

Breaking, unless we wait to merge this until `ImpossiblePolicy` is moved from an error to a warning, in which case non-breaking.  I also only observed the impact of this change in `Permissive` validation mode, which is an experimental feature in 3.x.  I'm not sure (not qualified to determine) whether this change has any impact in `Strict` validation mode, ie, on the stable 3.x validator.

I'm not sure whether this behavior matches the Lean validator, which doesn't yet fully match the Rust validator anyway.  I believe that neither before nor after this PR will the behavior exactly match the Dafny validator, but I propose that this is the correct/ideal behavior, which we want Dafny to match?  (That's a point for discussion below)

## Disclaimer

### DIFF
--- a/cedar-policy-validator/src/typecheck.rs
+++ b/cedar-policy-validator/src/typecheck.rs
@@ -2030,6 +2030,16 @@ impl<'a> Typechecker<'a> {
                     }
                 }
             }
+            (Some(EntityType::Specified(_)), None) => {
+                // One or more of the elements on the right is not an entity
+                // literal. The `in` is still valid, so typechecking succeeds
+                // with type Boolean.
+                TypecheckAnswer::success(
+                    ExprBuilder::with_data(Some(Type::primitive_boolean()))
+                        .with_same_source_loc(in_expr)
+                        .is_in(lhs_expr, rhs_expr),
+                )
+            }
             (Some(EntityType::Unspecified), Some(rhs)) => {
                 // It's perfectly valid for `principal` or `resource` to be `EntityType::Unspecified`
                 if rhs
@@ -2073,16 +2083,6 @@ impl<'a> Typechecker<'a> {
                             .is_in(lhs_expr, rhs_expr),
                     )
                 }
-            }
-            (Some(EntityType::Specified(_)), None) => {
-                // One or more of the elements on the right is not an entity
-                // literal. The `in` is still valid, so typechecking succeeds
-                // with type Boolean.
-                TypecheckAnswer::success(
-                    ExprBuilder::with_data(Some(Type::primitive_boolean()))
-                        .with_same_source_loc(in_expr)
-                        .is_in(lhs_expr, rhs_expr),
-                )
             }
         }
     }
--- a/cedar-policy-validator/src/typecheck.rs
+++ b/cedar-policy-validator/src/typecheck.rs
@@ -1916,7 +1916,10 @@ impl<'a> Typechecker<'a> {
 
     // Given an expression, if that expression is a literal or the `action`
     // variable, return it as an EntityUID. Return `None` otherwise.
-    fn euid_from_euid_literal_or_action(request_env: &RequestEnv, e: &Expr) -> Option<EntityUID> {
+    fn euid_from_euid_literal_or_action<T: Clone>(
+        request_env: &RequestEnv,
+        e: &Expr<T>,
+    ) -> Option<EntityUID> {
         match Typechecker::replace_action_var_with_euid(request_env, e)
             .as_ref()
             .expr_kind()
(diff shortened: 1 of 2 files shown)


cdisselkoen:
Revised, hopefully the new code is the correct/ideal behavior.  Looking for review.

To answer one of the questions from the PR description, this (should be) a backwards-compatible change, because the only edge cases it changes behavior for involve `[]` (the empty set), and the empty set is only valid under Permissive validation, which is an experimental feature in 3.x.  This (should) have no effect on the stable 3.x validator -- hoping a reviewer can confirm.

khieta (commented):
Motivated by this PR, I spent some time looking through the Rust code vs. the spec and I think the fix is actually simpler than what you have here: the Dafny function `tryGetEUIDs` returns `None` for an empty list while the corresponding Rust function `euids_from_euid_literals_or_action` returns `Some([])`. I think you can fix the behavior difference by just adjusting one of these functions to match the other.

cedar-policy-validator/src/typecheck.rs
```diff
@@ -1961,83 +1964,126 @@ impl<'a> Typechecker<'a> {
...
+                    // something on the RHS is unspecified, so we have to type `unspecified in RHS` as Bool,
+                    // because two unspecified entities are equal (and thus `in`) if they have the same `Eid`.
+                    TypecheckAnswer::success(
+                        ExprBuilder::with_data(Some(Type::primitive_boolean()))
+                            .with_same_source_loc(in_expr)
+                            .is_in(lhs_expr, rhs_expr),
+                    )
+                } else {
+                    // nothing on the RHS can be unspecified, so `unspecified in RHS` is always false
+                    TypecheckAnswer::success(
```
john-h-kastner-aws, line 2080: This might change strict validation for something like `principal in [resource.owner]` when `principal` is unspecified and `resource.owner` is some entity type.

john-h-kastner-aws (commented):
Not entirely confident this doesn't affect strict validation. As noted, that's could be fine if we make the `ImpossiblePolicy` change.

Kesha's sugestion seems worth looking into.

cedar-policy-validator/src/typecheck.rs
```diff
@@ -1961,83 +1964,126 @@ impl<'a> Typechecker<'a> {
...
+                // with type Boolean.
+                TypecheckAnswer::success(
                     ExprBuilder::with_data(Some(Type::primitive_boolean()))
                         .with_same_source_loc(in_expr)
                         .is_in(lhs_expr, rhs_expr),
-                ),
+                )
+            }
+            (Some(EntityType::Unspecified), Some(rhs)) => {
+                // It's perfectly valid for `principal` or `resource` to be `EntityType::Unspecified`
```
khieta, line 2044: This case is a little weird to me. It's changing behavior from returning an error to returning Bool or False. It seems like your edit is the correct behavior, but why wasn't this discrepancy picked up by DRT? Is there some other part of the code that was handling unspecified variables, as suggested by the original (possibly incorrect) comment?

cdisselkoen: It turns out that returning the `TypecheckFail` results in `typecheck_policy()` returning `false` instead of `true`, but that return value is discarded by its caller ([here](https://github.com/cedar-policy/cedar/blob/1ce81ec45f95723ba838f0a125070fe7de7f124e/cedar-policy-validator/src/lib.rs#L151)). Since no actual type errors are added to the set of type errors, validation is considered a success despite the `TypecheckFail` which is ignored.

So, my conclusion is that changing this Fail(Bool) to Success(Bool) does not affect anything, and in fact would better represent what is actually happening. Returning a Fail without adding a type error is functionally equivalent to returning a Success, at least in this instance.

cedar-policy-validator/src/typecheck.rs
cdisselkoen: I think you're right.  That doesn't mean this precision increase is invalid, just that it's currently breaking (unless/until we change `ImpossiblePolicy` from error to warning)

cdisselkoen: Actually I'm not sure if you're correct.  I believe `principal in [resource.owner]` when `principal` is unspecified and `resource.owner` is some entity type, might short-circuit out [here](https://github.com/cedar-policy/cedar/blob/2c456758391cfff0eb23d79d4c9d446ce09fd035/cedar-policy-validator/src/typecheck.rs#L1763) and never reach this function at all.

khieta: Agreed that earlier handling handles this case. But I'm not sure this new definition matches the spec -- the Dafny code will always return `AnyBool` in the case where the RHS cannot be converted to a list of Entity UIDs.

cdisselkoen:
Closing this in favor of #615 and #620.  See comments on those.

PR #615, opened 2024-01-31
increase validator precision in an edge case

cdisselkoen:
## Description of changes

This PR is an alternate way to address the problem in #603.  Instead of changing `type_of_var_in_entity_literals()` and `type_of_entity_literal_in_entity_literals()`, we merely change the definition of `Type::must_be_specified_entity()`, which will cause cases of `X in []` to be short-circuited out ([here](https://github.com/cedar-policy/cedar/blob/2c456758391cfff0eb23d79d4c9d446ce09fd035/cedar-policy-validator/src/typecheck.rs#L1763)) and never reach those functions in the first place.  This causes `X in []` to be typed `False` for all X, including unspecified X.  This matches the behavior of the Dafny as far as I understand it.

The edge case where this changes validator behavior is only possible when `[]` is a valid expression, which is only true under Permissive validation, which is only an experimental feature in Cedar 3.x, so this is not considered a breaking change from semver perspective, and could be released in a patch or minor of Cedar 3.x.

## Disclaimer

### DIFF
--- a/cedar-policy-validator/src/types.rs
+++ b/cedar-policy-validator/src/types.rs
@@ -482,14 +482,18 @@ impl Type {
     }
 
     /// Return true if we know that any value in this type must be a specified
-    /// entity. An unspecified entity has type `AnyEntity`, so `AnyEntity` might
-    /// not be specified. Other entity types must be specified.
+    /// entity.
     pub(crate) fn must_be_specified_entity(ty: &Type) -> bool {
         matches!(
             ty,
+            // An unspecified entity has type `AnyEntity`, so `AnyEntity` might
+            // not be specified. Other entity types must be specified.
             Type::EntityOrRecord(
                 EntityRecordKind::Entity(_) | EntityRecordKind::ActionEntity { .. }
             )
+            // It is vacuously true that any value in the type `Never` is a
+            // specified entity.
+            | Type::Never
         )
     }
 


khieta:
Unless I'm missing something, I don't think this does the same thing as #603 -- the `must_be_specified_entity` function is only used to type `==` expressions, not `in` expressions. I think what you want to update is the function `euids_from_euid_literals_or_action`.

cdisselkoen:
It's used in `typecheck_in` [here](https://github.com/cedar-policy/cedar/blob/2c456758391cfff0eb23d79d4c9d446ce09fd035/cedar-policy-validator/src/typecheck.rs#L1753)

khieta:
Thanks, I was indeed missing something 🤦‍♀️

khieta (approved):
Approving now that my confusion is resolved.

I agree that this change adjusts the precision as expected & only affects permissive validation. I prefer this PR over #603 because it's easier to audit, but I think we also want some of the changes from #603. In particular, returning `TypecheckFail` in `type_of_var_in_entity_literals` seems unintentional & I think we should change it to `TypecheckSuccess` with a general Bool type.

john-h-kastner-aws (approved):
I agree with Kesha. This change is a nicer fix to the particular difference we saw with in DRT, but we should look at pulling in some of the changes from the other PR.

PR #616, opened 2024-01-31
Expose policy and template EST methods in cedar-wasm

ncsze:
## Description of changes

Change broken off from a larger commit into smaller form from: https://github.com/Swolebrain/cedar/tree/tsify

Exposes `Policy` and `PolicySet` parsing and JSON conversion from `cedar-policy` to `cedar-wasm`.

Included is a `Policy` API `to_json_policy` that returns the same value as `to_json`, but with a `Policy` type instead of `Value` type.

## Issue #, if available

N/A

## Disclaimer


cedar-wasm/src/policies_and_templates.rs
```diff
@@ -0,0 +1,250 @@
...
+
+use tsify::Tsify;
+use wasm_bindgen::prelude::*;
+
+#[derive(Tsify, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub enum JsonToPolicyResult {
+    Success { policy_text: String },
+    Error { errors: Vec<String> },
```
aaronjeline, line 14: Probably want a more rich error type here, we started with strings and our in the process of moving to more and more structured.

cedar-wasm/src/policies_and_templates.rs
```diff
@@ -0,0 +1,250 @@
...
+pub fn policy_text_from_json(json_str: &str) -> JsonToPolicyResult {
+    let parsed_json = match serde_json::from_str(json_str) {
+        Ok(v) => v,
+        Err(e) => {
+            return JsonToPolicyResult::Error {
+                errors: vec![e.to_string()],
+            }
+        }
+    };
+    let policy = Policy::from_json(None, parsed_json);
```
aaronjeline, line 27: This provides no way to users to set the policy id, so the ids will all overlap, unless you edit it later

cedar-wasm/src/policies_and_templates.rs
```diff
@@ -0,0 +1,250 @@
...
+/// struct that defines the result for the syntax validation function
+pub enum PolicySetParseResult {
+    /// represents successful syntax validation
+    Success { policies: i32, templates: i32 },
+    /// represents a syntax error and encloses a vector of the errors
+    SyntaxError { errors: Vec<String> },
+}
+
+#[wasm_bindgen(js_name = "parsePolicySet")]
+pub fn parse_policy_set(input_policies_str: &str) -> PolicySetParseResult {
```
aaronjeline, line 81: At least for me, I assumed a function named `parse_policy_set` would return policies, but we're returning counts here?

cedar-wasm/src/policies_and_templates.rs
Swolebrain: is there a specific type you recommend?

cedar-wasm/src/policies_and_templates.rs
Swolebrain: i don't think there's a use case for policy id's in this stateless context

cedar-wasm/src/policies_and_templates.rs
```diff
@@ -0,0 +1,250 @@
...
+                slots: vec![],
+                parse_errors: Some(parse_errs.errors_as_strings()),
+            },
+            Ok(template) => match template.slots().count() {
+                1 | 2 => Self {
+                    text: template_str.to_string(),
+                    slots: template.slots().map(|slot| slot.to_string()).collect(),
+                    parse_errors: None,
+                },
+                _ => Self {
```
Swolebrain, line 128: Whatever error type we define for the above, we should use here too

cedar-wasm/src/policies_and_templates.rs
```diff
@@ -0,0 +1,250 @@
...
+    slots: Vec<String>,
+    parse_errors: Option<Vec<String>>,
+}
+
+#[wasm_bindgen]
+impl Template {
+    #[wasm_bindgen(constructor)]
+    pub fn new(template_str: &str) -> Template {
+        match cedar_policy::Template::from_str(template_str) {
+            Err(parse_errs) => Self {
```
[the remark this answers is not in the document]
Swolebrain: we should move this to a separate pr

cedar-wasm/src/policies_and_templates.rs
Swolebrain: On second thought, i think we might wanna keep this a string since there are two possible sources of errors here. Deserializing into serde_json::value can yield one type of Err, and then parsing the policy can yield another type. WDYT?

Swolebrain (approved):

ncsze:
Removed 2 unused imports in tests. Locally, the errors that failed the workflow just showed up as warnings, sorry to have missed those.

ncsze:
Rebased

cedar-wasm/src/policies_and_templates.rs
khieta: I also found this name confusing. I think `check_parse_policy` or something similar would make more sense.

cedar-wasm/src/policies_and_templates.rs
khieta: `String`s are probably ok here for ease of (de)serializability. You could also use a simplified version of our [`ParseError` enum](https://github.com/cedar-policy/cedar/blob/687d0f5d62b80556dfaa2b2c296703c50c332b4f/cedar-policy-core/src/parser/err.rs#L48). (Note: that enum is not (de)serializable as-is because the error variants contain a lot of extra information.)

cedar-wasm/src/policies_and_templates.rs
khieta: Yes, this will set the id to `policy0`. I think that's fine, since this function returns the policy source text (which has no way to represent ids).

khieta:
Overall looks fine. My main comment is on returning the internal type `cedar_policy_core::est::Policy` from our main API -- would prefer to avoid this if possible.

cedar-policy/src/api.rs
```diff
@@ -3157,6 +3157,11 @@ impl Policy {
         Ok::<_, PolicyToJsonError>(json)
     }
 
+    /// returns the json representation of a policy but strongly typed
+    pub fn to_json_policy(&self) -> Result<cedar_policy_core::est::Policy, impl std::error::Error> {
+        self.lossless.est()
+    }
```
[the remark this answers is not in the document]
khieta: Since `cedar_policy_core::est::Policy` is an "internal" datatype, it is subject to change. We did not design it to be particularly pretty -- it's primarily designed to allow nice (de)serialization from JSON. 

The stable interface is the JSON policy format itself. So, if possible, I think you might be better off using the JSON directly, and then performing conversion to an EST internally to check that it has the correct form.

cedar-policy/CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `wasm` Cargo feature for targeting Wasm
 - For the `partial-eval` experimental feature: added
 `Authorizer::evaluate_policies_partial()` (#474)
+- Exposed cedar-wasm functionality for policies and templates: `check_parse_policy_set`,
+  `policy_text_to_json`, and `policy_text_from_json`. (#616)
```
khieta, line 23: Not true of the current code, right? Maybe you want a separate CHANGELOG in `cedar-wasm`?

cedar-policy/src/api.rs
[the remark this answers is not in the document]
Swolebrain: I really don't think users would be better off using JSON format directly and not having a strongly typed option. But if that's what we need to do for now to unblock it, then so be it. However I would like to present my case a little better before i disagree and commit.

Remember that with wasm, people aren't actually using the `cedar_policy_core::est::Policy` type directly, so it's not quite the same as exposing it in `cedar-policy`. What people use is the typescript type generated from `cedar_policy_core::est::Policy`. In other words, any methods that might be in `cedar_policy_core::est::Policy`'s impl are not available to users of wasm. If in the future we need to break the shape of the struct, we can add a converter to the wasm package. It's not a one way door; but we do need to remember to do that if we break `cedar_policy_core::est::Policy`.

Question - isn't the format in docs.cedarpolicy.com derived from the `cedar_policy_core::est::Policy` struct shape? If not, how's it derived? If so, isn't the struct shape already exposed then? If we change the internal struct shape we'll have to create that conversion layer anyway anyway to keep backwards compatibility with the old format, or make the change during a MV update.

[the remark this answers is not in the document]
khieta: The `cedar_policy_core::est::Policy` data type looks similar to the JSON, but it's not as close as you might think. For example, the data type currently [splits expressions into two classes](https://github.com/cedar-policy/cedar/blob/0333786910b491f32f80075855d1e27733f9a958/cedar-policy-core/src/est/expr.rs#L39): extension function applications and non-extension-functions. Splitting things up this way makes parsing with serde a little simpler, but this is an internal detail that it not exposed in the JSON format itself.

So: it is entirely possible that we will change the shape of `cedar_policy_core::est::Policy`s without making a change to the JSON format as described in the docs.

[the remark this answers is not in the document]
cdisselkoen: If we were to let `tsify` transpile `cedar_policy_core::est::Policy` to TS, would users see the distinction between `Expr` and `ExprNoExt` (see Kesha's link)?  IMO it would be a bad user experience if users had to do something like `.expr_no_ext(...)` every time they want an expression that is not an extension function call (which is probably 99% of expressions people will create)

cedar-policy/CHANGELOG.md
ncsze: These are the current names for the `cedar-wasm` functions. What's not current anymore is templates, we aren't exposing those anymore in this PR. Is this CHANGELOG only for `cedar-policy`? I assumed it was for the whole project, but I see now there's a `cedar-policy-cli` changelog.

cedar-policy/src/api.rs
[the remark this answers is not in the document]
ncsze: Users would only see the distinction between `Expr` and `ExprNoExt` in the generated `.d.ts` type headers. The JSON itself looks like this for a policy with `Clause`/`Expr` types in its EST:
`permit(principal, action, resource) when { principal has "Email" && principal.Email == "a@a.com" };`
```
{
    "policy": {
        "effect": "permit",
        "principal": {
            "op": "All"
        },
        "action": {
            "op": "All"
        },
        "resource": {
            "op": "All"
        },
        "conditions": [
            {
                "kind": "when",
                "body": {
                    "&&": {
                        "left": {
                            "has": {
                                "left": {
                                    "Var": "principal"
                                },
                                "attr": "Email"
                            }
                        },
                        "right": {
                            "==": {
                                "left": {
                                    ".": {
                                        "left": {
                                            "Var": "principal"
                                        },
                                        "attr": "Email"
                                    }
                                },
                                "right": {
                                    "Value": "a@a.com"
                                }
                            }
                        }
                    }
                }
            }
        ]
    }
}
```

cedar-policy/CHANGELOG.md
khieta: The CHANGELOG here is only for the `cedar-policy` crate. We use one here and in `cedar-policy-cli` since these are the "public" parts of our project. We do not use them in other sub-folders since they contain "internal" crates that don't follow semver.

ncsze:
- Added more tsify attributes to complete the exported TS types for the EST.
- Added a build script to fix the exported types since `tsify` doesn't properly put quotes around fields that start with symbols (ie. `*`, `!` in `ExprNoExt`).
- Changed `policy_text_to_json` to use `cedar_policy_core::parser::parse_policy_or_template_to_est` to parse text into EST JSON.
- Gave `cedar-wasm` its own CHANGELOG

khieta (approved):
The changes look good to me, pending fixes for CI failures.

Note that the "SemVer Check" failure is just because this PR is one commit behind `main`. The "Cargo Build & Test" failure will require some more debugging.

ncsze:
Removed `#[cfg_attr(feature = "wasm", serde(skip))]` from `EntityUidJson`, replaced it with a `tsify` type override that then gets scrubbed by `build-wasm.sh`.

Swolebrain (approved):

PR #626, opened 2024-02-07
Error EST to AST conversion when there are slots in conditions

### LINKED ISSUE #606: Templates constructed from EST do not error on slots in condition

### Bug Category

Other

### Describe the bug

```rust
#[test]
fn est_template() {
    let est_json = serde_json::json!({
        "effect": "permit",
        "principal": { "op": "All" },
        "action": { "op": "All" },
        "resource": { "op": "All" },
        "conditions": [
            {
                "kind": "when",
                "body": {
                    "==": {
                        "left": { "Var": "principal" },
                        "right": { "Slot": "?principal" }
                    }
                }
            }
        ]
    });

    let tid = PolicyId::new("t0");
    // We should get an error here after trying to construct a template with a slot in the condition
    let template = Template::from_json(Some(tid.clone()), est_json).unwrap();

    let mut ps = PolicySet::new();
    ps.add_template(template).unwrap();
    let lid = PolicyId::new("l0");
    ps.link(
        tid,
        lid,
        [(SlotId::principal(), r#"User::"jane""#.parse().unwrap())].into(),
    )
    .unwrap();

    let auth = Authorizer::new();
    let r = Request::new(
        Some(r#"User::"jane""#.parse().unwrap()),
        Some(r#"Action::"take""#.parse().unwrap()),
        Some(r#"Thing::"thing""#.parse().unwrap()),
        Context::empty(),
        None,
    )
    .unwrap();
    assert_eq!(
        auth.is_authorized(&r, &ps, &Entities::empty()).decision,
        Decision::Allow
    );
}
```

shaobo-he-aws:
## Description of changes

fixes #606

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/est.rs
+++ b/cedar-policy-core/src/est.rs
@@ -243,11 +243,24 @@ impl Policy {
 }
 
 impl Clause {
+    fn filter_slots(e: ast::Expr, is_when: bool) -> Result<ast::Expr, FromJsonError> {
+        let slots: Vec<_> = e.slots().map(|s| *s).collect();
+        if let Some(slot) = slots.first() {
+            Err(FromJsonError::SlotsInConditionClause {
+                slot: *slot,
+                clausetype: if is_when { "when" } else { "unless" },
+            })
+        } else {
+            Ok(e)
+        }
+    }
     /// `id` is the ID of the policy the clause belongs to, used only for reporting errors
     fn try_into_ast(self, id: ast::PolicyID) -> Result<ast::Expr, FromJsonError> {
         match self {
-            Clause::When(expr) => expr.try_into_ast(id),
-            Clause::Unless(expr) => Ok(ast::Expr::not(expr.try_into_ast(id)?)),
+            Clause::When(expr) => Self::filter_slots(expr.try_into_ast(id)?, true),
+            Clause::Unless(expr) => {
+                Self::filter_slots(ast::Expr::not(expr.try_into_ast(id)?), false)
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


cedar-policy-core/src/est.rs
```diff
@@ -243,11 +243,24 @@ impl Policy {
 }
 
 impl Clause {
+    fn filter_slots(e: ast::Expr, is_when: bool) -> Result<ast::Expr, FromJsonError> {
+        let slots: Vec<_> = e.slots().map(|s| *s).collect();
+        if let Some(slot) = slots.first() {
```
cdisselkoen, line 248: I believe we can avoid collecting into a Vec here (and thus allocating memory).

```suggestion
        let slots = e.slots().copied();
        if let Some(slot) = slots.next() {
```

cdisselkoen (approved):

john-h-kastner-aws (approved):

cedar-policy-core/src/est.rs
shaobo-he-aws: It doesn't seem to work because `e` is still borrowed, which is moved in one of the if branches.

cdisselkoen: What if 
```
let first_slot = e.slots().copied().next();
if let Some(slot) = first_slot {
```
Since the type of `first_slot` is now `Option<Slot>`, it doesn't keep `e` borrowed

shaobo-he-aws: Right. Please see the latest commit.

andrewmwells-amazon (approved):

PR #632, opened 2024-02-08
refactoring in integration tests

khieta:
## Description of changes

* Deleted integration tests that fail to parse. With the new corpus test infrastructure, these will no longer be produced.
* Moved the `cedar_test_impl.rs` file (with some renaming and other additions) from `cedar-spec` so we can use the same trait for both integration testing & DRT.
* Refactoring in `integration_testing.rs` to make it easier to reuse the parsing code for the integration test file format.
* Removed the EST testing code from the integration tests. I'll add it to DRT instead.

## Disclaimer


khieta:
Marking this as ready for review now that 1) I have a PR up to fix the `cedar-spec` break ([cedar-spec#212](https://github.com/cedar-policy/cedar-spec/pull/212)) and 2) I have a PR up to add EST tests to DRT ([cedar-spec#215](https://github.com/cedar-policy/cedar-spec/pull/215)).

One question I'd like reviewer opinions on: as indicated by the semver check in CI, this PR is _actually a breaking change_ since it affects the "public" APIs in `cedar_policy::integration_testing` and `cedar_policy::cedar_test_impl`. I was planning on treating this PR as non-breaking since it's only editing testing code. But technically users could be depending on our testing data types... should I mark this PR as a breaking change intended for the next major release?

john-h-kastner-aws:
> One question I'd like reviewer opinions on: as indicated by the semver check in CI, this PR is _actually a breaking change_ since it affects the "public" APIs in `cedar_policy::integration_testing` and `cedar_policy::cedar_test_impl`. I was planning on treating this PR as non-breaking since it's only editing testing code. But technically users could be depending on our testing data types... should I mark this PR as a breaking change intended for the next major release?

It feel strange that the testing are exported from the cedar-policy crate directly. If we want to call this a breaking change, then I propose making the larger braking change, and ripping these functions out into a new crates, then that crate can follow semver without caring about cedar's version.

john-h-kastner-aws (approved):

khieta:
> I propose making the larger breaking change, and ripping these functions out into a new crates, then that crate can follow semver without caring about cedar's version.

Yeah, I agree that it makes sense to have this testing code in a separate crate. Do you think I could get away with moving it to a separate crate before the next major version? I'd like to get this refactoring integrated sooner rather than later to prevent annoying merge issues later.

PR #633, opened 2024-02-12
Core Entity `.uid()` returns a ref

cdisselkoen:
## Description of changes

Refactors Core's `Entity::uid()` method to return `&EntityUID` and not `EntityUID`, which is the more natural / idiomatic pattern.  This allows us to avoid cloning the UID for callers that really only needed `&EntityUID`.  Conversely, callers that actually needed owned `EntityUID` now have to call `.clone()` themselves.

Non-breaking because this is a Core API.  This PR does not make the corresponding change to the public API.

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/ast/entity.rs
+++ b/cedar-policy-core/src/ast/entity.rs
@@ -291,8 +291,8 @@ impl Entity {
     }
 
     /// Get the UID of this entity
-    pub fn uid(&self) -> EntityUID {
-        self.uid.clone()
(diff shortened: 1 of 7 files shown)


john-h-kastner-aws (approved):

khieta (approved):
Looks fine. Note that there's a downstream DRT build failure you'll need to fix.

cdisselkoen:
DRT fix is at https://github.com/cedar-policy/cedar-spec/pull/217

Issue #638, opened 2024-02-13
Remove `HierarchyNotRespected` validation error


khieta:
### Describe the improvement you'd like to request

According to [this comment](https://github.com/cedar-policy/cedar/blob/5f62c6df06b59abc5634d6668198a826839c6fb7/cedar-policy-validator/src/typecheck.rs#L1871) we return the `HierarchyNotRespected` error in some cases where we could soundly type an expression as `False`. This error was introduced (IIRC) to avoid having to update the Dafny type soundness proof.

Now that we've switched to using Lean, and tweaked the validation [property being tested](https://github.com/cedar-policy/cedar-spec/pull/210), I wonder if we could remove this error variant.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this internal improvement
- [ ] ⚠️ This feature might incur a breaking change

john-h-kastner-aws:
May depend on #539, but with that we definitely can. Haven't thought about what we can do given your tweak to the property being tested.

PR #639, opened 2024-02-13
Updates for cedar-spec#219

khieta:
## Description of changes

Updates comments to use "formal model" in places where we previously used "Dafny model". [cedar-spec#219](https://github.com/cedar-policy/cedar-spec/pull/219) deprecates the Dafny model in favor of a model in Lean.

I'm not sure if the alternate serialization used in `cedar-policy-core/src/ast/pattern.rs` is still needed now that we've switched Lean, but what we have now works so I'm hesitant to change it.

There is one other place we refer to "Dafny" in the code -- I've made an issue (#638) to follow up on that later.

## Disclaimer

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -19,7 +19,7 @@ I confirm that this PR (choose one, and delete the other options):
 
 I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):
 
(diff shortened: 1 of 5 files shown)


cdisselkoen (approved):

shaobo-he-aws (approved):

shaobo-he-aws:
> I'm not sure if the alternate serialization used in cedar-policy-core/src/ast/pattern.rs is still needed now that we've switched Lean, but what we have now works so I'm hesitant to change it.

We should keep it unless we change the deserializer in Lean, which I doubt will be an easy task given JSON doesn't have a character type.

PR #643, opened 2024-02-14
better error message for incorrect variable in policy scope

### LINKED ISSUE #562: Error message is misleading for incorrect variable in policy scopes
### Category

User level API changes

For this policy:
```
permit(principal, principal, resource);
```
The current error message is
```
the variable `principal` is invalid in this policy scope clause, the variable `action` is expected
```
which is misleading.  The variable `principal` isn't invalid in policy scope clauses.

Similar misleading messages occur in related cases:
```
permit(principal, action, principal);
permit(principal, action, action);
```

While we're here, I don't think "policy scope clause" is the official terminology, I think it's just "policy scope", but we should check

N/A

### Additional context

_No response_

cdisselkoen:
## Description of changes

Better error message in cases where `principal`, `action`, or `resource` are used multiple times or in the wrong order in the policy scope.

## Issue #, if available

Fixes #562

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/parser/cst_to_ast.rs
+++ b/cedar-policy-core/src/parser/cst_to_ast.rs
@@ -4766,7 +4766,7 @@ mod tests {
         assert_matches!(parse_policy_template(None, p_src), Err(e) => {
             expect_err(p_src, &e, &ExpectedErrorMessage::error_and_help(
                 "expected a variable that is valid in the policy scope; found: `foo`",
-                "must be one of `principal`, `action`, or `resource`"
+                "policy scopes must contain a `principal`, `action`, and `resource` element in that order",
             ));
             expect_source_snippet(p_src, &e, "foo");
         });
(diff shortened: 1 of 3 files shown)


cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -4768,28 +4768,39 @@ mod tests {
                 "expected a variable that is valid in the policy scope; found: `foo`",
                 "must be one of `principal`, `action`, or `resource`"
```
john-h-kastner-aws, line 4769: Let's update this error text match with the other error variant

john-h-kastner-aws (approved):

cedar-policy-core/src/parser/cst_to_ast.rs
cdisselkoen: Updated the help text to match, but I think this error text describes the error well.  Do the changes look good to you?

john-h-kastner-aws: Seems fine. This would require a bit of work beyond the immediate scope of this PR, but the two error variants could probably be unified into one variant IMO

khieta (approved):
LGTM

PR #653, opened 2024-02-20
Fixups for clippy lints

aaronjeline:
## Description of changes
Just fixing up some clippy lints
## Issue #, if available
N/A

## Disclaimer

### DIFF
--- a/cedar-policy-core/src/ast/policy.rs
+++ b/cedar-policy-core/src/ast/policy.rs
@@ -975,10 +975,27 @@ impl Annotations {
     pub fn iter(&self) -> impl Iterator<Item = (&AnyId, &Annotation)> {
         self.0.iter()
     }
+}
 
-    /// Consume this `Annotations`, iterating over all the annotations
-    pub fn into_iter(self) -> impl Iterator<Item = (AnyId, Annotation)> {
-        self.0.into_iter()
+/// Wraps the [`BTreeMap`]` into an opaque type so we can change it later if need be
+#[derive(Debug)]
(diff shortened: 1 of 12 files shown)


cedar-policy-core/src/est.rs
```diff
@@ -342,6 +342,8 @@ impl std::fmt::Display for Clause {
 
 // PANIC SAFETY: Unit Test Code
 #[allow(clippy::panic)]
+// PANIC SAFETY: Unit Test Code
+#[allow(clippy::indexing_slicing)]
```
john-h-kastner-aws, line 346: FWIW, I never get see clippy complaining about panics etc in tests. Is this some configuration you've set locally?

shaobo-he-aws (approved):

cedar-policy-core/src/ast/policy.rs
```diff
@@ -975,10 +975,27 @@ impl Annotations {
     pub fn iter(&self) -> impl Iterator<Item = (&AnyId, &Annotation)> {
         self.0.iter()
     }
+}
+
+/// Wraps the [`BTreeMap`]` into an opaque type so we can change it later if need be
+#[derive(Debug)]
+pub struct IntoIter(std::collections::btree_map::IntoIter<AnyId, Annotation>);
```
john-h-kastner-aws, line 982: What exactly are we getting by using this struct? Also, should probably give it a different name since we could conceivably want another `IntoIter` impl in the same module.

john-h-kastner-aws (approved):

PR #634, opened 2024-02-12
Allow annotation keys to be Cedar reserved words

cdisselkoen:
## Description of changes

Allows annotation keys to be Cedar reserved words.  Tests that this works both for human-format and JSON-format (EST) policies.  Fixes #623.

## Issue #, if available

#623 

## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [x] A bug fix or other functionality change requiring a patch to `cedar-policy`.

I confirm that this PR (choose one, and delete the other options):

- [x] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [x] Does not require updates because my change does not impact the Cedar Dafny model or DRT infrastructure.

(is that right?)

## Disclaimer

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #673, opened 2024-02-28
Fixing 2 bugs with IDs

aaronjeline:
## Description of changes
1. Entity types with the id `Set` were mistakenly interpreted as having the id `set`
2. A mistake the CFG could lead to a panic when parsing a schema that used `in` as an entity type name
## Issue #, if available
N/A

## Disclaimer

### DIFF
--- a/cedar-policy-validator/src/human_schema/test.rs
+++ b/cedar-policy-validator/src/human_schema/test.rs
@@ -948,6 +948,7 @@ mod parser_tests {
 #[cfg(test)]
 mod translator_tests {
     use cedar_policy_core::FromNormalizedStr;
+    use smol_str::ToSmolStr;
 
     use crate::{SchemaError, SchemaFragment, SchemaTypeVariant, TypeOfAttribute, ValidatorSchema};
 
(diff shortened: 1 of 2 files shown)


khieta (approved):

cedar-policy-validator/src/human_schema/test.rs
```diff
@@ -1267,4 +1268,158 @@ mod translator_tests {
...
+
+        let (schema, _) = SchemaFragment::from_str_natural(src).unwrap();
+        let ns = schema.0.get("").unwrap();
+        let foo = ns.entity_types.get("Foo").unwrap();
+        assert_eq!(foo.member_of_types, vec!["namespace".to_smolstr()]);
+    }
+
+    #[test]
+    fn entity_named_in() {
+        // This fails because `in` is reserved
```
shaobo-he-aws, line 1287: I don't think it should fail. `in` is a valid entity type name in Cedar.

aaronjeline: It is not an valid entity type name. Try it

john-h-kastner-aws: If we accept https://github.com/cedar-policy/rfcs/pull/48, then `id` should be a valid annotation (see https://github.com/cedar-policy/cedar/pull/634), so this fix is not ideal.

It is also an minor difference between the schema and policy parser. The policy parser will construct a CST containing invalid IDs and reject only when building a CST, but the schema parser is raising an error while parsing.

PR #677, opened 2024-02-29
Refactor the handling of negative numbers in CST to EST conversion

shaobo-he-aws:
Now it should be the same as CST to AST conversion

## Disclaimer


john-h-kastner-aws (commented):
Is there a behavior change here, or just reorganizing the code?

shaobo-he-aws:
> Is there a behavior change here, or just reorganizing the code?

It will lead to behavior change. For example, CST `-(1)` gets converted to EST literal `-1` before the change and it's EST negation expression `-(1)` after the change.

cdisselkoen:
Do we want that behavior change?  Wouldn't it be better to produce the literal `-1` when possible, for users trying to programmatically consume/edit policies?

shaobo-he-aws:
> Do we want that behavior change? Wouldn't it be better to produce the literal `-1` when possible, for users trying to programmatically consume/edit policies?

The roundtrip PBT fails without the fix. I think we should strive for consistency for CST to EST and CST to AST conversion.

aaronjeline (approved):

andrewmwells-amazon (approved):

andrewmwells-amazon:
> Do we want that behavior change? Wouldn't it be better to produce the literal -1 when possible, for users trying to programmatically consume/edit policies?

I think it's ok. I don't see why people would write -(1), so I doubt it affects anyone. And it's a fairly intuitive that we handle the parenthesized part before negation, so I don't think people will have trouble understanding the behavior.

PR #697, opened 2024-03-06
(3.1) Translate Schema --schema short option form should be -s

john-h-kastner-aws:
## Description of changes

Breaking change to the CLI, so lets get this in 3.1


aaronjeline (approved):
Good catch thanks

khieta (approved):

Issue #176, opened 2023-07-12
More usefull (or less confusing) parse errors on unexpected token


john-h-kastner-aws:
### Category

Other

### Describe the feature you'd like to request

An unexpected token causes the parser to emit an error message listing every token in would have accepted in that position.

E.g., on `permit(principal, action, resource;` (missing the closing paren) the error text is 
```
Error: cedar_policy_core::parser::to_cst_error

  × unexpected token `;`
   ╭─[a.cedar:1:1]
 1 │ permit(principal, action resource;
   ·                                  ┬
   ·                                  ╰── expected `!=`, `%`, `&&`, `(`, `)`, `*`, `+`, `,`, `-`, `.`, `/`, `:`, `::`, `<`, `<=`, `==`, `>`, `>=`, `[`, `]`, `{`, `||`, `}`, `else`, `has`, `in`, `like`, or `then`
   ╰────

```
The list of expected tokens isn't helpful, and potentially makes the error more confusing that it would be otherwise.

This was brought up recently due to a schema parsing error in #149 where this sort of overly verbose error was caused by an empty string used for an entity type name.

### Describe the solution you'd like

.

### Describe alternatives you've considered

.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

john-h-kastner-aws:
Further improved on in #668. With that PR `permit(principal, action, resource) when {` gives a much nicer list of tokens: `` expected `!`, `(`, `-`, `[`, `{`, `}`, `false`, identifier, `if`, number, `?principal`, `?resource`, string literal, or `true` ``.

Some cases where the unexpected token errors can still be improved are noted in the [comments on tests cases added with that PR](https://github.com/cedar-policy/cedar/blob/5ed3ff95fbe35141c9f1c9e042f26f3b8336f3df/cedar-policy-core/src/parser.rs#L1065).

A particular case that stand out is `permit(principal, action, resource) when { principal is`. We claim to expect `` expected `!`, `(`, `-`, `[`, `{`, `false`, identifier, `if`, number, `?principal`, `?resource`, string literal, or `true` ``, but only an `identifier` is actually valid. This is a side effect of #491 where we changed the CST grammar to parse the right operand of `is` as an `Add` rather than a `Name`. A similar issue applies `like` where the right operand must be a pattern literal, and record literals where attribute names must be literal names.

[...]

PR #698, opened 2024-03-06
Improve parser errors on unexpected tokens

john-h-kastner-aws:
## Description of changes

Improve the parser error reported for unexpected tokens/end-of-input when the parser expects an identifier. The error message used to list out the token for every special identifier. Also ensures we don't claim to expect some tokens that can never be in a valid AST. The `=`, `/`, and `%` tokens are always invalid, so we shouldn't claim to expect them even if they grammar seems to allow it.

New error:
```
  × failed to parse policy set
  ╰─▶ unexpected end of input
   ╭─[<stdin>:1:36]
 1 │ permit(principal, action, resource)
   ·                                    ▲
   ·                                    ╰── expected `;` or identifier
   ╰────
```

Old error: 

```
  × failed to parse policy set
  ╰─▶ unexpected end of input
   ╭─[<stdin>:1:1]
 1 │ permit(principal, action, resource)
   ·                                    ▲
   ·                                    ╰── expected `;`, `action`, `context`, `else`, `false`, `forbid`, `has`, identifier, `if`, `in`, `is`, `like`, `permit`, `principal`, `resource`, `then`, `true`, `unless`, or `when`
   ╰────
```

Improves on #176. There are sill some odd parse errors of this sort. 

- [x] A backwards-compatible change requiring a minor version bump to `cedar-policy` 

- [x] Does not require updates because my change does not impact the Cedar formal model or

### DIFF
--- a/cedar-policy-core/src/parser.rs
+++ b/cedar-policy-core/src/parser.rs
@@ -1100,6 +1100,8 @@ mod parse_tests {
         assert_labeled_span("permit(,,);", "expected `)` or identifier");
         assert_labeled_span("permit(principal,", "expected identifier");
         assert_labeled_span("permit(principal,action,", "expected identifier");
+        // Nothing will actually convert to an AST here.
+        assert_labeled_span("permit(principal,action,resource,", "expected identifier");
         // We still list out `if` as an expected token because it doesn't get
         // parsed as an ident in this position.
         assert_labeled_span(
--- a/cedar-policy-core/src/parser.rs
+++ b/cedar-policy-core/src/parser.rs
@@ -1090,16 +1090,34 @@ mod parse_tests {
             "permit(principal, action, resource)",
             "expected `;` or identifier",
         );
+        // AST actually requires `permit` or `forbid`, but grammar looks for any
+        // identifier.
         assert_labeled_span("@if(\"a\")", "expected `@` or identifier");
+        // AST actually requires `principal` (`action`, `resource`, resp.). In
+        // the `principal` case we also claim to expect `)` because an empty scope
+        // initially parses to a CST. The trailing comma rules this out in the others.
         assert_labeled_span("permit(", "expected `)` or identifier");
+        assert_labeled_span("permit(,,);", "expected `)` or identifier");
+        assert_labeled_span("permit(principal,", "expected identifier");
+        assert_labeled_span("permit(principal,action,", "expected identifier");
         // We still list out `if` as an expected token because it doesn't get
         // parsed as an ident in this position.
         assert_labeled_span(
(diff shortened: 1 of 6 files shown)


cedar-policy-core/src/parser.rs
```diff
@@ -1059,4 +1060,49 @@ mod parse_tests {
...
+            "expected identifier",
+        );
+
+        // We specifically want `when` or `unless`, but we previously listed all
+        // identifier tokens, so this is an improvement.
+        assert_labeled_span(
+            "permit(principal, action, resource)",
+            "expected `;` or identifier",
+        );
+        assert_labeled_span("@if(\"a\")", "expected `@` or identifier");
```
cdisselkoen, line 1093: this one should be `@`, `permit`, or `forbid`.  (doesn't have to be in this PR, but since we have comments on some of the other tests noting what error message would be even better)

cedar-policy-core/src/parser/err.rs
```diff
@@ -570,41 +570,111 @@ impl Diagnostic for ToCSTError {
...
+    /// reject any CST that uses the operator. To reduce confusion we filter
+    /// these from the list of expected tokens in an error message.
+    pub impossible_tokens: HashSet<&'static str>,
+
+    /// Both our policy and schema grammar have a generic identifier token
+    /// and some more specific identifier tokens that we use to parse specific
+    /// constructs. It is very often not useful to explicitly list out all of
+    /// these special identifier because the parser really just wants any
+    /// generic identifier. That it would accept these does not give any
+    /// usefully information.
```
cdisselkoen, line 593:
```suggestion
    /// useful information.
```

cedar-policy/CHANGELOG.md
```diff
@@ -57,6 +57,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 return a `RequestBuilder<&Schema>` so the `RequestBuilder<&Schema>::build`
 method checks the request against the schema provided and the
 `RequestBuilder<UnsetSchema>::build` method becomes infallible. (#559)
+* Improved "unexpected token" parse errors when the parsers expects an identifier. (#698)
```
cdisselkoen, line 60:
```suggestion
* Improved "unexpected token" parse errors when the parser expects an identifier. (#698)
```

cdisselkoen (approved):
Definitely an improvement. I agree that there are still more improvements that could be done in the future, but this PR is great.

cedar-policy/CHANGELOG.md
cdisselkoen: Touche

aaronjeline (approved):

Issue #680, opened 2024-03-01
`IntoIterator` for `Entities`


cdisselkoen:
### Category

User level API features/changes

### Describe the feature you'd like to request

The public `Entities` type should support `.into_iter()` and the `IntoIterator` trait.  Today, it only supports `.iter()` which gives `&Entity` as the item type; but some users may want to consume the `Entities` and have `Entity` as the item type.  In particular, I'm imagining that users might want to do this in order to add more `Entities` and then give the larger set of entities back to `Entities::from_entities()` (at least until #612 is fixed).

### Describe alternatives you've considered

The status quo works, but requires users to clone each `Entity` if they want to rebuild another `Entities` object.

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

cdisselkoen:
Actually [`Entities::add_entities()`](https://docs.rs/cedar-policy/latest/cedar_policy/struct.Entities.html#method.add_entities) is a reasonable alternative to `.into_iter()` followed by `::from_entities()`.  So this issue is not urgent.  It would probably still be good to provide `IntoIterator` though.

cre4ture:
I thought I give it a try ;-)

john-h-kastner-aws:
This was resolved by #713

PR #713, opened 2024-03-10
implement into_iter for api::Entities

cre4ture:
## Description of changes
- added IntoIterator trait implementation with function `into_iter()`
- added unittest for it

## Issue #, if available
 IntoIterator for Entities #680

## Disclaimer

### DIFF
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -50,7 +50,7 @@ use nonempty::NonEmpty;
 use ref_cast::RefCast;
 use serde::{Deserialize, Serialize};
 use smol_str::SmolStr;
-use std::collections::{hash_map, BTreeMap, BTreeSet, HashMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::convert::Infallible;
 use std::marker::PhantomData;
 use std::str::FromStr;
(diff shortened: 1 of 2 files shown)


cedar-policy/src/api.rs
```diff
@@ -4524,4 +4537,34 @@ mod test {
...
+        {
+        "uid": {"type":"Group","id":"admin"},
+        "attrs": {},
+        "parents": []
+        }
+        ]
+        "#;
+
+        let list = Entities::from_json_str(test_data, None).unwrap();
+        let mut list_out: Vec<String> = Vec::new();
```
aaronjeline, line 4562:
```suggestion
        let mut list_out: Vec<String> = list.into_iter().map(|entity| entity.uid().id().to_string()).collect();
```
Tiny nit, but slightly more idiomatic

aaronjeline (approved):
Looks great, thanks!

cedar-policy/src/api.rs
cre4ture: done. thanks

cedar-policy/src/api.rs
```diff
@@ -626,6 +626,19 @@ impl Entities {
     }
 }
 
+type ConvertInternToApiEntity = fn(ast::Entity) -> Entity;
+
+impl IntoIterator for Entities {
+    type Item = Entity;
+
+    type IntoIter =
```
shaobo-he-aws, line 634: I think this type is too complicated, if not idiomatic. IMO, we should follow what Rust standard library does. For instance, `HashMap`'s `IntoIter` type is a dedicated type [`std::collections::hash_map::IntoIter`](https://doc.rust-lang.org/std/collections/hash_map/struct.IntoIter.html). We should at least use an instantiation of this type instead.

shaobo-he-aws (changes requested):

cedar-policy/src/api.rs
john-h-kastner-aws: This is a good point. IIUC, this type becomes part of our public API, so if we don't wrap it would be a potentially braking change to make it return something other `std::iter::Map`

john-h-kastner-aws (approved):
Two small issue in CI, but otherwise looks good.

cedar-policy/src/api.rs
aaronjeline: Looks good. One tiny idiomatic change, can you rename it to `IntoIter` (like collections do in the standard library)

shaobo-he-aws (approved):

cedar-policy/src/api.rs
john-h-kastner-aws: Wouldn't also mean moving it into a module `mod entities { }` so that it's `entities::IntoIter`?

cre4ture:
> Two small issue in CI, but otherwise looks good.

CI should be green now.

shaobo-he-aws (approved):

john-h-kastner-aws (approved):

cre4ture:
cool. Thanks for merging. I was able to learn more about rust than I was initially expecting. :-)

Issue #539, opened 2023-12-21
Move `ImpossiblePolicy` from error to warning


khieta:
### Category

Cedar validation features

### Describe the feature you'd like to request

The validate function currently (v3.x) returns a [`ValidationResult`](https://docs.rs/cedar-policy/latest/cedar_policy/struct.ValidationResult.html), which consists of `validation_errors` and `validation_warnings`. One current validation error is the `TypeError` [`ImpossiblePolicy`](https://docs.rs/cedar-policy/latest/cedar_policy/enum.TypeErrorKind.html#variant.ImpossiblePolicy), which indicates that a policy will always evaluate to false (i.e., it will never fire). We had always intended for this to be more of a "warning" than an "error," but we didn’t add official support for warnings until recently (see #225). We would now like to move this error to be a warning instead.

### Benefits

* Since the `ImpossiblePolicy` error is returned on a best-effort basis (i.e., we may miss some "impossible" policies), it is sensitive to changes in our typing precision. This means that seemingly small changes to our typing algorithm could easily become breaking changes. For example: say that we decide to encode in the validator that an expression like `1 == 2` is always false (this is currently not checked). Then policies that previously validated will no longer be valid.
* Although having a policy that never applies is likely an error, it is not an issue that would lead to an authorization-time error, which is what the validator is intended to protect against. So this "feels" more like a warning than an error.
* There could be legitimate reasons for generating policies that never apply. For example, a user might be generating policies programmatically, expecting that "impossible" policies are ignored.
* This would set the precedent for adding more warnings to the validator. Easy example: check whether a policy always evaluates to true (i.e., will always fire).

### Impact

Some policies that previously would not validate will now validate, but with a warning. All previously-validated policies will remain valid.

### Describe alternatives you've considered

We could leave `ImpossiblePolicy` as an error, which is the status quo.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

shaobo-he-aws:
Implemented via #716

PR #716, opened 2024-03-11
Move `ImpossiblePolicy` from error to warning

khieta:
## Description of changes

* Deprecated `TypeErrorKind::ImpossiblePolicy`
* Added `ValidationWarningKind::ImpossiblePolicy` ⬅️ this requires a minor version bump
* Moved the definition of `ValidationWarning` to `validation_results.rs`
* Updated unit tests ⬅️ this accounts for most of the diff

As discussed in #539, `ImpossiblePolicy` is different from our other errors because it does not signal a potential authorization-time error. Instead, it indicates a policy that will not apply for any valid request.

As we increase the precision of the Cedar typechecker, `ImpossiblePolicy` errors will occur in more cases (a breaking change), while other types of errors will occur in fewer cases (a non-breaking change). So to make potential changes easier going forward, we've decided to demote `ImpossiblePolicy` to a warning. We consider this a _non-breaking_ change because all policies that previously validated will continue to validate.

## Issue #, if available

#539 

## Disclaimer


khieta:
Note: the corpus test failure is expected since the tests were generated when we expected "impossible policy" to be an error (and thus the test says that the policy should not validate)

cedar-policy-validator/src/rbac.rs
```diff
@@ -1149,6 +1150,48 @@ mod test {
...
 
+    fn assert_validate_policy_succeeds(schema: ValidatorSchema, policy: &Template) {
+        let validate = Validator::new(schema);
+        assert!(
+            validate
+                .validate_policy(&policy, ValidationMode::default())
+                .0
+                .next()
+                .is_none(),
+            "Did not expect any validation notes."
```
cdisselkoen, line 1161: is "notes" the generic term for warnings+errors?

cedar-policy-validator/src/rbac.rs
```diff
@@ -1308,29 +1335,15 @@ mod test {
...
 
-        let validate = Validator::new(schema.clone());
-        assert!(
-            validate
-                .validate_policy(&policy, ValidationMode::default())
-                .next()
-                .is_none(),
-            "Did not expect any validation notes."
-        );
+        assert_validate_policy_succeeds(schema.clone(), &policy);
```
cdisselkoen, line 1338: to avoid this `.clone()`, we could have `assert_validate_policy_succeeds()` take a `&Validator` which could be reused across multiple `assert_validate` calls

cedar-policy-validator/src/typecheck.rs
```diff
@@ -266,8 +266,15 @@ impl<'a> Typechecker<'a> {
     /// ensures that the policy expression has type boolean. If typechecking
     /// succeeds, then the method will return true, and no items will be
     /// added to the output list. Otherwise, the function returns false and the
-    /// output list is populated with any errors encountered while typechecking.
-    pub fn typecheck_policy(&self, t: &Template, type_errors: &mut HashSet<TypeError>) -> bool {
+    /// input `type_errors` list is populated with any errors encountered
+    /// while typechecking. In addition, warnings may be added to the `warnings`
+    /// list, although these will oot impact the boolean return value.
```
cdisselkoen, line 271:
```suggestion
    /// list, although these will not impact the boolean return value.
```

cedar-policy-validator/src/typecheck.rs
```diff
@@ -266,8 +266,15 @@ impl<'a> Typechecker<'a> {
...
+    pub fn typecheck_policy(
+        &self,
+        t: &Template,
+        type_errors: &mut HashSet<TypeError>,
+        warnings: &mut HashSet<ValidationWarningKind>,
+    ) -> bool {
```
cdisselkoen, line 277: this boolean return value used by anyone? or ignored by all callers?

should callers just check `type_errors.is_empty()` and `warnings.is_empty()` to be more explicit about whether they are checking for the absence of warnings or not?

cedar-policy-validator/src/typecheck/test_optional_attributes.rs
```diff
@@ -776,30 +778,17 @@ fn action_attrs_failing() {
         )],
     );
 
-    // Doesn't fail do to imprecision in ActionEntity LUB computation requiring `may_have_attr` to return true for ActionEntity types
+    // Doesn't fail due to imprecision in ActionEntity LUB computation requiring `may_have_attr` to return true for ActionEntity types
```
cdisselkoen, line 781: Despite the comment, this actually does fail, right? if "fail" means that validation fails, rather than that the test fails

cedar-policy-validator/src/rbac.rs
```diff
@@ -1149,6 +1150,48 @@ mod test {
         (principal_euid, action_euid, resource_euid, schema)
     }
 
+    fn assert_validate_policy_succeeds(schema: ValidatorSchema, policy: &Template) {
```
john-h-kastner-aws, line 1153: We can use `#[track_caller]` on these

cdisselkoen (approved):

cedar-policy-validator/src/rbac.rs
```diff
@@ -1149,6 +1150,48 @@ mod test {
...
+        let validate = Validator::new(schema);
+        assert!(
+            validate
+                .validate_policy(&policy, ValidationMode::default())
+                .0
```
john-h-kastner-aws, line 1158: Should this also check `.1` for warnings? Not strictly necessary for validation success, but it should be the case for the tests

john-h-kastner-aws (approved):

cedar-policy-validator/src/rbac.rs
khieta: I believe "notes" is the historical way to refer to errors... I'll update the terminology.

cedar-policy-validator/src/typecheck.rs
khieta: Return value is only used in tests -- I was on the fence about whether to remove it.

cedar-policy-validator/src/typecheck/test_optional_attributes.rs
john-h-kastner-aws: Yes. We changed this so that we know the attributes of action entities precisely (there typically being none, but code exists to handle attribute if they exist).

cedar-policy-validator/src/typecheck.rs
khieta: Decided to remove in the latest commit.

khieta: Scratch that -- turns out that the boolean value does matter sometimes. Will update comments accordingly.

john-h-kastner-aws: We can still remove it if we want, but it would be beyond the scope of this PR

john-h-kastner-aws (approved):

khieta:
Fyi: this PR will be on hold temporarily since I'll be OOTO until next Wednesday. I'm holding off on merging for now because I don't want to leave the corpus tests in a broken state.

khieta:
I've pulled in the latest changes from `main`, so this PR is ready to merge (pending re-approval). Other PRs should expect the `Run cargo test --verbose --features "integration-testing" -- --ignored` step of CI to fail until I can manually update the corpus tests. I expect to do this either this weekend or early next week.

john-h-kastner-aws (approved):

PR #718, opened 2024-03-12
Expose formatter and schema/entity/context parsing to wasm

ncsze:
## Description of changes

Broken off of/added to the changes under this larger CR: https://github.com/Swolebrain/cedar/commit/a1b511923e30020cd0e26b4bed6ab7cd3e8b2402

- Added policy formatting to wasm: `wasm_format_policies`
- Added and `check_parse_<schema/context/entities>` functions to wasm to allow schema to be parsed and checked for errors, as well as for entities and context to be checked against the schema.

## Issue #, if available

N/A

## Disclaimer

### DIFF
--- a/cedar-wasm/src/formatter.rs
+++ b/cedar-wasm/src/formatter.rs
@@ -1,0 +1,76 @@
+use cedar_policy_formatter::{policies_str_to_pretty, Config};
+use serde::{Deserialize, Serialize};
+
+use tsify::Tsify;
+use wasm_bindgen::prelude::*;
+
+#[derive(Tsify, Debug, Serialize, Deserialize)]
+#[tsify(from_wasm_abi, into_wasm_abi)]
+pub struct FormattingResult {
+    success: bool,
+    #[serde(rename = "formattedPolicy", skip_serializing_if = "Option::is_none")]
+    formatted_policy: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    error: Option<String>,
+}
+
+#[wasm_bindgen(js_name = "formatPolicies")]
+pub fn wasm_format_policies(
+    policies_str: &str,
+    line_width: i32,
+    indent_width: i32,
+) -> FormattingResult {
+    let line_width: usize = match line_width.try_into() {
+        Ok(width) => width,
+        Err(_) => {
+            return FormattingResult {
+                success: false,
+                formatted_policy: None,
+                error: Some("Input size error (line width)".to_string()),
+            }
+        }
+    };
+    let indent_width: isize = match indent_width.try_into() {
+        Ok(width) => width,
+        Err(_) => {
+            return FormattingResult {
+                success: false,
+                formatted_policy: None,
+                error: Some("Input size error (indent width)".to_string()),
+            }
+        }
+    };
+    let config = Config {
+        line_width,
+        indent_width,
+    };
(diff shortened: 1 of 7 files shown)


ncsze:
Changed the conversion to `usize`/`isize` in formatter to avoid this clippy error:
```
error: used `unwrap()` on a `Result` value
  --> cedar-wasm/src/formatter.rs:34:23
   |
34 |         indent_width: indent_width.unwrap(),
   |                       ^^^^^^^^^^^^^^^^^^^^^
   |
   = note: if this value is an `Err`, it will panic
```

ncsze:
Added TS type for context

shaobo-he-aws (approved):

cedar-wasm/src/formatter.rs
```diff
@@ -0,0 +1,76 @@
...
+    #[serde(rename = "formattedPolicy", skip_serializing_if = "Option::is_none")]
+    formatted_policy: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    error: Option<String>,
+}
+
+#[wasm_bindgen(js_name = "formatPolicies")]
+pub fn wasm_format_policies(
+    policies_str: &str,
+    line_width: i32,
```
andrewmwells-amazon, line 20: Can you explain a bit more why we have 3 types for these (`i32`, `usize` and `isize`). I'm guessing `i32` is what wasm uses. Can we just use that everywhere here?

cedar-wasm/src/schema_and_entities_and_context.rs
```diff
@@ -0,0 +1,251 @@
...
+    let parsed_schema = match Schema::from_str(schema_str) {
+        Ok(schema) => schema,
+        Err(err) => {
+            return CheckParseResult::SyntaxError {
+                errors: vec![err.to_string()],
+            }
+        }
+    };
+    match Entities::from_json_str(entities_str, Some(&parsed_schema)) {
+        Ok(_entities) => CheckParseResult::Success,
```
andrewmwells-amazon, line 42:
```suggestion
        Ok(_) => CheckParseResult::Success,
```

andrewmwells-amazon (changes requested):

cedar-wasm/src/formatter.rs
ncsze: The `cedar_policy_formatter::Config` type requires `usize` and `isize`. It seems like I can get away with making the inputs those types and not converting from `i32`, even though wasm only has `i32` and `i64` types internally. `wasm-pack` seems to build the function such that it works with the normal numbers in javascript. Thoughts on this @Swolebrain ?

Swolebrain: wouldn't that be a breaking change to the formatter though?

andrewmwells-amazon: Got it. I don't see why `indent_width` would be negative, but you're right that it would be a breaking change. Resolving.

andrewmwells-amazon (approved):

cedar-wasm/src/formatter.rs
shaobo-he-aws: The reason is that the library we used to implement the formatter makes it that way. So I just propagated the behavior.

shaobo-he-aws (approved):

PR #726, opened 2024-03-14
bump version numbers on `main`

cdisselkoen:

### DIFF
--- a/cedar-policy-cli/Cargo.toml
+++ b/cedar-policy-cli/Cargo.toml
@@ -2,7 +2,7 @@
 name = "cedar-policy-cli"
 edition = "2021"
 
-version = "3.0.0"
+version = "3.1.0"
 license = "Apache-2.0"
 categories = ["compilers", "config"]
 description = "CLI interface for the Cedar Policy language."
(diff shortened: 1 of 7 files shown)


khieta:
Definitely time to update these version numbers 🙂 But I think I'd prefer bumping them to the next major version (4.0) to avoid any confusion with the 3.1.0 version on crates.io.

aaronjeline (approved):

cdisselkoen:
@khieta what if we end up cutting the 3.2.x release branch from `main` like we did for 3.1.x?

I guess we'd just have to down-bump the version numbers on the 3.2.x release branch after reverting the PRs with breaking changes.  Is that the system we want?

khieta:
> what if we end up cutting the 3.2.x release branch from main like we did for 3.1.x?
>
> I guess we'd just have to down-bump the version numbers on the 3.2.x release branch after reverting the PRs with breaking changes. Is that the system we want?

Personally, I think this is fine. We are already used to doing version number doctoring before we release, so this isn't a big change from the status quo. (Note that with your current PR we'd have to change 3.1 -> 3.2 before release.) I feel like it makes more sense for the `main` branch to be 4.0 than 3.2 because it currently contains breaking changes that will have to wait until the next major release.

cdisselkoen:
Sure, you've sold me.

khieta (approved):

PR #737, opened 2024-03-19
Make non-json isAuthorized and validate public, rework wasm interfaces

ncsze:
## Description of changes
- Made `isAuthorized` and `validate` functions in the frontend public, as well as their related structs: `AuthorizationAnswer`, `AuthorizationCall`, `ValidateCall`, `ValidationSettings`, `ValidationMode`, `ValidateAnswer`, `ValidationNote`
- Changed `wasm_is_authorized` and `wasm_validate` to use non-json versions of `isAuthorized` and `validate`, and to return the data in `AuthorizationAnswer` and `ValidateAnswer` directly (instead of in a JSON string) in a wasm-specific result struct.
- Changed all wasm result interfaces to use untagged enums and to always contain `errors: Vec<String>` in their error case. In TS, users can now check for success with duck typing instead of the misleading `success` string. 

## Issue #, if available

N/A

## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [x] A backwards-compatible change requiring a minor version bump to `cedar-policy` (e.g., addition of a new API)

I confirm that this PR (choose one, and delete the other options):

- [x] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [x] Does not require updates because my change does not impact the Cedar formal model or DRT infrastructure.

PR #740, opened 2024-03-20
Cleanup in frontend code

khieta:
## Description of changes

This PR fixes a few inconsistencies in our frontend, which will become visible once we make these structures public in #737. I believe that all of these changes are non-breaking because they affect private data structures... but please double check my work.
* `ValidateCall` -> `ValidationCall` for consistency with `AuthorizationCall`
* `ValidateAnswer` -> `ValidationAnswer` for consistency with `AuthorizationAnswer`
* `ParseFailed` -> `Failure` since failures encompass more than just parsing (e.g., failure to convert a `Schema` to the internal `ValidatorSchema` type)
* "notes" -> "validation errors"
* Added support for returning validation warnings
* Updated `validate` to call the public function in `cedar_policy` instead of the internal one in `cedar_policy_validator`
* `ValidationMode` -> `ValidationEnabled` since "mode" usually refers to "strict" vs. "permissive"
* Updated the schema field in `ValidationCall` to be a `JsonValueWithNoDuplicateKeys` for consistency with `AuthorizationCall`

A few other things I'd like to change, but didn't because they will be breaking (will make in a followup PR):
* Rename `InterfaceResponse`, `InterfaceDiagnostics` to `Response`, `Diagnostics`. We are already redefining types (e.g., `ValidationError`) so I don't think this is too confusing.
* Get rid of `InterfaceResult` and instead use the `Failure` field of `AuthorizationAnswer`, `ValidationAnswer` to report errors. I don't think it's worth the extra infrastructure to distinguish between internal and non-internal errors.
* Rename `json_is_authorized` to `json_str_is_authorized` & update it to return a json string encoding of a `AuthorizationAnswer`. Similar for `json_validate`.

### DIFF
--- a/cedar-policy/src/frontend/is_authorized.rs
+++ b/cedar-policy/src/frontend/is_authorized.rs
@@ -24,8 +24,8 @@ use crate::api::EntityTypeName;
 use crate::api::PartialResponse;
 use crate::PolicyId;
 use crate::{
-    Authorizer, Context, Decision, Entities, EntityUid, Policy, PolicySet, Request, Response,
-    Schema, SlotId, Template,
+    Authorizer, Context, Decision, Entities, EntityUid, PolicySet, Request, Response, Schema,
+    SlotId,
 };
 use cedar_policy_core::jsonvalue::JsonValueWithNoDuplicateKeys;
 use itertools::Itertools;
@@ -55,7 +55,7 @@ fn is_authorized(call: AuthorizationCall) -> AuthorizationAnswer {
                     .into(),
             })
         }
-        Err(errors) => AuthorizationAnswer::ParseFailed { errors },
+        Err(errors) => AuthorizationAnswer::Failure { errors },
     }
 }
 
@@ -68,9 +68,7 @@ pub fn json_is_authorized(input: &str) -> InterfaceResult {
         |e| InterfaceResult::fail_internally(format!("error parsing call: {e:}")),
         |call| match is_authorized(call) {
             answer @ AuthorizationAnswer::Success { .. } => InterfaceResult::succeed(answer),
-            AuthorizationAnswer::ParseFailed { errors } => {
-                InterfaceResult::fail_bad_request(errors)
-            }
+            AuthorizationAnswer::Failure { errors } => InterfaceResult::fail_bad_request(errors),
         },
     )
 }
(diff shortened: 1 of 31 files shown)


cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -557,19 +564,7 @@ impl RecvdSlice {
             template_instantiations,
         } = self;
 
-        let policy_set = match policies {
-            PolicySpecification::Concatenated(policies) => match PolicySet::from_str(&policies) {
```
khieta, line 561: This parsing code, and the `parse_policy_set_from_individual_policies` function below were copied verbatim to `util.rs`.

andrewmwells-amazon (approved):

cedar-policy/src/frontend/utils.rs
```diff
@@ -98,14 +163,10 @@ impl InterfaceResult {
...
 #[cfg(test)]
 #[track_caller] // report the caller's location as the location of the panic, not the location in this function
-pub(crate) fn assert_is_failure(result: &InterfaceResult, internal: bool, err: &str) {
+pub(crate) fn assert_is_failure(result: &InterfaceResult, internal: bool) {
     use cool_asserts::assert_matches;
-    use itertools::Itertools;
 
-    assert_matches!(result, InterfaceResult::Failure { is_internal, errors } => {
-        assert!(
-            errors.iter().exactly_one().unwrap().contains(err),
```
john-h-kastner-aws, line 107: Can we keep this assertion?

john-h-kastner-aws (changes requested):

andrewmwells-amazon:
If you rebase, the downstream build should hopefully pass now

cedar-policy/src/frontend/utils.rs
khieta: I unified the code used to parse policies in `is_authorized.rs` and `validate.rs`, which resulted in some changed error messages (in particular, there was no longer exactly one error in some cases). Was easier to delete the assertion than alter it 😆 

I'll work on adding it back this afternoon.

khieta: Revised test cases. Lmk what you think.

cedar-policy/src/frontend/validate.rs
```diff
@@ -18,136 +18,129 @@
...
 #[cfg_attr(feature = "wasm", derive(tsify::Tsify))]
 #[cfg_attr(feature = "wasm", tsify(into_wasm_abi, from_wasm_abi))]
-struct ValidateCall {
+struct ValidationCall {
     #[serde(default)]
     #[serde(rename = "validationSettings")]
     validation_settings: ValidationSettings,
-    schema: cedar_policy_validator::SchemaFragment,
+    /// Schema in JSON format
+    schema: JsonValueWithNoDuplicateKeys,
```
john-h-kastner-aws, line 77: What's the benefit of this change?

cedar-policy/src/frontend/validate.rs
```diff
@@ -18,136 +18,129 @@
...
+            ValidationAnswer::Failure { errors } => InterfaceResult::fail_bad_request(errors),
         },
     )
 }
 
+/// Struct containing the input data for validation
 #[derive(Serialize, Deserialize)]
 #[cfg_attr(feature = "wasm", derive(tsify::Tsify))]
 #[cfg_attr(feature = "wasm", tsify(into_wasm_abi, from_wasm_abi))]
-struct ValidateCall {
```
john-h-kastner-aws, line 107: The `Validate` to `Validation` change looks nice, but I wonder if it's worth making a breaking change to the API. Depends on how much we care about keeping this interface stable.

john-h-kastner-aws (approved):

cedar-policy/src/frontend/validate.rs
khieta: I _think_ this isn't breaking since the `ValidateCall` type was only ever exposed indirectly via its serialized JSON format.

cedar-policy/src/frontend/validate.rs
khieta: Consistency with `AuthorizationCall`

john-h-kastner-aws (approved):

PR #744, opened 2024-03-22
Add the `euid` macro

shaobo-he-aws:
## Issue #, if available
#64

### DIFF
--- a/cedar-policy-core/cedar-policy-macros/src/lib.rs
+++ b/cedar-policy-core/cedar-policy-macros/src/lib.rs
@@ -1,0 +1,53 @@
+use cedar_policy_core::ast::{Eid, EntityUID, Name};
+use cedar_policy_core::parser::err::ParseErrors;
+use cedar_policy_core::FromNormalizedStr;
+use itertools::Itertools;
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse::Parse, punctuated::Punctuated, Token};
+use syn::{parse_macro_input, Ident, LitStr};
+
+#[derive(Clone)]
+struct Input {
+    path: Vec<Ident>,
+    id: LitStr,
+}
+
+impl Parse for Input {
+    fn parse(input: syn::parse::ParseStream) -> syn::Result<Self> {
+        let path = Punctuated::<Ident, Token![::]>::parse_separated_nonempty(input)?;
+        let _: Token![,] = input.parse()?;
+        let id: LitStr = input.parse()?;
+        Ok(Self {
+            path: path.into_iter().collect(),
+            id,
+        })
+    }
+}
+
+impl TryFrom<Input> for EntityUID {
+    type Error = ParseErrors;
+    fn try_from(value: Input) -> Result<Self, Self::Error> {
+        let name = Name::from_normalized_str(&value.path.into_iter().join("::"))?;
+        let eid = Eid::new(value.id.value());
+        Ok(Self::from_components(name, eid))
+    }
+}
+
+#[proc_macro]
+pub fn euid(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as Input);
+    let euid: EntityUID = match input.try_into() {
+        Ok(euid) => euid,
+        Err(_) => panic!("invalid euid"),
+    };
+    let euid_str = euid.to_string();
+    quote! { #euid_str.parse().unwrap() }.into()
+}
+
+#[test]
+fn tests() {
+    let t = trybuild::TestCases::new();
+    t.pass("tests/pass.rs");
+    //t.compile_fail("tests/fail.rs");
+}
--- a/cedar-policy-core/cedar-policy-macros/tests/fail.rs
+++ b/cedar-policy-core/cedar-policy-macros/tests/fail.rs
@@ -1,0 +1,8 @@
+use cedar_policy_core::ast::EntityUID;
+use cedar_policy_macros::euid;
+
+fn main() {
+    let _: EntityUID = euid!(f, "bar");
+    let _: EntityUID = euid!(Foo::Bar, "baz");
+    let _: EntityUID = euid!(Foo, "ha");
+}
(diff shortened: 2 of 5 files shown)


shaobo-he-aws:
The prototype expands `euid!(<EntityType>, <Eid>)` to `<EuidStr>.parse().unwrap()`, where `<EuidStr>` is the string representation of the `EntityUID` made of `<EntityType>` and `<Eid>`. So, it requires users to annotate the type of the return value like `let euid: EntityUID = euid!(...)`.

cedar-policy-core/cedar-policy-macros/tests/fail.rs
```diff
@@ -0,0 +1,8 @@
+use cedar_policy_core::ast::EntityUID;
+use cedar_policy_macros::euid;
+
+fn main() {
+    let _: EntityUID = euid!(f, "bar");
+    let _: EntityUID = euid!(Foo::Bar, "baz");
+    let _: EntityUID = euid!(Foo, "ha");
+}
```
cdisselkoen, line 8: Why is this file called `fail.rs`? what about these examples fails?

cdisselkoen (commented):
1) Is it possible to have a test that the `euid!()` macro fails at compile time if given invalid arguments (like, disallowed characters in the type name, such as `+`)?

2) A full migration can be out of scope for this PR, but can we at least adapt a few tests in Core to use this macro, just to proof-of-concept show that this macro works for the intended purpose?

cedar-policy-core/cedar-policy-macros/tests/pass.rs
```diff
@@ -0,0 +1,17 @@
+use cedar_policy_core::ast::{Eid, EntityUID};
+use cedar_policy_macros::euid;
+
+fn main() {
+    let e: EntityUID = euid!(Foo, "bar");
```
aaronjeline, line 5: If this returns `core::EntityUID` how do consumers of `cedar-policy` use it?

aaronjeline (changes requested):

john-h-kastner-aws:
I don't think this will catch a mistake like `let id : EntityID = euid!(Foo, "bar");`. It will check that `EntityUID::from_str("Foo::\"bar\"")` succeeds, but would then expand to `let id : EntityId = "Foo::\"bar\".parse().unwrap()`, which panics.

shaobo-he-aws:
> I don't think this will catch a mistake like `let id : EntityID = euid!(Foo, "bar");`. It will check that `EntityUID::from_str("Foo::\"bar\"")` succeeds, but would then expand to `let id : EntityId = "Foo::\"bar\".parse().unwrap()`, which panics.

Is your example like https://github.com/cedar-policy/cedar/pull/744/files#diff-818bcd662e3182821f9a69e7c30d881a7a89297b0c8bd3b3e82ed470d6b9e3b9R5?

cedar-policy-core/cedar-policy-macros/tests/fail.rs
shaobo-he-aws: This is actually related to your second question above. This file should contain macro usages that do not compile. However, it seems cargo will try to build each file in this package, which contains this file that is supposed to fail to compile.

cedar-policy-core/cedar-policy-macros/tests/pass.rs
shaobo-he-aws: I think this macro can only be used by cedar-policy because of circular dependencies. That is, if we use `core::EntityUID` to parse the input, we cannot export it to core since that would form a mutual reference. I probably miss something (e.g., crate vs. package) given my limited knowledge of Rust. I'm open to suggestions.

john-h-kastner-aws:
> > I don't think this will catch a mistake like `let id : EntityID = euid!(Foo, "bar");`. It will check that `EntityUID::from_str("Foo::\"bar\"")` succeeds, but would then expand to `let id : EntityId = "Foo::\"bar\".parse().unwrap()`, which panics.
> 
> Is your example like https://github.com/cedar-policy/cedar/pull/744/files#diff-818bcd662e3182821f9a69e7c30d881a7a89297b0c8bd3b3e82ed470d6b9e3b9R5?

I'm thinking about what would happen if the user accidentally wrote `EntityID` instead of `EntityUID`.

shaobo-he-aws:
> I'm thinking about what would happen if the user accidentally wrote `EntityID` instead of `EntityUID`.

I see. I think we can call `cedar_policy::EntityUid::from_str` explicitly. But we wouldn't be able to unit-test it in `cedar-policy-macros`.

cedar-policy/src/tests.rs
```diff
@@ -3443,3 +3443,37 @@ mod issue_619 {
...
+    use std::str::FromStr;
+
+    use crate::euid;
+
+    use crate::EntityId;
+    use crate::EntityUid;
+
+    #[test]
+    fn pass() {
+        let e: EntityUid = euid!(Foo, "bar");
```
john-h-kastner-aws, line 3457: Would it be possible let the macro be written like `euid!(Foo::"bar")`? The way you have it looks fine, but IMO it would be even nice to use the exact euid syntax.

shaobo-he-aws: I prefer the existing one because it explicitly requires users to adopt Rust strings as `eid`s, thus avoiding to reinterpret the `eid` component.

shaobo-he-aws: That being said, I'm still open to suggestions here.

cedar-policy-core/cedar-policy-macros/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+    let (basename, path) = ids.split_last().unwrap();
+    let euid: EntityUID = EntityUID::from_components(
+        Name::new(basename.clone(), path.into_iter().cloned()),
+        Eid::new(input.id.value()),
+    );
+    let euid_str = euid.to_string();
+    // PANIC SAFETY: `#euid_str` should parse
+    // Note that this comment is used to pass panic checks
+    // Users still need to annotate macro call sites to get rid of
+    // `clippy::unwrap_used`
```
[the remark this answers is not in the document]
shaobo-he-aws: There's no such annotation inside the quote, right?

aaronjeline:
This is something we are still interested in completing, but as of right now we have not gotten the time.

shaobo-he-aws:
Closed for now.

Issue #745, opened 2024-03-25
[4.0] Consistent public error types in the Rust API


cdisselkoen:
### Category

Error message improvements

### Describe the feature you'd like to request

Currently Cedar is inconsistent about what error data is public (accessible to Cedar consumers and thus part of our stable interface) and what error data is private (implementation details we're free to change at will, e.g., to improve error messages).  In fact, at least six different patterns are currently in use in different places in the Cedar codebase, as described below.  This issue requests that Cedar standardize all of its error types into __Pattern C__.  This will be a breaking change to some APIs.

## Pattern A

Methods that can return errors do not publicize what type they are returning, just `impl miette::Diagnostic`.  Callers (Cedar consumers) can use the methods on `miette::Diagnostic` to get the error's message, help text, source location, etc but no other info.  

[`Policy::to_json()` currently takes this approach:](https://docs.rs/cedar-policy/3.1.1/cedar_policy/struct.Policy.html#method.to_json)
```
pub fn to_json(&self) -> Result<serde_json::Value, impl miette::Diagnostic>
```

This option is maximally flexible for Cedar developers, as Cedar can make any changes to error structures that it wants, even in patch releases, as long as whatever is returned still implements `miette::Diagnostic` (which will always be the case).  But, this option is undesirable because it does not provide enough information to callers (Cedar consumers).  Callers cannot distinguish between different error cases in a structured way, and cannot get other information about the error in a structured way -- for instance, seeing which `EntityUid` was responsible for an `EntityDoesNotExist` error, except by inspecting the error message.

## Pattern B

Methods that can return errors return a named error type, but that type is a struct with private field(s).  Of course, the error type implements `miette::Diagnostic` so provides all of those methods, to get the error's message, help text, source location, etc.  The error type can also provide additional public methods if it wants.  

[`eval_expression()`](https://docs.rs/cedar-policy/3.1.1/cedar_policy/fn.eval_expression.html) (the public interface for the Cedar evaluator) currently takes this approach:
```
pub fn eval_expression(
    request: &Request,
    entities: &Entities,
    expr: &Expression,
) -> Result<EvalResult, EvaluationError>
```
where callers (Cedar consumers) just see `EvaluationError` as a struct with private fields in [the docs](https://docs.rs/cedar-policy/3.1.1/cedar_policy/struct.EvaluationError.html). 

[...]

PR #747, opened 2024-03-26
cedar#745 for PolicyToJsonError

cdisselkoen:
## Description of changes

Applies the recommendations in #745 for `PolicyToJsonError`.

Questions:

- Do we like how this looks? Do we want to change the recommendations in any way?
- We'll end up with a large number of these member structs that are `pub` but with private fields.  Should we organize them into submodules?  One submodule per error enum or something?

## Issue #, if available

#745

No changelog update, let's do one at the end for all of #745.

### DIFF
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -3400,9 +3400,9 @@ impl From<serde_json::Error> for PolicyToJsonError {
 
 /// Error types related to JSON processing
 pub mod json_errors {
+    use cedar_policy_core::est;
     use miette::Diagnostic;
     use thiserror::Error;
-    use cedar_policy_core::est;
 
     /// Error linking the JSON representation of a linked policy
     #[derive(Debug, Diagnostic, Error)]
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -2869,10 +2869,9 @@ impl Template {
     }
 
     /// Get the JSON representation of this `Template`.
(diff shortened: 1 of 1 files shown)


aaronjeline (approved):
The other option is returning `Box<dyn ..>` is there a reason against that?

cdisselkoen:
That would correspond roughly to Pattern A in #745.  #745 covers some reasons for recommending what it calls Pattern C, which is implemented here.  If we don't like how this looks, we can consider Patterns A or B

aaronjeline:
> That would correspond roughly to Pattern A in #745. #745 covers some reasons for recommending what it calls Pattern C, which is implemented here. If we don't like how this looks, we can consider Patterns A or B

Caching that discussion back into brain

john-h-kastner-aws (approved):
I like how this looks, but +1 to grouping errors into modules.
We'll end up with a lot of structs. The only thing to consider with that is that we might want to re-use error structs between different enums which could end up looking a bit odd when they're in different modules, but that's probably fine.

cdisselkoen:
Moved the two new error structs into a submodule.  Not sure if we want a submodule per top-level error type, or just by general category, or just one submodule for detailed error structs.  Current PR does the second option.  Can always move errors between submodules prior to 4.0 release.

cdisselkoen:
I'm unsure whether the Java CI failure is related to this PR or not

john-h-kastner-aws (approved):

cdisselkoen:
The Java CI failure is unrelated and will be addressed separately.

PR #748, opened 2024-03-26
reorg how advice is stored on `EvaluationError`

cdisselkoen:
## Description of changes

Pushes the `advice` down from `EvaluationError` into `EvaluationErrorKind`.  Only `TypeError` was using the dynamic capability, everyone else is fine with static.  This comes out from @khieta's comment https://github.com/cedar-policy/cedar/issues/745#issuecomment-2018696491; I'll do source_loc next.

Unfortunately, this change is breaking, because `EvaluationErrorKind` is public and `TypeError`'s fields are public.  After #745, this kind of change would not be breaking.

### DIFF
--- a/cedar-policy-core/src/evaluator.rs
+++ b/cedar-policy-core/src/evaluator.rs
@@ -465,12 +465,15 @@ impl<'e> Evaluator<'e> {
                                 // If arg1 is not an entity and arg2 is a set, then possibly
                                 // the user intended `arg2.contains(arg1)` rather than `arg1 in arg2`.
                                 // If arg2 is a record, then possibly they intended `arg2 has arg1`.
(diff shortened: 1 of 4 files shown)


khieta (approved):
This looks like an improvement to me. I especially like the `NonValue` case, where the help information is right next to the error definition.

john-h-kastner-aws (approved):

cdisselkoen:
Am I correct that both of the CI failures are expected / unrelated to this PR?

khieta:
Yep. I'm working on the integration test failure and @andrewmwells-amazon has up [a PR](https://github.com/cedar-policy/cedar-java/pull/117) to fix the CedarJava failure (which also turned out to be my fault 😅).

PR #751, opened 2024-03-27
improve error when `null` occurs in entity json data

### LINKED ISSUE #530: Improve error when `null` occurs in entity json data
### Category

Other

After discussion on https://github.com/cedar-policy/rfcs/pull/37 (rejected), we have decided that we do not want to support `null` JSON values in entity JSON data. They are already an error, so no change is required there, but we should improve the error message.

The current error is `data did not match any variant of untagged enum CedarValueJson` 

.

### Additional context

_No response_

cdisselkoen:
## Description of changes

As it says in the title.

## Issue #, if available

Fixes #530

Not breaking because the error type being changed is not re-exported in `cedar-policy`.

### DIFF
--- a/cedar-policy-core/src/entities/json/value.rs
+++ b/cedar-policy-core/src/entities/json/value.rs
@@ -94,6 +94,9 @@ pub enum CedarValueJson {
     Record(
         #[cfg_attr(feature = "wasm", tsify(type = "{ [key: string]: CedarValueJson }"))] JsonRecord,
     ),
+    /// JSON null, which is never valid, but we put this here in order to
+    /// provide a better error message.
+    Null,
 }
 
 /// Structure representing a Cedar record in JSON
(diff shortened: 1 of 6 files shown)


aaronjeline (approved):

cedar-policy-core/src/entities/json/value.rs
```diff
@@ -94,6 +94,9 @@ pub enum CedarValueJson {
     Record(
         #[cfg_attr(feature = "wasm", tsify(type = "{ [key: string]: CedarValueJson }"))] JsonRecord,
     ),
+    /// JSON null, which is never valid, but we put this here in order to
+    /// provide a better error message.
+    Null,
```
john-h-kastner-aws, line 99: Does this just implicitly deserailize from null without additional annotation?

john-h-kastner-aws (approved):

cedar-policy-core/src/entities/json/value.rs
cdisselkoen: It does seem so, as shown by this PR's tests

Issue #757, opened 2024-03-28
Cedar FFI Overhaul


khieta:
### Category

User level API features/changes

### Describe the feature you'd like to request

The code in `cedar-policy/src/frontend/` acts as our public interface for JSON-based FFI. It is currently used by the [Java](https://github.com/cedar-policy/cedar-java) and [WASM](https://github.com/cedar-policy/cedar/tree/main/cedar-wasm) bindings. We plan to use it for the differential testing harness too, in the future (see [#75](https://github.com/cedar-policy/cedar-spec/issues/75), [#76](https://github.com/cedar-policy/cedar-spec/issues/76), [#125](https://github.com/cedar-policy/cedar-spec/issues/125)). But there are several rough edges in our frontend interface as it is now. This issue proposes to overhaul our FFI interface. This will result in breaking changes that will need to be delayed to the next major release (as of writing, 4.0 or greater).

### Entry point for `is_authorized`

```rust
// Basic interface
is_authorized : AuthorizationCall -> AuthorizationAnswer

//  Input is a json encoding of `AuthorizationCall` and output is a json encoding of `AuthorizationAnswer`
is_authorized_json : serde_json::Value -> serde_json::Value

// Input and output are strings containing serialized JSON
is_authorized_json_str : &str -> String
```

### `AuthorizationCall` type

```rust
struct AuthorizationCall {
    principal: Option<EntityUid>,
    action: EntityUid,
    resource: Option<EntityUid>,
    context: HashMap<SmolStr, RestrictedExpression>,
    schema: Option<Schema>,
    validate_request: bool,
    entities: HashSet<Entity>,
    policies: PolicySet
}

struct EntityUid(serde_json::Value); // see <https://docs.cedarpolicy.com/policies/json-format.html>

struct RestrictedExpression(serde_json::Value); // see <https://docs.cedarpolicy.com/policies/json-format.html>

enum Schema {
    Human(String),
    Json(serde_json::Value), // see <https://docs.cedarpolicy.com/schema/json-schema.html>
}

struct Entity(serde_json::Value); // see <https://docs.cedarpolicy.com/policies/json-format.html>

struct PolicySet{
    static_policies: StaticPolicySet,
    templates: HashMap<SmolStr, Template>,
    template_links: HashSet<TemplateLink>,
}

enum StaticPolicySet {
    Concatenated(String),
    Set(HashSet<Policy>),
    Map(HashMap<SmolStr, Policy>),
}

enum Policy {
[...]

PR #760, opened 2024-03-29
new is_authorized signatures in FFI; support human schema format for request validation

cdisselkoen:
## Description of changes

- Starts applying the recommendations in #757, in particular, the new `is_authorized` signatures for FFI.  
- Enables the human schema format for request validation in FFI.  
- Adds tests for request validation in FFI (we didn't previously have any).
- Refactors `AuthorizationCall::get_components()` to do a better job collecting multiple errors before returning, rather than eagerly returning when the first error is encountered.

## Issue #, if available

#757 (but this PR implements only a small portion of #757's suggestions)


cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -324,88 +346,270 @@ fn constant_true() -> bool {
...
+        };
+        let (policies, entities) = match self.slice.try_into(schema.as_ref()) {
+            Ok((policies, entities)) => (Some(policies), Some(entities)),
+            Err(e) => {
+                errs.extend(e);
+                (None, None)
+            }
+        };
+        if errs.is_empty() {
+            // TODO: this ignores warnings
```
khieta, line 491: Do we want to adjust the APIs in #757 to allow returning non-fatal errors (i.e., warnings)?

cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -324,88 +346,270 @@ fn constant_true() -> bool {
...
+            let Some(q) = q else {
+                return Err(errs);
+            };
+            let Some(policies) = policies else {
+                return Err(errs);
+            };
+            let Some(entities) = entities else {
+                return Err(errs);
+            };
```
khieta, line 500: These cases shouldn't be possible with an empty `errs`, right?

khieta (commented):
Thanks for getting the ball rolling on #757 😄

You'll likely need to update the Java too -- you might want to try that before merging this PR to make sure it won't be too painful.

cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -303,13 +326,12 @@ struct AuthorizationCall {
         tsify(optional, type = "Record<string, CedarValueJson>")
     )]
     context: HashMap<String, JsonValueWithNoDuplicateKeys>,
-    /// Optional schema in JSON format.
+    /// Optional schema.
     /// If present, this will inform the parsing: for instance, it will allow
     /// `__entity` and `__extn` escapes to be implicit, and it will error if
     /// attributes have the wrong types (e.g., string instead of integer).
-    #[serde(rename = "schema")]
```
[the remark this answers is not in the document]
cdisselkoen: Correct. It was renaming from `schema` to `schema`

cedar-policy/src/frontend/is_authorized.rs
cdisselkoen: Correct.  But I thought returning `Err([])` in this impossible case was better than panicking

cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -324,88 +346,270 @@ fn constant_true() -> bool {
...
+            .map(|p| parse_entity_uid(p, "principal"))
+            .transpose()
+        {
+            Ok(None) => None,
+            Ok(Some(euid)) => Some(euid),
+            Err(e) => {
+                errs.push(e);
+                None
+            }
+        };
```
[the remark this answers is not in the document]
cdisselkoen: Factored some of the logic out; let me know what you think

cedar-policy/src/frontend/is_authorized.rs
cdisselkoen: We probably should

cdisselkoen:
I'm concerned that the downstream Java CI is not failing.  It should be, I think?

cdisselkoen:
I wonder if the downstream Java CI is somehow not pulling the correct version of `cedar-policy` when it runs its tests.  I.e., not pulling in the version of `cedar-policy` from this PR's branch.  I can't immediately tell from the CI log what version/hash of `cedar-policy` was used.

cedar-policy/src/frontend/is_authorized.rs
khieta: How about the following?
```rust
enum AuthorizationAnswer {
    Failure { errors: HashSet<serde_json::Value> },
    Success { 
        response: Response,
        warnings: HashSet<serde_json::Value>,
    },
}
```
The miette json format should support tagging an "error" with the severity "warning".

khieta (approved):

cedar-policy/src/frontend/is_authorized.rs
khieta: It still feels a little ugly to me 🙂 would prefer if you could refactor to remove, but it's not a blocking issue.

cedar-policy/src/frontend/is_authorized.rs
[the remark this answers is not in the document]
khieta: New structure looks better

cedar-policy/src/frontend/is_authorized.rs
cdisselkoen: Or, do we envision that `Response` (i.e., `Diagnostics`) will eventually evolve to include warnings?  Would it make more sense to put all warnings inside `Diagnostics`?

cedar-policy/src/frontend/is_authorized.rs
cdisselkoen: Refactored

cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -95,42 +91,42 @@ fn is_authorized_partial(call: AuthorizationCall) -> PartialAuthorizationAnswer
...
-        |call| match is_authorized_partial(call) {
-            answer @ PartialAuthorizationAnswer::Residuals { .. } => {
-                InterfaceResult::succeed(answer)
-            }
-            PartialAuthorizationAnswer::ParseFailed { errors } => {
-                InterfaceResult::fail_bad_request(errors)
-            }
-        },
-    )
+pub fn is_authorized_json_partial(
```
khieta, line 98: Nit
```suggestion
pub fn is_authorized_partial_json(
```

cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -95,42 +91,42 @@ fn is_authorized_partial(call: AuthorizationCall) -> PartialAuthorizationAnswer
...
+    let ans = is_authorized_partial(serde_json::from_value(json)?);
+    serde_json::to_value(ans)
 }
 
-/// Interface version of a `Response` that uses `InterfaceDiagnostics` for simpler (de)serialization
+/// Input and output are strings containing serialized JSON, in the shapes
+/// expected by [`is_authorized_json_partial()`]
+#[doc = include_str!("../../experimental_warning.md")]
+#[cfg(feature = "partial-eval")]
+pub fn is_authorized_json_str_partial(json: &str) -> Result<String, serde_json::Error> {
```
khieta, line 109:
```suggestion
pub fn is_authorized_partial_json_str(json: &str) -> Result<String, serde_json::Error> {
```

cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -305,27 +303,35 @@ impl TryFrom<PartialResponse> for InterfaceResidualResponse {
 #[cfg_attr(feature = "wasm", derive(tsify::Tsify))]
 #[cfg_attr(feature = "wasm", tsify(into_wasm_abi, from_wasm_abi))]
 pub enum AuthorizationAnswer {
-    /// Represents a failure to parse and call the authorizer
-    Failure {
-        /// Parsing errors
+    /// Represents a failure to parse or call the authorizer entirely
+    ParseFailed {
```
khieta, line 307: I prefer the more general name `Failure`. I'm imagining that we'll want to use these types for a general DRT interface, and other implementations of Cedar (e.g., the Lean) may return errors for more than just JSON parsing issues.

khieta (approved):

cedar-policy/src/frontend/is_authorized.rs
cdisselkoen: That's fair.  I changed the name to `ParseFailed` as part of the merge from `main` (a recent commit to `main` changed the name to `ParseFailed` -- see https://github.com/cedar-policy/cedar/blob/38754319c7cd3b50083f44c4098cfb10f27d540c/cedar-policy/src/frontend/is_authorized.rs#L324), but I am happy to change back to `Failure` for the reasons you give

cedar-policy/src/frontend/is_authorized.rs
khieta: Good question - I think that would also be reasonable.

cedar-policy/src/frontend/is_authorized.rs
cdisselkoen: Changed

cdisselkoen:
The Java (and DRT) downstream failures are indeed the result of this PR.  I will prep fixes to those before merging this.

cedar-policy/src/frontend/is_authorized.rs
cdisselkoen: Tried the version with warnings in `Diagnostics` and updated this PR; let me know what you think.

Also accounted for the possibility that we have both errors and warnings, by adding a `warnings` field to `Failure` here as well.

cdisselkoen: Having second thoughts.  In the fullness of time we probably want to support warnings generated by evaluating a particular policy, and also warnings generated when parsing/processing things like schemas and entity data.  The former are associated with a particular policy id and make sense to be in `Diagnostics`.  The latter are not and do not.  Perhaps we want a space for warnings both in `Diagnostics` (for warnings associated with evaluating a particular policy) and in `Failure` and `Success` generally.

cdisselkoen: Adjusted again in 684e8f65 on this PR

cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -656,9 +825,47 @@ impl RecvdSlice {
 #[cfg(test)]
 mod test {
     use super::*;
-    use crate::{frontend::utils::assert_is_failure, EntityUid};
     use cool_asserts::assert_matches;
-    use std::collections::HashMap;
+    use serde_json::json;
+
+    /// Assert that `is_authorized_json()` returns Allow with no errors
+    #[track_caller] // report the caller's location as the location of the panic, not the location in this function
```
andrewmwells-amazon, line 832: Is it considered a best practice to have this comment every time? It feels redundant to me.

cedar-policy/src/frontend/is_authorized.rs
```diff
@@ -370,51 +398,121 @@ fn constant_true() -> bool {
...
-fn parse_schema(
-    schema_json: Option<JsonValueWithNoDuplicateKeys>,
-) -> Result<Option<Schema>, Vec<String>> {
-    schema_json
-        .map(|v| Schema::from_json_value(v.into()))
-        .transpose()
-        .map_err(|e| vec![e.to_string()])
+#[derive(Debug, Serialize, Deserialize)]
+enum Schema {
+    /// Schema in Cedar's default, human-readable schema format. See <https://docs.cedarpolicy.com/schema/human-readable-schema.html>
```
andrewmwells-amazon, line 403: I thought we wanted to just call this `Cedar schema format` going forwards?

andrewmwells-amazon (approved):

cedar-policy/src/frontend/is_authorized.rs
cdisselkoen: Can remove :)

cedar-policy/src/frontend/is_authorized.rs
cdisselkoen: Will change.

andrewmwells-amazon (approved):

Issue #154, opened 2023-07-05
References between `commonTypes` 


john-h-kastner-aws:
### Category

Cedar validation features

### Describe the feature you'd like to request

The Cedar schema supports `commonTypes` definitions to help avoid duplicating attribute types in entity types shapes and actions contexts, but `commonTypes` _cannot_ currently be used inside other common types definitions.

The Cedar schema should support `commonTypes` defined in terms of other `commonTypes`. It should however, not support recursive types.

### Describe the solution you'd like

The schema parser will construct a directed graph of the type definitions, with an edge from each type to the other types it depends on. It will then check that the graph is acyclic. This can be done using our existing functions for checking that the entity hierarchy is a DAG. We can safely inline common type definitions once we know the dependency graph does not have cycles.

### Describe alternatives you've considered

If we depend on the order of `commonTypes` being significant, we could require that a type is declared before it is referenced, making cycles impossible to create.

The order of entries in a JSON map is not generally significant, so we would either need to break from this convention by making the order significant, or use an association list of pairs instead. Depending on the order of the map is not acceptable because it is likely to cause issues if the a schema document is processed by standard JSON tools which may not preserve order. Using an association list would make common types use a different structure than the entity types and actions list.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

john-h-kastner-aws:
Closed by #766

PR #766, opened 2024-03-29
Allow common type references in common type declarations

shaobo-he-aws:
## Issue #, if available
#154 


shaobo-he-aws:
I still need to add more tests but the basic idea is there. So I think it makes sense to start the review process.

cedar-policy-validator/src/err.rs
```diff
@@ -90,6 +90,9 @@ pub enum SchemaError {
     /// Cycle in the schema's action hierarchy.
     #[error("cycle in action hierarchy containing `{0}`")]
     CycleInActionHierarchy(EntityUID),
+    /// Cycle in the schema's action hierarchy.
+    #[error("cycle in common type references containing `{0}`")]
+    CycleInCommonTypeReferences(Name),
```
cdisselkoen, line 95: doc comment needs to be updated

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -52,6 +52,16 @@ pub struct SchemaFragment(
     pub  HashMap<Option<Name>, NamespaceDefinition>,
 );
 
+impl SchemaFragment {
+    /// Resolve common types in the schema
+    pub(crate) fn resolve_common_types(self) -> Result<Self> {
+        let mut schema = self.clone();
```
cdisselkoen, line 58: Is `.clone()` necessary?  Cloning the whole schema seems like something that could eventually be a performance problem

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -133,6 +143,193 @@ impl SchemaFragment {
...
+        Self { schema, graph }
+    }
+
+    /// Perform topological sort on the dependency graph
+    /// Let A -> B denote the RHS of type `A` refers to type `B` (i.e., `A`
+    /// depends on `B`)
+    /// topo_sort(A -> B -> C) produces [C, B, A]
+    /// If there is a cycle, a type name involving in this cycle is the error
+    /// It implements a variant of Kahn's algorithm
+    fn topo_sort(&self) -> std::result::Result<Vec<Name>, Name> {
```
cdisselkoen, line 175: This is a nontrivial graph algorithm, thoughts on using a dependency like `petgraph` (76 million downloads on crates) instead of hand-rolling?  I'm mostly just worried that if there are subtle bugs in this algorithm, code review / tests / fuzzing might not catch them.

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -133,6 +143,193 @@ impl SchemaFragment {
...
+        resolve_table: &HashMap<&Name, SchemaType>,
+        ty: SchemaType,
+        ns: &Option<Name>,
+    ) -> Result<SchemaType> {
+        match ty {
+            SchemaType::TypeDef { type_name } => {
+                let name = if type_name.is_unqualified() {
+                    type_name.prefix_namespace_if_unqualified(ns.clone())
+                } else {
+                    type_name.clone()
```
cdisselkoen, line 268: I think we don't need the `.clone()` here?  `type_name` should already be owned

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -133,6 +143,193 @@ impl SchemaFragment {
...
+            .topo_sort()
+            .map_err(|name| SchemaError::CycleInCommonTypeReferences(name))?;
+
+        let mut resolve_table = HashMap::new();
+
+        for name in sorted_names.iter() {
+            let ns: Option<Name> = if name.is_unqualified() {
+                None
+            } else {
+                Some(name.namespace().parse().unwrap())
```
cdisselkoen, line 318: I think clippy will yell at you for this `unwrap()` without a `PANIC SAFETY` comment

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -133,6 +143,193 @@ impl SchemaFragment {
...
+            };
+
+            let common_types = &mut self.schema.0.get_mut(&ns).unwrap().common_types;
```
cdisselkoen, line 321: Another `unwrap()` with no `PANIC SAFETY` comment

cdisselkoen (changes requested):

cedar-policy-validator/src/schema_file_format.rs
shaobo-he-aws: I tried to use this crate and eventually gave up on it. The reason is that it requires the node type of its hash graphs to implement `Copy`, which is kind of hard for Cedar's `Name`.

shaobo-he-aws: I also argue this algorithm is fairly standard. @andrewmwells-amazon what do you think?

cdisselkoen: Ok, I'm fine with this.  We did hand-roll transitive closure rather than using an existing graph library, so there's precedent.

john-h-kastner-aws:
This adds separate logic to resolve common type reference inside other common type declarations, but there's still the original code to handle common type reference in entity/context types only still exists. Can we combine these so that types are resolved everywhere in one pass?

cedar-policy-validator/src/schema_file_format.rs
shaobo-he-aws: Done.

cedar-policy-validator/src/schema_file_format.rs
shaobo-he-aws: Done.

shaobo-he-aws:
> This adds separate logic to resolve common type reference inside other common type declarations, but there's still the original code to handle common type reference in entity/context types only still exists. Can we combine these so that types are resolved everywhere in one pass?

I thought about it. This option implies an overhaul of type conversions (i.e., `SchemaFragment` to `ValidatorSchemaFragment` to `ValidatorSchema`). Any ideas about what a proper implementation is?

shaobo-he-aws:
I think a better design that keeps the `SchemaFragment` to `ValidatorSchemaFragment` to `ValidatorSchema` flow is to change `type_defs` of `ValidatorNamespaceDef` to `HashMap<Name, SchemaType>`, where type references in the values are fully qualified. Then, we do the topological sort and build the `Name` to `Type` bindings during `ValidatorSchemaFragment` to `ValidatorSchema` translation.

cedar-policy-validator/src/schema/namespace_def.rs
```diff
@@ -139,6 +140,192 @@ pub struct ActionFragment {
     pub(super) attributes: HashMap<SmolStr, PartialValueSerializedAsExpr>,
 }
 
+/// A common type reference resolver
+#[derive(Debug)]
+pub(crate) struct CommonTypeResolver<'a> {
```
john-h-kastner-aws, line 145: This struct would make more sense in the main `schema.rs` file or in its own submodule IMO

cedar-policy-validator/src/schema/namespace_def.rs
```diff
@@ -139,6 +140,192 @@ pub struct ActionFragment {
...
+                        o.insert(o.get() + 1);
+                    }
+                    std::collections::hash_map::Entry::Vacant(v) => {
+                        v.insert(1);
+                    }
+                }
+            }
+        }
+
+        // The set that contains type names with zero coming edges
```
john-h-kastner-aws, line 192:
```suggestion
        // The set that contains type names with zero incoming edges
```

john-h-kastner-aws (approved):
This design looks good. The breaking change in the error type is unfortunate if this is still targeting 3.2. Not sure what a good workaround would be though.

cedar-policy/src/api.rs
```diff
@@ -1437,6 +1437,9 @@ pub enum SchemaError {
     /// Cycle in the schema's action hierarchy.
     #[error("cycle in action hierarchy containing `{0}`")]
     CycleInActionHierarchy(EntityUid),
+    /// Cycle in the schema's common type references
+    #[error("cycle in common type references containing `{0}`")]
+    CycleInCommonTypeReferences(String),
```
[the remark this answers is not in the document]
cdisselkoen: The situations we would throw this now, were already errors in 3.1 -- what error variant would we get in 3.1?  Can/should we shoehorn this error into the old variant for 3.x, and make this new variant for 4.0?  Maybe even change the error message for the old variant in 3.x in this situation to be smart?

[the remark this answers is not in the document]
john-h-kastner-aws: I would be `UndeclaredCommonTypes` in 3.1. The easy work around I guess would be to make that message `undeclared common type or cycle in common type references`, but I don't see what we could do to intelligently specialize it without a breaking change to the contents of the enum variant.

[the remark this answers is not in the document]
john-h-kastner-aws: We could also make it an `UnsupportedFeature` or we could abuse `serde_json::error::Error::custom` since those both take free-form strings.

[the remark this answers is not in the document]
shaobo-he-aws: I prefer `serde_json::error::Error::custom` if we cherry pick it into 3.2.

cedar-policy-validator/src/schema/namespace_def.rs
shaobo-he-aws: Good idea. Done.

cedar-policy/src/api.rs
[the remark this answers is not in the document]
john-h-kastner-aws: That would be acceptable IMO

cedar-policy-validator/src/schema.rs
```diff
@@ -646,6 +649,192 @@ impl TryInto<ValidatorSchema> for NamespaceDefinitionWithActionAttributes {
     }
 }
 
+/// A common type reference resolver
+#[derive(Debug)]
+struct CommonTypeResolver<'a> {
+    /// Common type declarations to resolve
+    type_defs: &'a HashMap<Name, SchemaType>,
+    /// The dependency graph among common type names
```
andrewmwells-amazon, line 657:
```suggestion
    /// The dependency graph among common type names
    /// The graph contains a vertex for each `Name` and `graph.get(u)` gives the set of vertices `v` for which `(u,v)` is a directed edge in the graph
```

andrewmwells-amazon (approved):

andrewmwells-amazon (approved):

shaobo-he-aws:
SemVer CI failure is expected because it introduces a breaking change.

PR #749, opened 2024-03-27
Strengthen the types of `SchemaFragment`

shaobo-he-aws:
## Description of changes

Fields like `NamespaceDefinition::common_types` should have types like `HashMap<Id, SchemaType>` instead of `HashMap<SmolStr, SchemaType>`. So, this PR checks validity of these fields during JSON deserialization instead of post-processing `SchemaFragment`. Moving these checks ahead makes it easier to process `SchemaFrament` without incurring breaking changes. Furthermore, translating human-readable schema to JSON schema avoids redundant validity checks because the parser of the former already ensures such validity.

It's a breaking change because error variants `Parse*` were removed from `SchemaError`. They become custom errors thrown by serde. An example is as follows,

```
 × failed to parse schema from file schema.cedarschema.json
  ├─▶ failed to parse schema: invalid memberOfTypes: unexpected token `!` at line 8 column 13
  ╰─▶ invalid memberOfTypes: unexpected token `!` at line 8 column 13
```
v.s
```
 × failed to parse schema from file schema.cedarschema.json
  ╰─▶ parse error in entity type: unexpected token `!`
```

## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [x] A breaking change requiring a major version bump to `cedar-policy` (e.g., changes to the signature of an existing API).

I confirm that this PR (choose one, and delete the other options):

- [x] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [x] Requires updates, and I have made / will make these updates myself. (Please include in your description a timeline or link to the relevant PR in `cedar-spec`, and how you have tested that your updates are correct.)

Issue #488, opened 2023-12-11
Reuse CI workflows to build downstream dependencies


john-h-kastner-aws:
### Category

Internal refactors/changes

### Describe the feature you'd like to request

The [CI workflow used by this repository](https://github.com/cedar-policy/cedar/blob/main/.github/workflows/build_downstream_deps.yml) to build the downstream dependencies in cedar-spec, cedar-examples, and cedar-java duplicates the workflows found in those individual repositories. We should [reuse these workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows) so that CI doesn't break when a change made in a downstream repository isn't copied here.

Relevant workflows:

* cedar-examples: https://github.com/cedar-policy/cedar-examples/blob/main/.github/workflows/ci.yml
* cedar-java: https://github.com/cedar-policy/cedar-java/blob/main/.github/workflows/ci.yml
* cedar-spec: https://github.com/cedar-policy/cedar-spec/blob/main/.github/workflows/ci.yml

### Describe alternatives you've considered

.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

andrewmwells-amazon:
- [ ] tinytodo
- [ ] cedar-rust-hello-world
- [ ] cedar-example-use-cases
- [ ] cedar-java
- [ ] cedar-spec

john-h-kastner-aws:
The current reusable downstream builds turn out not to work for PRs coming from forks. See this failed workflow: https://github.com/cedar-policy/cedar/actions/runs/7451426259/job/20272524590

john-h-kastner-aws:
Fork build failure was fixed by https://github.com/cedar-policy/cedar-examples/pull/90

andrewmwells-amazon:
Java CI successfully fails for broken [PR](https://github.com/cedar-policy/cedar/actions/runs/8529036529/job/23363951815?pr=775) and [Fork](https://github.com/cedar-policy/cedar/actions/runs/8529220873/job/23364530934?pr=770).

khieta:
Reusable CI added for `cedar-spec` (the last hold out) in [cedar-spec#269](https://github.com/cedar-policy/cedar-spec/pull/269). Marking this issue as resolved.

PR #771, opened 2024-04-01
Fix schema serialization

khieta:
## Description of changes

Followup to #749 to fix json schema serialization.

* Added custom (de)serialization for `Id` and `Name`. This will affect our internal formats (e.g., used by DRT), but should not affect external formats, aside from error messages.
* This allowed me to remove some of the boilerplate in `schema_file_format.rs`

- [X] Requires updates, and I have made / will make these updates myself. (Please include in your description a timeline or link to the relevant PR in cedar-spec, and how you have tested that your updates are correct.)

### DIFF
--- a/cedar-policy-core/src/ast/id.rs
+++ b/cedar-policy-core/src/ast/id.rs
@@ -1,4 +1,4 @@
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 use smol_str::SmolStr;
 
 use crate::{parser::err::ParseErrors, FromNormalizedStr};
@@ -9,7 +9,7 @@ use crate::{parser::err::ParseErrors, FromNormalizedStr};
 /// `true | false | if | then | else | in | is | like | has`).
 //
 // For now, internally, `Id`s are just owned `SmolString`s.
-#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone, Hash, PartialOrd, Ord)]
+#[derive(Serialize, Debug, PartialEq, Eq, Clone, Hash, PartialOrd, Ord)]
 pub struct Id(SmolStr);
 
 impl Id {
@@ -60,6 +60,34 @@ impl std::str::FromStr for Id {
 impl FromNormalizedStr for Id {
     fn describe_self() -> &'static str {
         "Id"
+    }
+}
+
+struct IdVisitor;
+
+impl<'de> serde::de::Visitor<'de> for IdVisitor {
+    type Value = Id;
+
+    fn expecting(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        formatter.write_str("a valid id")
+    }
+
+    fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        Id::from_normalized_str(value)
+            .map_err(|err| serde::de::Error::custom(format!("invalid id `{value}`: {err}")))
+    }
+}
+
+/// Deserialize an `Id` using `from_normalized_str`
+impl<'de> Deserialize<'de> for Id {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        deserializer.deserialize_str(IdVisitor)
     }
 }
 
--- a/cedar-policy-core/src/ast/id.rs
+++ b/cedar-policy-core/src/ast/id.rs
@@ -81,7 +81,8 @@ impl<'de> serde::de::Visitor<'de> for IdVisitor {
     }
 }
 
-/// Deserialize an `Id` using `from_normalized_str`
+/// Deserialize an `Id` using `from_normalized_str`.
+/// This deserialization implementation is used in the JSON schema format.
 impl<'de> Deserialize<'de> for Id {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
--- a/cedar-policy-core/src/ast/name.rs
+++ b/cedar-policy-core/src/ast/name.rs
@@ -184,7 +184,7 @@ impl<'de> serde::de::Visitor<'de> for NameVisitor {
     type Value = Name;
 
     fn expecting(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        formatter.write_str("a name consisting of a path and id, separated by `::`")
+        formatter.write_str("a name consisting of an optional namespace and id")
     }
 
     fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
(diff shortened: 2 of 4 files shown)


cedar-policy-core/src/ast/name.rs
```diff
@@ -167,6 +178,34 @@ impl FromNormalizedStr for Name {
     }
 }
 
+struct NameVisitor;
+
+impl<'de> serde::de::Visitor<'de> for NameVisitor {
+    type Value = Name;
+
+    fn expecting(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        formatter.write_str("a name consisting of a path and id, separated by `::`")
```
cdisselkoen, line 187: maybe this comment could make it clear that the path (and thus the `::`) is optional?  Not sure if that's easy to concisely convey though

cdisselkoen (approved):
Nice.  Shares a bunch of Deserialize logic that wasn't previously shared.  Also the error messages are improved

khieta:
Update: ran the `json-schema-roundtrip` target locally for a couple hours and all appears to be good.

cedar-policy-core/src/ast/name.rs
khieta: Updated to `"a name consisting of an optional namespace and id"`. Couldn't find a concise way to work in the `::` but open to suggestions.

cedar-policy-core/src/ast/id.rs
```diff
@@ -63,6 +63,34 @@ impl FromNormalizedStr for Id {
     }
 }
 
+struct IdVisitor;
+
+impl<'de> serde::de::Visitor<'de> for IdVisitor {
```
john-h-kastner-aws, line 68: Add a comment that this impl is part of our public API now. Also for the other new impls used by json schema parsing.

As a side note, It'd be good if we could document more thoroughly how each `Serialize` and `Deserialize` implementation gets used. I've found my self wondering a few times if changing the internals of a struct would break part of our API, just DRT, or maybe nothing at all.

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -1340,31 +1277,31 @@ mod strengthened_types {
...
+        assert_matches!(schema, Err(err) if &err.to_string() == "invalid name `Action::`: unexpected end of input");
 
         let src = serde_json::json!(
         {
            "id": "action",
             "type": "::Action"
         });
         let schema: Result<ActionEntityUID, _> = serde_json::from_value(src);
-        assert_matches!(schema, Err(err) if &err.to_string() == "invalid action type: unexpected token `::`");
+        assert_matches!(schema, Err(err) if &err.to_string() == "invalid name `::Action`: unexpected token `::`");
```
john-h-kastner-aws, line 1304: Would it be easy to make this `invalid action name ...`?

john-h-kastner-aws (approved):
Looks good. Just would like some comments on deser impls

cedar-policy-validator/src/schema_file_format.rs
cdisselkoen: while we're here, this assert might be more pretty/readable written without the guard:
```
assert_matches!(schema, Err(err) => {
    assert_eq!(err.to_string, "invalid action name: ...");
});
```

khieta:
Another update: this PR breaks the Lean JSON parsing. Looks like CI didn't fail because we're not running the DRT implementation on the integration tests. I'll work on finishing off #488 before merging this PR.

cedar-policy-validator/src/schema_file_format.rs
khieta: > Would it be easy to make this invalid action name ...?

I think we'd need to add back the manual deserialization function for `ActionEntityUID` to get a customized error message here. I don't think it's worth it, but open to other opinions.

cedar-policy-core/src/ast/id.rs
khieta: I don't _think_ that any of the changes in this PR affect our public APIs.
* The `Id` and `Name` types are not publicly exposed, so changes to them should be safe.
* The only type with a `Deserialize` implementation in `cedar-policy` is `PolicyId`, which desugars to `SmolStr` (although possibly we should change that to be `Id`...).
* Most things in `cedar-policy/frontend` are represented directly as strings/JSON. The exceptions are `Decision`, `ValidationEnabled`, and `SchemaFragment`. The first two are not affected by this PR. The last is, but this PR fixes the deserialization to match what it was prior to #749

khieta: Good point though on better documentation. Once we update DRT to use only public interfaces, we should be able to significantly cut down on the types that support (de)serialization.

john-h-kastner-aws: My point is that this is `Deserialize` impl is now used to define the schema format (part of our api), but until now this `Deserialize` derive here only affected the DRT interface. It'd be easy to make a change here without thinking that it could break the json schema format

cdisselkoen: > although possibly we should change [PolicyId] to be Id

Today, we allow PolicyId's that are any string, not limited to Ids

cedar-policy-validator/src/schema_file_format.rs
john-h-kastner-aws: Seems alright as it is

khieta:
Ready for re-review.
* Added comments in `id.rs` and `name.rs` that de/serialization is used for the JSON schema format
* Updated to use the new downstream CI for `cedar-spec` (which now fails, as expected)
* Cleaned up some of the testing code

I'll now start working on the update to `cedar-spec` for this PR.

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -1406,38 +1351,174 @@ mod strengthened_types {
...
+        );
+    }
+}
+
+/// Check that (de)serialization works as expected.
+#[cfg(test)]
+mod test_json_roundtrip {
+    use super::*;
+
+    fn roundtrip(schema: SchemaFragment) {
```
john-h-kastner-aws, line 1404:
```suggestion
    #[track_caller]
    fn roundtrip(schema: SchemaFragment) {
```

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -1111,6 +1047,11 @@ mod strengthened_types {
         ActionEntityUID, ApplySpec, EntityType, NamespaceDefinition, SchemaFragment, SchemaType,
     };
 
+    /// Assert that `result` is an `Err`, and the error message matches `msg`
+    fn assert_error_matches<T: std::fmt::Debug>(result: Result<T, serde_json::Error>, msg: &str) {
```
john-h-kastner-aws, line 1051:
```suggestion
    #[track_caller]
    fn assert_error_matches<T: std::fmt::Debug>(result: Result<T, serde_json::Error>, msg: &str) {
```

john-h-kastner-aws (approved):

PR #790, opened 2024-04-16
chore: update copyright headers

khieta:
## Description of changes

Long overdue followup to #731. Also added the header to some files we missed.


cdisselkoen (approved):
ope, well, this is better and more thorough than my #791, which I'm just going to close now

shaobo-he-aws (approved):

PR #793, opened 2024-04-16
760's changes but for validation APIs

cdisselkoen:
## Description of changes

The same changes that #760 made for authorization APIs, but here for validation APIs.  In particular:
* New top-level signatures
* Enables the human schema format (!)
* More propagation of warnings

## Issue #, if available

#757 (but this is only a small portion of the work for #757)

(or rather, the changelog entry for #760 is general enough to cover it)

### DIFF
--- a/cedar-policy/src/frontend/utils.rs
+++ b/cedar-policy/src/frontend/utils.rs
@@ -15,7 +15,8 @@
  */
 
 //! Utility functions and types for JSON interface
-use crate::{Policy, PolicySet, Template};
+use crate::{Policy, SchemaWarning, Template};
+use cedar_policy_core::jsonvalue::JsonValueWithNoDuplicateKeys;
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, str::FromStr};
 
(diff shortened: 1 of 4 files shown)


cedar-policy/src/frontend/utils.rs
```diff
@@ -102,6 +103,47 @@ impl PolicySpecification {
...
+    Human(String),
+    /// Schema in Cedar's JSON schema format. See <https://docs.cedarpolicy.com/schema/json-schema.html>
+    #[serde(rename = "json")]
+    Json(JsonValueWithNoDuplicateKeys),
+}
+
+impl Schema {
+    pub(super) fn parse(
+        self,
+    ) -> Result<(crate::Schema, Box<dyn Iterator<Item = SchemaWarning>>), String> {
```
khieta, line 120: Nit: it would feel more consistent to have both `PolicySet::parse` and `Schema::parse`, or both `PolicySet::try_into` and `Schema::try_into`

khieta (approved):
Lgtm. Please keep #757 up-to-date with any changes you make so we can use that as a starting point for future documentation.

cedar-policy/src/frontend/utils.rs
cdisselkoen: Renamed to parse()

john-h-kastner-aws (approved):

PR #795, opened 2024-04-16
Implement -w/--write flag for cedar format

junlarsen:
## Description of changes

Adds a -w/--write flag to `cedar format` for automatically writing the formatted policy back into file, as described in #789 

## Issue #, if available

Fixes #789 

## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [x] A backwards-compatible change requiring a minor version bump to `cedar-policy` (e.g., addition of a new API).

I confirm that this PR (choose one, and delete the other options):

- [x] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [x] Does not require updates because my change does not impact the Cedar formal model or DRT infrastructure.

PR #798, opened 2024-04-17
Add -c/--check to cedar format for validating policy

junlarsen:
## Description of changes

Implements a new -c/--check flag to `cedar format` that will exit with exit code 1 if the formatted version of the policy did not match the original.

I've made the --write and --check options mutually exclusive, since you're probably not interested in writing the file at the same time as checking it. Let me know if you disagree and would like that changed.

## Issue #, if available

Fixes #796 

## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [x] A backwards-compatible change requiring a minor version bump to `cedar-policy` (e.g., addition of a new API).

I confirm that this PR (choose one, and delete the other options):

- [x] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [x] Does not require updates because my change does not impact the Cedar formal model or DRT infrastructure.

PR #799, opened 2024-04-17
Require --policies to be present when using --write

junlarsen:
## Description of changes

Adds a clap requirement for `--policies` to be present when using `--write`. Previously this would print the formatted policy to stdout (practically ignoring `--write`)

## Issue #, if available

N/A, I spotted this really quickly and figured it's a really small fix.

This only affects the `cedar format -w` which was introduced yesterday in #795, which is yet to be released.

### DIFF
--- a/cedar-policy-cli/src/lib.rs
+++ b/cedar-policy-cli/src/lib.rs
@@ -401,7 +401,7 @@ pub struct FormatArgs {
     pub indent_width: isize,
 
     /// Automatically write back the formatted policies to the input file.
-    #[arg(short, long, group = "action")]
+    #[arg(short, long, group = "action", requires = "policies_file")]
     pub write: bool,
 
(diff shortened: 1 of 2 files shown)


cedar-policy-cli/Cargo.toml
```diff
@@ -28,6 +28,7 @@ partial-validate = ["cedar-policy/partial-validate"]
 assert_cmd = "2.0"
 tempfile = "3"
 glob = "0.3.1"
+predicates = "3.1.0"
```
junlarsen, line 31: See https://github.com/cedar-policy/cedar/pull/798#discussion_r1569603483 for rationale

john-h-kastner-aws (approved):
Makes sense to me

shaobo-he-aws (approved):

cdisselkoen (approved):
Just needs a merge from `main`

john-h-kastner-aws (approved):

PR #800, opened 2024-04-19
Propagate miette error info through FFI

cdisselkoen:
## Description of changes

FFI returns rich JSON structures for errors, rather than a simple string.

This diverges from the plan proposed in #757 in two ways: 
1. This PR actually specifies the miette JSON structure produced, rather than just using `serde_json::Value` in our interface types.  This makes it easier to integrate with the FFI (and write tests), and makes us responsible for updates if the `miette` JSON structure ever changes, but as an upside the code will complain louder and earlier if the `miette` JSON structure does change.
2. `ValidationWarning` and `ValidationError` have the same structure, so this PR proposes just having a single struct `ValidationError`.  Errors and warnings can still be distinguished from each other in two ways: (1) via `miette`'s `severity` field; and (2) because `ValidationAnswer::Success` keeps them in separate lists `validation_errors` and `validation_warnings`.

Somewhat orthogonally, this PR also includes a pass through all of the `cedar` repo removing occurrences to the outdated terminology "instantiation" in favor of "linking".  This can be separated into a different PR if that's preferable.

## Issue #, if available

Part of #757 


cdisselkoen:
for review:

1. the `REVIEW` comment in `cedar-wasm/src/validator.rs`
2. The code currently has two possible panics in `cedar-policy/src/ffi/utils.rs`.  Do we want to: (1) justify these with PANIC_SAFETY; (2) bite the bullet and make this a `TryFrom` instead of `From`, with some error type indicating it's a purely internal error; or (3) re-architect to not use `miette`'s JSON at all and just write our own serialization from `miette::Report` to a very similar JSON structure, that we can (a) control better and (b) know won't panic

khieta:
When I last looked at the miette JSON format, it didn't seem like it was particularly well-documented (and thus maybe not widely used), so I'd lean towards a custom JSON format. We would presumably have a function that converts miette's format to ours, so that would be an indicator when something breaks.

cdisselkoen:
Yes, it doesn't seem to be well-documented, and in fact I just looked at some outputs experimentally when crafting the structure in this PR.

cedar-wasm/src/validator.rs
```diff
@@ -14,41 +14,61 @@
...
 use serde::{Deserialize, Serialize};
 use tsify::Tsify;
 use wasm_bindgen::prelude::*;
 
+/// `ffi::ValidationAnswer`, but with `serde(rename_all = "camelCase")`
+///
+/// REVIEW: do we want to just camelCase the fields in the normal
+/// `ValidationAnswer`, and adapt Java and others? It's confusing for
+/// Rust<->Wasm to use camelCased JSON field names but Rust<->Java to use
+/// snake_cased JSON field names
```
khieta, line 27: No strong feelings. Would it be weird if we set up serde `alias`es to support both snake and camel case (and strange combinations thereof)?

khieta: Various opinions on stack overflow: https://stackoverflow.com/questions/5543490/json-naming-convention-snake-case-camelcase-or-pascalcase

cdisselkoen: Serde `alias` is a good idea, but only helps for deserialization.  Serialization still has to pick one.

cdisselkoen:
Pushed another rev of this PR, now uses our own custom error JSON format and doesn't touch miette's JSON handler at all.  Gives us maximum control over the JSON shape.  Allows us to have `message` include all the `source`s (which seems to be the pre-existing behavior).

cedar-wasm/src/validator.rs
john-h-kastner-aws: aliasing does feel a little weird IMO. I prefer picking one convention for everything

cdisselkoen: TS folks seem to prefer camelCase in #805

Swolebrain: Yeah camelCase is definitely preferred in ts/js

cdisselkoen: #814 making the global change to camelCase in JSON throughout the `ffi` module

cdisselkoen:
Updated this PR given #814.  Made the cedar-wasm changes a bunch simpler.

cdisselkoen:
Updated this PR to account for the changes made in #824.  This PR does propose breaking changes to some of the choices made in #824:

* Errors can have multiple source locations, at least in the fullness of time.  (Cedar doesn't use this much currently, but `miette` supports it, and it seems like a good future-proofing measure to support it in the FFI now.)  This PR proposes to rename #824's `sourceLocation` to `sourceLocations` and make it a JSON array.
* Because there are many fields (up to 7) associated with each error now, this PR proposes to rename the main error message from `error` to `message`.
* This PR also introduces one more level of nesting for errors compared to #824 (which added `sourceLocation` as a peer of `error`): validation errors are now `{ policyId: ..., error: { message: ..., sourceLocations: [...], ...}`.  This could be flattened, but I thought it was valuable to have a consistent structure for errors which could be mirrored on the TS side, and the contents of `error` in this struct (`DetailedError` in the Rust in this PR) are used in many other places now, and there is not always a `policyId` associated with all of these errors (this struct is used for other kinds of errors as well, not just errors on policies).  I'm open to removing the nesting if yall think the flatter structure is more important than a consistent `DetailedError` type.
* `type: "error"` renamed to `type: "failure"`.  I think this is more clear, since the `"success"` case can still contain policy evaluation errors.  The top-level `Failure` is only for failures to parse or call the authorizer entirely.

chrnorm:
Chiming in with some of the work I've been doing on our [GitHub Actions integration](https://github.com/common-fate/cedar-validate-action) - the error structure proposed here (`{ policyId: ..., error: { message: ..., sourceLocations: [...], ...}`) would be very easy to work with and I am in favor of it.

cedar-policy/src/ffi/mod.rs
```diff
@@ -14,9 +14,11 @@
  * limitations under the License.
  */
 
+#![cfg_attr(feature = "wasm", allow(non_snake_case))]
```
khieta, line 17: Out of curiosity: what is this for? I thought wasm preferred camelCase to snake_case?

cedar-policy/src/ffi/utils.rs
```diff
@@ -17,12 +17,134 @@
...
+        }
+    }
+}
+
+impl<'a, E: miette::Diagnostic + ?Sized> From<&'a E> for DetailedError {
+    fn from(diag: &'a E) -> Self {
+        Self {
+            message: {
+                let mut s = diag.to_string();
+                let mut source = diag.source();
```
khieta, line 117: What is `.source()`? Not related to source locations?

khieta (approved):
Looking good! The only comments I left were clarifying questions for my own understanding.

Also, it's a little late in the process for me to mention this, but I'd prefer if the "instantiate" -> "link" changes were in a separate PR 🙂

john-h-kastner-aws (approved):
Looks good.

cedar-policy/src/ffi/mod.rs
cdisselkoen: Wasm does prefer camelCase.  Including, some of tsify's macros expand to functions with camelCase names, which clippy doesn't like.  This `allow` avoids warnings for these autogenerated camelCased functions.

cedar-policy/src/ffi/utils.rs
cdisselkoen: It's a method on the [`Error` trait in std](https://doc.rust-lang.org/std/error/trait.Error.html).  I'm not sure why, but I needed to add these `:` and `source()` in order to preserve previous behavior (where this is the string message that was previously passed in the FFI).

cdisselkoen:
Split the "instantiate" -> "link" changes into #826

cdisselkoen:
Now that #826 has been merged, I've rebased this PR on `main` and it is ready for re-review

khieta (approved):

PR #804, opened 2024-04-22
Implement `Diagnostic` for internal `cedar-policy-validator` types

john-h-kastner-aws:
## Issue #, if available

This makes it possible to write tests for these errors in the same style as our parser errors.

### DIFF
--- a/cedar-policy-validator/src/validation_result.rs
+++ b/cedar-policy-validator/src/validation_result.rs
@@ -76,9 +76,8 @@ impl ValidationResult {
 /// policy. The error contains a enumeration that specifies the kind of problem,
 /// and provides details specific to that kind of problem. The error also records
 /// where the problem was encountered.
-#[derive(Clone, Debug, Error)]
+#[derive(Clone, Debug, Error, Eq, PartialEq)]
 #[error("{error_kind}")]
-#[cfg_attr(test, derive(Eq, PartialEq))]
 pub struct ValidationError {
     location: SourceLocation,
     error_kind: ValidationErrorKind,
@@ -176,8 +175,7 @@ impl SourceLocation {
 
 /// Enumeration of the possible diagnostic error that could be found by the
 /// verification steps.
-#[derive(Debug, Clone, Diagnostic, Error)]
-#[cfg_attr(test, derive(Eq, PartialEq))]
+#[derive(Debug, Clone, Diagnostic, Error, Eq, PartialEq)]
 #[non_exhaustive]
 pub enum ValidationErrorKind {
     /// A policy contains an entity type that is not declared in the schema.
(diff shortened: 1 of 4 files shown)


cedar-policy-validator/src/validation_result.rs
```diff
@@ -289,15 +327,15 @@ pub struct UnspecifiedEntityError {
 }
 
 /// The structure for validation warnings.
-#[derive(Debug, Clone)]
-pub struct ValidationWarning<'a> {
-    pub(crate) location: SourceLocation<'a>,
+#[derive(Hash, Eq, PartialEq, Error, Debug, Clone)]
```
cdisselkoen, line 330: For `ValidationError` this PR hides `Eq` and `PartialEq` behind `cfg(test)`, presumably because the fact that `SourceLocation` factors into the equality is more useful for testing than in other settings.  We should be consistent and either do the same here, or just derive `Eq` and `PartialEq` directly on `ValidationError`.

cedar-policy/src/api.rs
```diff
@@ -1762,16 +1762,16 @@ impl Diagnostic for ValidationResult {
 /// and provides details specific to that kind of problem. The error also records
 /// where the problem was encountered.
 #[derive(Debug, Clone, Error)]
-#[error("validation error on {location}: {}", self.error_kind())]
+#[error("{error}")]
 pub struct ValidationError {
     location: SourceLocation,
```
cdisselkoen, line 1767: Do we still need this `location` field, if there's now a `location` embedded in the `cedar_policy_validator::ValidationError`?  Looks like in the `Diagnostic` impl for this type we're always deferring to the location in `error`

cedar-policy/src/api.rs
```diff
@@ -1887,17 +1883,17 @@ pub fn confusable_string_checker<'a>(
 }
 
 #[derive(Debug, Clone, Error)]
-#[error("validation warning on {location}: {kind}")]
+#[error("{warning}")]
 /// Warnings found in Cedar policies
 pub struct ValidationWarning {
     location: SourceLocation,
```
cdisselkoen, line 1889: same here

cdisselkoen (approved):
Looks good, a few nits/questions

cedar-policy/src/api.rs
john-h-kastner-aws: The `source_location()` function here needed to return a reference. Changed `SourceLocation` to be `RefCast` so that it can wrap the internal reference type.

cdisselkoen (approved):

cedar-policy/src/api.rs
cdisselkoen: Looks good

aaronjeline (approved):

PR #805, opened 2024-04-22
Add source location to cedar-wasm validate result

chrnorm:
## Description of changes
Adds the source location to the cedar-wasm output struct. Additionally removes the use of `#[serde(untagged)]` in the validate result, in favor of a "type" tag field.

The use of `untagged` made the resulting TypeScript result object difficult to use when working with the generated TypeScript SDK. Type tags are (IMO) fairly standard in the TypeScript ecosystem and are familiar to TS developers, despite the perhaps marginal serialization overhead.

## Issue #, if available
Related to #806 (although only implemented for validation)

(marked here given that the WASM build is not yet released)


cedar-wasm/src/validator.rs
```diff
@@ -22,10 +22,9 @@ use tsify::Tsify;
 use wasm_bindgen::prelude::*;
 
 #[derive(Tsify, Debug, Serialize, Deserialize)]
-#[serde(untagged)]
+#[serde(tag = "type", rename_all = "camelCase")]
```
Swolebrain, line 25: waiting on your feedback on my reply on the issue before we accept this change, for consistency's sake

chrnorm: Thanks, I've added [some feedback here](https://github.com/cedar-policy/cedar/issues/806#issuecomment-2071916686). What I will do is revert the change in this PR because it makes the result types inconsisten, I can raise a separate PR to adjust _all_ of the TypeScript result types together depending on the outcome on the issue discussion.

cedar-policy/src/api.rs
```diff
@@ -1852,6 +1852,12 @@ impl SourceLocation {
     pub fn range_end(&self) -> Option<usize> {
         self.source_loc.as_ref().map(parser::Loc::end)
     }
+
+    /// Returns a tuple of (offset, length) of the location.
+    /// Returns `None` if this location does not have a range.
+    pub fn offset_and_length(&self) -> Option<(usize, usize)> {
+        self.source_loc.as_ref().map(|loc| (loc.start(), loc.end()))
+    }
```
cdisselkoen, line 1860: The doc comment and method name say (offset, length), but the code does (start, end).

cdisselkoen (commented):
This PR conflicts with the proposal in #800, which proposes a different standardized error structure to use in all Cedar FFIs.  We should reconcile these or come to an agreement which approach is better.

cedar-policy/src/api.rs
chrnorm: Great pickup, thanks I missed this in the initial PR. I've fixed this up.

chrnorm: I see that Miette uses offset and length, whereas `cedar_policy_core::parser::Loc` uses start and end. I've changed the `SourceLocation` struct I added to use start and end in order to align with `cedar_policy_core::parser::Loc`.

cdisselkoen:
I think the CI failure is just a result of needing to merge from `main` -- #804 changed the definition of `SourceLocation` out from under this PR.

chrnorm:
I've just merged main, although I think #824 may superseed this so feel free to close this PR!

cdisselkoen:
Closing this in favor of #824.

PR #817, opened 2024-04-25
Some quick updates to the PE api

aaronjeline:
## Description of changes
Some quick refinements based on public feedback:
1. methods like `may_be_determining` and `must_be_determining` now return `Policy` instead of `PolicyID`. This makes iterating over them much more ergonomic
2. Methods like `get` now return a `Policy` instead of an `Expr`, allowing you to grab the policy annotations as well if you want them
3. Removed `nontrivial_residual_ids()` as it's a trivial map on `nontrivial_residuals` now

## Issue #, if available
N/A


cedar-policy/src/api.rs
```diff
@@ -872,48 +872,45 @@ impl PartialResponse {
...
-    pub fn definitely_satisfied(&self) -> impl Iterator<Item = &PolicyId> {
-        self.0.definitely_satisfied().map(PolicyId::ref_cast)
+    /// Returns the set of [`Policy`]s that were definitely satisfied
+    pub fn definitely_satisfied(&self) -> impl Iterator<Item = Policy> + '_ {
+        self.0.definitely_satisfied().map(Policy::from_ast)
     }
 
-    /// Returns the set of [`PolicyId`]s that encountered errors
+    /// Returns the set of [`Policy`]s that encountered errors
     pub fn definitely_errored(&self) -> impl Iterator<Item = &PolicyId> {
```
john-h-kastner-aws, line 881: Return type doesn't reflect change in comment

cedar-policy/benches/tinytodo/tinytodo.cedar
```diff
@@ -0,0 +1,56 @@
+// Policy 0: Any User can create a list and see what lists they owno
```
john-h-kastner-aws, line 1: Should these `benches` files be committed in this PR?

cedar-policy/src/api.rs
```diff
@@ -885,35 +885,32 @@ impl PartialResponse {
...
+        self.0.may_be_determining().map(Policy::from_ast)
     }
 
     /// Returns an under-approximation of the set of determining policies
     ///
-    /// This is all policies that must be determining for all possible substitutions of the unknowns
-    pub fn must_be_determining(&self) -> impl Iterator<Item = &PolicyId> {
-        self.0.must_be_determining().map(PolicyId::ref_cast)
+    /// This is all policies that must be determining for all possible substitutions of the unknowns.
+    /// Therefor, this set will end up only containing policies whose residuals are simple `true`
```
cdisselkoen, line 895:
```suggestion
    /// Therefore, this set will end up only containing policies whose residuals are simple `true`
```

cedar-policy/src/api.rs
```diff
@@ -885,35 +885,32 @@ impl PartialResponse {
...
+    /// statements (whether `permit` or `forbid` policies)
```
cdisselkoen, line 896: This comment might be a little deceiving, if I was unfamiliar with things, I might read this and think it means that it will actually return all policies whose residuals are simple `true`.  It might be good to clarify explicitly that it won't necessarily return all policies whose residuals are simple `true`; and/or to use the same language here and in `definitely_satisfied()` (there, we just say "definitely satisfied" and don't refer to "simple `true`", but we could).

cdisselkoen (approved):

cedar-policy/src/api.rs
aaronjeline: Hmm yeah, I'll workshop it a little bit more. I added that text because someone reported they were confused about what `must_be` would return.

aaronjeline: New revision

cedar-policy/src/api.rs
```diff
@@ -884,36 +885,39 @@ impl PartialResponse {
 
     /// Returns an over-approximation of the set of determining policies
     ///
-    /// This is all policies that may be determining for any substitution of the unknowns
-    pub fn may_be_determining(&self) -> impl Iterator<Item = &PolicyId> {
-        self.0.may_be_determining().map(PolicyId::ref_cast)
+    /// This is all policies that may be determining for any substitution of the unknowns.
+    /// Policies not in this set will not effect the final decision, regardless of any
```
cdisselkoen, line 889:
```suggestion
    /// Policies not in this set will not affect the final decision, regardless of any
```

cdisselkoen (approved):

shaobo-he-aws (approved):

john-h-kastner-aws (approved):

Issue #806, opened 2024-04-22
Improve generated TypeScript Result types for Cedar WASM


chrnorm:
### Category

User level API features/changes

### Describe the feature you'd like to request

The `cedar-wasm` and `ffi` crates currently use `#[serde(untagged)]` quite extensively -- I assume to minimize serialisation overhead.

Here's an example:

```rust
#[derive(Tsify, Debug, Serialize, Deserialize)]
#[serde(untagged)]
#[tsify(into_wasm_abi, from_wasm_abi)]
pub enum AuthorizationResult {
    Success { response: InterfaceResponse },
    Error { errors: Vec<String> },
}
```

This results in TypeScript types like the following:

```typescript
export type AuthorizationResult = { response: InterfaceResponse } | { errors: string[] };
```

These types are quite difficult to work with in TypeScript -- I work with TypeScript fairly frequently and rely on language server autocomplete to help me work with new SDKs. When using the my prerelease build of the Cedar WASM SDK I was surprised to find no autocomplete suggestions in my IDE for the result type above, which makes it more difficult to discover the available attributes on the result type.

<img width="431" alt="image" src="https://github.com/cedar-policy/cedar/assets/17420369/9e1caf65-eb92-42a5-ae09-12556f0457f0">

The approach I've seen in many libraries is to instead use a tagged representation of the struct. For example:

```
export type AuthorizationResult = { response: InterfaceResponse, type: "success" } | { errors: string[], type: "error" };
```

This helps immensely with the TypeScript developer experience:

<img width="432" alt="image" src="https://github.com/cedar-policy/cedar/assets/17420369/5f0d1df7-cdcc-49c1-acb5-1c370723fac3">
<img width="737" alt="image" src="https://github.com/cedar-policy/cedar/assets/17420369/9691dcfd-076a-45a1-87fd-623847bf40e4">

This is a matter of changing `#[serde(untagged)]` to `#[serde(tag = "type")]`, although we may want to limit this to the WASM crate only.

### Describe alternatives you've considered

I hit this problem when authoring a [GitHub Action for validating Cedar policies](https://github.com/common-fate/cedar-validate-action) and ended up forking the package and implementing `#[serde(tag = "type")]` myself on the structs I was using for validation

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

Swolebrain:
[...]

PR #824, opened 2024-04-26
Surface locations from validation errors, improve TS interfaces

Swolebrain:
## Description of changes

* Improves wasm interfaces in accordance with https://github.com/cedar-policy/cedar/issues/806
* Also subsumes changes in https://github.com/cedar-policy/cedar/pull/805, which needed to be rebased

## Issue #, if available

806

### DIFF
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -1814,6 +1814,15 @@ impl SourceLocation {
     pub fn range_end(&self) -> Option<usize> {
         self.0.source_loc().map(parser::Loc::end)
     }
+
+    /// Returns a tuple of (start, end) of the location.
+    /// Returns `None` if this location does not have a range.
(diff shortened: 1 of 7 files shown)


aaronjeline (approved):
Looks good. My only comment is on two changes:
You changed a `pub enum`s variant name, and all the changes from `untagged` to `tagged(type)`, doesn't this make this a breaking change?

Swolebrain:
> Looks good. My only comment is on two changes: You changed a `pub enum`s variant name, and all the changes from `untagged` to `tagged(type)`, doesn't this make this a breaking change?

Totally, but cedar-wasm is basically unreleased. People re using it by building directly from source, but it's not officially really released. I don't have strong opinions as to whether it's okay to do this though.

cdisselkoen (approved):
This conflicts with #800 which will be making more changes, but I think this PR is good on its own and we can take it and then revise #800 accordingly.

PR #826, opened 2024-05-01
Fix remaining mentions of template "instantiation" to refer to "linking"

cdisselkoen:
## Description of changes

Fix all* remaining mentions of template "instantiation" to refer to "linking".

*this PR designed to be non-breaking, so it did not fix JSON fields `instantiations` or `template_instantiations` in the public FFI interface, although it does update their doc comments.  Those fields are due to be renamed/refactored as part of the work for #757 anyway, which will be done separately.

## Issue #, if available

Split out from #800

### DIFF
--- a/cedar-policy-cli/sample-data/sandbox_c/README.md
+++ b/cedar-policy-cli/sample-data/sandbox_c/README.md
@@ -27,7 +27,8 @@ cargo run authorize \
 
 We should get `DENY`, as there is no policy that allows this.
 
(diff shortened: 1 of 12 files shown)


khieta:
Downstream cedar-spec failure should be fixed by [cedar-spec#300](https://github.com/cedar-policy/cedar-spec/pull/300)

john-h-kastner-aws (approved):

PR #827, opened 2024-05-01
Update from 'head' to 'scope' everywhere

john-h-kastner-aws:
## Description of changes

Follow up to #825, making the change everywhere. Changes the names of a couple function that are `pub` in `core`, so might also need a `cedar-spec` update, pending result of CI tests.

### DIFF
--- a/cedar-policy-core/src/est.rs
+++ b/cedar-policy-core/src/est.rs
@@ -20,10 +20,10 @@ mod err;
 pub use err::*;
 mod expr;
 pub use expr::*;
-mod scope_constraints;
-pub use scope_constraints::*;
 mod policy_set;
 pub use policy_set::*;
+mod scope_constraints;
+pub use scope_constraints::*;
 
 use crate::ast;
 use crate::entities::EntityUidJson;
--- a/cedar-policy-core/src/est.rs
+++ b/cedar-policy-core/src/est.rs
@@ -20,8 +20,8 @@ mod err;
 pub use err::*;
 mod expr;
 pub use expr::*;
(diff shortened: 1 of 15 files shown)


cedar-policy-core/src/est.rs
```diff
@@ -20,8 +20,8 @@ mod err;
 pub use err::*;
 mod expr;
 pub use expr::*;
-mod head_constraints;
-pub use head_constraints::*;
+mod scope_constraints;
+pub use scope_constraints::*;
```
cdisselkoen, line 24: nit: could reorder these so they remain alphabetical

cedar-policy-core/src/parser/err.rs
```diff
@@ -180,11 +180,11 @@ pub enum ToASTErrorKind {
     #[error("this policy is missing the `{0}` variable in the scope")]
     MissingScopeConstraint(Var),
     /// Returned when a policy has an extra scope clause. This is not valid syntax
-    #[error("this policy has an extra constraint in the scope: `{0}`")]
+    #[error("this policy has an extra scope constraint in the scope: `{0}`")]
```
cdisselkoen, line 183: redundant: "extra scope constraint in the scope"

cdisselkoen (approved):

cedar-policy-core/src/parser/err.rs
john-h-kastner-aws: Yeah, that was removed in #825, but found its way back in there

shaobo-he-aws (approved):

PR #645, opened 2024-02-16
better error message for single =

cdisselkoen:
## Description of changes

Improves parse error messages when single `=` is used as a relational operator (ie, anywhere `==` or `<` etc would be expected).  This includes the case #452 was complaining about, but this PR's fix is more general and covers `=` used in `when` clauses as well.  Here's an example of what the error message now looks like for #452's example:

image

## Issue #, if available

Fixes #452

## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [x] A bug fix or other functionality change requiring a patch to `cedar-policy`.

I confirm that this PR (choose one, and delete the other options):

- [x] Does not update the CHANGELOG because my change does not significantly impact released code.

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [x] Does not require updates because my change does not impact the Cedar formal model or DRT infrastructure.

## Disclaimer

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #828, opened 2024-05-01
Rewrite of `EntitiesError`

aaronjeline:
## Description of changes
Rewrite of entities error to bring inline with our general error strategy
## Issue #, if available
#645 


shaobo-he-aws (approved):

cedar-policy/src/api.rs
```diff
@@ -79,8 +76,31 @@ pub mod entities {
             self.inner.size_hint()
         }
     }
+
+    /// Errors around entities
+    pub mod err {
+        pub use cedar_policy_core::entities::err::*;
```
john-h-kastner-aws, line 82: I'm a little wary of `pub use` with `*` from core, since it makes it easy to accidentally expose types we don't intend to. I see why it's nice not to explicitly `use` each struct though. Anyone else have thoughts?

cedar-policy/src/tests.rs
```diff
@@ -2774,9 +2778,15 @@ mod schema_based_parsing_tests {
             }
         ]);
         let r = Entities::from_json_value(json, None).err().unwrap();
-        let expected_euid: cedar_policy_core::ast::EntityUID = r#"User::"alice""#.parse().unwrap();
         match r {
-            EntitiesError::Duplicate(euid) => assert_eq!(euid, expected_euid),
```
john-h-kastner-aws, line 2779: We could update some of these tests to use `test_utils::expect_err` while were making changes

cedar-policy/src/api.rs
aaronjeline: Yeah, I honestly think it's worth just spelling it all out. Revision on the way

cedar-policy/src/api.rs
```diff
@@ -73,8 +73,45 @@ pub mod entities {
             self.inner.size_hint()
         }
     }
+
+    /// Errors around entities
+    pub mod err {
```
john-h-kastner-aws, line 78: Based on #830 maybe we want to move these to `err.rs`? @khieta

john-h-kastner-aws (approved):
Looks good. We'll probably want to take a second look at the exact structure and placement of sub modules once the whole error refactor is done

cedar-policy-core/src/ast/request.rs
```diff
@@ -382,6 +379,21 @@ pub enum ContextCreationError {
...
+        })
+    }
+}
+
+/// Error type for an expression that is not a record
+#[derive(Debug)]
+pub struct NotARecord {
+    /// Expression which is not a record
+    expr: Box<RestrictedExpr>,
+}
```
cdisselkoen, line 395: I know I'm late and this has already been merged, but since this is an error type, it would probably make sense for it to derive `Error`, and have the error message defined here (and `ContextCreationError::NotARecord` just have `error(transparent)`)

PR #829, opened 2024-05-01
forgot the standard derives on one enum

cdisselkoen:
## Description of changes

Fixup of #800.  I believe this is breaking because it will switch the Severity codes from `Error`, `Warning`, etc to `error`, `warning`, etc.


khieta (approved):
Probably not actually breaking since `Severity` was introduced in another unreleased (also breaking) PR

john-h-kastner-aws (approved):

PR #830, opened 2024-05-02
Refactoring in api.rs

khieta:
## Description of changes

This PR performs some refactoring to reduce the size of our api.rs file. For the most part, the changes are just copy-and-paste, but I did need to add some new `pub(crate)` functions to handle the new layer of indirection. I’ll add review comments to any new functions added.
* Moved `EntityUid` and `PolicyId` to a separate api/id.rs file.
* Moved all error definitions to a separate api/err.rs file. (My theory is that this change will make it easier to keep track of the upcoming changes for #745.)
* Moved tests in api.rs to join their friends in the tests.rs file.

The new files are included in api.rs using:
```rust
mod id;
pub use id::*;

mod err;
pub use err::*;
```
I believe this leaves the API unchanged from a user's perspective. Lmk if I'm wrong about that.

The changes in this PR introduce two new clippy warnings:
* [module_name_repetitions](https://rust-lang.github.io/rust-clippy/master/index.html#/module_name_repetitions) for both `PolicyId` and `EntityId` since these types end with "id" (the name of the containing module). I think we can safely ignore these because the types are `pub use`d in api.rs, so users should write `cedar_policy::PolicyId` instead of `cedar_policy::id::PolicyId`.
* [redundant_pub_crate](https://rust-lang.github.io/rust-clippy/master/index.html#/redundant_pub_crate) for the `fold_partition` function. The reason I made it `pub(crate)` instead of private is that it's also used in testing code. I _think_ it is still effectively private with the current setting.

### DIFF
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -20,6 +20,13 @@
     clippy::missing_errors_doc,
     clippy::similar_names
 )]
+
+mod id;
+pub use id::*;
+
+mod err;
+pub use err::*;
+
 pub use ast::Effect;
 pub use authorizer::Decision;
 use cedar_policy_core::ast;
@@ -30,35 +37,25 @@ use cedar_policy_core::ast::{
 }; // `ContextCreationError` is unsuitable for `pub use` because it contains internal types like `RestrictedExpr`
 use cedar_policy_core::authorizer;
 use cedar_policy_core::entities::{
-    ContextJsonDeserializationError, ContextSchema, Dereference, JsonDeserializationError,
-    JsonDeserializationErrorContext,
+    ContextSchema, Dereference, JsonDeserializationError, JsonDeserializationErrorContext,
 };
 use cedar_policy_core::est;
 use cedar_policy_core::est::{Link, PolicyEntry};
 use cedar_policy_core::evaluator::Evaluator;
 #[cfg(feature = "partial-eval")]
 use cedar_policy_core::evaluator::RestrictedEvaluator;
-pub use cedar_policy_core::evaluator::{EvaluationError, EvaluationErrorKind};
 pub use cedar_policy_core::extensions;
 use cedar_policy_core::extensions::Extensions;
 use cedar_policy_core::parser;
-pub use cedar_policy_core::parser::err::ParseErrors;
+use cedar_policy_core::parser::err::ParseErrors;
 use cedar_policy_core::FromNormalizedStr;
-pub use cedar_policy_validator::human_schema::SchemaWarning;
 use cedar_policy_validator::RequestValidationError; // this type is unsuitable for `pub use` because it contains internal types like `EntityUID` and `EntityType`
-pub use cedar_policy_validator::{
-    TypeErrorKind, UnsupportedFeature, ValidationErrorKind, ValidationWarningKind,
-};
 use itertools::{Either, Itertools};
 use miette::Diagnostic;
-use nonempty::NonEmpty;
 use ref_cast::RefCast;
-use serde::{Deserialize, Serialize};
 use smol_str::SmolStr;
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
-use std::convert::Infallible;
 use std::str::FromStr;
-use thiserror::Error;
 
 /// Extended functionality for `Entities` struct
 pub mod entities {
@@ -153,12 +150,12 @@ impl Entity {
         // the `Entities` object is created
         // INVARIANT(UidOfEntityNotUnspecified): by invariant on `EntityUid`
         Ok(Self(ast::Entity::new(
-            uid.0,
+            uid.into_inner(),
             attrs
                 .into_iter()
                 .map(|(k, v)| (SmolStr::from(k), v.0))
                 .collect(),
-            parents.into_iter().map(|uid| uid.0).collect(),
+            parents.into_iter().map(EntityUid::into_inner).collect(),
             &Extensions::all_available(),
         )?))
     }
@@ -172,9 +169,9 @@ impl Entity {
         // the `Entities` object is created
         // INVARIANT(UidOfEntityNotUnspecified): by invariant on `EntityUid`
         Self(ast::Entity::new_with_attr_partial_value(
-            uid.0,
+            uid.into_inner(),
             HashMap::new(),
-            parents.into_iter().map(|uid| uid.0).collect(),
+            parents.into_iter().map(EntityUid::into_inner).collect(),
         ))
     }
 
@@ -190,7 +187,7 @@ impl Entity {
     /// ```
     pub fn with_uid(uid: EntityUid) -> Self {
         // INVARIANT(UidOfEntityNotUnspecified): by invariant on `EntityUid`
-        Self(ast::Entity::with_uid(uid.0))
+        Self(ast::Entity::with_uid(uid.into_inner()))
     }
 
     /// Get the Uid of this entity
@@ -205,7 +202,7 @@ impl Entity {
     /// ```
     pub fn uid(&self) -> EntityUid {
         // INVARIANT: By invariant on self and `EntityUid`: Our Uid can't be unspecified
-        EntityUid(self.0.uid().clone())
+        EntityUid::new(self.0.uid().clone())
     }
 
     /// Get the value for the given attribute, or `None` if not present.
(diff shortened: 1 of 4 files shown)


cedar-policy/src/api/id.rs
```diff
@@ -0,0 +1,289 @@
...
+    /// ```
+    pub fn namespace(&self) -> String {
+        self.0.namespace()
+    }
+
+    /// Construct an [`EntityTypeName`] from the internal type.
+    /// This function is only intended to be used internally.
+    pub(crate) fn new(ty: ast::Name) -> Self {
+        Self(ty)
+    }
```
khieta, line 134: New function added by this PR

cedar-policy/src/api/id.rs
```diff
@@ -0,0 +1,289 @@
...
+    /// This function is only intended to be used internally.
+    pub(crate) fn into_inner(self) -> ast::EntityUID {
+        self.0
+    }
+
+    /// Deconstruct an [`EntityUid`] to get the internal type.
+    /// This function is only intended to be used internally.
+    pub(crate) fn as_inner(&self) -> &ast::EntityUID {
+        &self.0
+    }
```
khieta, line 226: New functions added by this PR

cedar-policy/src/api/id.rs
```diff
@@ -0,0 +1,289 @@
...
+    /// This function is only intended to be used internally.
+    pub(crate) fn into_inner(self) -> ast::PolicyID {
+        self.0
+    }
+
+    /// Deconstruct an [`PolicyId`] to get the internal type.
+    /// This function is only intended to be used internally.
+    pub(crate) fn as_inner(&self) -> &ast::PolicyID {
+        &self.0
+    }
```
khieta, line 267: New functions added by this PR

cedar-policy/src/api.rs
```diff
@@ -3579,6 +2839,13 @@ impl Expression {
             vec![src_expr],
         ))
     }
+
+    /// Deconstruct an [`Expression`] to get the internal type.
+    /// This function is only intended to be used internally.
+    #[cfg(test)]
+    pub(crate) fn into_inner(self) -> ast::Expr {
+        self.0
+    }
```
khieta, line 2848: New function added by this PR

cedar-policy/src/api.rs
```diff
@@ -3667,6 +2934,13 @@ impl RestrictedExpression {
             [src_expr],
         ))
     }
+
+    /// Deconstruct an [`RestrictedExpression`] to get the internal type.
+    /// This function is only intended to be used internally.
+    #[cfg(test)]
+    pub(crate) fn into_inner(self) -> ast::RestrictedExpr {
+        self.0
+    }
```
khieta, line 2943: New function added by this PR

khieta:
Marking this as ready for review now that (most of) the CI has passed. I don't think the Java failures are the fault of this PR -- I think we just need to review & merge [cedar-java#133](https://github.com/cedar-policy/cedar-java/pull/133). Will re-run CI once that PR is merged.

cedar-policy/src/api.rs
```diff
@@ -2124,19 +1557,10 @@ impl EntityUid {
...
-    #[cfg(test)]
-    pub(crate) fn from_strs(typename: &str, id: &str) -> Self {
-        Self::from_type_name_and_id(
-            EntityTypeName::from_str(typename).unwrap(),
-            EntityId::from_str(id).unwrap(),
-        )
-    }
 }
 
 impl FromStr for EntityUid {
```
cdisselkoen, line 1566: Should we move this `FromStr` impl too, since we moved the definition of `EntityUid`?

cedar-policy/src/api.rs
```diff
@@ -2688,7 +2040,7 @@ fn is_static_or_link(
 
 /// Like [`itertools::Itertools::partition_map`], but accepts a function that can fail.
 /// The first invocation of `f` that fails causes the whole computation to fail
-fn fold_partition<T, A, B, E>(
+pub(crate) fn fold_partition<T, A, B, E>(
```
cdisselkoen, line 2043: This was recently introduced -- I'm wondering if it's actually used in more than one file?  Or if we can just move it to the place it is used.

cedar-policy/src/api/err.rs
```diff
@@ -0,0 +1,569 @@
...
+    /// Action that had the attribute with the error
+    pub uid: EntityUid,
+    /// Attribute that had the error
+    pub attr: SmolStr,
+    /// Underlying evaluation error
+    #[diagnostic(transparent)]
+    pub err: EvaluationError,
+}
+
+impl From<ast::EntityAttrEvaluationError> for EntityAttrEvaluationError {
```
cdisselkoen, line 231: Let's do a consistency pass for adding `doc(hidden)` to all of these `From` impls that convert from private to public types.  Right now, some do and some do not.  (This can be a followup PR or in this PR)

cedar-policy/src/api/id.rs
cdisselkoen, line 226: Thoughts on making this an `impl AsRef` instead of (or in addition to) `as_inner`?

I guess the same goes for `into_inner()` it could be an `impl From` instead / in addition

cedar-policy/src/api/id.rs
cdisselkoen, line 267: same comment on `as_inner` and `into_inner` here

cdisselkoen (approved):
Definitely approve of the idea, making api.rs more organized.

cedar-policy/src/api.rs
```diff
@@ -2124,19 +1557,10 @@ impl EntityUid {
     pub fn from_json(json: serde_json::Value) -> Result<Self, impl miette::Diagnostic> {
```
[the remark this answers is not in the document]
khieta: I left this in api.rs because it can return an error, and I was trying to maintain the hierarchy id.rs -> err.rs -> api.rs. I'm imagining that we'll want to update the return types here to use the types that will be defined (eventually) in err.rs.

Same answer for the comment on the `FromStr` impl below.

cedar-policy/src/api.rs
khieta: It's currently used (indirectly) by `PolicySet::to_json` and in the tests. One option is to move the relevant tests for this function back to api.rs, how do you feel about that?

cedar-policy/src/api/err.rs
khieta: Easy enough to do in this PR -- I'll update this thread once I'm done.

cedar-policy/src/api/id.rs
khieta: Rust noob question: Can I make the `AsRef` and `From` impls `pub(crate)`? I don't want them exposed through `cedar_policy`'s main interface since they expose internal types.

cedar-policy/src/api/err.rs
khieta: Done 🙂

cedar-policy/src/api.rs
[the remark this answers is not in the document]
cdisselkoen: I'm pretty sure Rust allows cycles in module imports (within the same crate), if that makes a difference :)

cedar-policy/src/api.rs
cdisselkoen: No strong feelings either way.

cedar-policy/src/api/id.rs
cdisselkoen: I don't believe you can.  We put up with this for `From` impls on error types, but mark them `doc(hidden)`.  We could do the same here.  But I am also perfectly fine with keeping this as `into_inner()` and `as_inner()`, was just asking for thoughts 😄

john-h-kastner-aws (approved):
We should add `allow` directives for the new clippy lints if we're OK with them.

khieta:
Latest updates:
* Moved `SlotId` to the `id.rs` file (missed it on my first pass)
* Added `#[allow(clippy...)]` to ignore the warnings introduced by my changes
* Moved remaining `EntityUid` and `PolicyId`-related functions to `id.rs`
* Switched to using implementations of `AsRef` and `From` (with `doc(hidden)`) instead of explicit `into_inner` and `as_inner` functions


The only potentially contentious change is this last one. Because `AsRef` was already defined for `PolicyId` using the `str` type, users may need to add annotations to distinguish between the different variants. I needed to do this in one doc comment in my last commit.

andrewmwells-amazon (approved):
LGTM

PR #836, opened 2024-05-03
Add `help` message for `NonLitExtConstructor`

john-h-kastner-aws:

### DIFF
--- a/cedar-policy-validator/src/type_error.rs
+++ b/cedar-policy-validator/src/type_error.rs
@@ -289,6 +289,7 @@ pub enum TypeErrorKind {
     #[error("empty set literals are forbidden in policies")]
     EmptySetForbidden,
     #[error("extension constructors may not be called with non-literal expressions")]
+    #[diagnostic(help("consider applying extension constructors to attributes when constructing entity or context data"))]
     NonLitExtConstructor,
     /// To pass strict validation a policy cannot contain an `in` expression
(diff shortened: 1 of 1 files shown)


cdisselkoen (approved):

cedar-policy-validator/src/type_error.rs
```diff
@@ -289,6 +289,7 @@ pub enum TypeErrorKind {
     #[error("empty set literals are forbidden in policies")]
     EmptySetForbidden,
     #[error("extension constructors may not be called with non-literal expressions")]
+    #[diagnostic(help("consider applying extension constructors to attributes when constructing entity or context data"))]
```
khieta, line 292: Maybe this? In the current wording, I'm not sure what "attributes" refers to.
```suggestion
    #[diagnostic(help("consider applying extension constructors to literal values when constructing entity or context data"))]
```

khieta: (Also, yay for one line PRs! So easy to review 😄)

khieta (approved):

cedar-policy-validator/src/type_error.rs
cdisselkoen: I think the point here was not to emphasize "use literal values" (that was clear from the error message itself), but to emphasize "do it when constructing the entity or context data [and not in the policy]".  This could probably still be clearer

PR #837, opened 2024-05-03
FFI nits

khieta:
## Description of changes

A variety of small updates to the FFI to move us closer to completing #757. Downstream breaks expected.

* In the validation FFI, changed `policy_set` to `policies` for consistency with the authorization FFI.
* Standardized on using `PolicyId` for policy ids instead of `SmolStr`.
* Re-exported `JsonValueWithNoDuplicateKeys` in the public FFI API.
* Switched to using camelCase for `ValidationMode` and `Decision`. This will impact the current serialization of decisions & will likely lead to downstream breaks.
* Updated `ValidationSettings` to match the proposal in #757. Also updated the logic so that the "enabled" flag is actually used for something.
* Since this PR already makes a change that requires updating the integration tests (namely "Allow" → "allow"), I snuck in another breaking change to switch the integration tests to use camelCase.

## Issue #, if available

#757

**Note**: Also require updates to `cedar-java` and `cedar-integration-tests`.


cedar-policy/src/ffi/is_authorized.rs
```diff
@@ -352,24 +351,27 @@ impl TryFrom<crate::PartialResponse> for ResidualResponse {
             decision: partial_response.decision(),
             satisfied: partial_response
                 .definitely_satisfied()
-                .map(|p| p.id().clone())
+                .map(|p| p.id().to_smolstr())
```
cdisselkoen, line 354: Not sure if `.to_smolstr()` is smart enough to realize that `PolicyId` just wraps `SmolStr` itself (through a couple layers), and just clone the `SmolStr` here (O(1)) instead of using `Display` and allocating a new `SmolStr`.  Do we care?  Alternative is to make a dedicated method on `PolicyId` that does the right thing, even if only `pub(crate)`

cedar-policy/src/ffi/is_authorized.rs
```diff
@@ -1906,12 +1908,12 @@ mod partial_test {
...
             assert_eq!(errors.len(), 0, "{errors:?}");
-            let actual_residuals: HashSet<_> = response.nontrivial_residual_ids().collect();
+            let actual_residuals: HashSet<&str> = response.nontrivial_residual_ids().map(|id| id.as_str()).collect();
             for id in &expected_residuals {
-                assert!(actual_residuals.contains(&PolicyId::from_str(id).ok().unwrap()), "expected nontrivial residual for {id}, but it's missing")
+                assert!(actual_residuals.contains(id), "expected nontrivial residual for {id}, but it's missing")
             }
             for id in &actual_residuals {
-                assert!(expected_residuals.contains(id.to_string().as_str()),"found unexpected nontrivial residual for {id}")
+                assert!(expected_residuals.contains(id),"found unexpected nontrivial residual for {id}")
```
cdisselkoen, line 1916:
```suggestion
                assert!(expected_residuals.contains(id), "found unexpected nontrivial residual for {id}")
```

cdisselkoen (approved):

cedar-policy/src/api.rs
```diff
@@ -1060,7 +1061,10 @@ impl From<authorizer::Response> for Response {
 }
 
 /// Used to select how a policy will be validated.
-#[derive(Default, Eq, PartialEq, Copy, Clone, Debug)]
+#[derive(Default, Eq, PartialEq, Copy, Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "wasm", derive(tsify::Tsify))]
+#[cfg_attr(feature = "wasm", tsify(into_wasm_abi, from_wasm_abi))]
+#[serde(rename_all = "camelCase")]
 #[non_exhaustive]
```
john-h-kastner-aws, line 1068: This the experimental modes visible in the ffi api where they weren't before. That probably a good thing, but worth thinking about whether that could go wrong.

cedar-policy/src/ffi/is_authorized.rs
```diff
@@ -444,18 +446,18 @@ pub enum PartialAuthorizationAnswer {
 pub struct AuthorizationCall {
     /// The principal taking action
     #[cfg_attr(feature = "wasm", tsify(type = "string|{type: string, id: string}"))]
-    principal: Option<JsonValueWithNoDuplicateKeys>,
+    principal: Option<serde_json::Value>,
```
john-h-kastner-aws, line 449: Not sure I like this. We went through some effort previously to catch this duplicate key case, and I think that was the right thing to do.

cedar-policy/src/ffi/is_authorized.rs
```diff
@@ -154,10 +153,10 @@ pub struct Response {
 #[cfg_attr(feature = "wasm", tsify(into_wasm_abi, from_wasm_abi))]
 #[serde(rename_all = "camelCase")]
 pub struct Diagnostics {
-    /// `PolicyId`s of the policies that contributed to the decision.
+    /// Ids of the policies that contributed to the decision.
     /// If no policies applied to the request, this set will be empty.
-    #[cfg_attr(feature = "wasm", tsify(type = "Set<String>"))]
-    reason: HashSet<PolicyId>,
+    #[cfg_attr(feature = "wasm", tsify(type = "Set<String>"))] // REVIEW: still needed?
```
khieta, line 158: Can we delete this?

cedar-policy/src/ffi/is_authorized.rs
khieta: My thinking was that we want the FFI to use simple types where possible, and users can decide how to handle duplicates when they construct the `serde_json::Value`.

The other option is to move `JsonValueWithNoDuplicateKeys` from core to the public API -- would you prefer that?

cedar-policy/src/ffi/is_authorized.rs
khieta: I added a `pub(crate)` `into_smolstr` function in the latest commit, which helps get rid of one clone. It's not useful in this particular case though since the id comes from `Policy::id()`, which necessarily returns a reference, since under the hood policies store an `Arc<Template>`. There needs to be a clone somewhere along the call chain, and it seems reasonable to make it here.

cdisselkoen: Makes sense.

andrewmwells-amazon (approved):

cedar-policy/src/api.rs
andrewmwells-amazon: I think they should be visible in the FFI.

cedar-policy/src/ffi/is_authorized.rs
andrewmwells-amazon: IIRC, we decided downstream users should handle duplicates as they see fit. So in principle I'm fine with moving `JsonValueWithNoDuplicateKeys` to the public API. I'm not sure this will help anyone using the FFI though because they'll still likely use an equivalent of `serde_json::Value`on the other side of the interface.

shaobo-he-aws (approved):

khieta:
Returning to working on this PR! 😄 

The latest commit re-exports the `JsonValueWithNoDuplicateKeys` type, as discussed in comment threads and in-person. Remaining outstanding question on my end is: Are we ok with the switch from `PolicyId` →`SmolStr`, or would we prefer to standardize on  `PolicyId`?

cdisselkoen:
Slightly prefer `PolicyId`, but happy with either one.  Either way, I expect we'll keep `tsify(type = "string")`, so we're not really getting additional semver flexibility (the `PolicyId` type isn't really opaque at the FFI layer).  So using the `PolicyId` name would just be for documentation purposes, as I understand it, and because serde would do the right thing automatically, making it slightly more convenient to implement the FFI layer.

shaobo-he-aws (approved):

khieta:
Latest version standardizes on using `PolicyId` instead of `SmolStr`. The downstream fix for `cedar-spec` is in [cedar-spec#346](https://github.com/cedar-policy/cedar-spec/pull/346) -- I'll work on the updates needed for `cedar-integration-tests` and `cedar-java` tomorrow.

cdisselkoen (approved):

PR #863, opened 2024-05-13
Add pub applies_to_{principals, resources} APIs to validator

andrewmwells-amazon:
## Description of changes
Add pub `applies_to_{principals, resources}` APIs to validator

### DIFF
--- a/cedar-policy-validator/src/schema/action.rs
+++ b/cedar-policy-validator/src/schema/action.rs
@@ -66,6 +66,16 @@ impl ValidatorActionId {
     pub fn context_type(&self) -> Type {
         self.context.clone()
     }
+
+    /// The `EntityType`s that can be the `principal` for this action.
+    pub fn applies_to_principals(&self) -> impl Iterator<Item = &EntityType> {
+        self.applies_to.principal_apply_spec.iter()
+    }
+
+    /// The `EntityType`s that can be the `resource` for this action.
+    pub fn applies_to_resources(&self) -> impl Iterator<Item = &EntityType> {
+        self.applies_to.resource_apply_spec.iter()
+    }
 }
 
 impl TCNode<EntityUID> for ValidatorActionId {
--- a/cedar-policy/CHANGELOG.md
+++ b/cedar-policy/CHANGELOG.md
@@ -34,6 +34,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `ValidatorActionId::applies_to_principals` and `ValidatorActionId::applies_to_resources`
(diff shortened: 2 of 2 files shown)


andrewmwells-amazon:
I'll cherry-pick onto the 3.2 branch

cedar-policy-validator/src/schema/action.rs
```diff
@@ -66,6 +66,16 @@ impl ValidatorActionId {
     pub fn context_type(&self) -> Type {
         self.context.clone()
     }
+
+    /// The `EntityType`s that can be the `principal` for this action.
+    pub fn applies_to_principals(&self) -> impl Iterator<Item = &EntityType> {
```
khieta, line 71: IIRC, `EntityType` is a core type. I think the public API exposes `EntityTypeName` instead.

cedar-policy/CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `ValidatorActionId::applies_to_principals` and `ValidatorActionId::applies_to_resources`
```
john-h-kastner-aws, line 37: We haven't been tracking changes to internal APIs here

andrewmwells-amazon: ack

cedar-policy-validator/src/schema/action.rs
john-h-kastner-aws: I think this is just exposing it from the validator crate for internal use. If we want this to be a public API (which would be reasonable), we'd need to do a bit more work to expose the `ValidatorActionId` itself.

khieta: Haha, just saw John's other comment -- this isn't the public API.

john-h-kastner-aws (approved):

cedar-policy-validator/src/schema/action.rs
andrewmwells-amazon: I think this is what I want. `ValidatorSchema.action_entities()` returns `Entities` from `cedar-policy-core` so it's weird / confusing to return `EntityTypeName` from `cedar-policy` instead. I could see refactoring more validator APIs in a future PR, but since I'm adding right before the release I want these changes to be minimal.

khieta (approved):

cedar-policy-validator/src/schema/action.rs
khieta: Technically no rush to get this in 3.2 though -- we've been allowing major & minor changes in internal crates in patches.

Issue #862, opened 2024-05-13
Policy formatter drops empty lines in multi-line strings


john-h-kastner-aws:
### Bug Category

Other

### Describe the bug

Given a policy

```cedar
permit (principal, action, resource)
when { "

a
" };
```

The formatter drops the new-line character and rewrites the policy to

```cedar
permit (principal, action, resource)
when { "
a
" };
```

This is probably because the [`RcDoc::as_string`](https://docs.rs/pretty/latest/pretty/struct.RcDoc.html#method.as_string) function cannot be used with strings containing line breaks. Possibly related to #265.

### Reproduction steps

```console
[jkastner@dev-dsk-jkastner-1a-3309db3b cedar]$ cargo run format <<<'
permit (principal, action, resource)
when { "

a"};'

permit (principal, action, resource)
when { "
a" };
```

shaobo-he-aws:
This bug should be captured with the fix in https://github.com/cedar-policy/cedar/pull/865.

john-h-kastner-aws:
To fix the formatted policy, I don't see a way to keep the literal newlines while using the current `pretty` formatter. We instead need to replace them with escapes. This should be as simple as printing the `debug_escape()` of the string because we use Rust's escaping rules. Although, this may influence other escapes. For example, I think a string containing a tab literals would have it replaced with `\t` which is not horrible, but ideally it shouldn't do that.

PR #865, opened 2024-05-13
Fix formatter soundness check

shaobo-he-aws:
## Description of changes

There were two issues with the original function.

1. It iterates over templates instead of policies.
2. It zips over iterations coming from hashmaps, whose orders are non-deterministic.

### DIFF
--- a/cedar-policy-formatter/src/pprint/fmt.rs
+++ b/cedar-policy-formatter/src/pprint/fmt.rs
@@ -14,11 +14,14 @@
  * limitations under the License.
  */
 
+use std::collections::BTreeMap;
+
 use miette::{miette, Result, WrapErr};
 
-use cedar_policy_core::ast::{PolicySet, Template};
+use cedar_policy_core::ast::PolicySet;
 use cedar_policy_core::parser::parse_policyset;
 use cedar_policy_core::parser::{err::ParseErrors, text_to_cst::parse_policies};
+use smol_str::ToSmolStr;
 
 use crate::token::get_comment;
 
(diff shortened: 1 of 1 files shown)


john-h-kastner-aws (approved):
Can we add a test case?

shaobo-he-aws:
> Can we add a test case?

Good idea. Working on it.

cedar-policy-formatter/src/pprint/fmt.rs
```diff
@@ -135,6 +143,49 @@ mod tests {
...
+        let p1 = r#"permit (principal, action, resource)
+        when { "
+        
+        a
+        " };"#;
+        let p2 = r#"permit (principal, action, resource)
+        when { "
+        a
+        " };"#;
+        assert!(soundness_check(p2, &parse_policyset(p1).unwrap()).is_err());
```
khieta, line 157: Probably a dumb question, but why does the change in this PR help identify the issue in #862? Looking at the diff, it seems like the only change is to add a comparison of policy ids, which doesn't seem related.

Or is this test case unrelated to the changed code?

cedar-policy-formatter/src/pprint/fmt.rs
```diff
@@ -69,7 +77,7 @@ fn soundness_check(ps: &str, ast: &PolicySet) -> Result<()> {
                 .eq_shape(p.non_scope_constraints()))
         {
             return Err(miette!(
-                "policies differ in meaning or annotations:\noriginal: {p}\nformatted: {f_p}"
+                "policies differ in policy ids or meaning or annotations:\noriginal: {p}\nformatted: {f_p}"
```
khieta, line 80: Nit: probably easier for debugging if you separate these cases.

cedar-policy-formatter/src/pprint/fmt.rs
john-h-kastner-aws: `templates()` to `policies()` I think

khieta: Aha, thanks. @shaobo-he-aws can you add a description to the PR so it's easy not to miss this.

shaobo-he-aws: Done.

cedar-policy-formatter/src/pprint/fmt.rs
shaobo-he-aws: I think I'll leave it to future work

PR #871, opened 2024-05-14
more useful error messages in `soundness_check`

khieta:
## Description of changes

I intend to backport this PR + #865 as a "fix" for #862 in older versions of Cedar. These PRs will ensure that string literals are not silently changed by the formatter. Instead, if a string literal is changed, the formatter will produce an error & redirect the user to our issue page. This may be annoying, but will not lead to unexpected authorization behavior. It also shouldn't be too common, since it only affects policies with string literals containing empty newlines.

The actual fix for #862 will come in #870. I will likely not backport that PR since I expect it to touch many lines of code. If I finish quickly it might be included in Cedar v3.2. Otherwise, it will wait for v4.0.


shaobo-he-aws (approved):
Looks great to me.

john-h-kastner-aws (approved):

PR #874, opened 2024-05-15
Removing `error` extension function

aaronjeline:
## Description of changes
Reduces the precision of partial evaluation when reducing expressions that are lazily evaluated (`if` `and` `or` subexpressions). The primary motivation for this is removing the `error` extension function, which is impossible to type.

Increases in precision without re-introducing `error` could come later.

## Issue #, if available
N/A

### DIFF
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -322,7 +322,7 @@ impl Expr {
     }
 
     /// Create a ternary (if-then-else) `Expr`.
-    /// Takes `Arc`s instead of owned `Expr`s
+    /// Takes `Arc`s instead of owned `Expr`s.
     /// `test_expr` must evaluate to a Bool type
     pub fn ite_arc(test_expr: Arc<Expr>, then_expr: Arc<Expr>, else_expr: Arc<Expr>) -> Self {
         ExprBuilder::new().ite_arc(test_expr, then_expr, else_expr)
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -835,7 +835,7 @@ impl<T> ExprBuilder<T> {
     }
 
     /// Create a ternary (if-then-else) `Expr`.
-    /// Takes `Arc`s instead of owned `Expr`s
+    /// Takes `Arc`s instead of owned `Expr`s.
     /// `test_expr` must evaluate to a Bool type
     pub fn ite_arc(
         self,
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -321,6 +321,13 @@ impl Expr {
         ExprBuilder::new().ite(test_expr, then_expr, else_expr)
     }
 
+    /// Create a ternary (if-then-else) `Expr`.
+    /// Takes `Arc`s instead of owned `Expr`s
+    /// `test_expr` must evaluate to a Bool type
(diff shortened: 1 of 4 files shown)


cedar-policy/CHANGELOG.md
```diff
@@ -25,6 +25,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+- Reduced precision of partial evaluation when evaluating lazy
+  expressions
```
khieta, line 29: Probably makes more sense under the "changed" header than "removed" header. It also may be unclear to users what lazy expressions means in this context, so I'd suggest just
```suggestion
- Reduced precision of partial evaluation for `||`, `&&`,  and conditional expressions.
```

khieta (approved):
Looks fine to me. A little sad to see the precision loss in the updated test cases -- maybe open an issue to consider precision optimizations in the future?

cedar-policy/CHANGELOG.md
```diff
@@ -25,6 +25,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Removed the `error` extension function
```
khieta, line 30:
```suggestion
- Removed the `error` extension function, which was previously used during partial evaluation.
```

cedar-policy-core/src/ast/expr.rs
```diff
@@ -321,6 +321,13 @@ impl Expr {
         ExprBuilder::new().ite(test_expr, then_expr, else_expr)
     }
 
+    /// Create a ternary (if-then-else) `Expr`.
+    /// Takes `Arc`s instead of owned `Expr`s
+    /// `test_expr` must evaluate to a Bool type
+    pub fn ite_arc(test_expr: Arc<Expr>, then_expr: Arc<Expr>, else_expr: Arc<Expr>) -> Self {
```
shaobo-he-aws, line 327: Nit: can we make it `pub(crate)`?

cedar-policy/CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+- Reduced precision of partial evaluation for `||`, `&&`,  and conditional expressions.
```
shaobo-he-aws, line 28: Maybe we want to be more specific here? An example would help.

cedar-policy-core/src/ast/expr.rs
aaronjeline: Rest of the functions are not `pub(crate)`, why this one?

shaobo-he-aws: I think the signature of this function is not idiomatic as public functions. I was thinking another function as well but just commented on this one.

shaobo-he-aws (approved):

cedar-policy-core/src/ast/expr.rs
```diff
@@ -321,6 +321,13 @@ impl Expr {
         ExprBuilder::new().ite(test_expr, then_expr, else_expr)
     }
 
+    /// Create a ternary (if-then-else) `Expr`.
+    /// Takes `Arc`s instead of owned `Expr`s
```
cdisselkoen, line 325:
```suggestion
    /// Takes `Arc`s instead of owned `Expr`s.
```

cdisselkoen (approved):

aaronjeline:
Issue for re-increasing precision: https://github.com/cedar-policy/cedar/issues/880

cdisselkoen (approved):

PR #876, opened 2024-05-15
Implement #745 for `api::SchemaError`

shaobo-he-aws:

### DIFF
--- a/cedar-policy-validator/src/err.rs
+++ b/cedar-policy-validator/src/err.rs
@@ -22,6 +22,7 @@ use cedar_policy_core::{
 };
 use itertools::Itertools;
 use miette::Diagnostic;
+use smol_str::SmolStr;
 use thiserror::Error;
 
 use crate::human_schema;
@@ -140,28 +141,28 @@ pub enum SchemaError {
     #[diagnostic(help(
         "any entity types appearing anywhere in a schema need to be declared in `entityTypes`"
     ))]
-    UndeclaredEntityTypes(HashSet<String>),
+    UndeclaredEntityTypes(HashSet<Name>),
     /// Undeclared action(s) used in the `memberOf` field of an action.
     #[error("undeclared action(s): {0:?}")]
     #[diagnostic(help("any actions appearing in `memberOf` need to be declared in `actions`"))]
-    UndeclaredActions(HashSet<String>),
+    UndeclaredActions(HashSet<SmolStr>),
     /// This error occurs in either of the following cases (see discussion on #477):
     ///     - undeclared common type(s) appearing in entity or context attributes
     ///     - common type(s) (declared or not) appearing in declarations of other common types
     #[error("undeclared common type(s), or common type(s) used in the declaration of another common type: {0:?}")]
     #[diagnostic(help("any common types used in entity or context attributes need to be declared in `commonTypes`, and currently, common types may not reference other common types"))]
-    UndeclaredCommonTypes(HashSet<String>),
+    UndeclaredCommonTypes(HashSet<Name>),
     /// Duplicate specifications for an entity type. Argument is the name of
     /// the duplicate entity type.
     #[error("duplicate entity type `{0}`")]
-    DuplicateEntityType(String),
+    DuplicateEntityType(Name),
     /// Duplicate specifications for an action. Argument is the name of the
     /// duplicate action.
     #[error("duplicate action `{0}`")]
-    DuplicateAction(String),
+    DuplicateAction(SmolStr),
     /// Duplicate specification for a reusable type declaration.
     #[error("duplicate common type `{0}`")]
-    DuplicateCommonType(String),
+    DuplicateCommonType(Name),
     /// Cycle in the schema's action hierarchy.
     #[error("cycle in action hierarchy containing `{0}`")]
     CycleInActionHierarchy(EntityUID),
@@ -192,7 +193,7 @@ pub enum SchemaError {
     /// An action entity (transitively) has an attribute of unsupported type (`ExprEscape`, `EntityEscape` or `ExtnEscape`).
     /// This error variant should only be used when `PermitAttributes` is enabled.
     #[error("action `{0}` has an attribute with unsupported JSON representation: {1}")]
-    UnsupportedActionAttribute(EntityUID, String),
+    UnsupportedActionAttribute(EntityUID, SmolStr),
     /// Error when evaluating an action attribute
     #[error(transparent)]
     #[diagnostic(transparent)]
--- a/cedar-policy/src/api/err.rs
+++ b/cedar-policy/src/api/err.rs
@@ -33,6 +33,7 @@ pub use cedar_policy_validator::{
 use miette::Diagnostic;
 use ref_cast::RefCast;
 use smol_str::SmolStr;
+use smol_str::ToSmolStr;
 use std::collections::HashSet;
 use thiserror::Error;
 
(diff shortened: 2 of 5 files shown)


cedar-policy/src/api/err.rs
```diff
@@ -121,65 +326,63 @@ pub enum SchemaError {
...
-    CycleInCommonTypeReferences(Name),
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    CycleInCommonTypeReferences(CycleInCommonTypeReferencesError),
     /// The schema file included an entity type `Action` in the entity type
     /// list. The `Action` entity type is always implicitly declared, and it
     /// cannot currently have attributes or be in any groups, so there is no
     /// purposes in adding an explicit entry.
     #[error("entity type `Action` declared in `entityTypes` list")]
     ActionEntityTypeDeclared,
```
john-h-kastner-aws, line 370: In think these unit enum variants should also contain an empty struct so that we can add details inside the struct in the future

cedar-policy/src/api/err.rs
```diff
@@ -112,6 +113,210 @@ pub enum ReAuthorizeError {
...
+
+impl TransitiveClosureError {
+    /// Getter of hierarchy kind
+    pub fn get_kind(&self) -> HierarchyKind {
+        self.kind.clone()
+    }
+
+    /// Getter of error message
+    pub fn get_message(&self) -> SmolStr {
+        self.message.clone()
```
john-h-kastner-aws, line 150: I think I'd rather be conservative with what internal details are `pub` for error types. 

`message` specifically feels like it should be definitly hidden implementation detail.

cedar-policy-validator/src/err.rs
```diff
@@ -140,28 +141,28 @@ pub enum SchemaError {
     #[diagnostic(help(
         "any entity types appearing anywhere in a schema need to be declared in `entityTypes`"
     ))]
-    UndeclaredEntityTypes(HashSet<String>),
```
john-h-kastner-aws, line 143: Can we `pub use` this `SchemaError` and then make it the only `SchemaError` struct?

shaobo-he-aws: Do you recall why we needed this wrapper type originally?

john-h-kastner-aws: IIRC, it was because the internal enum variants with things like `UndeclaredENtityTypes(HashSet<Name>)` where `Name` was the internal type. This problem shouldn't exist with the opaque struct pattern we're using.

shaobo-he-aws: Given that semver-check can't check re-exports of another crate (see https://github.com/obi1kenobi/cargo-semver-checks/issues/2), I think we'd better keep this type to avoid nasty semver issues.

aaronjeline (approved):
Agree with johns comment

cedar-policy/src/api/err.rs
```diff
@@ -115,6 +116,197 @@ pub enum ReAuthorizeError {
...
+#[derive(Debug, Diagnostic, Error, Clone)]
+#[error("transitive closure computation/enforcement error on {kind} hierarchy: {message}")]
+pub struct TransitiveClosureError {
+    kind: HierarchyKind,
+    message: SmolStr,
+}
+
+impl TransitiveClosureError {
+    /// Getter of hierarchy kind
+    pub fn get_kind(&self) -> HierarchyKind {
```
cdisselkoen, line 147: This should probably either return `&HierarchyKind`, or we should just make `HierarchyKind` `derive(Copy)`

cedar-policy/src/api/err.rs
```diff
@@ -115,6 +116,197 @@ pub enum ReAuthorizeError {
...
+#[derive(Debug, Clone, Diagnostic, Error)]
+#[error("undeclared {kind}(s): {constructs:?}")]
+pub struct UndeclaredSchemaConstructError {
+    kind: SchemaConstructKind,
+    constructs: HashSet<SmolStr>,
+}
+
+impl UndeclaredSchemaConstructError {
+    /// Getter of schema construct kind
+    pub fn get_kind(&self) -> SchemaConstructKind {
```
cdisselkoen, line 183: Same here, we should probably either return `&SchemaConstructKind` or `derive(Copy)`

cedar-policy/src/api/err.rs
```diff
@@ -115,6 +116,197 @@ pub enum ReAuthorizeError {
...
+        self.kind.clone()
+    }
+
+    /// Getter of schema constructs
+    pub fn get_constructs(&self) -> impl Iterator<Item = SmolStr> + '_ {
+        self.constructs.iter().cloned()
```
cdisselkoen, line 189: Probably no reason to `.cloned()` here instead of returning `Item = &SmolStr`

cedar-policy/src/api/err.rs
```diff
@@ -115,6 +116,197 @@ pub enum ReAuthorizeError {
...
+impl DuplicateSchemaConstructError {
+    /// Getter of schema construct kind
+    pub fn get_kind(&self) -> SchemaConstructKind {
+        self.kind.clone()
+    }
+
+    /// Getter of schema constructs
+    pub fn get_construct(&self) -> SmolStr {
+        self.construct.clone()
+    }
```
cdisselkoen, line 210: Same here, probably no reason to `.clone()` in these; by returning a `&` we avoid the clone, and caller can still call `.clone()` if they need an owned one

cedar-policy/src/api/err.rs
```diff
@@ -115,6 +116,197 @@ pub enum ReAuthorizeError {
...
+#[error("cycle in action hierarchy containing `{node}`")]
+pub struct CycleInActionHierarchyError {
+    node: EntityUid,
+}
+
+impl CycleInActionHierarchyError {
+    /// Getter of the action in the cycle
+    pub fn get_node_in_cycle(&self) -> SmolStr {
+        self.node.to_smolstr()
+    }
```
cdisselkoen, line 224: Would it make sense to return the `EntityUid` here?

cedar-policy/src/api/err.rs
```diff
@@ -115,6 +116,197 @@ pub enum ReAuthorizeError {
...
+#[error("action `{action_uid}` has an attribute that is an empty set")]
+pub struct ActionAttributesContainEmptySetError {
+    action_uid: EntityUid,
+}
+
+impl ActionAttributesContainEmptySetError {
+    /// Getter of the action uid
+    pub fn get_action(&self) -> SmolStr {
+        self.action_uid.to_smolstr()
+    }
```
cdisselkoen, line 266: Would it make sense to return the `EntityUid` here?

cedar-policy/src/api/err.rs
```diff
@@ -115,6 +116,197 @@ pub enum ReAuthorizeError {
...
+pub struct UnsupportedActionAttributeError {
+    action_uid: EntityUid,
+    escape_type: SmolStr,
+}
+
+impl UnsupportedActionAttributeError {
+    /// Getter of the action uid
+    pub fn get_action(&self) -> SmolStr {
+        self.action_uid.to_smolstr()
+    }
```
cdisselkoen, line 283: Would it make sense to return the `EntityUid` here?

cedar-policy/src/api/err.rs
shaobo-he-aws: I think we'd better not expose implementation details here, which include what nodes are.

cedar-policy/src/api/err.rs
shaobo-he-aws: Done.

cedar-policy/src/api/err.rs
shaobo-he-aws: Done.

cedar-policy/src/api/err.rs
shaobo-he-aws: Done.

cedar-policy/src/api/err.rs
shaobo-he-aws: Done.

cedar-policy/src/api/err.rs
shaobo-he-aws: Done.

cedar-policy/src/api/err.rs
shaobo-he-aws: Done.

cedar-policy-validator/src/err.rs
john-h-kastner-aws: E.g., there could an internal struct `UndelaredEntityType` that implements `Diagnostic` and defines the error message text. Then we have the internal `SchemaError` struct hold that with `diagnostic(transparent)` and the external `SchemaError` holds an external wrapper struct `UndeclaredEntityType` that holds the internal `UndeclaredEntityStruct`, both with `diagnostic(transparant)`.

cedar-policy/src/api/err.rs
cdisselkoen: I think it's sensible for a `CycleInActionHierarchyError` to have a getter which tells users the `EntityUid` of the action that was in the cycle.  I agree we shouldn't expose implementation details about nodes, but I think returning an `EntityUid` is much more helpful to users than a string.

shaobo-he-aws:
Pushed a new design.

1. Replaced the `SchemaError` in `cedar-policy` with the one in `cedar-policy-validator`. I think it's the best for maintainability, even at the cost of messing up semver checks.
2. Group all sub-error types of `SchemaError` in a submodule of `err` and re-export it in `cedar-policy`.

cedar-policy-validator/src/err.rs
shaobo-he-aws: Please check the latest changes, @john-h-kastner-aws

cedar-policy-validator/src/err.rs
```diff
@@ -112,204 +105,333 @@ impl HumanSyntaxParseError {
...
+    #[error("unsupported feature used in schema: {0}")]
+    #[diagnostic(transparent)]
+    pub struct UnsupportedFeatureError(#[from] pub(crate) UnsupportedFeature);
+
+    #[derive(Debug, Diagnostic, Error)]
+    pub enum SchemaError {
+        /// Error thrown by the `serde_json` crate during serialization
+        #[error(transparent)]
+        #[diagnostic(transparent)]
+        // no #[from], because if you just have a serde_json::Error you should choose between JsonDeserialization and JsonSerialization appropriately
```
john-h-kastner-aws, line 236: Should the `JsonDeserialization` variant also not have `#[from]`?

shaobo-he-aws: Good catch. Fixed.

john-h-kastner-aws (approved):

PR #534, opened 2023-12-21
Updating changelog

aaronjeline:
## Description of changes
Adds 2.4.3 info to changelog
## Issue #, if available
N/A
## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [X] A change "invisible" to users (e.g., documentation, changes to "internal" crates like `cedar-policy-core`, `cedar-validator`, etc.)

I confirm that this PR (choose one, and delete the other options):

- [X] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [X] Does not require updates because my change does not impact the Cedar Dafny model or DRT infrastructure.

## Disclaimer

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #854, opened 2024-05-09
Add parsing APIs to FFI


khieta:
### Category

User level API features/changes

### Describe the feature you'd like to request

The FFI currently focuses on authorization and validation. It also supports parsing indirectly via the `enabled` flag in `ValidationSettings` (`enabled=false` means just parse the schema and policy set), but it would be convenient to support parsing more directly. The WASM and Java bindings both need this functionality, and built their own support. I suspect we just want to pull what WASM is doing directly into the `ffi` code -- see [cedar-wasm/policies_and_templates.rs](https://github.com/cedar-policy/cedar/blob/main/cedar-wasm/src/policies_and_templates.rs) and [cedar-wasm/schema_and_entities_and_context.rs](https://github.com/cedar-policy/cedar/blob/main/cedar-wasm/src/schema_and_entities_and_context.rs).

Making a separate issue from #757 because if I keep increasing the scope of that issue it will never get done 😄 

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

cdisselkoen:
+1 I think most of the functions that are today Wasm-only should be pulled into `ffi` in general, so that other bindings can make use of the same functions

khieta:
Also possibly in scope: APIs to convert between different formats (e.g. JSON vs. Cedar-syntax schemas)

john-h-kastner-aws:
Note from #882: the wasm version of these functions return strings, but with this change we should update them to use the `DetailedError` struct.

khieta:
For future reference: parsing APIs were added in #1079 and the other utilities mentioned were added in #1087.

PR #882, opened 2024-05-17
Remove `ParseErrors::errors_as_strings` (fix #534)

john-h-kastner-aws:
## Description of changes

There were three callers of this function in the wasm crate, but this function is simple enough that in lining it is not a problem. Possibly those calls should be updated to use rich information from miette.

## Issue #, if available

#534

- [ ] 


cdisselkoen (approved):
This looks good to me, although I do think the Wasm functions should use the detailed errors in the same manner as #800, but it's acceptable to defer that to #854.

aaronjeline (approved):

Issue #754, opened 2024-03-28
Improve performance for attribute access error on large records


john-h-kastner-aws:
### Describe the improvement you'd like to request

Attempting to access an non-existing attribute on a large record causes unexpectedly high latency for authorization requests. 

In the following benchmarks, I construct a context record containing 100,000 attributes and then evaluate two policies, one that accesses an existing attribute, and another that access an attribute that does not exist, resulting in an evaluation error.  The error case has much worse performance, taking over a millisecond while the non-error case is measured in nanoseconds. 

```
is_authorized large_context_record/get-attr
                        time:   [739.76 ns 745.21 ns 752.30 ns]
                        change: [-12.561% -11.919% -11.215%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 17 outliers among 100 measurements (17.00%)
  1 (1.00%) low severe
  3 (3.00%) low mild
  5 (5.00%) high mild
  8 (8.00%) high severe

Benchmarking is_authorized large_context_record/get-attr err: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 9.2s, enable flat sampling, or reduce sample count to 50.
is_authorized large_context_record/get-attr err
                        time:   [1.8045 ms 1.8104 ms 1.8164 ms]
                        change: [-1.7403% -1.1954% -0.6567%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild
```

This behavior is due to error case rebuilding the context record as a `Vec` of its keys on this line:

https://github.com/cedar-policy/cedar/blob/57d3463937aafd907ea22602bf85dd63e8764d14/cedar-policy-core/src/evaluator.rs#L774

Replacing this line with `vec![]` and re-running the benchmarks shows much better times, but we would like to keep the nicer error message if possible.

```
is_authorized large_context_record/get-attr
                        time:   [733.44 ns 733.63 ns 733.92 ns]
                        change: [-2.3506% -1.5131% -0.8134%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) high mild
  3 (3.00%) high severe
is_authorized large_context_record/get-attr err
                        time:   [849.06 ns 855.66 ns 869.37 ns]
                        change: [-99.953% -99.953% -99.953%] (p = 0.00 < 0.05)
[...]

Issue #833, opened 2024-05-02
Make exact error message text deterministic


john-h-kastner-aws:
### Describe the improvement you'd like to request

This is specifically motivated by error messages whose text is generated by iterating over a `HashMap` (see #832). The exact exact text of these error messages depends on the non-deterministic order of iteration, so we can't write tests against these errors that check for specific error text. In place of a `HashMap` error messages should hold a `BTreeMap` or `Vec`, or else sort the elements before printing.

There may be other way that error messages aren't deterministic, so I'm defining the scope of this issue to include auditing our errors for other ways this might happen.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this internal improvement
- [ ] ⚠️ This feature might incur a breaking change

PR #887, opened 2024-05-20
Improve performance on a record error case

aaronjeline:
## Description of changes
Improves a performance issue with errors for accessing a record attribute that doesn't exist. The issue exists for both entities and records, this PR fixes both. It also adds a new benchmark target to ensure we don't regress.

image

Screenshot 2024-05-20 at 9 30 33 AM

## Issue #, if available
#754 

- [X] A non-breaking change

### DIFF
--- a/cedar-policy-core/src/evaluator/err.rs
+++ b/cedar-policy-core/src/evaluator/err.rs
@@ -22,6 +22,9 @@ use smol_str::SmolStr;
 use std::sync::Arc;
 use thiserror::Error;
 
+// How many attrs will we store in an error before cutting off for performance reason
+const TOO_MANY_ATTRS: usize = 5;
+
 /// Enumeration of the possible errors that can occur during evaluation
 //
 // CAUTION: this type is publicly exported in `cedar-policy`.
@@ -183,16 +186,22 @@ impl EvaluationError {
     }
 
     /// Construct a [`EntityAttrDoesNotExist`] error
-    pub(crate) fn entity_attr_does_not_exist(
+    pub(crate) fn entity_attr_does_not_exist<'a>(
         entity: Arc<EntityUID>,
         attr: SmolStr,
-        available_attrs: Vec<SmolStr>,
+        available_attrs: impl IntoIterator<Item = &'a SmolStr>,
+        total_attrs: usize,
         source_loc: Option<Loc>,
     ) -> Self {
         evaluation_errors::EntityAttrDoesNotExistError {
             entity,
             attr,
-            available_attrs,
+            available_attrs: available_attrs
+                .into_iter()
+                .take(TOO_MANY_ATTRS)
+                .cloned()
+                .collect::<Vec<_>>(),
+            total_attrs,
             source_loc,
         }
         .into()
@@ -204,14 +213,20 @@ impl EvaluationError {
     }
 
     /// Construct a [`RecordAttrDoesNotExist`] error
-    pub(crate) fn record_attr_does_not_exist(
+    pub(crate) fn record_attr_does_not_exist<'a>(
         attr: SmolStr,
-        available_attrs: Vec<SmolStr>,
+        available_attrs: impl IntoIterator<Item = &'a SmolStr>,
+        total_attrs: usize,
         source_loc: Option<Loc>,
     ) -> Self {
         evaluation_errors::RecordAttrDoesNotExistError {
             attr,
-            available_attrs,
+            available_attrs: available_attrs
+                .into_iter()
+                .take(TOO_MANY_ATTRS)
+                .cloned()
+                .collect(),
+            total_attrs,
             source_loc,
         }
         .into()
@@ -364,8 +379,10 @@ pub mod evaluation_errors {
         pub(crate) entity: Arc<EntityUID>,
         /// Name of the attribute it didn't have
         pub(crate) attr: SmolStr,
-        /// Available attributes on the entity
+        /// (First five) Available attributes on the entity
         pub(crate) available_attrs: Vec<SmolStr>,
+        /// Total number of attributes on the entity
+        pub(crate) total_attrs: usize,
         /// Source location
         pub(crate) source_loc: Option<Loc>,
     }
@@ -376,13 +393,20 @@ pub mod evaluation_errors {
         fn help<'a>(&'a self) -> Option<Box<dyn std::fmt::Display + 'a>> {
             if self.available_attrs.is_empty() {
                 Some(Box::new("entity does not have any attributes"))
+            } else if self.available_attrs.len() == self.total_attrs {
+                Some(Box::new(format!(
+                    "Available attributes: {:?}",
+                    self.available_attrs
+                )))
             } else {
                 Some(Box::new(format!(
-                    "available attributes: {:?}",
-                    self.available_attrs
+                    "available attributes: [{}, ... ({} more attributes) ]",
+                    self.available_attrs.iter().join(","),
+                    self.total_attrs - self.available_attrs.len()
                 )))
             }
         }
+
         fn code<'a>(&'a self) -> Option<Box<dyn std::fmt::Display + 'a>> {
             None
         }
@@ -430,8 +454,10 @@ pub mod evaluation_errors {
     pub struct RecordAttrDoesNotExistError {
         /// Name of the attribute we tried to access
         pub(crate) attr: SmolStr,
-        /// Available attributes on the record
+        /// (First five) Available attributes on the record
         pub(crate) available_attrs: Vec<SmolStr>,
+        /// The total number of attrs this record has
+        pub(crate) total_attrs: usize,
         /// Source location
         pub(crate) source_loc: Option<Loc>,
     }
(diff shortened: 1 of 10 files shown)


cedar-policy-core/src/evaluator/err.rs
```diff
@@ -183,16 +186,23 @@ impl EvaluationError {
...
         attr: SmolStr,
-        available_attrs: Vec<SmolStr>,
+        available_attrs: impl IntoIterator<Item = &'a SmolStr>,
+        total_attrs: usize,
         source_loc: Option<Loc>,
     ) -> Self {
+        let available_attrs = if total_attrs < TOO_MANY_ATTRS {
+            Some(available_attrs.into_iter().cloned().collect::<Vec<_>>())
+        } else {
+            None
```
cdisselkoen, line 199: In the `else` case don't we still want to take the first `TOO_MANY_ATTRS`?

cedar-policy-core/src/evaluator/err.rs
```diff
@@ -364,8 +380,10 @@ pub mod evaluation_errors {
         pub(crate) entity: Arc<EntityUID>,
         /// Name of the attribute it didn't have
         pub(crate) attr: SmolStr,
-        /// Available attributes on the entity
-        pub(crate) available_attrs: Vec<SmolStr>,
+        /// (First Five) Available attributes on the entity
```
cdisselkoen, line 383:
```suggestion
        /// (First five) Available attributes on the entity
```

cedar-policy-core/src/evaluator/err.rs
```diff
@@ -374,13 +392,21 @@ pub mod evaluation_errors {
...
-                    self.available_attrs
-                )))
+            match &self.available_attrs {
+                Some(attrs) => {
+                    if attrs.is_empty() {
+                        Some(Box::new("entity does not have any attributes"))
+                    } else {
+                        let mut attrs = attrs.iter().collect::<Vec<_>>();
+                        attrs.sort();
+                        Some(Box::new(format!("Available attributes: {:?}", attrs)))
```
cdisselkoen, line 402: I think this needs to check whether `self.total_attrs < TOO_MANY_ATTRS` and if there's too many, say "(and `n` other attributes)" where `n` is `total_attrs - TOO_MANY_ATTRS`

cedar-policy-core/src/evaluator/err.rs
```diff
@@ -374,13 +392,21 @@ pub mod evaluation_errors {
...
+                    }
+                }
+                // If there's too many attributes we won't display them, as we'd have to sort the entire list to have a deterministic error message.
+                None => Some(Box::new(format!(
+                    "There are {} attributes on thsi entity",
```
cdisselkoen, line 407:
```suggestion
                    "there are {} attributes on this entity",
```

cedar-policy-core/src/evaluator/err.rs
```diff
@@ -442,11 +470,17 @@ pub mod evaluation_errors {
         fn help<'a>(&'a self) -> Option<Box<dyn std::fmt::Display + 'a>> {
             if self.available_attrs.is_empty() {
                 Some(Box::new("record does not have any attributes"))
-            } else {
+            } else if self.available_attrs.len() == self.total_attrs {
```
cdisselkoen, line 473: Maybe `>=`?

cedar-policy/Cargo.toml
```diff
@@ -59,14 +59,24 @@ crate_type = ["rlib", "cdylib"]
 cool_asserts = "2.0"
 criterion = "0.5"
 globset = "0.4"
-cedar-policy-core = { version = "=4.0.0", features = ["test-util"], path = "../cedar-policy-core" }
+cedar-policy-core = { version = "=4.0.0", features = [
+    "test-util",
+], path = "../cedar-policy-core" }
```
cdisselkoen, line 64: ?

cedar-policy-core/src/evaluator/err.rs
aaronjeline: We can't, because it's likely coming from a hashmap, so there's no way to deterministically get the same five.

cedar-policy/Cargo.toml
aaronjeline: I am not sure why my editor made this format, will delete

cedar-policy-core/src/evaluator/err.rs
aaronjeline: No, if the available attrs is equal to the total_attrs, then we know we have all of them don't need the ... syntax.

cedar-policy-core/src/evaluator/err.rs
aaronjeline: No, there's an invariant that's thread through here that's not made clear. If `available_attrs` is `Some` then it contains all of the attrs, as we can't sort the entire attr list if it's too big

cedar-policy-core/src/evaluator/err.rs
cdisselkoen: I guess if available_attrs is 4 and total_attrs is 3, that shouldn't happen, so I shouldn't worry about that case.  Resolving this.

cedar-policy-core/src/evaluator/err.rs
cdisselkoen: Oof. That's unfortunate. I feel like a nondeterministic error message would be a better experience for users than this? but I guess we have other tooling/tests that can't tolerate nondeterministic error messages?

aaronjeline: I had thought we'd settled on wanting error messages to be deterministic. I agree that in this specific case the non-det results in a more useful error message.

john-h-kastner-aws: I opened #833 based on some annoying non-deterministic in schema parsing errors but I don't think we discussed it. IMO we should keep determinism in error text as a goal,  but it doesn't need to be absolute if it means we can't maintain performance and error message quality.

cedar-policy/Cargo.toml
```diff
@@ -67,6 +67,14 @@ proptest = "1.0.0"
 name = "cedar_benchmarks"
 harness = false
 
+[[bench]]
```
john-h-kastner-aws, line 70: Do we need to add another `[[bench]]` target for every new benchmark? Could become annoying after a while.

cedar-policy-core/src/evaluator/err.rs
```diff
@@ -183,16 +186,23 @@ impl EvaluationError {
     }
 
     /// Construct a [`EntityAttrDoesNotExist`] error
-    pub(crate) fn entity_attr_does_not_exist(
+    pub(crate) fn entity_attr_does_not_exist<'a>(
         entity: Arc<EntityUID>,
         attr: SmolStr,
-        available_attrs: Vec<SmolStr>,
+        available_attrs: impl IntoIterator<Item = &'a SmolStr>,
```
john-h-kastner-aws, line 192: Maybe this could used `ExactSizeIterator`

cedar-policy-core/src/evaluator/err.rs
john-h-kastner-aws: I'd just make `available_attrs` something more verbose then. like `first_five_available_attrs`

cedar-policy-core/src/evaluator/err.rs
```diff
@@ -374,13 +392,21 @@ pub mod evaluation_errors {
...
+                    if attrs.is_empty() {
+                        Some(Box::new("entity does not have any attributes"))
+                    } else {
+                        let mut attrs = attrs.iter().collect::<Vec<_>>();
+                        attrs.sort();
+                        Some(Box::new(format!("Available attributes: {:?}", attrs)))
+                    }
+                }
+                // If there's too many attributes we won't display them, as we'd have to sort the entire list to have a deterministic error message.
+                None => Some(Box::new(format!(
```
john-h-kastner-aws, line 406: What would be the impact of using a `BTreeMap` for entity attributes?

cedar-policy/Cargo.toml
cdisselkoen: That's how it works for fuzz targets in cedar-spec

andrewmwells-amazon (approved):

cedar-policy/Cargo.toml
```diff
@@ -59,14 +59,26 @@ crate-type = ["rlib", "cdylib"]
 cool_asserts = "2.0"
 criterion = "0.5"
 globset = "0.4"
-cedar-policy-core = { version = "=4.0.0", features = ["test-util"], path = "../cedar-policy-core" }
+cedar-policy-core = { version = "=4.0.0", features = [
+    "test-util",
+], path = "../cedar-policy-core" }
+# NON-CRYPTOGRAPHIC random number generators
+oorandom = "11.1"
```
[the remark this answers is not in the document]
aaronjeline: Randomizing the keys used in the benchmark tests

PR #889, opened 2024-05-20
Standardize PolicySetErrors

andrewmwells-amazon:
## Description of changes
Standardize `PolicySetErrors`

## Issue #, if available
#745

### DIFF
--- a/cedar-policy/src/api/err.rs
+++ b/cedar-policy/src/api/err.rs
@@ -525,59 +525,59 @@ pub mod policy_set_error_structs {
 
     /// There was a duplicate [`PolicyId`] encountered in either the set of
     /// templates or the set of policies.
-    #[derive(Debug, Diagnostic, PartialEq, Eq, Error, Clone)]
+    #[derive(Debug, Diagnostic, Error)]
     #[error("duplicate template or policy id `{id}`")]
     pub struct AlreadyDefined {
         pub(crate) id: PolicyId,
     }
 
     /// Expected a static policy, but a template-linked policy was provided
-    #[derive(Debug, Diagnostic, PartialEq, Eq, Error, Clone)]
+    #[derive(Debug, Diagnostic, Error)]
     #[error("expected a static policy, but a template-linked policy was provided")]
     pub struct ExpectedStatic {}
 
     /// Expected a template, but a static policy was provided.
-    #[derive(Debug, Diagnostic, PartialEq, Eq, Error, Clone)]
+    #[derive(Debug, Diagnostic, Error)]
     #[error("expected a template, but a static policy was provided")]
     pub struct ExpectedTemplate {}
 
     /// Error when removing a static policy that doesn't exist
-    #[derive(Debug, Diagnostic, PartialEq, Eq, Error, Clone)]
+    #[derive(Debug, Diagnostic, Error)]
     #[error("unable to remove static policy `{policy_id}` because it does not exist")]
     pub struct PolicyNonexistentError {
         pub(crate) policy_id: PolicyId,
     }
 
     /// Error when removing a static policy that doesn't exist
-    #[derive(Debug, Diagnostic, PartialEq, Eq, Error, Clone)]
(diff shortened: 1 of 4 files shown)


cedar-policy/src/api/err.rs
```diff
@@ -517,45 +517,122 @@ impl From<cedar_policy_validator::ValidationWarning> for ValidationWarning {
...
+    #[derive(Debug, Diagnostic, PartialEq, Eq, Error, Clone)]
+    #[error("duplicate template or policy id `{id}`")]
+    pub struct AlreadyDefined {
+        pub(crate) id: PolicyId,
+    }
+
+    /// Expected a static policy, but a template-linked policy was provided
+    #[derive(Debug, Diagnostic, PartialEq, Eq, Error, Clone)]
+    #[error("expected a static policy, but a template-linked policy was provided")]
+    pub struct ExpectedStatic {}
```
andrewmwells-amazon, line 537: Not sure if we want empty structs for these two. I did it to be consistent, but can be persuaded otherwise.

cedar-policy/src/api/err.rs
```diff
@@ -564,26 +641,28 @@ pub enum PolicySetError {
     #[error("Error serializing a policy to JSON: {0}")]
     #[diagnostic(transparent)]
     ToJson(#[from] PolicyToJsonError),
-    /// Errors encountered in JSON ser/de
-    #[error("Error serializing or deserializing from JSON: {0})")]
+    /// Errors encountered in JSON ser/de of the policy set (as opposed to individual policies)
+    #[error("Error serializing / deserializing PolicySet to / from JSON: {0})")]
```
andrewmwells-amazon, line 645: Not sure if we want to implement `Diagnostic` for this. IMO `serde_json::Error` is probably better for this case because we'll automatically get improvements to `serde_json` error reporting. The main downside is it's less consistent. (Note, if I understand the code correctly this will only happen if the `PolicySet` isn't in our expected JSON format. We should never fail `PolicySet -> JSON`.)

cedar-policy/src/api/err.rs
john-h-kastner-aws: I think they're nice. Makes it easier to add more detail to the error message in the future for pretty much no cost now.

shaobo-he-aws: I'm convinced by @john-h-kastner-aws as well. We should add structs for empty variants.

shaobo-he-aws (approved):

cedar-policy/src/api/err.rs
```diff
@@ -517,45 +517,122 @@ impl From<cedar_policy_validator::ValidationWarning> for ValidationWarning {
...
+
+    /// Error when unlinking a template
+    #[derive(Debug, Diagnostic, PartialEq, Eq, Error, Clone)]
+    #[error("unable to unlink policy `{policy_id}` because it does not exist")]
+    pub struct LinkNonexistentError {
+        pub(crate) policy_id: PolicyId,
+    }
+
+    /// Error when removing a link that is not a link
+    #[derive(Debug, Diagnostic, PartialEq, Eq, Error, Clone)]
```
john-h-kastner-aws, line 580: Add these derives to `PolicySetError` also? Not much use otherwise.

john-h-kastner-aws (approved):

cedar-policy/src/api/err.rs
andrewmwells-amazon: `serde_json::Error` doesn't derive these, so `PolicySetError` can't either. I was thinking they'd still be useful for testing, but since none of our tests use them and you can't construct the structs outside of this crate I think we can remove them.

cedar-policy/src/api/err.rs
andrewmwells-amazon: Leaving as `serde_json::Error for now at least.

shaobo-he-aws (approved):

PR #861, opened 2024-05-10
Fix formatter comment reordering inside euid and on if-then-else

john-h-kastner-aws:
## Description of changes

Fix #787 and a similar comment reordering issue I found for if-then-else expressions.

Based on the existing tests and some time running the `formatter` fuzz target locally, I'm fairly confident that not adding `end_comment` in `VariableDef` won't break anything, but review is of course appreciated. 

As a side effect of the `VariableDef` change, a line break is introduced in the policy scope for some policies in test.cedar. I don't understand how that was introduced, but I'm willing to accept it.

## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [x] A backwards-compatible change requiring a minor version bump to `cedar-policy` (e.g., addition of a new API).

I confirm that this PR (choose one, and delete the other options):

- [x] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [x] Does not require updates because my change does not impact the Cedar formal model or DRT infrastructure.

PR #856, opened 2024-05-10
Use `insta` for snapshot testing of formatter output

john-h-kastner-aws:
## Description of changes

Use [insta](https://insta.rs/) to make testing the formatter easier.

Our existing test for the formatter are all taking some Cedar source file, formatting it, and asserting that it's equal to some formatted policy generated by a previous commit. Updating these test requires tedious manual work to review and update the files. `inst` automates most of this.

## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [x] A change "invisible" to users (e.g., documentation, changes to "internal" crates like `cedar-policy-core`, `cedar-validator`, etc.)

I confirm that this PR (choose one, and delete the other options):

- [x] Does not update the CHANGELOG because my change does not significantly impact released code.

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [x] Does not require updates because my change does not impact the Cedar formal model or DRT infrastructure.

PR #894, opened 2024-05-21
Cherry-pick formatter fixes to release/3.1.x

shaobo-he-aws:
## Description of changes

We need #856 for smooth cherry-picking.

* #867 
* #861 
* #870


khieta:
I thought our policy was to support the latest minor in supported majors? If that's the case, then the fixes only need to be in 3.2.x.

aaronjeline (approved):

shaobo-he-aws:
> I thought our policy was to support the latest minor in supported majors? If that's the case, then the fixes only need to be in 3.2.x.

I thought the same but saw #865 is included in a 3.1.x release.

shaobo-he-aws:
Closed as we don't support 3.1 as of now.

PR #908, opened 2024-05-23
Minor edits to `ParseError`

khieta:
## Description of changes

Another round of edits to the parsing-related errors. I expect to make more changes to the public API in a future PR (in particular: I think we want to better hide the terms "AST" and "CST" in the public API).
* **[Affects public API]** Added `pub use` for `RestrictedExpressionParseError` and `ParseError`, which could previously be returned by public functions, but were never exported themselves 😅 Added "CAUTION" notes to these error types like the ones added in #860.
* Removed the `ParseLiteral` and `RestrictedExpr` cases from `ParseError`. `RestrictedExpr::from_str()` and `Literal::from_str()` now have their own error types.
* Updated the error message for `UnescapeError`s to not print the `EscapeError` variant (which comes from Rust's lexer).

(This PR will be batched with the other breaking changes that are part of #745.)

### DIFF
--- a/cedar-policy-core/src/parser.rs
+++ b/cedar-policy-core/src/parser.rs
@@ -40,7 +40,7 @@ use smol_str::SmolStr;
 use std::collections::HashMap;
 
 use crate::ast;
-use crate::ast::RestrictedExprParseError;
+use crate::ast::RestrictedExpressionParseError;
 use crate::est;
 
 /// simple main function for parsing policies
@@ -243,7 +243,7 @@ pub(crate) fn parse_expr(ptext: &str) -> Result<ast::Expr, err::ParseErrors> {
 /// `FromStr` impl or its constructors
 pub(crate) fn parse_restrictedexpr(
     ptext: &str,
-) -> Result<ast::RestrictedExpr, RestrictedExprParseError> {
+) -> Result<ast::RestrictedExpr, RestrictedExpressionParseError> {
     let expr = parse_expr(ptext)?;
     Ok(ast::RestrictedExpr::new(expr)?)
 }
@@ -286,24 +286,19 @@ pub(crate) fn parse_name(name: &str) -> Result<ast::Name, err::ParseErrors> {
 ///
 /// Private to this crate. Users outside Core should use `Literal`'s `FromStr` impl
 /// or its constructors
-pub(crate) fn parse_literal(val: &str) -> Result<ast::Literal, err::ParseErrors> {
+pub(crate) fn parse_literal(val: &str) -> Result<ast::Literal, err::LiteralParseError> {
     let mut errs = err::ParseErrors::new();
     let cst = text_to_cst::parse_primary(val)?;
     let Some(ast) = cst.to_expr(&mut errs) else {
-        return Err(errs);
+        return Err(err::LiteralParseError::Parse(errs));
     };
     if errs.is_empty() {
-        match ast.into_expr_kind() {
+        match ast.clone().into_expr_kind() {
             ast::ExprKind::Lit(v) => Ok(v),
-            _ => Err(
-                err::ParseError::ParseLiteral(err::ParseLiteralError::ParseLiteral(
-                    val.to_string(),
-                ))
-                .into(),
-            ),
+            _ => Err(err::LiteralParseError::InvalidLiteral(ast)),
         }
     } else {
-        Err(errs)
+        Err(err::LiteralParseError::Parse(errs))
     }
 }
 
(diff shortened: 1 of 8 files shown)


shaobo-he-aws (approved):

cedar-policy-core/src/parser.rs
```diff
@@ -1146,4 +1141,41 @@ mod parse_tests {
...
+        // test strings with valid escapes
+        // convert a string `s` to `<double-quote> <escaped-form-of-s> <double-quote>`
+        // and test if the resulting string literal AST contains exactly `s`
+        // for instance, "\u{1F408}"" is converted into r#""\u{1F408}""#,
+        // the latter should be parsed into `Literal(String("🐈"))` and
+        // `🐈` is represented by '\u{1F408}'
+        let test_valid = |s: &str| {
+            let r = parse_literal(&format!("\"{}\"", s.escape_default()));
+            assert!(r.is_ok());
+            assert_eq!(r.unwrap(), ast::Literal::String(s.into()));
```
john-h-kastner-aws, line 1156: While we're in here, this should give a nicer error. 

```suggestion
            assert_eq!(r.unwrap(), Ok(ast::Literal::String(s.into())));
```

john-h-kastner-aws (approved):

cedar-policy-core/src/parser.rs
```diff
@@ -286,24 +286,19 @@ pub(crate) fn parse_name(name: &str) -> Result<ast::Name, err::ParseErrors> {
...
+pub(crate) fn parse_literal(val: &str) -> Result<ast::Literal, err::LiteralParseError> {
     let mut errs = err::ParseErrors::new();
     let cst = text_to_cst::parse_primary(val)?;
     let Some(ast) = cst.to_expr(&mut errs) else {
-        return Err(errs);
+        return Err(err::LiteralParseError::Parse(errs));
     };
     if errs.is_empty() {
-        match ast.into_expr_kind() {
+        match ast.clone().into_expr_kind() {
```
cdisselkoen, line 296: maybe we can avoid this `.clone()`?  Cloning an AST might be expensive and it seems we only need to clone a `Lit` in the happy path?

cedar-policy-core/src/parser.rs
cdisselkoen: I think you need to remove the `.unwrap()` as well

cedar-policy-core/src/parser/err.rs
```diff
@@ -43,52 +43,33 @@ pub(crate) type RawErrorRecovery<'a> = lalr::ErrorRecovery<RawLocation, RawToken
...
+// adding public methods.
 #[derive(Clone, Debug, Diagnostic, Error, PartialEq, Eq)]
 pub enum ParseError {
-    /// Error from the CST parser.
+    /// Error from the text -> CST parser
     #[error(transparent)]
     #[diagnostic(transparent)]
     ToCST(#[from] ToCSTError),
-    /// Error in the CST -> AST transform, mostly well-formedness issues.
+    /// Error from the CST -> AST transform; mostly well-formedness issues.
```
cdisselkoen, line 57: I'm not sure the comment about well-formedness issues is useful/helpful

```suggestion
    /// Error from the CST -> AST transform
```

cedar-policy-core/src/parser/err.rs
```diff
@@ -685,6 +666,10 @@ pub fn expected_to_string(expected: &[String], config: &ExpectedTokenConfig) ->
 }
 
 /// Multiple parse errors.
+//
+// CAUTION: this type is publicly exported in `cedar-policy`.
+// Don't make fields `pub`, don't make breaking changes, and use caution when
+// adding public methods.
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 pub struct ParseErrors(pub Vec<ParseError>);
```
cdisselkoen, line 674: "Don't make fields `pub`" is confusing when the type already has a pub field

cdisselkoen (approved):

cedar-policy-core/src/parser/err.rs
cdisselkoen: Maybe we should have this field be `pub(crate)` instead of `pub`, now that it's publically exported?  Or should this type be an exception?

cedar-policy-core/src/parser.rs
khieta: Thanks for the spot check. Fixed in the latest commit.

cedar-policy-core/src/parser/err.rs
khieta: Changed to `pub(crate)` in the latest commit. I plan to make the field private in a future PR.

khieta: On second thought... undid this in the latest commit since I don't want to fix the resulting build failures. I promise that I'll take care of it in a PR in the next few days!

cdisselkoen (approved):

PR #922, opened 2024-05-29
Ensure that parse errors are nonempty (part 1/2)

khieta:
## Description of changes

Part 1 (of 2) of updating the `ParseErrors` structure to use `NonEmpty` under the hood. This makes this error type consistent with the parse error type for human-readable schemas. It also makes sense -- if the input fails to parse, it should return at least one error explaining what went wrong (instead of "unknown error"). I suspect that our current code does this (i.e., it never returns "unknown error"), but sticking this invariant in the type system makes things cleaner 🙂  

I'm splitting this change into two PRs because it will be quite large, and I want to make it easier to review. This first PR:
* Changes the inner type of `ParseErrors` from a (public) `Vec` to (private) `NonEmpty`.
* Adds a new parse error variant `Unknown` to account for the (probably unreachable?) case where no error is returned. I intend to remove this variant in the next PR.
* Updates the policy parsing functions in `cst_to_ast.rs` to return a `Result<_, ParseErrors>` instead of using an `Option<_>` + mutable list of errors.
* Updates the remaining functions in `cst_to_ast.rs` to use a vector of `ToASTError`s instead of `ParseError`s. This is a temporary measure -- in the next PR, all functions in this file will be updated as in the bullet above.

_Public-facing changes:_
* Made the `ParseErrors` field private
* Some changes to the traits implemented by `ParseErrors`
* Renamed one variant of `ParseError` (`MissingNodeData` -> `EmptyNodeInvariantViolation`)

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -82,62 +85,67 @@ impl Node<Option<cst::Policies>> {
...
     pub fn with_generated_policyids(
         &self,
-    ) -> Option<impl Iterator<Item = (ast::PolicyID, &Node<Option<cst::Policy>>)>> {
+    ) -> Result<impl Iterator<Item = (ast::PolicyID, &Node<Option<cst::Policy>>)>, ParseErrors>
+    {
         // if `self` doesn't have data, nothing we can do here, just propagate
         // the `None`; we don't need to signal an error, because one was already
         // signaled when the `Node` without data was created
-        let policies = self.as_inner()?;
+        let policies = self.ok_or_missing()?;
```
cdisselkoen, line 93: Comment is awkward now, it says "we don't need to signal an error", but it does signal an error.  Moreover the old code never threw `MissingNodeData` and the new code does; I'm not sure that's a desirable behavior change?

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -82,62 +85,67 @@ impl Node<Option<cst::Policies>> {
...
-            Some(pset)
-        } else {
-            None
+        match all_errs.split_first() {
+            None => Ok(pset),
+            Some((first, rest)) => {
+                let mut first = first.clone();
+                rest.iter()
+                    .for_each(|errs| first.extend(errs.iter().cloned()));
+                Err(first)
```
cdisselkoen, line 147: maybe make this a helper function `.flatten()` with type `Vec<ParseErrors> -> ParseErrors`?

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -147,103 +155,80 @@ impl Node<Option<cst::Policy>> {
...
+                                Some(ToASTError::new(
+                                    ToASTErrorKind::UnexpectedTemplate { slot: slot.clone() },
+                                    err.source_loc().clone(),
+                                ))
+                            }
+                            _ => None,
+                        },
+                        _ => None,
+                    })
+                    .collect::<Vec<_>>();
```
cdisselkoen, line 203: I think we can avoid this `.collect()`, and just pass an iterator straight to `.extend()`

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -147,103 +155,80 @@ impl Node<Option<cst::Policy>> {
...
-        &self,
-        id: ast::PolicyID,
-        errs: &mut ParseErrors,
-    ) -> Option<ast::Template> {
-        // if `self` doesn't have data, nothing we can do here, just propagate
-        // the `None`; we don't need to signal an error, because one was already
-        // signaled when the `Node` without data was created
-        let policy = self.as_inner()?;
+    pub fn to_policy_template(&self, id: ast::PolicyID) -> Result<ast::Template, ParseErrors> {
+        let policy = self.ok_or_missing()?;
```
cdisselkoen, line 213: Same comment as above, we didn't use to throw `MissingNodeData` here ever, now we do, is that desirable?

cedar-policy-core/src/parser/err.rs
```diff
@@ -670,52 +676,61 @@ pub fn expected_to_string(expected: &[String], config: &ExpectedTokenConfig) ->
...
-    /// Constructs a new, empty `ParseErrors`.
-    pub fn new() -> Self {
-        ParseErrors(Vec::new())
+    /// Construct a new `ParseErrors` from another `NonEmpty` type
+    pub(crate) fn new_from_nonempty(errs: NonEmpty<ParseError>) -> Self {
+        let (first, rest) = errs.split_first();
+        let mut nv = NonEmpty::singleton(first.clone());
+        let mut v = rest.to_vec();
+        nv.append(&mut v);
+        Self(nv)
```
cdisselkoen, line 702: why not just `Self(errs)` and avoid the cloning and allocation?

cedar-policy-core/src/parser/err.rs
```diff
@@ -803,14 +806,25 @@ impl<T: Into<ParseError>> From<T> for ParseErrors {
...
+    /// error if necessary.
     fn from(errs: Vec<ParseError>) -> Self {
-        ParseErrors(errs)
+        match errs.split_first() {
+            None => ParseErrors::singleton(ParseError::ToAST(ToASTError::new(
+                ToASTErrorKind::Unknown,
+                Loc::new(0, Arc::from("")),
+            ))),
+            Some((first, rest)) => ParseErrors::new(first.clone(), rest.iter().cloned()),
+        }
```
cdisselkoen, line 818: any way to avoid cloning all the errors?  maybe `Vec::drain()`?  (minor nit though, as performance on the error path isn't as important)

cedar-policy-core/src/parser/err.rs
```diff
@@ -803,14 +806,25 @@ impl<T: Into<ParseError>> From<T> for ParseErrors {
...
 }
 
-impl<T: Into<ParseError>> FromIterator<T> for ParseErrors {
-    fn from_iter<I: IntoIterator<Item = T>>(errs: I) -> Self {
-        ParseErrors(errs.into_iter().map(Into::into).collect())
+impl From<Vec<ToASTError>> for ParseErrors {
+    fn from(errs: Vec<ToASTError>) -> Self {
+        errs.into_iter()
+            .map(Into::into)
+            .collect::<Vec<ParseError>>()
```
cdisselkoen, line 826: Can we avoid `.collect()`ing this intermediate `Vec`?  `ParseErrors` does have a `.from_iter()`

cedar-policy-core/src/parser/text_to_cst.rs
```diff
@@ -50,21 +50,23 @@ fn parse_collect_errors<'a, P, T>(
     let mut errs = Vec::new();
     let result = parse(parser, &mut errs, &Arc::from(text), text);
 
-    let mut errors: err::ParseErrors = errs
+    let errors = errs
         .into_iter()
         .map(err::ToCSTError::from_raw_err_recovery)
-        .collect();
+        .map(Into::into)
+        .collect::<Vec<err::ParseError>>();
```
cdisselkoen, line 57: I think we can avoid this `.collect()` here and have it only in the `Err(e)` case.  The local variable `errors` here would just be an iterator.

cedar-policy-core/src/parser/cst_to_ast.rs
john-h-kastner-aws: > old code never threw MissingNodeData and the new code does

The `simple-parser` fuzz target asserts this fact, so we will at the very least need to update that. 


IIUC, the node data is `None` only when we use lalrpops error recovery mechanism. 
Which is only here
https://github.com/cedar-policy/cedar/blob/11ae14a8f1437b06d08e59ffcd4b535b8ca75f72/cedar-policy-core/src/parser/grammar.lalrpop#L192
and here
https://github.com/cedar-policy/cedar/blob/11ae14a8f1437b06d08e59ffcd4b535b8ca75f72/cedar-policy-core/src/parser/grammar.lalrpop#L115

If we wanted to we could probably refactor so that the CST nodes that don't use this mechanism can't be `None`.

khieta: Ah, I meant to remove this comment throughout the code as I update to using `ok_or_missing`, but looks like I missed this one.

I think that throwing the error is the more sensible behavior, since the alternative could result in returning a `None` result, but no parse errors. Per my understanding, the `MissingNodeData` error can only be thrown if CST parsing failed, which should mean that the CST->AST conversion is never reached -- I'll update the error message for this variant from "data should not be empty" to "invalid input data: unable to continue processing" (open to better suggestions!)

cedar-policy-core/src/parser.rs
```diff
@@ -150,41 +120,26 @@ pub fn parse_policy_template_to_est_and_ast(
...
-        return Err(errs);
-    };
-    if errs.is_empty() {
-        let est = cst_node.try_into()?;
-        Ok((est, ast))
-    } else {
-        Err(errs)
-    }
+    let ast = cst.to_policy_template(id)?;
+    let est = cst.ok_or_missing()?.clone().try_into()?;
```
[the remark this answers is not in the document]
khieta: The reason is that `ok_or_missing` takes `&self` to `Result<&T,_>`, and `try_into` expects an owned input. Is the right solution is to define another `ok_or_missing` with owned inputs/outputs? What is the Rust-appropriate name for this variant?

cedar-policy-core/src/parser/cst_to_ast.rs
john-h-kastner-aws: I'd also try to rename the error variant and `ok_or_missing` function to something more self explanatory. It's not at all clear why the data would be "Missing" just based on the name.

cedar-policy-core/src/parser/cst_to_ast.rs
khieta: Based on the compiler error message, it seems this is required since the input to `extend` is an iterator over `errs` (which is what's being extended).

cedar-policy-core/src/parser/err.rs
khieta: Great question. No idea why I did that - will fix in the next commit.

cedar-policy-core/src/est.rs
```diff
@@ -145,7 +142,6 @@ impl TryFrom<cst::Policy> for Policy {
             resource,
             conditions,
             annot_success,
-            errs.is_empty(),
```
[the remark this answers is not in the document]
cdisselkoen: I'm not convinced that we actually always have an error here.  `errs` was constructed from a possibly-empty vector.  I'm wondering whether this check should be replaced with a check on whether the `ParseErrors` is `Unknown`, in order to maintain the previous semantics.

cedar-policy-core/src/parser.rs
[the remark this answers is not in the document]
cdisselkoen: Sure, that seems like a fine solution.  I don't love the name `ok_or_missing` anyway, so maybe they could both use better names.

cedar-policy-core/src/parser/cst_to_ast.rs
cdisselkoen: > the node data is `None` only when we use lalrpops error recovery mechanism.

My knowledge is from a few refactors ago, but I didn't think this was the case.  The CST->AST transform frequently (last I looked) uses a pattern where it encounters a problem, and instead of immediately throwing the error, just adds the error to a `Vec`, makes the node `None`, and keeps going, in order to try to find other possible (unrelated) errors.  The old `.as_inner()` line here basically says "if we have a `None`, this method doesn't have any meaningful way to find more errors for you" and returns.

cedar-policy-core/src/est.rs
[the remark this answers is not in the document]
khieta: Yep, realizing that now. Will comment again once I've pushed a fixed version.

cedar-policy-core/src/parser/cst_to_ast.rs
john-h-kastner-aws: I don't think that's right. It frequently returns `None`, but I don't think it every makes a CST node `None` that wasn't already

cdisselkoen: Ack. You're right.

shaobo-he-aws (approved):

cedar-policy-core/src/est.rs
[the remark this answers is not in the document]
khieta: Ok, new version pushed.

cedar-policy-core/src/est.rs
```diff
@@ -188,14 +184,14 @@ impl TryFrom<cst::Cond> for Clause {
...
+                errs.extend(expr_errs);
                 None
             }
         };
 
-        if let (Some(expr), true) = (expr, errs.is_empty()) {
+        if let Some(expr) = expr {
             Ok(match is_when {
-                // PANIC SAFETY errs should be non empty if is_when is None
+                // PANIC SAFETY expr should be Some if is_when is None
```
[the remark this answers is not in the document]
khieta: New version pushed. Same strategy as your comment earlier in this file.

cedar-policy-core/src/parser.rs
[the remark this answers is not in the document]
khieta: New names:
```rust
pub fn try_as_inner(&self) -> Result<&T, ToASTError> {
  ...
}

pub fn try_into_inner(self) -> Result<T, ToASTError> {
  ...
}
```

shaobo-he-aws (approved):

cedar-policy-core/src/parser/cst_to_ast.rs
khieta: I agree with John's comment: I don't think there is any way the text->CST conversion will return an empty node without an error. So I've changed the name & description of `MissingNodeData` to the following:
```rust
/// Returned when a CST node is empty during CST to AST/EST conversion.
/// This should have resulted in an error during the text to CST
/// conversion, which will terminate parsing. So it should be unreachable
/// in later stages.
#[error("internal invariant violated. Parsed data node should not be empty")]
EmptyNodeInvariantViolation,
```

cedar-policy-core/src/parser/err.rs
khieta: Switched to using `from_iter()` , which includes an extra collect, but no clone.

(But note that this function will be deleted in a future PR.)

cedar-policy-core/src/parser/err.rs
khieta: The `from_iter` function also collects a `Vec` so probably no point. I'm going to leave this comment unaddressed for now since I expect to delete this function in a future PR.

khieta:
Thanks for the comments! This PR is ready for re-review.

Unfortunately, it looks like this PR _does_ break downstream DRT. I will have a PR up to `cedar-spec` to fix this shortly.

cedar-policy-core/src/est.rs
```diff
@@ -163,15 +152,19 @@ impl TryFrom<cst::Policy> for Policy {
                 conditions,
                 annotations: annotations.into_iter().map(|(k, v)| (k, v.val)).collect(),
             }),
-            _ => Err(errs),
+            (_, _, _, _, Err(mut cond_errs), _, _) => {
+                cond_errs.extend(errs);
+                Err(cond_errs)
+            }
+            _ => Err(errs.into()),
```
cdisselkoen, line 159: Nice

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -3694,13 +3662,12 @@ mod tests {
...
-            let e = text_to_cst::parse_policy(src)
+            text_to_cst::parse_policy(src)
                 .expect("parse_error")
-                .to_policy_template(ast::PolicyID::from_string("i0"), &mut errs);
-            if e.is_none() {
-                panic!("Failed to create a policy template: {:?}", errs);
-            }
+                .to_policy_template(ast::PolicyID::from_string("i0"))
+                .unwrap_or_else(|errs| {
+                    panic!("Failed to create a policy template: {:?}", errs);
```
cdisselkoen, line 3669: maybe `miette::Report::new(errs)`

cedar-policy-core/src/est.rs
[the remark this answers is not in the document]
cdisselkoen: The new version here continues to call `.is_empty()`, but `.is_empty()` [will always return false](https://docs.rs/nonempty/latest/nonempty/struct.NonEmpty.html#method.is_empty), so that's not a meaningful check, and in fact might (? not sure) have different behavior than before this PR in an undesirable way, because even if `errs` was created from an empty `Vec`, it will report `.is_empty() == false` due to the `Unknown` error being added.

cedar-policy-core/src/est.rs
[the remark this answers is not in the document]
cdisselkoen: Same comment about `is_empty()`

cdisselkoen (approved):
Looks good, pending the comment threads from my earlier review

cedar-policy-core/src/parser/err.rs
```diff
@@ -822,27 +844,18 @@ impl<T: Into<ParseError>> Extend<T> for ParseErrors {
 
 impl IntoIterator for ParseErrors {
     type Item = ParseError;
-    type IntoIter = std::vec::IntoIter<Self::Item>;
+    type IntoIter = iter::Chain<iter::Once<Self::Item>, std::vec::IntoIter<Self::Item>>;
 
     fn into_iter(self) -> Self::IntoIter {
-        self.0.into_iter()
+        iter::once(self.0.head).chain(self.0.tail)
```
cdisselkoen, line 850: One more comment: `NonEmpty` does implement `IntoIterator` itself, so this can be just
```suggestion
        self.0.into_iter()
```
and the `type IntoIter` would be `<NonEmpty<ParseError> as IntoIterator>::IntoIter`

shaobo-he-aws (approved):

cdisselkoen (approved):

Issue #878, opened 2024-05-17
Box error types


cdisselkoen:
### Describe the improvement you'd like to request

Our code uses `Result<T, SomeErr>` a lot, both internally and in its public APIs.  In many cases these `SomeErr` types are large enums, large in the physical sense (many bytes) due to information stored in each error variant.  For one example, Core's `EntitiesError` is 376 bytes; a single `EntityUID` is 120 bytes.  This means that our Result types are very large as well, which requires large amounts of stack space be allocated and de-allocated as we call functions, and also perhaps hurts our CPU cache performance.  (And it causes Clippy warnings in some cases.)

As a proposed general solution for this, we should transition from `Result<T, SomeErr>` to `Result<T, Box<SomeErr>>` everywhere -- at least in internal APIs, but possibly in our public APIs as well, while we're already overhauling error types for #745.   Because `Box` is small (conceptually just a pointer), `Result<T, Box<SomeErr>>` will have size very close to the size of `T`.  If this has noticeable performance implications at all, it will likely be to improve the performance of the happy-path (due to smaller Result types) at the expense of some performance on the error path (due to more heap allocation and pointer-chasing), which is an acceptable/desirable tradeoff.

### Describe alternatives you've considered

Boxing the types in each variant; for instance `EntitiesError::SomeVariantName(Box<SomeErr>)`.  This is more verbose (requires us writing `Box` more places) and also impedes matching on the error more than the proposed solution, I think.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this internal improvement
- [X] ⚠️ This feature might incur a breaking change

cdisselkoen:
See discussion in #928.  We will keep this issue on the backlog for now, as it represents a pretty large effort for an unclear benefit. But, if in the future we are looking for ways to improve performance and identify this as relevant to performance metrics we care about, we can revisit this issue.

tustvold:
FWIW Since Rust 1.87 clippy will complain about these error variants - https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err

PR #928, opened 2024-05-30
Box `EvaluationError` when it appears in `Result`

cdisselkoen:
## Description of changes

This PR is a trial for deciding whether we want to go all-in on the recommendation in #878.  It applies the recommendation in #878 to `EvaluationError` only.

- Ergonomics of matching on errors are slightly worse, as we see in the tests in this PR.
- Ergonomics for internal developers are slightly worse, as we see in a couple places: 
    - for one example, `IntegerOverflowError` is `Into<EvaluationError>`, but is not `Into<Box<EvaluationError>>`, so the `.into()` in places like evaluator.rs:360-ish no longer works automatically.  We could re-enable the `.into()` by writing the `impl From<IntegerOverflowError> for Box<EvaluationError>`, but this PR currently just makes the `Box::new` explicit instead.
    - this situation is slightly worse in places where we rely on `?` to apply `Into::into` automatically: for instance, around evaluator.rs:172 where we not only need to add explicit `Box::new` but also a `.into()` call.  The issue is that types like `ExtensionFunctionLookupError` are throwable with `?` in functions returning `EvaluationError`, but are not throwable with `?` in functions returning `Box<EvaluationError>`, because of the missing `Into` impl.  We also can't write the blanket `impl<T: Into<EvaluationError>> From<T> for Box<EvaluationError>` because it conflicts with the built-in blanket `impl From<T> for Box<T>`, and `EvaluationError` is one of the `impl Into<EvaluationError>` types.  We could manually write the `From` impl for each error struct individually, but this PR just puts up with the extra `.into()` and `Box::new` at the `?` site.
- We should benchmark whether there are any noticeable performance differences (either on happy paths or error paths).

## Issue #, if available

Partially addresses #878 

### DIFF
--- a/cedar-policy-cli/src/lib.rs
+++ b/cedar-policy-cli/src/lib.rs
@@ -627,7 +627,9 @@ pub fn evaluate(args: &EvaluateArgs) -> (CedarExitCode, EvalResult) {
             }
         },
     };
-    match eval_expression(&request, &entities, &expr).wrap_err("failed to evaluate the expression")
+    match eval_expression(&request, &entities, &expr)
+        .map_err(|boxed| *boxed)
+        .wrap_err("failed to evaluate the expression")
     {
         Err(e) => {
             println!("{e:?}");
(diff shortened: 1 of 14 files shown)


andrewmwells-amazon:
My vote is to keep the status quo until people complain.

cdisselkoen:
After offline discussion, we will not take this change right now, as it's unclear it's worth it, and completing #878 for all the error types looks like a pretty large effort.  We'll leave #878 open as a backlog issue in case we are looking to improve performance in the future and identify #878 as relevant to performance metrics we care about.

PR #946, opened 2024-06-03
Remove `RestrictedExpressionError` from `EvalutationError`

### LINKED ISSUE #187: Separate `RestrictedExpressionError` from `EvaluationError`
### Category

Other

A `RestrictedExpressionError` is a kind of `EvaluationError` that is only reported when evaluating a restricted expression. Because it is included in the main `EvaluationError` enum, anyone evaluating an expression and matching on the result would need to know to ignore this case.

As far as I can tell, `RestrictedExpressionError` is only included as a variant of `EvaluationError` due to the structure of [a few test cases](https://github.com/cedar-policy/cedar/blob/main/cedar-policy-core/src/evaluator.rs#L3808), so we should be able to modify those test case and then separate these error types.

john-h-kastner-aws:
Fix #187

Turns out to be an easy fix, so good to get in for the 4.0 error refactoring.

### DIFF
--- a/cedar-policy-core/src/ast/restricted_expr.rs
+++ b/cedar-policy-core/src/ast/restricted_expr.rs
@@ -633,26 +633,6 @@ impl Diagnostic for RestrictedExprError {
     }
 }
 
-impl RestrictedExprError {
-    /// Get the `Loc` of this error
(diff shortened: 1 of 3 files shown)


khieta (approved):
Nice.

cdisselkoen (approved):

Issue #500, opened 2023-12-14
`Schema` should have a `.to_json()` method


cdisselkoen:
### Category

User level API changes

### Describe the feature you'd like to request

Cedar currently has a way to parse schemas from JSON, but should also have a way to serialize schemas back to JSON.

### Describe alternatives you've considered

N/A

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

oflatt:
I gave this a shot in #949, but it looks like we would need to convert back to a `cedar_policy_validator::SchemaFragment` and then serialize.

john-h-kastner-aws:
Copying a note from the linked PR:

> Perhaps a better solution would to take an approach similar to the LosslessPolicy struct we use when serializing policies.

In essence this would mean we store the `serde_json::Value` used to build the schema so we can regurgitate that when asked for it. If the schema was parsed from the human format, then we would store the schema source text, and we do have a function to take the human schema syntax tree and convert it to JSON.

cdisselkoen:
Some notes on this:
1. We do already have a `.to_json_value()` and `.to_json_string()` for `SchemaFragment`.
2. It would be nice to have, not only a `.to_json_*()` on `Schema`, but also a `Display` impl that prints the `Schema` in Cedar format.
3. Possibly the best way to implement both of these at once would be to simply provide a method that converts `Schema` into `SchemaFragment`, i.e., constructs a single `Fragment` that contains all of the information from the (possibly multi-fragment) `Schema`.

cdisselkoen:
See also #682

PR #949, opened 2024-06-03
Expose json serialization of schemas

oflatt:
## Description of changes
As far as I can tell, the json format for schemas is stable. So, this PR exposes methods to print and pretty-print json strings from schemas. I ran into this while writing my first cedar code in rust.

### DIFF
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -48,6 +48,7 @@ use cedar_policy_validator::RequestValidationError; // this type is unsuitable f
 use itertools::{Either, Itertools};
 use miette::Diagnostic;
 use ref_cast::RefCast;
+use schema_error::JsonSerializationError;
 use smol_str::SmolStr;
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::io::Read;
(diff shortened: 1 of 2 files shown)


cdisselkoen:
This fixes existing issue #500.  I think we should consider some of the questions mentioned in #500 before merging this (because changing the behavior later might be difficult or painful for folks relying on it).  Specifically, it would be great to leave default-valued fields implicit, rather than explicitly writing the value of every default-valued field, which I think this implementation will do (although I could be wrong).

cedar-policy/src/api.rs
```diff
@@ -1382,6 +1383,18 @@ impl Schema {
         ))
     }
 
+    /// Serialize the `Schema` to a JSON value.
+    pub fn to_json_str(&self) -> Result<String, schema_error::SchemaError> {
+        serde_json::to_string(&self.0)
```
john-h-kastner-aws, line 1388: Unfortunately this will not give the public JSON schema format. IIRC the `Serialize` impl for `cedar_policy_validator::ValidatorSchema` is only to serialize the schema to send it to the Lean.

This function would need to first convert back to a `cedar_policy_validator::SchemaFragment` and then serialize. We don't have any way to do that conversion atm. Perhaps a better solution would to take an approach similar to the `LosslessPolicy` struct we use when serializing policies.

john-h-kastner-aws (changes requested):

cedar-policy/src/api.rs
cdisselkoen: Actually, along these lines, we should also have at least one test for each of the new public methods, just to sanity-check that we're getting the results we expect.  (We can also add the JSON serialization to the roundtrip PBT targets, but that would obviously be a separate PR to `cedar-spec`.)

oflatt:
Given the scope of changes needed to make this work, closing this PR for now. Updated #500

PR #954, opened 2024-06-04
Revise APIs of `PrincipalOrResourceConstraint`

shaobo-he-aws:
## Description of changes

* Replace `PrincipalOrResourceConstraint::iter_euids` with `PrincipalOrResourceConstraint::get_euid`
  * The original function returns an iterator that works on at most one element. So, this PR replaces it with a function returning an `Option` so that we can pattern match the return value more conveniently.
* Wrap `Name` and `EntityUID` with `Arc`
   * They are mostly read only but can contain lots of bytes

### DIFF
--- a/cedar-policy-core/src/ast/policy.rs
+++ b/cedar-policy-core/src/ast/policy.rs
@@ -1567,23 +1567,17 @@ impl PrincipalOrResourceConstraint {
         }
     }
 
-    /// Get an iterator over all of the entity uids in this constraint.
-    pub fn iter_euids(&'_ self) -> impl Iterator<Item = &'_ EntityUID> {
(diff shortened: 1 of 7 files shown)


john-h-kastner-aws (approved):
lgtm

cdisselkoen (approved):
Looks good.  `Name` and `EntityUID` were already O(1) clone (they already contain `Arc`s internally I believe), but I'm willing to believe this `Arc` is still a good idea

andrewmwells-amazon (approved):

andrewmwells-amazon:
Looks like the downstream DRT will need to be fixed but the Java failure on Mac seems to be a glitch.

PR #956, opened 2024-06-05
Removing constant folding of `!`

aaronjeline:
## Description of changes
Removes the constant folding `!` in the CST->AST parser.
The motivation for this change is that the CST->AST parser does this simplification, but the EST->AST parser does not. We'd like these parsers to line up, so we either need to remove it from one or add it to the other. I'm voting for removing it for simplicity, having all these little simplifications throughout the parser bugs me. They are hard to audit/verify and make sure we've gotten them right, and hard to ensure consistency between our different parsers. (It's bitten us here and in the past)
## Issue #, if available
N/A

### DIFF
--- a/cedar-policy-core/src/parser/cst_to_ast.rs
+++ b/cedar-policy-core/src/parser/cst_to_ast.rs
@@ -1235,31 +1235,19 @@ impl Node<Option<cst::Unary>> {
     fn to_expr_or_special(&self) -> Result<ExprOrSpecial<'_>> {
         let unary = self.try_as_inner()?;
 
-        // A thunk to delay the evaluation of `item`
-        let maybe_item = || unary.item.to_expr_or_special();
-
         match unary.op {
-            None => maybe_item(),
-            Some(cst::NegOp::Bang(0)) => maybe_item(),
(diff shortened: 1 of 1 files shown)


cedar-policy-core/src/parser/cst_to_ast.rs
john-h-kastner-aws, line 1: Have you looked at what we're doing for `-`?

john-h-kastner-aws (approved):
I like this change.

Some additional context: This difference was found by the `convert-policy-cedar-to-json` fuzz target. For the target to pass we need to have the EST and CST parsing paths produce the same AST, so either folding `!` or not. The transformation feels unnecessary, and I would prefer not to duplicate it and maintain it in two places, so removing it in the CST->AST function seems appropriate.

cedar-policy-core/src/parser/cst_to_ast.rs
aaronjeline: Yeah, `-` doesn't fold

cdisselkoen (approved):
This makes me wish we had a generic AST->AST constant folding pass though

PR #960, opened 2024-06-06
Add entities visualizer to CLI

oflatt:
## Description of changes
This PR adds a new command line option to the Cedar CLI interface called `visualize`. It accepts an entity json and visualizes it using the graphviz `dot` format. I made it for personal use, but I'm putting up a PR in case we want it to be a supported feature. It also lacks testing currently.

### DIFF
--- a/cedar-policy-cli/src/lib.rs
+++ b/cedar-policy-cli/src/lib.rs
@@ -99,6 +99,8 @@ pub enum Commands {
     TranslatePolicy(TranslatePolicyArgs),
     /// Translate JSON schema to natural schema syntax and vice versa (except comments)
     TranslateSchema(TranslateSchemaArgs),
+    /// Visualize a set of JSON entities to the graphviz format.
+    Visualize(VisualizeArgs),
     /// Create a Cedar project
     New(NewArgs),
 }
@@ -379,6 +381,12 @@ pub struct AuthorizeArgs {
     pub timing: bool,
 }
 
+#[derive(Args, Debug)]
+pub struct VisualizeArgs {
+    #[arg(long = "entities", value_name = "FILE")]
+    pub entities_file: String,
+}
+
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq, ValueEnum)]
 pub enum PolicyFormat {
     /// The standard human-readable Cedar policy format, documented at <https://docs.cedarpolicy.com/policies/syntax-policy.html>
@@ -653,6 +661,25 @@ pub fn link(args: &LinkArgs) -> CedarExitCode {
         CedarExitCode::Failure
     } else {
         CedarExitCode::Success
+    }
+}
+
+pub fn visualize(args: &VisualizeArgs) -> CedarExitCode {
+    match load_entities(&args.entities_file, None) {
+        Ok(entities) => match entities.to_dot_str() {
+            Ok(dot_str) => {
+                println!("{}", dot_str);
+                CedarExitCode::Success
+            }
+            Err(e) => {
+                eprintln!("{}", e);
+                CedarExitCode::Failure
+            }
+        },
+        Err(e) => {
+            eprintln!("{}", e);
+            CedarExitCode::Failure
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


oflatt:
I'm concerned that if we expose this visualizer to customers they will think it's not polished enough.
For example, here's a result I just got. It calculates the transitive closure instead of just showing the parents in the data:

entities

oflatt:
Conclusion from standup: We will provide this feature, but with documentation that says it is best effort and not well tested.

cedar-policy-cli/src/lib.rs
```diff
@@ -649,6 +658,25 @@ pub fn link(args: &LinkArgs) -> CedarExitCode {
...
 
+pub fn visualize(args: &VisualizeArgs) -> CedarExitCode {
+    match load_entities(&args.entities_file, None) {
+        Ok(entities) => match entities.to_dot_str() {
+            Ok(dot_str) => {
+                println!("{}", dot_str);
+                CedarExitCode::Success
+            }
+            Err(e) => {
+                eprintln!("{}", e);
```
cdisselkoen, line 669: Let's follow the pattern set elsewhere in this file and print as a `miette::Report` in Debug format.
```suggestion
                let report = miette::Report::new(e);
                eprintln!("{report:?}");
```

cedar-policy-cli/src/lib.rs
```diff
@@ -649,6 +658,25 @@ pub fn link(args: &LinkArgs) -> CedarExitCode {
...
+                CedarExitCode::Failure
+            }
+        },
+        Err(e) => {
+            eprintln!("{}", e);
```
cdisselkoen, line 674: Same here

cdisselkoen (approved):

cedar-policy-cli/src/lib.rs
oflatt: On closer reading of the code, this error should actually never occur. The `write_str` in the dot code should always succeed I believe.
What do we do with errors that should not occur? I can make a new `EntitiesError` variant, or re-write the code to not use `write_str` and instead use a vec of strings and a join at the end.

cdisselkoen: I think it's fine to have an error that "should never occur" in this case, since it's easy to report it correctly in the hypothetical case that it did occur.  But rewriting to avoid `write_str` would be fine too.

john-h-kastner-aws: Any reason not to switch it to using `push_str` so we can get rid of the `Result`? I'd like that better if we're adding it to our public API.

cedar-policy/src/api.rs
```diff
@@ -712,6 +712,14 @@ impl Entities {
     pub fn write_to_json(&self, f: impl std::io::Write) -> std::result::Result<(), EntitiesError> {
         self.0.write_to_json(f)
     }
+
+    #[doc = include_str!("../experimental_warning.md")]
+    /// Visualize an `Entities` object in the graphviz `dot`
+    /// format. Entity visualization is best-effort and not well tested.
+    /// Feel free to submit an issue if you are using this feature and would like it improved.
+    pub fn to_dot_str(&self) -> std::result::Result<String, std::fmt::Error> {
+        self.0.to_dot_str()
```
john-h-kastner-aws, line 721: I wonder if `to_dot_str()` is implementable in using just our public API. It probably should be.

john-h-kastner-aws (approved):
Looks good, but would like to see a (hopefully) small refactoring of `to_dot_str` to remove the `Result` before making it public.

cedar-policy-cli/src/lib.rs
oflatt: Good idea! Done

cedar-policy/src/api.rs
oflatt: Good question, I think it is. I'll leave that to a future refactor.

john-h-kastner-aws (approved):

cdisselkoen (approved):

PR #963, opened 2024-06-07
`ParseError` cleanup

### LINKED ISSUE #564: Invalid constraint operator error should distinguish between action and principal/resource constraint
### Category

Other

With the introduction of `is`, the `action` scope constraint accepts fewer operator than principal an resource constraints, but this error always tells you that `` policy scope constraints must either `==`, `in`, `is`, or `_ is _ in _` ``.

The `help` in the following example should be something like `` action policy scope constraints must either `==` or `in` ``

```
jkastner@88665a03aa97 cedar % cargo run check-parse <<<'permit(principal, action != Action::"foo", resource);'
    Finished dev [unoptimized + debuginfo] target(s) in 0.65s
     Running `target/debug/cedar check-parse`
  × failed to parse policy set
  ╰─▶ not a valid policy scope constraint: !=
   ╭─[<stdin>:1:1]
 1 │ permit(principal, action != Action::"foo", resource);
   ·                   ───────────────────────
   ╰────
  help: policy scope constraints must either `==`, `in`, `is`, or `_ is _ in _`
```

.

### Additional context

_No response_

khieta:
## Description of changes

As I was starting to work on #745 for `ParseError`, I noticed several small things to fix (summarized below). I put these changes in their own PR so that my next PR will be quick to review.
* Removed some error variants that were unused, or that were better captured by a different error variant.
* Re-used some `ParseError` structures in `FromJsonError` in an effort to keep error messages consistent. 
* Removed a chunk of outdated code from `cedar-testing/tests/cedar-policy-cli/main.rs` -- we no longer save ill-formed tests into the corpus.
* Various nits in doc comments and error messages.

Sorry for the large PR (again) 😊

### Details about errors combined/removed

In `ToASTErrorKind`:
* `MissingScopeConstraint` **renamed** to `MissingScopeVariable`
* `ExtraScopeConstraints` **renamed** to `ExtraScopeElement`
* `InvalidScopeConstraintVariable` **renamed** to `InvalidScopeVariable`
* `InvalidConstraintOperator` **renamed** to `InvalidScopeOperator`
* `InvalidActionScopeOperator` **added** to resolve #564
* `InvalidMethodName` **renamed** to `UnknownMethod`
* `NotAFunction` **renamed** to `UnknownFunction`
* `IsInvalidName` **renamed** to `InvalidIsType`
* `AnnotationInvariantViolation` **removed**  because it was unused
* `InvalidAttributesInRecordLiteral` **removed** because it was always returned along with another error, and doesn't add any useful addition information IMO. See tests in `cst_to_ast.rs` for an example of how this changes error output.
* `UnsupportedEntityLiterals` **renamed** to `InvalidEntityLiteral`
* `InvalidPath` **removed** because it could never be returned (see comment on the `maybe_to_var` function in `cst_to_ast.rs`). I don't think it was adding anything useful regardless.
* `HasNonLiteralRHS` **removed** because it was unused
* `InvalidExpression` **removed** because it was redundant with `ArbitraryVariable` (the removed variant is what was used in the EST code)
* `InvalidIs` (which had two variants) **replaced** with `IsInActionScope` and `IsWithEq`
* `RefCreation` **renamed** to `WrongEntityArgument`

In `FromJsonError`:
* `MultiplicationByNonConstant` **removed** because it was unused
* `UnknownExtFunc` **renamed** to `UnknownExtensionFunction`

_Note:_ I believe that none of the changes made in this PR are to publicly exposed structures. Regardless, this PR belongs with the rest of the error refactoring in v4.0.


cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -532,9 +528,6 @@ impl Node<Option<cst::VariableDef>> {
             let eref = rel_expr.to_ref_or_slot(var)?;
             match (op, &vardef.entity_type) {
                 (cst::RelOp::Eq, None) => Ok(PrincipalOrResourceConstraint::Eq(eref)),
-                (cst::RelOp::Eq, Some(_)) => Err(self.to_ast_err(ToASTErrorKind::InvalidIs(
```
john-h-kastner-aws, line 535: This will lose a useful help message for some errors.

khieta: Here's an example of the change:
```
msg: "`is` cannot appear in the scope at the same time as `==`"
help: "try moving `is` into a `when` condition"
```
```
msg: "invalid policy scope constraint",
help: "policy scope constraints can only include `==`, `in`, `is`, or `_ is _ in _`"
```

My feeling was that the new message was as helpful as the old one, but happy to be overruled. Maybe it would be best to combine the original error message with the new help text?

john-h-kastner-aws: I feel like it's worth having a bit of extra code to specialize the message of `is` since it can sometimes appear in the scope

cedar-policy-core/src/parser/err.rs
```diff
@@ -129,139 +129,132 @@ impl ToASTError {
     }
 }
 
+const POLICY_SCOPE_HELP: &str =
+    "policy scopes must contain a `principal`, `action`, and `resource` element in that order";
+
 /// Details about a particular kind of `ToASTError`.
 #[derive(Debug, Diagnostic, Error, Clone, PartialEq, Eq)]
 #[non_exhaustive]
 pub enum ToASTErrorKind {
```
john-h-kastner-aws, line 138: Can you list out which error variants you removed and which you combined? Kind of hard to parse out of the diff.

khieta: Sorry for being lazy with my original notes 😄 Added the requested info to the PR description.

john-h-kastner-aws: thanks

cedar-policy-core/src/parser/cst_to_ast.rs
john-h-kastner-aws: Also, #564 if it's easy to fix while you're in this code messing about with that error variant

khieta: Nice call out. I'll make the changes in this thread on Monday.

khieta: Had some time today. Here are examples of error messages in the latest commit:

```
permit(principal > User::"alice", action, resource);
```
**Error message:** invalid operator in the policy scope: >
**Help text:** policy scope clauses can only use `==`, `in`, `is`, or `_ is _ in _`

```
permit(principal, action != Action::"view", resource);
```
**Error message:** invalid operator in the action scope: !=
**Help text:** action scope clauses can only use `==` or `in`

```
permit(principal is User == User::"Alice", action, resource);
```
**Error message:** `is` cannot be used together with `==`
**Help text:** try using `_ is _ in _`

```
permit(principal, action is Action, resource);
```
**Error message:** `is` cannot appear in the action scope
**Help text:** try moving `action is ..` into a `when` condition

john-h-kastner-aws: Looks good!

cedar-policy-core/src/parser/err.rs
```diff
@@ -369,14 +364,10 @@ pub enum ToASTErrorKind {
...
-    /// Returns when a policy scope has incorrect EntityUIDs/Template Slots
-    #[error(transparent)]
-    #[diagnostic(transparent)]
-    RefCreation(#[from] RefCreationError),
-    /// Returned when an `is` appears in an invalid position in the policy scope
+    /// Returned when a policy scope has incorrect entity uids or template slots
     #[error(transparent)]
     #[diagnostic(transparent)]
-    InvalidIs(#[from] InvalidIsError),
+    RefCreation(#[from] parse_errors::RefCreation),
```
john-h-kastner-aws, line 370: `RefCreation` has never made much sense as an error name to me. Maybe change if you have a good idea for something better.

john-h-kastner-aws (approved):

cedar-policy-core/src/parser/err.rs
khieta: The name comes from the `cst::Ref` type that describes a "reference to an entity". But we don't use this terminology anywhere else, so I agree that it's confusing as an error name. Changed in the latest commit to `WrongEntityArgument` since it (at least currently) can only be returned when a scope clause uses incorrect entity arguments (e.g., a template slot of an entity uid). I left the remaining "ref" notation in `cst_to_ast.rs` unchanged avoid exploding the size of the diff. Worth making an issue to fix later?

cedar-policy-core/src/parser/err.rs
```diff
@@ -405,74 +396,105 @@ impl ToASTErrorKind {
...
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub enum Ref {
-    /// A single entity uids
-    Single,
-    /// A list of entity uids
-    Set,
-    /// A template slot
-    Template,
-}
+/// Error subtypes for `ToASTErrorKind`
```
cdisselkoen, line 430:
```suggestion
/// Error subtypes for [`ToASTErrorKind`]
```

cedar-testing/tests/cedar-policy-cli/main.rs
```diff
@@ -82,32 +80,6 @@ fn perform_integration_test_from_json(jsonfile: impl AsRef<Path>) {
...
-            assert!(
-                parse_errs
-                    .iter()
-                    .any(|e| e.to_string().ends_with("not a function")),
-                "unexpected parse errors in test {}: {}",
-                jsonfile.display(),
-                parse_errs,
-            );
-            continue;
-        };
```
cdisselkoen, line 110: Nice.  Does this mean that programmatically generated policies are now guaranteed to be parseable?

cdisselkoen (approved):

cedar-testing/tests/cedar-policy-cli/main.rs
khieta: Afaik, the issue still exists in our generators. The difference is that we no longer save invalid policy files to the generated integration tests. See this function: https://github.com/cedar-policy/cedar-spec/blob/8b6aa6956ca0720de51c4dbe710d1b162648f1e0/cedar-drt/fuzz/src/dump.rs#L196

cdisselkoen (approved):

PR #968, opened 2024-06-10
Improve parser tests

### LINKED ISSUE #739: Update old parse error tests to check that the correct error is reported

Some of our parse error tests assert that an expression fails to parse, but do not assert that it fails with the correct error kind.  We previously had an issue where a regression in the json schema parser was not noticed because of a test written in this way. We updated those tests, but we should also update these to avoid a similar regression in policy parsing.

For example, here we assert that exactly one error is raised, but we could assert that the error is due to `1` not being a valid attribute.
https://github.com/cedar-policy/cedar/blob/0f9df9589146900877c0508eb684ca9d4ab1d137/cedar-policy-core/src/parser/cst_to_ast.rs#L3340-L3353

_No response_

### Additional context

_No response_

khieta:
## Description of changes

Continuation of my nit-picking in the parsing code. This PR:
* Checks the error message for any tests that expect errors (#739). Most of the work here was in the `text_to_cst.rs` file, but there were a couple additions to `parser.rs` as well.
* Updates the error testing infrastructure to consider label text, which is where message like ``` "expected `&&`, `||`, or `}`" ``` end up
* Moves some tests from `cst_to_ast.rs` to `test_to_cst.rs`, since the relevant tests were actually testing functionality in that file.
* Flattens tests in `parser.rs` into a single `tests` module.

**TLDR;** This code does not add new tests or delete old tests -- it just improves current tests and moves things around.

## Issue #, if available

Resolves #739 


john-h-kastner-aws:
Test updates look good, but IMO many of the tests moved to `parser.rs` are testing `cst_to_ast.rs` logic.

john-h-kastner-aws:
>  If we leave them in that file, then I think we should switch to the lower-level functions. Would you prefer that over moving the tests?

I don't see a need to change the tests at all. Switching to lower-level functions is definitely a waste of time. Moving the tests at least doesn't take any time, but it churn in the code without  a benefit in my opinion.

khieta:
Ok, new version pushed with my picky edits removed 😄

john-h-kastner-aws (approved):

shaobo-he-aws (approved):
LGTM

PR #970, opened 2024-06-11
Expose an opaque `ParseError` type

### LINKED ISSUE #745: [4.0] Consistent public error types in the Rust API

khieta:
## Description of changes

This PR addresses #745 for `ParseErrors` by exposing the bare minimum level of detail: users can use the `miette::Diagnotic` attached to `ParseErrors`, or use `.iter()` to get an iterator over individual `ParseError`s. Each `ParseError` will expose no internal information outside of its `miette::Diagnostic`.

I think the ideal solution here is to flatten the structure of `ParseError` to remove the distinction between CST and AST errors and get rid of the "Kind" pattern, similar to what has been done for the validation & evaluation errors... but this is a large amount of work for small benefit (I find it unlikely that users will need to programmatically match against parse errors). Assuming this PR is approved, I'll make a backlog issue for the larger change. Note that this larger change can be made in a backwards-compatible way in the future.

_(Part of the batch of breaking error structure changes for 4.0)_

_(Adds an undocumented breaking change made in #908.)_

### DIFF
--- a/cedar-policy/src/api/err.rs
+++ b/cedar-policy/src/api/err.rs
@@ -24,11 +24,11 @@ pub use cedar_policy_core::evaluator::{evaluation_errors, EvaluationError};
 pub use cedar_policy_core::extensions::{
     extension_function_lookup_errors, ExtensionFunctionLookupError,
 };
-pub use cedar_policy_core::parser::err::{ParseError, ParseErrors};
 use cedar_policy_core::{ast, authorizer, est};
 pub use cedar_policy_validator::human_schema::{schema_warnings, SchemaWarning};
 pub use cedar_policy_validator::{schema_errors, SchemaError};
 use miette::Diagnostic;
+use ref_cast::RefCast;
 use smol_str::SmolStr;
 use thiserror::Error;
 
@@ -730,6 +730,30 @@ impl From<ast::UnexpectedSlotError> for PolicySetError {
         Self::ExpectedStatic(policy_set_errors::ExpectedStatic::new())
     }
 }
+
+/// Represents one or more [`ParseError`]s encountered when parsing a policy or
+/// expression.
+/// By default, the `Diagnostic` and `Error` implementations will only print the
+/// first error. If you want to see all errors, use `.iter()` or `.into_iter()`.
+#[derive(Debug, Diagnostic, Error)]
+#[error(transparent)]
+#[diagnostic(transparent)]
+pub struct ParseErrors(#[from] cedar_policy_core::parser::err::ParseErrors);
+
+impl ParseErrors {
+    /// Get every [`ParseError`] associated with this [`ParseErrors`] object.
+    /// The returned iterator is guaranteed to be nonempty.
+    pub fn iter(&self) -> impl Iterator<Item = &ParseError> {
+        self.0.iter().map(ParseError::ref_cast)
+    }
+}
+
+/// Errors that can occur when parsing policies or expressions.
+#[derive(Debug, Diagnostic, Error, RefCast)]
+#[repr(transparent)]
+#[error(transparent)]
+#[diagnostic(transparent)]
+pub struct ParseError(#[from] cedar_policy_core::parser::err::ParseError);
 
 /// Errors that can happen when getting the JSON representation of a policy
 #[derive(Debug, Diagnostic, Error)]
--- a/cedar-policy/src/api/err.rs
+++ b/cedar-policy/src/api/err.rs
@@ -749,11 +749,17 @@ impl ParseErrors {
 }
 
 /// Errors that can occur when parsing policies or expressions.
+/// Marked as `non_exhaustive` to support adding additional error information
+/// in the future without a major version bump.
 #[derive(Debug, Diagnostic, Error, RefCast)]
 #[repr(transparent)]
 #[error(transparent)]
 #[diagnostic(transparent)]
-pub struct ParseError(#[from] cedar_policy_core::parser::err::ParseError);
+#[non_exhaustive]
+pub struct ParseError {
+    #[from]
+    inner: cedar_policy_core::parser::err::ParseError,
+}
 
 /// Errors that can happen when getting the JSON representation of a policy
 #[derive(Debug, Diagnostic, Error)]
--- a/cedar-policy-core/src/parser/err.rs
+++ b/cedar-policy-core/src/parser/err.rs
@@ -45,10 +45,6 @@ pub(crate) type RawErrorRecovery<'a> = lalr::ErrorRecovery<RawLocation, RawToken
 type OwnedRawParseError = lalr::ParseError<RawLocation, String, RawUserError>;
 
 /// Errors that can occur when parsing Cedar policies or expressions.
-//
-// CAUTION: this type is publicly exported in `cedar-policy`.
-// Don't make fields `pub`, don't make breaking changes, and use caution when
-// adding public methods.
 #[derive(Clone, Debug, Diagnostic, Error, PartialEq, Eq)]
 pub enum ParseError {
     /// Error from the text -> CST parser
@@ -696,10 +692,6 @@ pub fn expected_to_string(expected: &[String], config: &ExpectedTokenConfig) -> 
 
 /// Represents one or more [`ParseError`]s encountered when parsing a policy or
 /// template.
-//
-// CAUTION: this type is publicly exported in `cedar-policy`.
-// Don't make fields `pub`, don't make breaking changes, and use caution when
-// adding public methods.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct ParseErrors(NonEmpty<ParseError>);
 
--- a/cedar-policy/CHANGELOG.md
+++ b/cedar-policy/CHANGELOG.md
@@ -37,6 +37,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the rich data provided by `miette::Diagnostic`, for instance `.help()` and
   `labels()`. Callers can continue using the same behavior by calling
   `.iter().map(ToString::to_string)`. (#882, resolving #543)
+- Removed `ParseError::primary_source_span`. Callers should use the location
+  information provided by `miette::Diagnostic` via `.labels()` and
+  `.source_code()` instead. (#908)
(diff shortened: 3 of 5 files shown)


cedar-policy/src/api/err.rs
```diff
@@ -731,6 +731,30 @@ impl From<ast::UnexpectedSlotError> for PolicySetError {
...
+        self.0.iter().map(ParseError::ref_cast)
+    }
+}
+
+/// Errors that can occur when parsing policies or expressions.
+#[derive(Debug, Diagnostic, Error, RefCast)]
+#[repr(transparent)]
+#[error(transparent)]
+#[diagnostic(transparent)]
+pub struct ParseError(#[from] cedar_policy_core::parser::err::ParseError);
```
john-h-kastner-aws, line 756: Regarding exposing `ParseError` as an enum in the future, we should fine based on https://predr.ag/blog/turning-rust-struct-to-enum-is-not-always-breaking/ (by the author of cargo-semver-checks, so I assume they know what they're talking about)

john-h-kastner-aws (approved):
Looks good to me

shaobo-he-aws (approved):

shaobo-he-aws (approved):

cedar-policy/src/api/err.rs
```diff
@@ -731,6 +731,36 @@ impl From<ast::UnexpectedSlotError> for PolicySetError {
...
+    /// Get every [`ParseError`] associated with this [`ParseErrors`] object.
+    /// The returned iterator is guaranteed to be nonempty.
+    pub fn iter(&self) -> impl Iterator<Item = &ParseError> {
+        self.0.iter().map(ParseError::ref_cast)
+    }
+}
+
+/// Errors that can occur when parsing policies or expressions.
+/// Marked as `non_exhaustive` to support adding additional error information
+/// in the future without a major version bump.
```
cdisselkoen, line 753: I don't think we technically need `non_exhaustive` for that, since the fields are already private, but it doesn't hurt

cdisselkoen (approved):

PR #980, opened 2024-06-13
Improve parse error for inverting `in` and `is`

### LINKED ISSUE #755: Better help message for error `principal in Group is Type`
### Category

Error message improvements

For the following policy:
```
permit(principal in Group::"friends" is User, action, resource);
```
The error message is currently:
```
expected an entity uid or matching template slot, found an `is` expression
```
with no help message.  We should have a help message that suggests `principal is User in Group::"friends"`.

Status quo is not terrible.

### Additional context

_No response_

khieta:
## Description of changes

Resolving an easy parsing-related issue in penance for opening some other issues.

Adds custom error text in the case where a policy uses `_ in _ is _` instead of `_ is _ in _` in the policy scope. This pattern is harder to address in the policy condition because it is currently rule out by the lalrpop parser, and updating the parser to support this pattern (for the purpose of returning a more useful error) is nontrivial.

This PR adds some extra work to parsing the policy scope, which may hurt performance, but I suspect the impact will be small since scope clauses are typically small. We also have not typically considered the parser part of the critical path for performance.

## Issue #, if available

Resolves #755 (depending on what you consider to be the scope of this issue)

### DIFF
--- a/cedar-policy-core/src/parser/cst_to_ast.rs
+++ b/cedar-policy-core/src/parser/cst_to_ast.rs
@@ -588,12 +588,25 @@ impl Node<Option<cst::VariableDef>> {
 
         if let Some((op, rel_expr)) = &vardef.ineq {
             let action_constraint = match op {
-                cst::RelOp::In => match rel_expr.to_refs(ast::Var::Action)? {
-                    OneOrMultipleRefs::Single(single_ref) => {
-                        Ok(ActionConstraint::is_in([single_ref]))
+                cst::RelOp::In => {
+                    // special check for the syntax `_ in _ is _`
+                    match rel_expr.to_expr() {
+                        Ok(expr) => {
+                            if matches!(expr.expr_kind(), ast::ExprKind::Is { .. }) {
+                                return Err(rel_expr
+                                    .to_ast_err(ToASTErrorKind::IsInActionScope)
+                                    .into());
+                            }
+                        }
+                        Err(_) => (), // ignore
                     }
-                    OneOrMultipleRefs::Multiple(refs) => Ok(ActionConstraint::is_in(refs)),
-                },
+                    match rel_expr.to_refs(ast::Var::Action)? {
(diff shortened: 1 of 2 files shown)


cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -3840,11 +3863,24 @@ mod tests {
...
+                ExpectedErrorMessageBuilder::error("when `is` and `in` are used together, `is` must come first")
+                    .help("try `_ is _ in _`")
+                    .exactly_one_underline(r#"principal in Group::"friends" is User"#)
+                    .build(),
+            ),
+            (
+                r#"permit(principal, action in Group::"action_group" is Action, resource);"#,
+                ExpectedErrorMessageBuilder::error("`is` cannot appear in the action scope")
+                    .help("try moving `action is ..` into a `when` condition")
+                    .exactly_one_underline(r#"Group::"action_group" is Action"#)
```
cdisselkoen, line 3875: This underline is strange, because this text isn't part of the original policy. Does the test pass?

cdisselkoen (approved):

cedar-policy-core/src/parser/cst_to_ast.rs
cdisselkoen: The underline is still a little strange because the help text talks about `action is`, but `action` is not part of the underline

cdisselkoen: might be clearer to underline the whole `action` clause?

john-h-kastner-aws (approved):

cedar-policy-core/src/parser/cst_to_ast.rs
john-h-kastner-aws: Agreed

khieta: Done

john-h-kastner-aws (approved):

PR #989, opened 2024-06-17
Use public `ParseErrors` in the `Parse` variant of `RestrictedExpressionParseError`

cdisselkoen:
## Description of changes

Previously, we just re-exported Core's `RestrictedExpressionParseError`, so the `Parse` variant contained Core's `ParseErrors`. This PR does the appropriate wrapping so that the `Parse` variant contains the public `ParseErrors` instead.

## Issue #, if available

Mentioned in https://github.com/cedar-policy/cedar/issues/745#issuecomment-2168026032

### DIFF
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -3026,7 +3026,9 @@ impl FromStr for RestrictedExpression {
 
     /// create a `RestrictedExpression` using Cedar syntax
     fn from_str(expression: &str) -> Result<Self, Self::Err> {
-        ast::RestrictedExpr::from_str(expression).map(RestrictedExpression)
+        ast::RestrictedExpr::from_str(expression)
+            .map(RestrictedExpression)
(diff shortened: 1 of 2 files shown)


khieta (approved):
Looks fine. I didn't do this in #989 because I didn't think it mattered -- the contents of `RestrictedExpressionParseError::Parse` are private. But I suppose if we changed the public `ParseErrors` definition to be something other than a wrapper around the internal version, then it could lead to inconsistent error messages.

cdisselkoen:
I don't think it's private -- contents of an `enum` variant are always the same visibility as the enum itself (which is a main reason why we are doing the error-structs thing in the first place).  @john-h-kastner-aws called this out as visible from the public docs in https://github.com/cedar-policy/cedar/issues/745#issuecomment-2168026032.

khieta:
TIL

john-h-kastner-aws (approved):

PR #966, opened 2024-06-10
lock url to 2.5.0

john-h-kastner-aws:
## Description of changes

Fixes cargo-deny CI failure while we sort out license details

## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [ ] A change "invisible" to users (e.g., documentation, changes to "internal" crates like `cedar-policy-core`, `cedar-validator`, etc.)

I confirm that this PR (choose one, and delete the other options):

- [ ] Does not update the CHANGELOG because my change does not significantly impact released code.

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [ ] Does not require updates because my change does not impact the Cedar formal model or DRT infrastructure.

PR #990, opened 2024-06-17
Cherry-pick #966 to 3.2.x

cdisselkoen:
## Description of changes

#966 

Also updates `itertools` build dependency, matching the dependency version on `main`


khieta (approved):

khieta:
Fyi integration test failures are because the integration tests currently don't support prior releases. Here's an issue to track: https://github.com/cedar-policy/cedar-integration-tests/issues/5. But if the CI failures are too noisy, we could also disable these checks when merging to release branches.

john-h-kastner-aws (approved):

PR #983, opened 2024-06-13
Implement RFC 55

aaronjeline:
## Description of changes
(partially) Implements [RFC 55](https://github.com/cedar-policy/rfcs/blob/main/text/0055-remove-unspecified.md)
This does not implement the backwards compatibility functions, saving that for a second PR as this one is already large.
## Issue #, if available
N/A
## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [X] A breaking change requiring a major version bump to `cedar-policy` (e.g., changes to the signature of an existing API).

I confirm that this PR (choose one, and delete the other options):

- [x] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [X] Does not require updates because my change does not impact the Cedar formal model or DRT infrastructure.

PR #993, opened 2024-06-17
Replace internal `::new()` with `From`/`Into` for `EntityUid` and `EntityTypeName`

cdisselkoen:
## Description of changes

I thought this would be more consistent with how we handle other internal/external pairs.

Not breaking, it only removes `pub(crate)` methods.

### DIFF
--- a/cedar-policy/src/api/id.rs
+++ b/cedar-policy/src/api/id.rs
@@ -134,12 +134,6 @@ impl EntityTypeName {
     pub fn namespace(&self) -> String {
         self.0.name().namespace()
     }
-
-    /// Construct an [`EntityTypeName`] from the internal type.
-    /// This function is only intended to be used internally.
-    pub(crate) fn new(ty: ast::EntityType) -> Self {
-        Self(ty)
-    }
 }
 
 /// This `FromStr` implementation requires the _normalized_ representation of the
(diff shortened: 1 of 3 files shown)


cedar-policy/src/api/id.rs
```diff
@@ -160,6 +155,29 @@ impl std::fmt::Display for EntityTypeName {
...
+            ast::EntityType::Specified(name) => Ok(name.into()),
+            ast::EntityType::Unspecified => Err(UnspecifiedCannotBeConvertedToEntityTypeName),
+        }
+    }
+}
+
+#[doc(hidden)]
+#[derive(Debug, Error)]
+#[error("`Unspecified` cannot be converted to `EntityTypeName`")]
+pub struct UnspecifiedCannotBeConvertedToEntityTypeName;
```
khieta, line 179: Makes sense to wait to merge this until after #983 since that PR make the unspecified error irrelevant.

khieta (approved):
Seems fine, modulo fixes to CI (appears to be a `cargo fmt` issue?)

cedar-policy/src/api/id.rs
cdisselkoen: Yes, I will wait.  It also changes the contained type from `Name` to `EntityType` which breaks a bunch of the changes in this PR.

shaobo-he-aws (approved):

cdisselkoen:
I think this is ready for re-review, and that the CI failures for DRT and Java are due to #983.

cedar-policy/src/api/id.rs
cdisselkoen: Updated

shaobo-he-aws (approved):

john-h-kastner-aws (approved):

PR #996, opened 2024-06-18
various error cleanup

khieta:
## Description of changes

While working on writing tests for the FFI edits, I found I wanted to make some adjustments to the error messages. Then before I knew it, I had a bunch of changes in my local repo -- this PR contains a (harmless?) subset of them. My next PR will have more interesting changes, but requires this one to be merged first.
* Added `expect_err` checks to some existing tests to better keep track of how error messages are displayed.
* Added some new tests to `cedar-policy/src/test.rs` to check error behavior not covered by other tests (most tests are related to request validation & context creation).
* Deleted `JsonDeserializationError::ContextCreation` since it was unused.
* Deleted `FromJsonError::PolicySet`, `FromJsonError::Linking`, and `impl TryFrom<PolicySet> for ast::PolicySet` since they were unused & redundant with the [`cedar-policy` definitions](https://github.com/cedar-policy/cedar/blob/9eb19064ac0bb2c1081ad0300b5b27190a25ae16/cedar-policy/src/api.rs#L1683).
* Changed the error type of `typecheck_partial_value` to be `ExtensionFunctionLookupError`, which is more specific than the previous error type `GetSchemaTypeError`.
* Moved most of the content from the `entities` module from `api.rs` to `api/err.rs` so we can keep all the re-exported errors in one place.
* Removed some indirection from the public `ContextJsonError` type. I noticed that we had a version of `ContextCreationError` in the public API, but were using the internal version under the hood in `ContextJsonError`, which was leading to inconsistent error messages with changes I was making for the followup PR mentioned above.
* Minor fixes in error messages.

_(Breaking changes are only in unreleased code, I think)_

### DIFF
--- a/cedar-policy-core/src/est/policy_set.rs
+++ b/cedar-policy-core/src/est/policy_set.rs
@@ -51,9 +51,7 @@ impl PolicySet {
                     None
                 }
             })
-            .collect::<Vec<_>>()
-            .first()
-            .cloned();
+            .next();
 
         let maybe_link = self
             .links
@@ -69,9 +67,7 @@ impl PolicySet {
                     None
                 }
             })
-            .collect::<Vec<_>>()
-            .first()
-            .cloned();
+            .next();
 
         maybe_static_policy.or(maybe_link)
     }
@@ -90,9 +86,7 @@ impl PolicySet {
                     None
                 }
             })
-            .collect::<Vec<_>>()
-            .first()
-            .cloned()
+            .next()
     }
 }
 
--- a/cedar-policy-core/src/est/policy_set.rs
+++ b/cedar-policy-core/src/est/policy_set.rs
@@ -14,9 +14,7 @@
  * limitations under the License.
  */
 
-use super::FromJsonError;
 use super::Policy;
-use crate::ast;
 use crate::ast::EntityUID;
 use crate::ast::{PolicyID, SlotId};
 use crate::entities::json::err::JsonDeserializationErrorContext;
@@ -27,6 +25,7 @@ use std::collections::HashMap;
 
 /// An EST set of policies
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
+#[serde(rename_all = "camelCase")]
 pub struct PolicySet {
     /// The set of templates in a policy set
     pub templates: Vec<PolicyEntry>,
@@ -64,35 +63,6 @@ struct TemplateLinkContext;
 impl crate::entities::json::DeserializationContext for TemplateLinkContext {
     fn static_context() -> Option<JsonDeserializationErrorContext> {
         Some(JsonDeserializationErrorContext::TemplateLink)
-    }
-}
-
-impl TryFrom<PolicySet> for ast::PolicySet {
-    type Error = FromJsonError;
-
-    fn try_from(value: PolicySet) -> Result<Self, Self::Error> {
-        let mut ast_pset = ast::PolicySet::default();
-
-        for PolicyEntry { id, policy } in value.templates {
-            let ast = policy.try_into_ast_template(Some(id))?;
-            ast_pset.add_template(ast)?;
-        }
-
-        for PolicyEntry { id, policy } in value.static_policies {
-            let ast = policy.try_into_ast_policy(Some(id))?;
-            ast_pset.add(ast)?;
-        }
-
-        for Link {
-            id,
-            template,
-            slots: env,
-        } in value.links
-        {
-            ast_pset.link(template, id, env)?;
-        }
-
-        Ok(ast_pset)
     }
 }
 
--- a/cedar-policy-core/src/est/policy_set.rs
+++ b/cedar-policy-core/src/est/policy_set.rs
@@ -15,8 +15,8 @@
  */
 
 use super::Policy;
-use crate::ast::EntityUID;
-use crate::ast::{PolicyID, SlotId};
+use super::PolicySetFromJsonError;
+use crate::ast::{self, EntityUID, PolicyID, SlotId};
 use crate::entities::json::err::JsonDeserializationErrorContext;
 use crate::entities::json::EntityUidJson;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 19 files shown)


john-h-kastner-aws (approved):
Looks good.

A couple comments on some existing error messages.

cedar-policy-validator/src/coreschema.rs
```diff
@@ -307,11 +306,11 @@ pub enum RequestValidationError {
...
     },
-    /// Error computing the type of the `Context`; see the contained error type
-    /// for details about the kinds of errors that can occur
-    #[error("context is not valid: {0}")]
+    /// Error computing the type of the `Context`.
+    /// Currently the only possible error is due to an extension function lookup failure.
+    #[error("invalid context: {0}")]
     #[diagnostic(transparent)]
-    TypeOfContext(GetSchemaTypeError),
+    TypeOfContext(ExtensionFunctionLookupError),
```
cdisselkoen, line 313: I'm surprised that the other variants of `GetSchemaTypeError` aren't possible here.  Shouldn't most of them be possible to encounter when trying to typecheck a `Context`?

Also, even if they aren't possible today, does it seem possible that future changes could leave us wanting to be able to return the other variants of `GetSchemaTypeError` here?  This change would box us in from being able to return those other variants in the future.

cedar-policy-validator/src/types.rs
```diff
@@ -551,7 +551,7 @@ impl Type {
         &self,
         value: &PartialValue,
         extensions: Extensions<'_>,
-    ) -> Result<bool, GetSchemaTypeError> {
+    ) -> Result<bool, ExtensionFunctionLookupError> {
```
cdisselkoen, line 554: Same with these three changes

cedar-policy/src/tests.rs
```diff
@@ -4862,7 +4915,394 @@ mod policy_set_est_tests {
...
+            Some(EntityUid::from_strs("Resource", "resource")),
+            Context::empty(),
+            Some(&schema),
+        )
+        .unwrap_err();
+        expect_err(
+            "",
+            &Report::new(err),
+            &ExpectedErrorMessageBuilder::error(
+                r#"context `<first-class record with 0 fields>` is not valid for `Action::"action"`"#,
```
cdisselkoen, line 5143: it would probably be more helpful to say
```
context `{}` is not valid for `Action::"action"`
```

cdisselkoen (approved):

cedar-policy/src/tests.rs
```diff
@@ -4862,7 +4915,394 @@ mod policy_set_est_tests {
...
+            Context::from_json_value(json!({"foo": 123}), None)
+                .expect("context creation should have succeeded"),
+            Some(&schema),
+        )
+        .unwrap_err();
+        expect_err(
+            "",
+            &Report::new(err),
+            &ExpectedErrorMessageBuilder::error(
+                r#"context `<first-class record with 1 fields>` is not valid for `Action::"action"`"#,
```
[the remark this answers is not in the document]
khieta: Not sure. I agree it's a weird error message -- adding a test gives us a way to track for the future. I'm planning on opening an issue to improve. Will post a link here once I've done that.

cedar-policy-validator/src/coreschema.rs
cdisselkoen: What about the heterogeneous-set case?  do the public constructors for `Context` rule out heterogeneous sets?  I would be surprised if they do -- actually I think it's a bug if they do, since in unvalidated Cedar, heterogeneous sets should be allowed

cedar-policy/src/tests.rs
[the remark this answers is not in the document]
khieta: #998

cedar-policy/src/tests.rs
```diff
@@ -4862,7 +4915,394 @@ mod policy_set_est_tests {
...
+            Some(EntityUid::from_strs("Resource", "resource")),
+            Context::empty(),
+            Some(&schema),
+        )
+        .unwrap_err();
+        expect_err(
+            "",
+            &Report::new(err),
+            &ExpectedErrorMessageBuilder::error(
+                r#"principal type `Resource` is not valid for `Action::"action"`"#,
```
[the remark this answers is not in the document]
khieta: Added this suggestion to #998

cedar-policy/src/tests.rs
khieta: Punting to future work 🙂 #998

khieta:
By popular demand, I've moved the `est::PolicySet` parsing logic back to `est/policy_set.rs`. This ended up being more complicated than I expected since we need to maintain some invariants about the public `PolicySet` type. Careful review of the latest commit would be appreciated.

cedar-policy-validator/src/coreschema.rs
khieta: Different code paths. The heterogenous set error can only be returned when trying to infer the type of a set. This will never happen for context validation because the schema says exactly what type the set should contain. A heterogeneous set will cause `typecheck_restricted_expr` to return `Ok(false)`. (So no errors, but there's a problem with the type of the context.)

khieta: Which is also making me wonder: since we are masking other errors in the `typecheck_restricted_expr` function anyways, maybe we should just drop the `Result` type and return false in case of an `ExtensionFunctionLookupError`?

khieta: Reverted this change in the last commit. I think I can put up a PR later today with some improvements in this part of the code

shaobo-he-aws (approved):

cedar-policy-core/src/est/policy_set.rs
```diff
@@ -36,6 +35,67 @@ pub struct PolicySet {
...
+            .iter()
+            .filter_map(|entry| {
+                if &entry.id == id {
+                    Some(entry.policy.clone())
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>()
+            .first()
```
cdisselkoen, line 55:
```suggestion
            .next()
```

cedar-policy-core/src/est/policy_set.rs
```diff
@@ -36,6 +35,67 @@ pub struct PolicySet {
...
+                        let unwrapped_est_vals: HashMap<SlotId, EntityUidJson> =
+                            link.slots.iter().map(|(k, v)| (*k, v.into())).collect();
+                        template.link(&unwrapped_est_vals).ok()
+                    })
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>()
+            .first()
```
cdisselkoen, line 73:
```suggestion
            .next()
```

cedar-policy-core/src/est/policy_set.rs
```diff
@@ -36,6 +35,67 @@ pub struct PolicySet {
...
+            .iter()
+            .filter_map(|entry| {
+                if &entry.id == id {
+                    Some(entry.policy.clone())
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>()
+            .first()
```
cdisselkoen, line 94:
```suggestion
            .next()
```

cedar-policy-validator/src/coreschema.rs
```diff
@@ -310,7 +310,6 @@ pub enum RequestValidationError {
     /// Error computing the type of the `Context`; see the contained error type
     /// for details about the kinds of errors that can occur
     #[error("context is not valid: {0}")]
-    #[diagnostic(transparent)]
```
cdisselkoen, line 313: should we keep this?

cedar-policy/src/api.rs
```diff
@@ -1680,54 +1645,66 @@ impl FromStr for PolicySet {
...
-            let slots = slots
-                .into_iter()
-                .map(|(key, value)| (key.into(), EntityUid::new(value)))
-                .collect();
-            pset.link(PolicyId::new(template), PolicyId::new(id), slots)?;
-        }
-
-        Ok(pset)
+    fn from_est(est: &est::PolicySet) -> Result<Self, PolicySetError> {
+        let ast: ast::PolicySet = est.clone().try_into()?;
```
cdisselkoen, line 1649: If we're going to unconditionally `.clone()` the `est` in this function, I'd rather take an owned `est` (like the current code); callers can call `.clone()` if they need the clone, but if they don't need ownership after this call, we avoid the clone.  (And looking at the callers, in this case we can indeed avoid the clone)

cdisselkoen (approved):
Looks good, I agree the `.expect()`s are fine

cedar-policy-core/src/est/policy_set.rs
khieta: Thanks for the review lol. Figured I was doing something weird here.

cedar-policy/src/api.rs
cdisselkoen: I'd still take an owned argument, and consume it at the last time it's used in this function (i.e., when we're constructing the lossless policy).  Or is does its last use only require a reference?

khieta: Constructing the lossless policies only requires a reference. (Although behind the scenes every policy ends up cloned via `.get_template()` or `.get_policy()`.)

cdisselkoen: Ick. Ok, probably nothing to do in this PR; if the cloning becomes a performance problem, we'll address it then.

cdisselkoen (approved):

PR #1000, opened 2024-06-18
use `#[serde(rename_all = "camelCase")]` in more places

cdisselkoen:
## Description of changes

I think there's only one instance where this is a functional change and not just a refactor, which is for `ValidatorActionId::attribute_types`, but that struct is not part of the official schema JSON format (do we only use the `Serialize`/`Deserialize` for DRT?).  I can revert that one if we'd rather a pure refactor.


aaronjeline (approved):

khieta (approved):
The `ValidatorActionId::attribute_types` change won't affect anything since the Lean JSON parser doesn't currently use the action attribute field (since it doesn't support action attributes in general)

Issue #599, opened 2024-01-24
Source locations for parsing errors in entities-JSON and policy-JSON formats


cdisselkoen:
### Describe the improvement you'd like to request

#174 covers the work for source locations in parsing errors for the schema-JSON format.  This issue covers the work for source locations in parsing errors for the entities-JSON and policy-JSON formats.  The issues are related because they'll all require getting source spans from `serde_json` somehow.  #477 has laid important groundwork for adding source locations to errors in general.

### Describe alternatives you've considered

We've basically committed to doing this already, since we've added source locations to many of our other error types (e.g. in #582)

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this internal improvement
- [ ] ⚠️ This feature might incur a breaking change

cdisselkoen:
Possibly useful for this: https://docs.rs/serde_spanned/0.6.7/serde_spanned/

PR #1009, opened 2024-06-19
Add checks for error source

khieta:
## Description of changes

The more interesting error message change promised in #996. 

While messing with the FFI, I noticed that some error messages were displaying with redundant information, e.g.:
Screenshot 2024-06-17 at 11 09 06 AM

The reason for this is that given a definition like:
```rust
    #[error("unable to link template: {inner}")]
    pub struct LinkingError {
        #[from]
        #[diagnostic(transparent)]
        pub(crate) inner: ast::LinkingError,
    }
```

The `ast::LinkingError` error is automatically set as the "source" field for the error. So there's no need to include `: {inner}` in the error message text. This PR fixes the error message above (along with several other errors using this pattern) and updates the error testing infrastructure to check the "source" field in addition to help text and underlines.

**Bonus:** the best error message I found that used this pattern was:
Screenshot 2024-06-17 at 1 58 55 PM

### DIFF
--- a/cedar-policy-core/src/entities.rs
+++ b/cedar-policy-core/src/entities.rs
@@ -961,11 +961,14 @@ mod json_parsing_tests {
         let eparser: EntityJsonParser<'_, '_> =
             EntityJsonParser::new(None, Extensions::all_available(), TCComputation::ComputeNow);
         assert_matches!(eparser.from_json_value(json.clone()), Err(e) => {
-            expect_err(&json, &miette::Report::new(e), &ExpectedErrorMessageBuilder::error(
-                r#"error during entity deserialization: in uid field of <unknown entity>, expected a literal entity reference, but got `"test_entity_type::\"Alice\""`"#,
-            ).help(
-                r#"literal entity references can be made with `{ "type": "SomeType", "id": "SomeId" }`"#,
-            ).build());
+            expect_err(
+                &json,
+                &miette::Report::new(e),
+                &ExpectedErrorMessageBuilder::error("error during entity deserialization")
+                    .source(r#"in uid field of <unknown entity>, expected a literal entity reference, but got `"test_entity_type::\"Alice\""`"#)
+                    .help(r#"literal entity references can be made with `{ "type": "SomeType", "id": "SomeId" }`"#)
+                    .build()
+            );
         });
     }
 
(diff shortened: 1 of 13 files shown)


cedar-policy-core/src/entities.rs
```diff
@@ -1805,9 +1823,14 @@ mod json_parsing_tests {
         let eparser: EntityJsonParser<'_, '_> =
             EntityJsonParser::new(None, Extensions::all_available(), TCComputation::ComputeNow);
         assert_matches!(eparser.from_json_str(json), Err(e) => {
-            expect_err(json, &miette::Report::new(e), &ExpectedErrorMessageBuilder::error(
-                r#"error during entity deserialization: the key `bar` occurs two or more times in the same JSON object at line 11 column 25"# // TODO: put the line-column information in `Diagnostic::labels()` instead of printing it in the error message
-            ).build());
+            // TODO: put the line-column information in `Diagnostic::labels()` instead of printing it in the error message
```
cdisselkoen, line 1826: #599

cedar-policy-core/src/test_utils.rs
```diff
@@ -116,13 +121,24 @@ impl<'a> ExpectedErrorMessageBuilder<'a> {
         }
     }
 
+    /// Add expected underlined text. The error message will be expected to have
+    /// exactly one miette label, and the underlined portion should be `snippet`.
+    /// The label text is expected to match `label`.
```
cdisselkoen, line 126: Doc comment should be updated

cedar-policy-validator/src/schema/namespace_def.rs
```diff
@@ -499,6 +499,7 @@ impl ValidatorNamespaceDef {
                 }
             }
             if !actions_with_attributes.is_empty() {
+                actions_with_attributes.sort(); // for deterministic error messages
```
cdisselkoen, line 502: might be good to reference #833 in this comment

john-h-kastner-aws: I fixed this in other places by using a `BTreeMap`, so maybe use that for uniformity.

john-h-kastner-aws (approved):

cedar-policy-core/src/entities.rs
john-h-kastner-aws: easier special case of that issue since we'd just need to convert line column back to a single index.

cedar-policy-validator/src/schema/namespace_def.rs
khieta: Added a ref to the issue in the comment since that was the easier "fix."

Issue #579, opened 2024-01-10
Schema files should be able to refer to types in the empty namespace


john-h-kastner-aws:
### Category

Cedar validation features/changes

### Describe the feature you'd like to request

In the following schema, there is an `id` common type without a namespace. We should be able to refer to it inside of a namespace, but the schema currently doesn't parse because we implicitly qualify `id` with the current namespace, and then look for `Demo::id`. When this type isn't found, schema parsing fails.

```json
{
  "Demo": {
    "entityTypes": {
      "User": {
        "memberOfTypes": [],
        "shape": {
          "type": "Record",
          "attributes": {
            "id": { "type": "id" },
          }
        }
      }
    },
    "actions": {}
  },
  "": {
    "commonTypes": {
      "id": {
        "type": "String"
      },
    }
    "entityTypes": {},
    "actions": {}
  }
}
```

We should fix this in two ways:

* When `id` doesn't exist in the current namespace, we should fall back to looking for it in the empty/global namespace. This schema would then parse as expected.
* It should be possible to explicitly refer to types in the empty namespace. A possible syntax for this might be `{"type": "::id"}` (we should think about other possible syntax for this, and we also need a syntax for the human-readable schema format). This would allow referring to this definition of `id` even when another one is defined in this namespace.

The example in this issue uses `commonTypes`, but we should ensure that the behavior is consistent between `commonTypes`, `entityTypes` and `actions` which can all be qualified with a namespace.

john-h-kastner-aws:
We've seen a couple different people encounter this issue now. Should make it a priority to get this fixed.

shaobo-he-aws:
> We've seen a couple different people encounter this issue now. Should make it a priority to get this fixed.

We need to pin down some syntax for qualified names that refer to the empty namespace. Maybe an amendment to RFC 24?

cdisselkoen:
The only syntax proposal I see in this thread is `::id` (for both the JSON and human formats). Is this:
* Basically settled, go with it
* Not settled but let's discuss here on this issue and make a decision quickly
* Not settled and we should have an RFC for discussion

shaobo-he-aws:
> * Not settled and we should have an RFC for discussion

I prefer this path because I propose we leverage the soon-to-be-reserved `__cedar` namespaces. For instance, we could use something like `__cedar::schema::top_level::id`. My objection to `::id` is that it's not a valid `Name`.

cdisselkoen:
[...]

PR #1016, opened 2024-06-21
introduce `RawName` wrapper type

cdisselkoen:
## Description of changes

Introduces the `RawName` wrapper type, to make explicit in the type system the difference between may-not-yet-be-fully-qualified `Name` (we now call this `RawName`) and guaranteed-to-be-fully-qualified `Name` (which is just `Name`).

This PR also starts to separate out the operations of (1) adding explicit namespaces to type references, and (2) inlining common type definitions. These two operations are currently conflated / performed simultaneously, and this PR begins to separate them for clarity.

Also cleans up a few comments, including a couple places where comments weren't updated to reflect #983. 

This is a non-functional-change refactor in preparation for #579. But I thought this change was worth separating into its own PR, to review it separately and keep diff sizes down.

## Issue #, if available

Towards #579 (but this is a non-functional-change refactor that does not resolve the issue yet)

### DIFF
--- a/cedar-policy-validator/src/schema/raw_name.rs
+++ b/cedar-policy-validator/src/schema/raw_name.rs
@@ -1,0 +1,93 @@
+/*
+ * Copyright Cedar Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+use cedar_policy_core::ast::{Id, Name};
+use cedar_policy_core::parser::Loc;
+use serde::{Deserialize, Serialize};
+
+/// A newtype which indicates that the contained `Name` may not yet be
+/// fully-qualified.
+///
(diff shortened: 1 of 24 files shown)


aaronjeline (approved):
I like this change, good sanity check

cedar-policy-validator/src/schema/raw_name.rs
```diff
@@ -0,0 +1,93 @@
...
+        self.0.is_unqualified()
+    }
+
+    /// Convert this `RawName` to a `Name` by adding the given `ns` as its
+    /// prefix, or by no-op if `ns` is `None`.
+    ///
+    /// Note that if the `RawName` already had a non-empty explicit namespace,
+    /// no additional prefixing will be done, even if `ns` is `Some`.
+    pub fn qualify_with(self, ns: Option<&Name>) -> Name {
+        self.0.prefix_namespace_if_unqualified(ns)
```
john-h-kastner-aws, line 78: Do we still want this function on `Name`? Feels like it should only exist for `RawName` if possible. Or, if we need both, they should probably use the same name.

john-h-kastner-aws (approved):
Seems legit.

cedar-policy-validator/src/schema/raw_name.rs
cdisselkoen: `Name` is a Core type, and I think DRT relies on that function being available on Core `Name`. We could switch DRT over to this `RawName` type as well at some point, at which time we should probably move `RawName` into Core.  That seems out of scope for this PR in my opinion, which is already a large enough diff.

Which name do you prefer?

john-h-kastner-aws: either is fine

cdisselkoen: Renamed the Core function to `qualify_with` in the latest rev

john-h-kastner-aws (approved):

PR #1019, opened 2024-06-21
update cedar lang versions

khieta:
## Description of changes

It appears that we've forgotten to do this for the last few releases 🙂 Please check that you agree with the versions I've assigned. Most of our recent releases did not change the language version, but I think that the implementation of [RFC57](https://github.com/cedar-policy/rfcs/pull/57) did.

After this PR is merged, I'll take care of the corresponding updates to `cedar-docs`.

### DIFF
--- a/cedar-policy/CHANGELOG.md
+++ b/cedar-policy/CHANGELOG.md
@@ -5,7 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+The "Cedar Language Version" refers to the language version as documented in the [Cedar Policy Language Guide](https://docs.cedarpolicy.com/other/doc-history.html). The language version may differ from the Rust crate version because a breaking change for the Cedar Rust API may or may not be a breaking change for the Cedar language.
+
 ## [Unreleased]
+Cedar Language Version: 4.0
 
 ### Added
 - JSON representation for Policy Sets, along with methods like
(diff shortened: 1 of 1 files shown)


shaobo-he-aws (approved):

cedar-policy/CHANGELOG.md
```diff
@@ -148,6 +155,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   format. (#734, resolving #681)
 
 ## [3.1.2] - 2024-03-29
+Cedar Language Version: 3.2.0
```
mwhicks1, line 159: It seems strange that the Rust create is a patch version update but the language is a minor version. Why is that?

cedar-policy/CHANGELOG.md
```diff
@@ -73,6 +77,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed Wasm build script to be multi-target in JS ecosystem (#933)
 
 ## [3.2.0] - 2024-05-17
+Cedar Language Version: 3.2.0
```
cdisselkoen, line 80: I think #766 might qualify as a language minor, making this 3.3.0?

cdisselkoen (approved):

cedar-policy/CHANGELOG.md
khieta: I don't recall the exact reasoning, but I think RFC#57 (which allows multiplication of arbitrary expressions) was needed to fix some weird parsing behavior, so we decided it was a "bug fix". From a language standpoint, it seems more like a "minor change" to me since it expands the class of parseable Cedar programs.

Definitely a grey area though... open to counter arguments.

cedar-policy/CHANGELOG.md
khieta: Ah, I agree.

cdisselkoen (approved):

john-h-kastner-aws (approved):

Issue #942, opened 2024-05-31
Partial evaluation APIs are not ergonomic


shaobo-he-aws:
### Describe the improvement you'd like to request

Consider a simple use case where only the resource is unknown.

```Rust
let partial_request = RequestBuilder::default()
    .action(Some(...) // concrete action
    .principal(Some(...)) // concrete principal
    .build();
let partial_response = authorizer
    .is_authorized_partial(&partial_request, &policies, &entities);
```

To use `PartialResponse::reauthorize` when users have a concrete resource entity, they need to do two things,

1. Create a binding from `"resource"` to the concrete entity value
2. Make a dummy request like the following because `reauthorize` method requires it
```Rust
Request::new(None, None, None, Context::empty(), None).expect("should be a valid request")
```

1 is intuitive, though knowing to use `"resource"` as the key is not. 2 is unintuitive because the partial request is derived from an existing request. So, it's confusing what the role of the new request is. Personally, a natural workflow is to get the partial response and reauthorize it by providing the "missing" pieces.

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this internal improvement
- [ ] ⚠️ This feature might incur a breaking change

cdisselkoen:
Mostly addressed by #947.  I think the only part of this issue that remains unresolved is that the API requires you know to use the string "resource" as the magic name for the resource unknown?

shaobo-he-aws:
I just discovered that this issue is not addressed properly. #947 is not correct.

Issue #541, opened 2023-12-21
Refactor `Context` to enforce its invariant statically


aaronjeline:
### Category

Internal refactors/changes

### Describe the feature you'd like to request

`cedar_policy_core::ast::request::Context` looks like:
```rust
/// `Context` field of a `Request`
#[derive(Debug, Clone, PartialEq, Serialize)]
pub struct Context {
    /// Context is internally represented as a `PartialValue`
    ///
    /// Context is serialized as a `RestrictedExpr`, for partly historical reasons.
    //
    // INVARIANT(ContextRecord): This must be a `Record`: either
    // `PartialValue::Value(Value::Record)`, or
    // `PartialValue::Residual(Expr::Record)`, or an appropriate unknown
    #[serde(flatten)]
    context: PartialValueSerializedAsExpr,
}
```
We should enforce this invariant statically if possible

### Describe alternatives you've considered

Status quo, enforce it dynamically. 

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

shaobo-he-aws:
I'd like to prioritize this issue as it's been shown to help multiple PRs.

khieta:
One PR this would have helped is #1027. (Note that the PR also updates the `ContextRecord` invariant to accurately reflect that a `Context` can never be purely unknown).  

I think the right solution here is to store the context contents as a hash map (or similar data structure) instead of a record expression.

PR #1026, opened 2024-06-25
Fix `reauthorize`

shaobo-he-aws:
## Issue #, if available

#942

### DIFF
--- a/cedar-policy-core/src/ast/request.rs
+++ b/cedar-policy-core/src/ast/request.rs
@@ -350,6 +350,14 @@ impl From<Context> for PartialValue {
     }
 }
 
+impl From<PartialValue> for Context {
+    fn from(value: PartialValue) -> Self {
+        Self {
+            context: value.into(),
+        }
+    }
+}
+
 impl std::default::Default for Context {
     fn default() -> Context {
         Context::empty()
--- a/cedar-policy-core/src/ast/request.rs
+++ b/cedar-policy-core/src/ast/request.rs
@@ -231,6 +231,13 @@ impl Context {
         }
     }
 
+    /// Create a `Context` from a `PartialValue` without checking its validity
+    pub(crate) fn from_partial_value_unchecked(val: PartialValue) -> Self {
+        Self {
+            context: val.into(),
+        }
+    }
+
     /// Create a `Context` from a `RestrictedExpr`, which must be a `Record`.
     ///
     /// `extensions` provides the `Extensions` which should be active for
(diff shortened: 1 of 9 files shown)


cedar-policy-core/src/ast/request.rs
```diff
@@ -350,6 +350,14 @@ impl From<Context> for PartialValue {
     }
 }
 
+impl From<PartialValue> for Context {
+    fn from(value: PartialValue) -> Self {
+        Self {
+            context: value.into(),
+        }
+    }
+}
```
cdisselkoen, line 359: Should this fail if the `PartialValue` is not a record or unknown?

cedar-policy/src/api/err.rs
```diff
@@ -134,14 +134,7 @@ pub enum ReauthorizationError {
...
     #[diagnostic(transparent)]
-    PolicySet(#[from] PolicySetError),
-}
-
-#[doc(hidden)]
-impl From<cedar_policy_core::ast::PolicySetError> for ReauthorizationError {
-    fn from(e: cedar_policy_core::ast::PolicySetError) -> Self {
-        Self::PolicySet(e.into())
-    }
+    Reauthorization(#[from] cedar_policy_core::authorizer::ReauthorizationError),
```
cdisselkoen, line 137: Since `cedar_policy::ReauthorizationError` is a public enum, this contained struct needs to be either re-exported or wrapped.  (In a public enum, all data contained in each variant is always `pub`.)

cedar-policy-core/src/ast/request.rs
shaobo-he-aws: Removed it.

cedar-policy/src/api/err.rs
shaobo-he-aws: Pushed a fix.

cdisselkoen (approved):
LGTM

andrewmwells-amazon (approved):

cedar-policy-core/src/authorizer/partial_response.rs
```diff
@@ -333,6 +337,127 @@ impl PartialResponse {
...
+
+        // We need to replace unknowns in the partial context as well
+        if let Some(ref ctx) = context {
+            if let PartialValue::Residual(residual) = ctx.as_ref() {
+                let expr = residual.substitute(mapping);
+                let extns = Extensions::all_available();
+                let eval = RestrictedEvaluator::new(&extns);
+                let partial_value =
+                    eval.partial_interpret(BorrowedRestrictedExpr::new_unchecked(&expr))?;
+                context = Some(Context::from_partial_value_unchecked(partial_value));
```
john-h-kastner-aws, line 449: The safety of this `_unchecked` constructor isn't immediately obvious. It depends on the earlier `Context::from_partial_value_unchecked(val.clone().into())` maintaining the invariant and then that `partial_interpret` on `Record` residual still being a `Record`.

That all seems legit, but IMO merits some additional explanation in comments. (We could also refactor to remove the need for the invariant as suggested in #541,  but that may be complicated.)

PR #1027, opened 2024-06-25
Context `merge` function

### LINKED ISSUE #1013: Add a merge API to `cedar_policy::Context`
### Category

As a developer working on translation of user-defined rust types to `cedar_policy::Context`, I find that I often want to reach for a `merge` method to merge two contexts `ctx1.merge(ctx2)`. A use-case for this is when translating nested structs of attributes into a nested `Context`.

Consider a top-level aggregate of multiple user-defined Contexts.
```rust
#[derive(Builder, Debug, Clone, PartialEq, Eq)]
#[builder(pattern = "owned")]
pub struct MyContext {
    pub foo: FooContext,
    pub bar: BarContext,
}
```

Where sub-contexts may have other nested user-defined Contexts that I need to translate into a `cedar_policy::Context`, are defined as:

```rust
// Contains other Contexts I need to translate, recursively.
pub struct FooContext {
    pub ctx_a: Option<FooContextA>,
    pub ctx_b: Option<FooContextB>,
}

// A leaf Context.
pub struct BarContext {
    pub key: String,
    pub value: String,
}
```
A merge API could help write code like this to stitch together sub-context translation results:

```rust
impl MyContext {
    pub fn try_into_cedar_context(self) -> Result<cedar_policy::Context, MyContextToCedarContextError> {
            let foo_ctx = self.foo.try_into_cedar_context()?;
            let bar_ctx = self.bar.try_into_cedar_context()?;
            Ok(foo_ctx.merge(bar_ctx));
    }
}

// Example impl for a non-top-level context.
impl FooContext {
    // It would be good to run a fold/collect to reduce over the contexts. Rough idea, may not compile.
    pub fn try_into_cedar_context(self) -> Result<cedar_policy::Context, MyContextToCedarContextError> {
          Ok(vec![self.ctx_a, self.ctx_b].iter()
                    .fold(Context::new(), |mut acc, ctx| { 
                         acc.merge(ctx);
                         acc
                  })?);
    }
}
```

Aside from the `merge` proposal another good alternative could be implementing the `extend` trait like `HashMap` does, https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.extend.

The workaround that I ended up doing was to implement non-top-level translations to a `HashMap<String, RestrictedExpression>`. Then I can build the top-level context by merging the restricted expression maps and passing that to `Context::from_pairs(result)`. 

```rust
// Translation of a leaf context.
impl BarContext {
    pub fn try_into_cedar_context(self) -> Result<HashMap<String, RestrictedExpression, MyContextToCedarContextError> {
          Ok(HashMap::from([
               ("key".to_string(), RestrictedExpression::new_string(self.key)),
               ("value".to_string(), RestrictedExpression::new_string(self.value)),
           ])?);
    }
}
```

At the top-level I converted the sub-contexts to a `RestrictedExpression` record, and then used `Context::from_pairs`:

```rust
impl MyContext {
    pub fn try_into_cedar_context(self) -> Result<cedar_policy::Context, MyContextToCedarContextError> {
         let foo_expr_map = self.foo.try_into_cedar_context()?;
         let bar_expr_map = self.bar.try_into_cedar_context()?;
         Ok(Context::from_pairs([
             ("foo".to_string(), RestrictedExpression::new_record(foo_expr_map.into_iter())?),
             ("bar".to_string(), RestrictedExpression::new_record(bar_expr_map.into_iter())?)
         ]))
    }
}
```

### Additional context

_No response_

khieta:
## Description of changes

Adds two new methods to the public API:
* `Context::merge` as suggested in #1013 
* `Context::values` which gets the key/value pairs out of the context. This function seemed like the most natural way to implement `merge`, and could be useful in general.

Makes a breaking error API change (because I can't help myself):
* Renamed `DuplicateKeyInRecordLiteralError` to `DuplicateKeyError` and added a "context" field to allow more precise error messages. The reason for this change is just that I didn't like the error message "duplicate key _ in a record literal" for the `Context` functions.

## Issue #, if available

Resolves #1013

### DIFF
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -424,7 +424,7 @@ impl Expr {
     /// Create an `Expr` which evaluates to a Record with the given (key, value) pairs.
     pub fn record(
         pairs: impl IntoIterator<Item = (SmolStr, Expr)>,
-    ) -> Result<Self, RecordConstructionError> {
+    ) -> Result<Self, ExpressionConstructionError> {
         ExprBuilder::new().record(pairs)
     }
 
@@ -995,18 +995,16 @@ impl<T> ExprBuilder<T> {
     pub fn record(
         self,
         pairs: impl IntoIterator<Item = (SmolStr, Expr<T>)>,
-    ) -> Result<Expr<T>, RecordConstructionError> {
+    ) -> Result<Expr<T>, ExpressionConstructionError> {
         let mut map = BTreeMap::new();
         for (k, v) in pairs {
             match map.entry(k) {
                 btree_map::Entry::Occupied(oentry) => {
-                    return Err(
-                        expression_construction_errors::DuplicateKeyInRecordLiteralError {
-                            key: oentry.key().clone(),
-                            context: "in record literal",
-                        }
-                        .into(),
-                    );
+                    return Err(expression_construction_errors::DuplicateKeyError {
+                        key: oentry.key().clone(),
+                        context: "in record literal",
+                    }
+                    .into());
                 }
                 btree_map::Entry::Vacant(ventry) => {
                     ventry.insert(v);
@@ -1155,16 +1153,14 @@ impl<T: Clone> ExprBuilder<T> {
 // Don't make fields `pub`, don't make breaking changes, and use caution
 // when adding public methods.
 #[derive(Debug, PartialEq, Eq, Clone, Diagnostic, Error)]
-pub enum RecordConstructionError {
+pub enum ExpressionConstructionError {
     /// The same key occurred two or more times
     #[error(transparent)]
     #[diagnostic(transparent)]
-    DuplicateKeyInRecordLiteral(
-        #[from] expression_construction_errors::DuplicateKeyInRecordLiteralError,
-    ),
+    DuplicateKeyInRecordLiteral(#[from] expression_construction_errors::DuplicateKeyError),
 }
 
-/// Error subtypes for [`RecordConstructionError`]
+/// Error subtypes for [`ExpressionConstructionError`]
 pub mod expression_construction_errors {
     use miette::Diagnostic;
     use smol_str::SmolStr;
@@ -1177,20 +1173,20 @@ pub mod expression_construction_errors {
     // when adding public methods.
     #[derive(Debug, PartialEq, Eq, Clone, Diagnostic, Error)]
     #[error("duplicate key `{key}` {context}")]
-    pub struct DuplicateKeyInRecordLiteralError {
+    pub struct DuplicateKeyError {
         /// The key which occurred two or more times
         pub(crate) key: SmolStr,
         /// Information about where the duplicate key occurred (e.g., "in record literal")
         pub(crate) context: &'static str,
     }
 
-    impl DuplicateKeyInRecordLiteralError {
+    impl DuplicateKeyError {
         /// Get the key which occurred two or more times
         pub fn key(&self) -> &str {
             &self.key
         }
 
-        /// Get a new error with an updated `context` field
+        /// Make a new error with an updated `context` field
         pub fn set_context(self, context: &'static str) -> Self {
             Self { context, ..self }
         }
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -424,7 +424,7 @@ impl Expr {
     /// Create an `Expr` which evaluates to a Record with the given (key, value) pairs.
     pub fn record(
         pairs: impl IntoIterator<Item = (SmolStr, Expr)>,
-    ) -> Result<Self, ExpressionConstructionError> {
+    ) -> Result<Self, RecordConstructionError> {
         ExprBuilder::new().record(pairs)
     }
 
@@ -995,7 +995,7 @@ impl<T> ExprBuilder<T> {
     pub fn record(
         self,
         pairs: impl IntoIterator<Item = (SmolStr, Expr<T>)>,
-    ) -> Result<Expr<T>, ExpressionConstructionError> {
+    ) -> Result<Expr<T>, RecordConstructionError> {
         let mut map = BTreeMap::new();
         for (k, v) in pairs {
             match map.entry(k) {
@@ -1003,6 +1003,7 @@ impl<T> ExprBuilder<T> {
                     return Err(
                         expression_construction_errors::DuplicateKeyInRecordLiteralError {
                             key: oentry.key().clone(),
+                            context: "in record literal",
                         }
                         .into(),
                     );
@@ -1154,8 +1155,8 @@ impl<T: Clone> ExprBuilder<T> {
 // Don't make fields `pub`, don't make breaking changes, and use caution
 // when adding public methods.
 #[derive(Debug, PartialEq, Eq, Clone, Diagnostic, Error)]
-pub enum ExpressionConstructionError {
-    /// The same key occurred two or more times in a single record literal
+pub enum RecordConstructionError {
+    /// The same key occurred two or more times
     #[error(transparent)]
     #[diagnostic(transparent)]
     DuplicateKeyInRecordLiteral(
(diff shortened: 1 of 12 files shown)


shaobo-he-aws (approved):

cedar-policy-core/src/ast/expr.rs
```diff
@@ -1155,12 +1154,10 @@ impl<T: Clone> ExprBuilder<T> {
...
 #[derive(Debug, PartialEq, Eq, Clone, Diagnostic, Error)]
 pub enum ExpressionConstructionError {
-    /// The same key occurred two or more times in a single record literal
+    /// The same key occurred two or more times
     #[error(transparent)]
     #[diagnostic(transparent)]
-    DuplicateKeyInRecordLiteral(
-        #[from] expression_construction_errors::DuplicateKeyInRecordLiteralError,
-    ),
+    DuplicateKeyInRecordLiteral(#[from] expression_construction_errors::DuplicateKeyError),
```
cdisselkoen, line 1160: probably rename the variant here too while we're at it

cedar-policy-core/src/ast/expr.rs
```diff
@@ -1169,23 +1166,30 @@ pub mod expression_construction_errors {
...
-    impl DuplicateKeyInRecordLiteralError {
-        /// Get the key which occurred two or more times in the record literal
+    impl DuplicateKeyError {
+        /// Get the key which occurred two or more times
         pub fn key(&self) -> &str {
             &self.key
         }
+
+        /// Make a new error with an updated `context` field
+        pub fn set_context(self, context: &'static str) -> Self {
```
cdisselkoen, line 1190: I'm not sure if this is general idiomaticity guideline or not, but I would expect `set_` to be a method that takes `&mut Self`, and `with_` to be a method that takes `self` and returns `Self`, like `Value::with_maybe_source_loc()` in Core.  Take or leave this advice

cedar-policy-core/src/ast/request.rs
```diff
@@ -309,28 +316,28 @@ impl Context {
...
 
-    /// Iterate over the (key, value) pairs in the `Context`; or return `None`
-    /// if the `Context` is purely unknown
+    /// Get an iterator over the (key, value) pairs in the `Context`, cloning
+    /// only if necessary.
     //
-    // PANIC SAFETY: This is safe due to the invariant on `self.context`, `self.context` must always be a record
-    pub fn iter<'s>(&'s self) -> Option<Box<dyn Iterator<Item = (&SmolStr, PartialValue)> + 's>> {
+    // PANIC SAFETY: This is safe due to the invariant (ContextRecord) on `self.context`
+    pub fn values(self) -> Box<dyn Iterator<Item = (SmolStr, PartialValue)>> {
```
cdisselkoen, line 323: Should we call this `.into_iter()` instead?  Maybe even have it be an implementation of the `IntoIterator` trait?

cedar-policy/src/api.rs
```diff
@@ -3384,6 +3384,34 @@ impl Context {
...
+    /// constructors like [`Context::from_pairs`] or [`Context::from_json_str`].
+    pub fn values(self) -> HashMap<String, RestrictedExpression> {
+        self.0
+            .values()
+            .map(|(k, v)| {
+                (
+                    k.to_string(),
+                    match v {
+                        ast::PartialValue::Value(val) => RestrictedExpression(
+                            ast::RestrictedExpr::new_unchecked(ast::Expr::from(val)),
```
cdisselkoen, line 3405: might be useful to have a `ast::RestrictedExpr::from(val)` function.  From this vantage point it's a little harder to see why `new_unchecked` is safe here

cdisselkoen (approved):

cedar-policy-core/src/ast/expr.rs
john-h-kastner-aws, line 1190: I wouldn't make this `pub`. `pub(crate)` probably?

cedar-policy-core/src/ast/request.rs
```diff
@@ -309,28 +316,28 @@ impl Context {
...
-                record
-                    .iter()
-                    .map(|(k, v)| (k, PartialValue::Value(v.clone()))),
-            )),
-            PartialValue::Residual(expr) => match expr.expr_kind() {
-                ExprKind::Record(map) => Some(Box::new(
-                    map.iter()
-                        .map(|(k, v)| (k, PartialValue::Residual(v.clone()))),
-                )),
-                ExprKind::Unknown(_) => None,
```
john-h-kastner-aws, line 333: How is the `Unknown` case handled now? Kind of worried it might hit the `panic!("internal invariant violation: expected a record, got {kind:?}"),` case

khieta: Great question! I believe the `Context::from_expr` function is used on all paths for building a context, and this function must return a record value or record residual due to the handing of records in `RestrictedEvaluator::partial_interpret_internal`. I updated the comment about the `ContextRecord` invariant to reflect this, but would appreciate if you would double check my reasoning.

cedar-policy-core/src/ast/request.rs
khieta: I'll try this out tomorrow

cedar-policy-core/src/ast/request.rs
john-h-kastner-aws: Looked at the code closer and I agree that this match is safe given the invariant. 

#1026 introduced `from_partial_value_unchecked` which makes the invariant depend on those call sites. I'd be happier if that didn't need the `_unchecked` constructor, but I am convinced that the invariant is maintained. We already have #541 for future refactoring.

cedar-policy-core/src/entities/json/err.rs
```diff
@@ -80,10 +80,10 @@ pub enum JsonDeserializationError {
     #[error(transparent)]
     #[diagnostic(transparent)]
     MissingImpliedConstructor(MissingImpliedConstructor),
-    /// The same key appears two or more times in a single record literal
+    /// The same key appears two or more times in a single record
     #[error(transparent)]
     #[diagnostic(transparent)]
-    DuplicateKeyInRecordLiteral(DuplicateKeyInRecordLiteral),
+    DuplicateKey(DuplicateKey),
```
aaronjeline, line 86: This is a nicer name

aaronjeline (approved):

cedar-policy-core/src/ast/request.rs
khieta: Latest version adds additional comments to `from_partial_value_unchecked` to clarify 🙂

cedar-policy-core/src/ast/request.rs
khieta: Latest version adds an implementation of `IntoIterator`. I don't think we can provide an `.iter()` function that doesn't perform cloning atm. I think the best solution would be to address #541 by storing the context as a hash map instead of a record expression.

cedar-policy-core/src/ast/request.rs
shaobo-he-aws: Would it make sense to implement #541 in this PR if it's convenient?

cedar-policy/src/api.rs
```diff
@@ -3387,30 +3387,50 @@ impl Context {
...
                     k.to_string(),
                     match v {
-                        ast::PartialValue::Value(val) => RestrictedExpression(
-                            ast::RestrictedExpr::new_unchecked(ast::Expr::from(val)),
-                        ),
+                        ast::PartialValue::Value(val) => {
+                            RestrictedExpression(ast::RestrictedExpr::from(val))
+                        }
                         ast::PartialValue::Residual(exp) => {
                             RestrictedExpression(ast::RestrictedExpr::new_unchecked(exp))
```
cdisselkoen, line 3416: maybe a short comment justifying the `_unchecked`, like you have elsewhere in this PR?

cdisselkoen (approved):
Looks great

cedar-policy-core/src/ast/request.rs
cdisselkoen: Nicely done :)

Want to add a comment to #541 with your suggestion for implementation?

cedar-policy/src/api.rs
```diff
@@ -3384,6 +3384,54 @@ impl Context {
         cedar_policy_validator::context_schema_for_action(&schema.0, action.as_ref())
             .ok_or_else(|| ContextJsonError::missing_action(action.clone()))
     }
+
+    /// Combine two [`Context`]s, returning an error if they contain overlapping keys
+    pub fn merge(self, context: Self) -> Result<Self, ContextCreationError> {
```
shaobo-he-aws, line 3389: Sorry for being late. I was wondering if we can generalize the argument to `impl Iterator<SmolStr, ...>`, since `Context` implements the conversion.

cedar-policy-core/src/ast/request.rs
john-h-kastner-aws: I expect that change to be a bit annoying, so I'd rather see it in a follow up PR. Up to Kesha if wants to do it ofc.

john-h-kastner-aws (approved):
looks good!

cedar-policy-core/src/ast/request.rs
khieta: Yeah, that will be a somewhat involved change, so I'll leave it for future work.

khieta:
I believe I've addressed all comments now. Will wait until tomorrow to merge because I will need to make some downstream fixes (caused by the change `Context::iter` -> `Context::into_iter`)

khieta:
Fix for DRT failure is here: https://github.com/cedar-policy/cedar-spec/pull/376. Need a re-approval before I can merge this PR.

Issue #977, opened 2024-06-13
Some `Diagnostic` implementations set `labels` without `source_code`


khieta:
### Describe the improvement you'd like to request

Some of our `Diagnostic` implementations set the `labels` function without setting `source_code`. These two fields are related (see the excerpt from `miette` below), so it doesn't make sense to set one without the other. The relevant error types seem to print fine (for some reason), but users who are consuming the `miette` errors programmatically may observe unexpected behavior. For example: I was trying to update [this testing code](https://github.com/cedar-policy/cedar/blob/b3a6eadd1f4f7923f5ac759b17f011ecef07d31a/cedar-policy-core/src/test_utils.rs#L225) to use the error's source text instead of the passed in source (which may or may not match the error), but found that the source text didn't exist for `ToCSTError`s.

```rust
    /// Source code to apply this `Diagnostic`'s [`Diagnostic::labels`] to.
    fn source_code(&self) -> Option<&dyn SourceCode> {
        None
    }

    /// Labels to apply to this `Diagnostic`'s [`Diagnostic::source_code`]
    fn labels(&self) -> Option<Box<dyn Iterator<Item = LabeledSpan> + '_>> {
        None
    }
```

Here's the list of error types doing this, from a quick scan through the code:
* [`UnexpectedSlotError`](https://github.com/cedar-policy/cedar/blob/b3a6eadd1f4f7923f5ac759b17f011ecef07d31a/cedar-policy-core/src/ast/policy.rs#L1424)
* [`ToCSTError`](https://github.com/cedar-policy/cedar/blob/b3a6eadd1f4f7923f5ac759b17f011ecef07d31a/cedar-policy-core/src/parser/err.rs#L552) 
* [`human_schema_errors::ParseError`](https://github.com/cedar-policy/cedar/blob/b3a6eadd1f4f7923f5ac759b17f011ecef07d31a/cedar-policy-validator/src/human_schema/err.rs#L155)
* [`ToJsonSchemaError`](https://github.com/cedar-policy/cedar/blob/b3a6eadd1f4f7923f5ac759b17f011ecef07d31a/cedar-policy-validator/src/human_schema/err.rs#L390C21-L390C38)
* [`schema_warnings::ShadowsBuiltinWarning`](https://github.com/cedar-policy/cedar/blob/b3a6eadd1f4f7923f5ac759b17f011ecef07d31a/cedar-policy-validator/src/human_schema/err.rs#L437C25-L437C46)
[...]

PR #1036, opened 2024-07-02
More tasks from #745

aaronjeline:
## Description of changes
Cleans up two remaining error tasks
* `EntityUid::from_json` returns a concrete type instead of an existential
* `HumanSchemaError::Parse` is more transparent
## Issue #, if available
#745

### DIFF
--- a/cedar-policy-validator/src/err.rs
+++ b/cedar-policy-validator/src/err.rs
@@ -41,10 +41,10 @@ pub enum HumanSchemaError {
 #[error("error parsing schema: {errs}")]
 pub struct HumanSyntaxParseError {
     /// Underlying parse error(s)
-    pub errs: human_schema::parser::HumanSyntaxParseErrors,
+    errs: human_schema::parser::HumanSyntaxParseErrors,
     /// Did the schema look like it was intended to be JSON format instead of
     /// human?
-    pub suspect_json_format: bool,
+    suspect_json_format: bool,
 }
 
 impl Diagnostic for HumanSyntaxParseError {
@@ -104,9 +104,15 @@ impl HumanSyntaxParseError {
         }
     }
 
-    #[cfg(test)]
-    pub(crate) fn inner(&self) -> &human_schema::parser::HumanSyntaxParseErrors {
+    /// Underlying parse error(s)
+    pub fn inner(&self) -> &human_schema::parser::HumanSyntaxParseErrors {
         &self.errs
+    }
+
+    /// Did the schema look like it was intended to be JSON format instead of
+    /// human?
+    pub fn suspect_json_format(&self) -> bool {
+        self.suspect_json_format
     }
 }
 
--- a/cedar-policy-validator/src/err.rs
+++ b/cedar-policy-validator/src/err.rs
@@ -41,10 +41,10 @@ pub enum HumanSchemaError {
 #[error("error parsing schema: {errs}")]
 pub struct HumanSyntaxParseError {
     /// Underlying parse error(s)
-    errs: human_schema::parser::HumanSyntaxParseErrors,
+    pub errs: human_schema::parser::HumanSyntaxParseErrors,
     /// Did the schema look like it was intended to be JSON format instead of
     /// human?
-    suspect_json_format: bool,
+    pub suspect_json_format: bool,
 }
 
 impl Diagnostic for HumanSyntaxParseError {
--- a/cedar-policy-validator/src/human_schema/err.rs
+++ b/cedar-policy-validator/src/human_schema/err.rs
@@ -652,35 +652,6 @@ impl Diagnostic for NoPrincipalOrResource {
     }
 }
 
-// impl Diagnostic for ToJsonSchemaError {
-//     fn labels(&self) -> Option<Box<dyn Iterator<Item = LabeledSpan> + '_>> {
-//         match self {
-//             ToJsonSchemaError::DuplicateDeclarations { loc1, loc2, .. }
-//             | ToJsonSchemaError::DuplicateContext { loc1, loc2 }
-//             | ToJsonSchemaError::DuplicatePrincipalOrResource { loc1, loc2, .. }
-//             | ToJsonSchemaError::DuplicateKeys { loc1, loc2, .. } => Some(Box::new(
-//                 vec![
-//                     LabeledSpan::underline(loc1.span),
-//                     LabeledSpan::underline(loc2.span),
-//                 ]
-//                 .into_iter(),
-//             )),
-//             ToJsonSchemaError::DuplicateNameSpaces { loc1, loc2, .. } => {
-//                 Some(Box::new([loc1, loc2].into_iter().filter_map(|loc| {
-//                     Some(LabeledSpan::underline(loc.as_ref()?.span))
-//                 })))
-//             }
-//             ToJsonSchemaError::UnknownTypeName(node) => Some(Box::new(std::iter::once(
-//                 LabeledSpan::underline(node.loc.span),
-//             ))),
-//             ToJsonSchemaError::UseReservedNamespace(loc)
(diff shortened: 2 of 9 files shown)


cedar-policy-validator/src/err.rs
```diff
@@ -41,10 +41,10 @@ pub enum HumanSchemaError {
 #[error("error parsing schema: {errs}")]
 pub struct HumanSyntaxParseError {
     /// Underlying parse error(s)
-    errs: human_schema::parser::HumanSyntaxParseErrors,
+    pub errs: human_schema::parser::HumanSyntaxParseErrors,
     /// Did the schema look like it was intended to be JSON format instead of
     /// human?
-    suspect_json_format: bool,
+    pub suspect_json_format: bool,
 }
```
cdisselkoen, line 48: Is this struct publicly exported in cedar-policy?  If it is, I'd prefer not having `pub` here and instead using getters

cedar-policy/src/api/err.rs
```diff
@@ -213,7 +213,11 @@ pub mod human_schema_errors {
     #[derive(Debug, Error, Diagnostic)]
     #[error(transparent)]
     #[diagnostic(transparent)]
-    pub struct ParseError(#[from] pub(super) cedar_policy_validator::HumanSyntaxParseError);
+    pub struct ParseError {
+        /// The errors reported by the parser
+        #[from]
+        pub errors: cedar_policy_validator::HumanSyntaxParseError,
```
cdisselkoen, line 219: Is this error struct publicly exported?  If not, let's wrap in a struct at this layer (there are other examples of that pattern in this file)

cedar-policy/src/api/id.rs
```diff
@@ -17,7 +17,7 @@
 //! This module defines the publicly exported identifier types including
 //! `EntityUid` and `PolicyId`.
 
-use crate::entities_json_errors::JsonDeserializationError;
+pub use crate::entities_json_errors::JsonDeserializationError;
```
cdisselkoen, line 20: Can we avoid this `pub`? I think that will result in this type being exposed in two different places in `cedar-policy` (it's already public in the `entities_json_errors` module)

cedar-policy/src/api/err.rs
aaronjeline: Not sure what you're referring to

cedar-policy-validator/src/human_schema/err.rs
```diff
@@ -354,80 +363,324 @@ impl Diagnostic for ToJsonSchemaErrors {
...
-        loc2: Option<Loc>,
-    },
-    /// Invalid type name
-    #[error("Unknown type name: `{}`", .0.node)]
-    UnknownTypeName(Node<SmolStr>),
-    #[error("Use reserved namespace `__cedar`")]
-    UseReservedNamespace(Loc),
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    DuplicateNameSpaces(#[from] DuplicateNameSpaces),
```
cdisselkoen, line 393:
```suggestion
    DuplicateNamespaces(#[from] DuplicateNamespaces),
```

cedar-policy-validator/src/human_schema/err.rs
```diff
@@ -354,80 +363,324 @@ impl Diagnostic for ToJsonSchemaErrors {
...
+    /// Error raised when there are duplicate `principal` or `resource` declarations
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    DuplicatePrincipalOrResource(#[from] DuplicatePrincipalOrResource),
+    /// Error raised when there are duplicate `principal` or `resource` declarations
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    NoPrincipalOrResource(#[from] NoPrincipalOrResource),
 
     /// Error raised when there are duplicate namespace IDs
```
cdisselkoen, line 390: I thought we allow reopening a namespace in the human schema format?

cedar-policy-validator/src/human_schema/err.rs
```diff
@@ -354,80 +363,324 @@ impl Diagnostic for ToJsonSchemaErrors {
...
+        self.key.as_ref()
+    }
+}
+
+impl Diagnostic for DuplicateKeys {
+    fn labels(&self) -> Option<Box<dyn Iterator<Item = LabeledSpan> + '_>> {
+        let l1 = LabeledSpan::underline(self.loc1.span);
+        let l2 = LabeledSpan::underline(self.loc2.span);
+        Some(Box::new(vec![l1, l2].into_iter()))
+    }
```
cdisselkoen, line 425: implementations of `labels()` should also implement `source_code()`. #977

cedar-policy-validator/src/human_schema/err.rs
```diff
@@ -354,80 +363,324 @@ impl Diagnostic for ToJsonSchemaErrors {
...
+    fn labels(&self) -> Option<Box<dyn Iterator<Item = LabeledSpan> + '_>> {
+        Some(Box::new(
+            vec![LabeledSpan::underline(self.0.loc.span)].into_iter(),
+        ))
+    }
+}
+
+/// Use of a reserved namespace
+#[derive(Debug, Clone, Error, PartialEq, Eq)]
+#[error("Use reserved namespace `__cedar`")]
```
cdisselkoen, line 544:
```suggestion
#[error("this uses the reserved namespace `__cedar`")]
```

cedar-policy-validator/src/human_schema/err.rs
```diff
@@ -354,80 +363,324 @@ impl Diagnostic for ToJsonSchemaErrors {
...
+    fn labels(&self) -> Option<Box<dyn Iterator<Item = LabeledSpan> + '_>> {
+        Some(Box::new(
+            vec![LabeledSpan::underline(self.0.span)].into_iter(),
+        ))
+    }
 }
 
 impl ToJsonSchemaError {
+    /// Creates a duplicate keys error
     pub fn duplicate_keys(key: SmolStr, loc1: Loc, loc2: Loc) -> Self {
```
cdisselkoen, line 557: Since this type is publicly exposed, end-users can see and use these constructors, and we have to commit to stability for their arguments.  I think it would be better if we could avoid (at least fully `pub`) constructors, so that in the future we could add or change arguments as necessary?

cedar-policy-validator/src/human_schema/err.rs
```diff
@@ -354,80 +363,324 @@ impl Diagnostic for ToJsonSchemaErrors {
...
+//                 LabeledSpan::underline(node.loc.span),
+//             ))),
+//             ToJsonSchemaError::UseReservedNamespace(loc)
+//             | ToJsonSchemaError::NoPrincipalOrResource { loc, .. } => {
+//                 Some(Box::new(std::iter::once(LabeledSpan::underline(loc.span))))
+//             }
+//         }
+//     }
+// }
+
```
cdisselkoen: can probably remove this commented-out code?

cedar-policy/src/api/err.rs
cdisselkoen: It is publicly exported, so this comment doesn't apply

cedar-policy-validator/src/human_schema/err.rs
john-h-kastner-aws: agreed. I think our error constructors are generally not public

cedar-policy/src/api/err.rs
john-h-kastner-aws, line 219: If we're making the inner errors a public error type, do we still want this `ParseError` wrapper?

If we do, it might be more consistent with our other errors to make it `pub struct ParseErrors(...)` instead of a record struct.

aaronjeline:
Superseded by #1129

PR #1037, opened 2024-07-03
Fix latent bug in `contains_at_least_two`

aaronjeline:
## Description of changes
The function `contains_at_least_two` is used in the construction of IP Address parsing errors to provide better error messages. 
Currently it has a bug, if called with a multibyte character as the search query, it can end up slicing a `&str` across a non-char boundary, resulting in a panic. 
As currently implemented, it is only ever called with `':'` or `'.'` as queries, which are both single-byte chars, meaning the panic is currently  unreachable from public APIs. That said, this is an undocumented invariant, which could be exposed or violated easily if this code was refactored to allow for easier sharing.

As an example: the string `"\u{f1b}"` and the query `'\u{f1b}'` would have caused a panic in the old version of the code

In addition this adds a Kani verification harness, which is how the bug was found.
## Issue #, if available
N/A

### DIFF
--- a/cedar-policy-core/src/extensions/ipaddr.rs
+++ b/cedar-policy-core/src/extensions/ipaddr.rs
@@ -262,7 +262,7 @@ fn contains_at_least_two(s: &str, c: char) -> bool {
 #[cfg(kani)]
 mod proof {
 
-    /// Prove that `contains_two_correct` does not panic for strings with length <= 4 
+    /// Prove that `contains_two_correct` does not panic for strings with length <= 4
     #[kani::proof]
     #[kani::unwind(5)]
     fn contains_at_least_two_correct() {
--- a/cedar-policy-core/src/extensions/ipaddr.rs
+++ b/cedar-policy-core/src/extensions/ipaddr.rs
@@ -242,12 +242,42 @@ fn contains_at_least_two(s: &str, c: char) -> bool {
     let idx = s.find(c);
     match idx {
         Some(i) => {
-            // PANIC SAFETY `i` is guaranteed to be < `s.len()`, so this won't panic
+            // For this slicing operation not to panic, two preconditoins must be met:
+            // 1) `i + 1` must be <= `s.len()`
+            //     This is met because `i` comes from `s.find()` meaning it must be < `s.len()`
+            // 2) `i + 1` must be a character boundary
+            //     This is met because we use `c.len_utf()` which is guaranteed to be the length of
+            //     `c`, since Rust strings use UTF8
+            // The above is checked by [`proof::contains_at_least_two_correct`], which proves
+            // safety for strings up to length 4
+            // PANIC SAFETY: (see above)
             #[allow(clippy::indexing_slicing)]
-            let idx = s[i + 1..].find(c);
+            let idx = s[i + c.len_utf8()..].find(c);
             idx.is_some()
         }
         None => false,
+    }
+}
+
+#[cfg(kani)]
+mod proof {
+
+    /// Prove that `contains_two_correct` does not panic for strings with length <= 4 
+    #[kani::proof]
+    #[kani::unwind(5)]
+    fn contains_at_least_two_correct() {
+        let mut buf: Vec<u8> = vec![];
+        let len: usize = kani::any();
+        kani::assume(len <= 4);
+        for i in 0..len {
+            let c = kani::any();
+            buf.push(c);
+        }
+        if let Ok(s) = std::str::from_utf8(buf.as_slice()) {
+            let pat = kani::any();
+            // Just don't panic
+            let _ = super::contains_at_least_two(s, pat);
+        }
     }
 }
 
@@ -1134,5 +1164,10 @@ mod tests {
         assert!(contains_at_least_two(":::", ':'));
         assert!(contains_at_least_two("::", ':'));
         assert!(!contains_at_least_two(":", ':'));
+    }
+
+    #[test]
+    fn test_contains_two_multibyte() {
+        assert!(!contains_at_least_two("\u{f1b}", '\u{f1b}'));
     }
 }
--- a/cedar-policy-core/src/extensions/ipaddr.rs
+++ b/cedar-policy-core/src/extensions/ipaddr.rs
@@ -269,7 +269,7 @@ fn contains_at_least_two(s: &str, c: char) -> bool {
 #[cfg(kani)]
 mod proof {
 
-    /// Prove that `contains_two_correct` does not panic for strings with length <= 4
+    /// Prove that `contains_two_correct` does not panic for strings with length <= 6
     #[kani::proof]
     #[kani::unwind(7)]
     fn contains_at_least_two_correct() {
--- a/cedar-policy-core/src/extensions/ipaddr.rs
+++ b/cedar-policy-core/src/extensions/ipaddr.rs
@@ -275,7 +275,7 @@ mod proof {
     fn contains_at_least_two_correct() {
         let buf: [u8; 6] = kani::any();
         let len: usize = kani::any();
-        kani::assume(len <= 4);
+        kani::assume(len <= 6);
         let slice = &buf[0..len];
         if let Ok(s) = std::str::from_utf8(slice) {
             let pat = kani::any();
--- a/cedar-policy-core/src/extensions/ipaddr.rs
+++ b/cedar-policy-core/src/extensions/ipaddr.rs
@@ -242,7 +242,7 @@ fn contains_at_least_two(s: &str, c: char) -> bool {
     let idx = s.find(c);
     match idx {
         Some(i) => {
-            // For this slicing operation not to panic, two preconditoins must be met:
+            // For this slicing operation not to panic, two preconditions must be met:
             // 1) `i + 1` must be <= `s.len()`
             //     This is met because `i` comes from `s.find()` meaning it must be < `s.len()`
             // 2) `i + 1` must be a character boundary
--- a/cedar-policy-core/src/extensions/ipaddr.rs
+++ b/cedar-policy-core/src/extensions/ipaddr.rs
@@ -242,7 +242,7 @@ fn contains_at_least_two(s: &str, c: char) -> bool {
     let idx = s.find(c);
     match idx {
         Some(i) => {
-            // For this slicing operation not to panic, two preconditoins must be met:
+            // For this slicing operation not to panic, two preconditions must be met:
             // 1) `i + c.len_utf()` must be <= `s.len()`
             //     This is met because:
             //     i := s.find(c) meaning:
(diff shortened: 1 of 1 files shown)


cedar-policy-core/src/extensions/ipaddr.rs
```diff
@@ -242,15 +242,45 @@ fn contains_at_least_two(s: &str, c: char) -> bool {
     let idx = s.find(c);
     match idx {
         Some(i) => {
-            // PANIC SAFETY `i` is guaranteed to be < `s.len()`, so this won't panic
+            // For this slicing operation not to panic, two preconditoins must be met:
```
cdisselkoen, line 245:
```suggestion
            // For this slicing operation not to panic, two preconditions must be met:
```

cdisselkoen (approved):

cedar-policy-core/src/extensions/ipaddr.rs
```diff
@@ -242,15 +242,45 @@ fn contains_at_least_two(s: &str, c: char) -> bool {
     let idx = s.find(c);
     match idx {
         Some(i) => {
-            // PANIC SAFETY `i` is guaranteed to be < `s.len()`, so this won't panic
+            // For this slicing operation not to panic, two preconditions must be met:
+            // 1) `i + 1` must be <= `s.len()`
```
john-h-kastner-aws, line 246:
```suggestion
            // 1) `i + c.len_utf8()` must be <= `s.len()`
```

which is less obviously correct

cedar-policy-core/src/extensions/ipaddr.rs
```diff
@@ -242,15 +242,45 @@ fn contains_at_least_two(s: &str, c: char) -> bool {
...
+            //     This is met because `i` comes from `s.find()` meaning it must be < `s.len()`
+            // 2) `i + 1` must be a character boundary
```
john-h-kastner-aws, line 248:
```suggestion
            // 2) `i + c.len_utf8()` must be a character boundary
```

and this also depends on `i` being the character boundary where  `c` was found

aaronjeline:
Some performance stats:
```
        Command being timed: "cargo kani"
        User time (seconds): 1611.86
        System time (seconds): 81.92
        Percent of CPU this job got: 100%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 28:13.66
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 105030976
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 57922306
```

shaobo-he-aws:
I think a more general and interesting question is how we can avoid this issue programatically. In this particular example, we need to add a precondition that `c` can only be `:` or `.`. I was thinking if we can add a type such that it can only be constructed via the two chars (e.g., the constructor returns an `Option`). So upon the call site of this function, we construct this type and use unwrap, which is fine and acts a warning to the developer.

shaobo-he-aws:
I think I oppose this change as it slows down parsing IP addresses. IMO, we should not try to generalize a function that we know its application is limited.

aaronjeline:
> I think a more general and interesting question is how we can avoid this issue programatically. In this particular example, we need to add a precondition that c can only be : or .. I was thinking if we can add a type such that it can only be constructed via the two chars (e.g., the constructor returns an Option). So upon the call site of this function, we construct this type and use unwrap, which is fine and acts a warning to the developer.

Yeah, the original change looked something like this (actually avoids needing an unwrap since we only call it with constants)

```
enum SafeChar { 
  Colon, 
  Period
}

impl From<SafeChar> for char { 
 ...
}

fn contains_at_least_two(s : &str, c : SafeChar) {
  let c : char = c.into();
  ...
}
```



But I found this kinda ugly, and proffered the general solution.

aaronjeline:
> I think I oppose this change as it slows down parsing IP addresses. IMO, we should not try to generalize a function that we know its application is limited.

It only slows it down on the error path, and `len_utf8()` is constant time. [source](https://doc.rust-lang.org/src/core/char/methods.rs.html#1741)

cedar-policy-core/src/extensions/ipaddr.rs
```diff
@@ -242,15 +242,45 @@ fn contains_at_least_two(s: &str, c: char) -> bool {
     let idx = s.find(c);
```
john-h-kastner-aws, line 242: I occurred to me that we could use `split_once` to avoid slicing entirely

andrewmwells-amazon (approved):

andrewmwells-amazon:
> I think I oppose this change as it slows down parsing IP addresses. IMO, we should not try to generalize a function that we know its application is limited.

I agree with Shaobo in the sense that I wouldn't have considered this worth generalizing, but now that it is generalized I don't see a problem in taking it.

zhassan-aws:
I'm a bit alarmed by Kani's excessive memory usage for this harness (>100 GB). Digging a bit into it, it seems that Kani is struggling with the string indexing operation (`s[i + c.len_utf8()..]`). The checked version of slice indexing (`std::str::get`) seems significantly easier for Kani to reason about:
```diff
$ git diff
diff --git a/cedar-policy-core/src/extensions/ipaddr.rs b/cedar-policy-core/src/extensions/ipaddr.rs
index d4341aa0..b49f6d13 100644
--- a/cedar-policy-core/src/extensions/ipaddr.rs
+++ b/cedar-policy-core/src/extensions/ipaddr.rs
@@ -252,7 +252,7 @@ fn contains_at_least_two(s: &str, c: char) -> bool {
             // safety for strings up to length 4
             // PANIC SAFETY: (see above)
             #[allow(clippy::indexing_slicing)]
-            let idx = s[i + c.len_utf8()..].find(c);
+            let idx = s.get(i + c.len_utf8()..).unwrap().find(c);
             idx.is_some()
         }
         None => false,
```

```bash
$ /usr/bin/time --verbose cargo kani
...
SUMMARY:
 ** 0 of 734 failed (20 unreachable)

VERIFICATION:- SUCCESSFUL
Verification Time: 58.97311s

Complete - 1 successfully verified harnesses, 0 failures, 1 total.
        Command being timed: "cargo kani"
        User time (seconds): 95.89
        System time (seconds): 11.43
        Percent of CPU this job got: 89%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 1:59.82
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 1462956
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 12973
        Minor (reclaiming a frame) page faults: 2544311
        Voluntary context switches: 19990
        Involuntary context switches: 523
        Swaps: 0
        File system inputs: 2336560
        File system outputs: 3189648
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

zhassan-aws:
I filed this Kani issue for investigating the memory blow-up: https://github.com/model-checking/kani/issues/3328

aaronjeline:
> > I think I oppose this change as it slows down parsing IP addresses. IMO, we should not try to generalize a function that we know its application is limited.
> 
> I agree with Shaobo in the sense that I wouldn't have considered this worth generalizing, but now that it is generalized I don't see a problem in taking it.

Yeah, again I tried narrowing the domain first using enums, but that was more code than changing `+1` to `+ c.utf8_len()`

aaronjeline:
> I'm a bit alarmed by Kani's excessive memory usage for this harness (>100 GB). Digging a bit into it, it seems that Kani is struggling with the string indexing operation (`s[i + c.len_utf8()..]`). The checked version of slice indexing (`std::str::get`) seems significantly easier for Kani to reason about:

Took this and changed the harness slightly, got much better performance and verified strings up to length `6`:

```
        Command being timed: "cargo kani"
        User time (seconds): 122.97
        System time (seconds): 10.84
        Percent of CPU this job got: 101%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 2:11.99
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 3235992
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 0
        Minor (reclaiming a frame) page faults: 4057004
        Voluntary context switches: 6140
        Involuntary context switches: 278
        Swaps: 0
        File system inputs: 0
        File system outputs: 3187888
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
        Page size (bytes): 4096
        Exit status: 0
```

cedar-policy-core/src/extensions/ipaddr.rs
```diff
@@ -242,15 +242,49 @@ fn contains_at_least_two(s: &str, c: char) -> bool {
...
     match idx {
         Some(i) => {
-            // PANIC SAFETY `i` is guaranteed to be < `s.len()`, so this won't panic
+            // For this slicing operation not to panic, two preconditoins must be met:
```
cdisselkoen, line 245:
```suggestion
            // For this slicing operation not to panic, two preconditions must be met:
```

cedar-policy-core/src/extensions/ipaddr.rs
```diff
@@ -242,15 +242,49 @@ fn contains_at_least_two(s: &str, c: char) -> bool {
...
             idx.is_some()
         }
         None => false,
     }
 }
 
+#[cfg(kani)]
+mod proof {
+
+    /// Prove that `contains_two_correct` does not panic for strings with length <= 4
```
cdisselkoen, line 272: Comment out of date with the `kani::unwind` annotation

cdisselkoen (approved):

cdisselkoen (approved):

cedar-policy-core/src/extensions/ipaddr.rs
```diff
@@ -242,15 +242,49 @@ fn contains_at_least_two(s: &str, c: char) -> bool {
...
+#[cfg(kani)]
+mod proof {
+
+    /// Prove that `contains_two_correct` does not panic for strings with length <= 6
+    #[kani::proof]
+    #[kani::unwind(7)]
+    fn contains_at_least_two_correct() {
+        let buf: [u8; 6] = kani::any();
+        let len: usize = kani::any();
+        kani::assume(len <= 4);
```
john-h-kastner-aws, line 278:
```suggestion
        kani::assume(len <= 6);
```

cdisselkoen:
I agree with @andrewmwells-amazon that now that this generalized version exists, seems fine to take it instead of trying to restrict the domain

shaobo-he-aws (approved):

Issue #950, opened 2024-06-04
CLI should support reading JSON formatted policy sets (as opposed to single policies)


john-h-kastner-aws:
### Category

CLI features/changes

### Describe the feature you'd like to request

The Cedar CLI supports JSON formatted policies by passing the `--policy-format json` argument, but this expects exactly one policy in the JSON policy. 

We recently added a JSON format for policy sets in #783, but did not add support for this format to the CLI. We should update the CLI to accept this format. It ideally should continue to accept the single policy format to avoid making an unnecessary breaking change.

### Describe alternatives you've considered

.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

memark:
The actual deserializing the policy set from json appears to be easy.

About the non-breaking behaviour, the current logic looks like this:

1. try parsing as policy
1. if fails, try parsing as template
1. if fails, return (template parsing) error

Where in this small cascade do we want to add the "try parsing as policy set" step?

john-h-kastner-aws:
The procedure I had in mind would look at JSON attributes to decided what it needs to parse.  Something like:


Parse policies as a `serde_json::Value` using `serde_json::from_str`. Error if it is not a JSON object. Then parse based on contents of object:

1. If it contains any of the attributes required for JSON policies, then try to parse as a JSON policy.
2. If it contains any of the attributes required for JSON policy set, then try to parse as a JSON policy set.
3. Otherwise, error.

memark:
Ok, sounds reasonable.

If we determine that it's not a policy set, should we then call the same logic as above (parse as policy, parse as template, error)?

muse254:
I would like to start development on this given we are heading the 'check if set and parse as set' path. 
@memark what did you mean by coupling? 

If I was to implement it that logic would remain as is, so far as I can tell
> If we determine that it's not a policy set, should we then call the same logic as above (try parse as policy, try parse as template, error)?

Granted I might be missing a lot of context but happy to try this on for size

memark:
>I would like to start development on this given we are heading the 'check if set and parse as set' path.

I'm actually planning to develop this myself. That's why I'm investigating and discussing the solution here. :)

>@memark what did you mean by coupling?

I mean that the CLI would now know the structure of the json schema, and use it to decide what parser to call. As of right now, the CLI knows nothing about this.

muse254:
[...]

PR #1038, opened 2024-07-04
Refactor json policy tests

memark:
## Description of changes

Preparatory refactoring work for issue below. Next PR will add new functionality.

Split a test into two, add another one, rename some functions.

## Issue #, if available

https://github.com/cedar-policy/cedar/issues/950


cedar-policy-cli/src/lib.rs
```diff
@@ -1189,7 +1193,7 @@ fn read_json_policy(filename: Option<impl AsRef<Path> + std::marker::Copy>) -> R
         Report::new(err).with_source_code(NamedSource::new(name, json_source.clone()))
     };
 
-    match Policy::from_json(None, json.clone()).map_err(err_to_report) {
```
memark, line 1192: This had no effect, since it was matched with `Err(_)` further down.

khieta (approved):
Thanks for the additional test cases!

Just a minor comment: I think it'd be fine to have just one `json-format` sandbox like before. Or you could call the new sandboxes `json-authorize` and `json-check-parse` - I think that `authorize` and `check_parse` may be too general. (Although, to be fair, our current sandbox naming is all over the place.)

shaobo-he-aws (approved):

andrewmwells-amazon (approved):
LGTM

memark:
> Just a minor comment: I think it'd be fine to have just one `json-format` sandbox like before. Or you could call the new sandboxes `json-authorize` and `json-check-parse` - I think that `authorize` and `check_parse` may be too general. (Although, to be fair, our current sandbox naming is all over the place.)

@khieta, in my view, the ability to read json policies is kind of a cross-cutting concern, not specific to any one subcommand. Hence I would prefer to test it using only `check-parse`, which is closer to the metal than `authorize`, but I felt it a bit out of place to remove the `authorize` test. What are you thoughts here?

Out of your alternatives, I'd prefer the two separate json-prefixed sandboxes.

khieta (approved):
New names look good, thanks!

PR #1046, opened 2024-07-08
disallow `partial_evaluation` extension in schemas

cdisselkoen:
## Description of changes

Disallows the `partial_evaluation` extension type in schemas.  This extension is only for partial evaluation and doesn't make sense in schemas.

I'm unclear if this is a breaking change, because while this was previously allowed in schemas, I'm unclear if you could successfully use the resulting schema for anything (like validating policies using the `partial_evaluation` type, validating requests, schema-based parsing of entities, etc).

### DIFF
--- a/cedar-policy-validator/src/schema_file_format.rs
+++ b/cedar-policy-validator/src/schema_file_format.rs
@@ -490,26 +490,6 @@ impl<N> SchemaType<N> {
         }
     }
 
-    /// Is this [`SchemaType`] an extension type, or does it contain one
-    /// (recursively)? Returns `None` if this is a `TypeDef` because we can't
-    /// easily properly check the type of a typedef, accounting for namespaces,
-    /// without first converting to a [`crate::types::Type`].
-    pub fn is_extension(&self) -> Option<bool> {
-        match self {
-            Self::Type(SchemaTypeVariant::Extension { .. }) => Some(true),
-            Self::Type(SchemaTypeVariant::Set { element }) => element.is_extension(),
-            Self::Type(SchemaTypeVariant::Record { attributes, .. }) => attributes
-                .values()
-                .try_fold(false, |a, e| match e.ty.is_extension() {
-                    Some(true) => Some(true),
-                    Some(false) => Some(a),
-                    None => None,
-                }),
-            Self::Type(_) => Some(false),
-            Self::TypeDef { .. } => None,
-        }
-    }
-
     /// Is this [`SchemaType`] an empty record? This function is used by the `Display`
     /// implementation to avoid printing unnecessary entity/action data.
     pub fn is_empty_record(&self) -> bool {
--- a/cedar-policy-validator/src/schema_file_format.rs
+++ b/cedar-policy-validator/src/schema_file_format.rs
@@ -17,6 +17,7 @@
 use cedar_policy_core::{
     ast::{Id, Name},
     entities::CedarValueJson,
+    extensions::Extensions,
     FromNormalizedStr,
 };
 use serde::{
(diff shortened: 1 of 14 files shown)


john-h-kastner-aws (approved):
>I'm unclear if you could successfully use the resulting schema for anything

At the very least, you could use `partial_evaluation` in some entity type you never use anywhere else.

I'd also expect that equality between two `partial_evaluation` attributes would validate (`principal.a == resource.a`). 

---

Also, could we take this opportunity to refactor a bit to ensure that the human and json formats accept the same set of extension? Perhaps by having both check against this:
https://github.com/cedar-policy/cedar/blob/32f2f0ec9bbc28ef5e300b7d20d6ce2a5197b0c1/cedar-policy-validator/src/human_schema/ast.rs#L34

shaobo-he-aws:
It looks good to me. Nit: should we add something like `Extensions::all_available_for_schema` in additional to existing `all_available` so that we do not have to special-case `partial_evaluation` everywhere?

shaobo-he-aws (approved):

cdisselkoen:
> At the very least, you could use `partial_evaluation` in some entity type you never use anywhere else. I'd also expect that equality between two `partial_evaluation` attributes would validate (`principal.a == resource.a`).

Given this, I think we should consider this breaking.  The existing behavior is probably benign, just strange; I can't see it causing any real problems (please correct me if wrong).  So I don't see a reason to consider this a bug or backport the fix to 3.x / 2.x.

We should probably step back and consider whether it's even worth making this change for 4.0, given that it is technically a breaking change and should be called out as such in changelog and docs.  I'm inclined to think yes, but we should make the decision intentionally

cdisselkoen:
> should we add something like `Extensions::all_available_for_schema` in addition to existing `all_available`

This is a good suggestion.  While looking at implementing it I realized that existing code is abusing `Extensions::ext_names()` (which returns the names of extensions) and assuming it represents all available extension _types_, which is how we got into this mess in the first place -- `partial_evaluation` was never intended to be a valid extension _type_.  Perhaps we could fix that as well, or even perhaps fixing that is enough, instead of special-casing partial_evaluation in schemas.

john-h-kastner-aws:
> So I don't see a reason to consider this a bug or backport the fix to 3.x / 2.x.

The argument to change it would be that this is morally part of the partial eval experimental feature (even if we forgot to hide it). I don't see any urgent need to fix, but we could fix it and still be nice by keeping the current behavior when the `partial-evaluation` is turned on.

In the end, there's not harm caused by this bug, and it's less work for us and users if we don't backport, so my final vote would be to not do it. 

> We should probably step back and consider whether it's even worth making this change for 4.0, given that it is technically a breaking change and should be called out as such in changelog and docs.

I'm pretty confident we should make it.

> partial_evaluation was never intended to be a valid extension type. Perhaps we could fix that as well, or even perhaps fixing that is enough, instead of special-casing partial_evaluation in schemas.

This might also change policy validation. [This file](https://github.com/cedar-policy/cedar/blob/main/cedar-policy-validator/src/extensions/partial_evaluation.rs) exists, so I expect you can call `unknown` in a policy and have it validate. I don't think we want that, unless there's a partial eval use case I'm not aware of

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -490,26 +490,6 @@ impl<N> SchemaType<N> {
         }
     }
 
-    /// Is this [`SchemaType`] an extension type, or does it contain one
```
john-h-kastner-aws, line 493: Looks like the downstream DRT failure is because it needs this function. Keep it unless you've got something else in mind?

cdisselkoen: Just thought this was unused, so removed it. Didn't think to check whether DRT used it.

john-h-kastner-aws (approved):
Looks good. My only other comment is that the human schema parser still parsers extension types based on it's own hard coded list, so it could become out of sync with the JSON parser (though only to become more restrictive).

cdisselkoen:
Working on that

cdisselkoen:
Pushed another revision where the human schema parser now also depends on the active `Extensions` instead of a hardcoded list. Review appreciated, it was a somewhat larger change.

john-h-kastner-aws (approved):

PR #1047, opened 2024-07-08
Remove custom serialization for PatternElem under the arbitrary feature

Brandon-Rozek:
## Description of changes
Removed the custom implementation of Serialize for PatternElem under the arbitrary feature.

This was originally created for the Dafny as it didn't support unicode escape sequences.

## Issue #, if available
Issue #1042 
## Checklist for requesting a review

The change in this PR is:
- [x] A change "invisible" to users (e.g., documentation, changes to "internal" crates like `cedar-policy-core`, `cedar-validator`, etc.)

I confirm that this PR:
- [x] Does not update the CHANGELOG because my change does not significantly impact released code.

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec):
- [x] Requires updates, and I have made / will make these updates myself.

I will add a comment linking to the PR in cedar-spec as soon as possible

Issue #1042, opened 2024-07-05
(De-)serialization of PatternElem under the Arbitrary feature


Brandon-Rozek:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Other

### Describe the bug

De-serializing using `serde_json` during fuzzing leads to a panic when `non_scope_constraints` contain unicode escape sequences in the `LIKE` sub-expression.

### Expected behavior

De-serialization to not lead to a panic

### Reproduction steps

1. Install Cedar and Cedar-drt
2. In the fuzz source generate a random policy when implementing `Arbitrary`
```
let schema: Schema = Schema::arbitrary(SETTINGS.clone(), u)?;
let hierarchy = schema.arbitrary_hierarchy(u)?;
let policy = schema.arbitrary_policy(&hierarchy, u)?;
```
3. Serialize and de-serialize using serde
```
let nsc: ast::Expr = policy.non_scope_constraints().to_owned();
let nsc_string = serde_json::to_string(&nsc)
    .expect("serialization failed");
let nsc_rt: ast::Expr = serde_json::from_str(nsc_string.as_ref())
    .expect("deserialization failed"); // Panic occurs here
```

### Code Snippet

```rust
// Example failure case
let expr: Expr = Expr::like(Expr::var(Var::Principal), vec![PatternElem::Char('\0')]);
let expr_string = serde_json::to_string(&expr)
    .expect("serialization failed");
let expr_rt: Expr = serde_json::from_str(expr_string.as_ref())
    .expect("deserialization failed"); // Panic occurs here
```

### Log output

```
deserialization failed: Error("invalid type: integer `0`, expected a character", line: 1, column: 459)
```

### Operating  System

Linux

### Additional information and screenshots

From some investigation, it looks like this might be because of the Serialize implementation of `PatternElem` in the file `cedar/cedar-policy-core/src/ast/pattern.rs`. It converts the character to a `u32` primitive.

The serializer is defined under the `cfg(feature = "arbitrary")` which is active during fuzzing. There isn't a corresponding `Deserialize` implementation which undoes the `u32` conversion.

[...]

PR #1050, opened 2024-07-09
Add Derive(Deserialize) to AST structs

Brandon-Rozek:
## Description of changes
Added Derive(Deserialize) to the AST structs used within `AuthorizationRequest`.

This allows us to test roundtrip equality which catches issues like #1042 

Note that this assumes that PR #1047 is merged first. 

## Issue #, if available
N/A

- This PR by itself does not affect DRT, but #1047 which this is based on does.

### DIFF
--- a/cedar-policy-core/src/ast/entity.rs
+++ b/cedar-policy-core/src/ast/entity.rs
@@ -293,7 +293,7 @@ impl<'a> arbitrary::Arbitrary<'a> for Eid {
 }
 
 /// Entity datatype
-#[derive(Debug, Clone, Serialize)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct Entity {
     /// UID
     uid: EntityUID,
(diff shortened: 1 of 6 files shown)


john-h-kastner-aws (commented):
Not sure I like this change. Needing to maintain a deserialize implementation makes it harder to change the internal representation used for these structs, and we don't need the deserialization impl for anything other than these tests

emina:
Ah, it's interesting to see that we didn't have or need this deserialization code before.  

From a testing point of view, it would be nice to retain the code so we can catch bugs (if any) in the de/serialization on the Rust side.  Though I understand John's concerns about maintaining more code just for the purpose of testing.  

Curious to see what others think.

john-h-kastner-aws:
I don't think the de-serialization make this _so_ much worse. Really, the fact that DRT depends on the serialization (and hence our internal representation) of these structures is the more annoying part.

aaronjeline:
Closed

PR #969, opened 2024-06-11
Implement RFC 52

shaobo-he-aws:
## Description of changes

Implement [RFC 52](https://github.com/cedar-policy/rfcs/blob/main/text/0052-reserved-namespaces.md)

The main idea of this PR is to use stronger types to describe valid names specified by the RFC.

## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [x] A breaking change requiring a major version bump to `cedar-policy` (e.g., changes to the signature of an existing API).

I confirm that this PR (choose one, and delete the other options):

- [x] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [x] Requires updates, and I have made / will make these updates myself. (Please include in your description a timeline or link to the relevant PR in `cedar-spec`, and how you have tested that your updates are correct.)

Issue #948, opened 2024-06-03
CLI doesn't use source locations on schema errors


john-h-kastner-aws:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Other

### Describe the bug

For some human format schema parsing errors (at least those coming from lalrpop) should know the source location where the error occurred, but the CLI isn't printing them. 

For example, we might get this error is we try to declare the entity `User` twice

```
  × Error parsing schema: Duplicate declarations: `User`
  ╰─▶ Duplicate declarations: `User`
```

even though the schema parsing code captures the exact location

https://github.com/cedar-policy/cedar/blob/afd6a651808ae0090996c6eef74ad8c7b837f292/cedar-policy-validator/src/human_schema/err.rs#L393-L400

### Expected behavior

The CLI should include the source label

### Reproduction steps

.

shaobo-he-aws:
I believe this is already handled by #1124.

Issue #174, opened 2023-07-12
Report source location for more schema parsing errors


john-h-kastner-aws:
### Category

Cedar validation features

### Describe the feature you'd like to request

Some schema parsing errors report where the error occurred in the source JSON file. All of these errors are generated by serde when it parses JSON text into our intermediate schema representation. We want to provide similar information for the errors generated when converting the intermediate structures into the final form used for validation.

For this example schema an error is reported because the entity type `B` appears in a `memberOftypesList` without being defined, but the error does not say where this `memberOfTypes` list is in the schema document. 

```json
{
  "entityTypes": {
    "A": { "memberOfTypes": [ "B" ] }
  },
  "actions": {} 
}
```

Serde does not have an obvious mechanism for passing source location through to deserializeed structures. I looked into getting at the source locations a bit and found this: https://github.com/serde-rs/json/issues/637

It definitely looks likes it's possible. There's a [crate linked](https://crates.io/crates/json-spanned-value) that does something like what we want. It's not quite exactly what we need, but the implementation shows how we could do this ourselves.

(Original discussed in #149)

### Describe the solution you'd like

.

### Describe alternatives you've considered

.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

john-h-kastner-aws:
With work on the custom schema format well under way (#557), this issues is less of a priority. The design for that parser should support reporting source information from the beginning.

That said, we're interested in supporting source information on _other_ json parsing errors (#599). If we can solve the problem of getting source information out of serde in those parsers, then we should be able to apply the same technique here without too much additional work.

cdisselkoen:
Possibly useful for this: https://docs.rs/serde_spanned/0.6.7/serde_spanned/

PR #1060, opened 2024-07-11
Resolve types at fragment-combine time

### LINKED ISSUE #579: Schema files should be able to refer to types in the empty namespace

### LINKED ISSUE #711: De-duplicate code for checking if entity types exist in schema

The human-readable schema format included some duplicate checks that check for the existence of an entity type when it is used in a schema. This checks is already done when construct a `ValidatorSchema` object, so it shouldn't be necessary to also perform the check when parsing the schema.  

_No response_

### Additional context

_No response_

cdisselkoen:
## Description of changes

Large change to typename resolution during schema construction.

Previously, typenames were always implicitly qualified with the current namespace.  This applied to both typenames _being defined_, and typenames _being referenced_.  Unfortunately, this prevented users from writing an unqualified typename `Foo` to refer to a definition `Foo` that appears in the empty namespace: the appearance of an unqualified typename `Foo` was assumed to refer to `NS::Foo` instead (if the current namespace was `NS`), even if there was no `Foo` defined in `NS`, causing an error.  This is #579.

Now, an unqualified typename reference `Foo` may resolve to either `NS::Foo` or empty-namespace `Foo` (assuming `NS` is the current namespace), depending on which of these are defined, with `NS::Foo` taking priority.  Importantly, this means we cannot resolve the unqualified typename reference `Foo` until all fragments have been combined together, because either `NS::Foo` or empty-namespace `Foo` could be defined in a different fragment.

To support this, this PR introduces a new type `ConditionalName` representing a name that could resolve to any of a list of candidate fully-qualified names, in a priority order, depending on which of them exist at fragment-combine time.  Full typename resolution is deferred until fragment-combine time, at which point the `ConditionalName`s are all resolved into fully-qualified `Name`s.

Relatedly, human-syntax processing code currently (before this PR) has its own typename-resolution algorithm, for determining when an unqualified typename refers to an entity type, a common type, a primitive type, or an extension type.  For entity and common types, it only considers types defined in the current namespace, not the empty namespace, so this equally suffers from #579.  However, now (in this PR) this processing also cannot be done until all fragments have been combined together, because either entity or common types in either the current or empty namespace could be defined in other fragments, and the presence of those definitions could change how the typename resolves.

This PR removes the typename-resolution code that previously applied only to the human-syntax processing.  Instead, both human-syntax and JSON-syntax schemas rely on the same typename resolution algorithm, which runs at fragment-combine time.  This increases consistency between the two formats and hopefully avoids a class of bugs that could appear in the future if resolution were handled differently in the two formats.

Another consequence of unifying the typename-resolution algorithms used by the human syntax and JSON syntax is that `__cedar::String`, `__cedar::ipaddr`, etc become valid aliases for `String`, `ipaddr`, etc (respectively) in the JSON syntax.  (Previously, these were only valid aliases in the human syntax; they are not needed for expressiveness in the JSON syntax.). Consequently, this PR also lifts the restriction on defining common types named `Long`, `String`, etc in the JSON syntax (before this PR, those were errors).  (so now, after this PR, the `__cedar::` aliases may actually be needed in the JSON syntax for any JSON schema that uses the newfound capability to define a common type like `Long` or `String`.)  I see no problem with any of the changes described in this paragraph, and they're all backward-compatible, but I'm calling these changes out for awareness and in case reviewers have other opinions.

Another change in this PR worth calling out, is that in the new typename resolution code, the `__cedar` namespace is significantly less special: at fragment-combine time, we simply add a completely normal schema fragment containing the definitions in the `__cedar` namespace.  We also add aliases (completely normal common-type definitions) to the empty namespace that map `Bool` to `__cedar::Bool`, etc, but not overriding any user definition of `Bool` etc in the empty namespace (in that case, they will need to use the `__cedar` prefix to reference the primitive type `Bool`).  This seems like a much better organization to me, with typename resolution not needing to special-case primitive/extension types or the `__cedar` namespace nearly as much.

This PR includes a large number of tests, outlined in the comments on the new `test_579` module (and inspired by the discussion on #579).  As of this writing, the PR contains only tests in groups A, B, and C.  I believe the current code will fail tests in some of the other categories -- for instance, referring to an empty-namespace entity type in an entity parent list.

I recommend reviewing the files in this PR in a different order than GitHub displays them: start with the changes to `raw_name.rs` (the definitions of `RawName` and `ConditionalName`) and `schema.rs` (which contains the code for combining fragments into a `ValidatorSchema`).  Then the other changes may make more sense in context.

## Issue #, if available

Fixes #579 and fixes #711


cedar-policy-validator/src/err.rs
```diff
@@ -326,15 +331,51 @@ pub mod schema_errors {
...
+
+        /// Combine all the errors into a single `TypeResolutionError`.
+        ///
+        /// Unlike `join()`, this cannot fail, because `NonEmpty` guarantees
+        /// there is at least one error to join.
+        pub(crate) fn join_nonempty(errs: NonEmpty<TypeResolutionError>) -> Self {
+            // PANIC SAFETY: Since the input is `NonEmpty`, there must be at least one error, so `NonEmpty::collect()` must return `Some`
+            #[allow(clippy::expect_used)]
+            Self(
+                NonEmpty::collect(errs.into_iter().flat_map(|err| err.0))
```
khieta, line 374: I think you can do something like this to avoid the `expect`
```suggestion
                errs.flat_map(|err| err.0)
```

cedar-policy-validator/src/err.rs
```diff
@@ -150,11 +150,19 @@ pub enum SchemaError {
...
+    /// entity or context attribute. (because it refers to an entity type
+    /// or common type that was not declared.)
     #[error(transparent)]
     #[diagnostic(transparent)]
-    UndeclaredCommonTypes(#[from] schema_errors::UndeclaredCommonTypesError),
+    TypeResolution(#[from] schema_errors::TypeResolutionError),
+    /// This error occurs when we cannot resolve a common-type reference.
+    /// Normally, such errors would be `TypeResolution` instead; I think
+    /// this variant should never occur, but not positive, and hard to guarantee
+    /// structurally.
```
khieta, line 162: Elsewhere (e.g. [here](https://github.com/cedar-policy/cedar/blob/b653e6c0627423b24bed50ee664b0302c512f16a/cedar-policy-core/src/parser/err.rs#L228)) we've named these error variants `...InvariantViolation` and included help text asking the user to file as issue.

khieta (commented):
Partial review for now since I'll be OOTO tomorrow

cedar-policy-validator/src/schema/raw_name.rs
```diff
@@ -105,3 +167,143 @@ impl std::str::FromStr for RawName {
...
+                "neither `{}` nor `{}` refers to anything that has been declared",
+                self.possibilities[0], self.possibilities[1]
+            ),
+            _ => format!(
+                "none of these have been declared: {}",
+                self.possibilities
+                    .iter()
+                    .map(|p| format!("`{p}`"))
+                    .join(", ")
+            ),
```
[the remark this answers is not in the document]
cdisselkoen: You're correct.  There at one point was a version of this code where there could be 3 possibiltiies (namespace, empty namespace, __cedar namespace) but in the latest commit on this PR, there is only ever 1 or 2 possibilities.  Still seems like good future-proofing to support more than 2 though.

cedar-policy-validator/src/schema.rs
```diff
@@ -2571,23 +2767,1487 @@ mod test {
...
+/// We test all possible (position, scenario) pairs where:
+///
+/// position is all places a typename can occur in a schema:
+/// A. Inside a context attribute type
+/// B. Inside an entity attribute type
+/// C. Inside the body of a common-type definition
+/// D. As an entity parent type
+/// E. In an action `appliesTo` declaration
+/// F. Inside an action attribute type?
+/// G. In an action parent declaration?
```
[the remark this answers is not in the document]
cdisselkoen: There are only tests for A-C right now actually, as noted in the PR description.  I believe the current code will fail some of the D-G tests.

cedar-policy-validator/src/schema.rs
```diff
@@ -2571,23 +2767,1487 @@ mod test {
...
+            entity MyType;
+            {mytype_use}
+        }}
+        "#
+        )
+    }
+
+    /// Makes a schema for all the XXa1 test cases, where different XX plug in a
+    /// different schema construct that uses `MyType` (e.g., with a function
+    /// like `A1_json()`).
```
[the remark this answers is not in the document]
cdisselkoen: Good catch

cedar-policy-validator/src/schema.rs
```diff
@@ -2571,23 +2767,1487 @@ mod test {
...
+    #[test]
+    fn A1c_human() {
+        assert_parse_error_human(
+            &c_human(A1_human()),
+            &ExpectedErrorMessageBuilder::error("failed to resolve type: MyType")
+                .help(
+                    "neither `NS1::MyType` nor `MyType` refers to anything that has been declared",
+                )
+                // TODO: why don't we have an underline here? (Uncommenting this produces test failure)
+                //.exactly_one_underline("MyType")
```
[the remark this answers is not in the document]
cdisselkoen: Investigated this briefly and I think it's just #174 again -- the JSON structures don't carry source locations, and since the human processing works by first converting to the JSON structures, we lose the source locations at that point.  Note this is a regression, as previously some errors were detected during typename resolution before converting to JSON, and came with source locations, whereas now, typename resolution happens after converting to JSON and thus loses source locations.

cedar-policy-validator/src/err.rs
cdisselkoen: Nice, TIL

cedar-policy-validator/src/schema.rs
```diff
@@ -2571,23 +2767,1487 @@ mod test {
             }
         });
         let schema = ValidatorSchema::from_json_value(src.clone(), Extensions::all_available());
-        assert_matches!(schema, Err(SchemaError::CommonTypeNameConflict(CommonTypeNameConflictError(n))) if n == "String".parse().unwrap());
+        assert_matches!(schema, Ok(_));
     }
 }
 
+/// Tests described in https://github.com/cedar-policy/cedar/issues/579
+///
```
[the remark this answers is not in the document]
cdisselkoen: Done

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -476,10 +498,26 @@ pub enum SchemaType<N> {
...
+        /// this variant can never be constructed by the `Deserialize`
+        /// implementation, as any structure that would fit this would
+        /// first fit `CommonTypeRef` above.
+        /// This is intended; the JSON syntax in question is specified to
+        /// resolve to only a common type.
+        /// This variant (`EntityOrCommonTypeRef`) exists only for the use
+        /// of the human-schema parser, so that we can represent all valid
+        /// human-syntax schemas in this format.
+        #[serde(rename = "type")]
+        type_name: N,
```
john-h-kastner-aws, line 514: Will we every try to serialize this variant? That could be a problem if it's holding something that's ultimately meant to be an entity type.

cedar-policy-validator/src/schema/test_579.rs
```diff
@@ -0,0 +1,1431 @@
...
+    assert_parses_successfully_human(&b2_human(C1_human()));
+}
+#[test]
+fn C1c_human() {
+    assert_parse_error_human(
+        &c_human(C1_human()),
+        &ExpectedErrorMessageBuilder::error("failed to resolve type: MyType")
+            .help("neither `NS1::MyType` nor `MyType` refers to anything that has been declared")
+            // TODO: why don't we have an underline here? (Uncommenting this produces test failure)
+            //.exactly_one_underline("MyType")
```
john-h-kastner-aws, line 983: Could be the same cause as #948

cedar-policy-validator/src/err.rs
```diff
@@ -326,15 +331,51 @@ pub mod schema_errors {
...
+        pub(crate) fn join(errs: impl IntoIterator<Item = TypeResolutionError>) -> Option<Self> {
+            NonEmpty::collect(errs.into_iter().flat_map(|err| err.0)).map(Self)
+        }
+
+        /// Combine all the errors into a single `TypeResolutionError`.
+        ///
+        /// Unlike `join()`, this cannot fail, because `NonEmpty` guarantees
+        /// there is at least one error to join.
+        pub(crate) fn join_nonempty(errs: NonEmpty<TypeResolutionError>) -> Self {
+            // PANIC SAFETY: Since the input is `NonEmpty`, there must be at least one error, so `NonEmpty::collect()` must return `Some`
```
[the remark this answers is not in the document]
john-h-kastner-aws: Looks like you found this already

cedar-policy-validator/src/human_schema/to_json_schema.rs
```diff
@@ -81,22 +84,23 @@ fn is_valid_ext_type(ty: &Id, extensions: Extensions<'_>) -> bool {
...
-        &Namespace {
-            name: None,
-            decls: vec![],
+/// Convert a `Type` into the JSON representation of the type.
+pub fn human_type_to_json_type(ty: Node<Type>) -> SchemaType<RawName> {
+    match ty.node {
+        Type::Set(t) => SchemaType::Type(SchemaTypeVariant::Set {
+            element: Box::new(human_type_to_json_type(*t)),
+        }),
+        Type::Ident(p) => SchemaType::EntityOrCommonTypeRef {
```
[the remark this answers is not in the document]
cdisselkoen: In this PR's implementation, `Long` and `Bool` will always be common type refs; we define them as common-type aliases for `__cedar::Long` and `__cedar::Bool` respectively.

cedar-policy-validator/src/human_schema/fmt.rs
```diff
@@ -183,6 +184,23 @@ impl NameCollisionsError {
...
+///
+/// REVIEW: The existing code throws an error if any fully-qualified name in a
+/// non-empty namespace is a valid common type and also a valid entity type.
+/// 1) I think this check is more conservative than necessary? Schemas are
+/// allowed to shadow an entity type with a common type declaration in the same
+/// namespace; see RFCs 24 and 70. What the human syntax can't express is if, in
+/// that situation, we then specifically refer to the shadowed entity type name.
+/// But it's harder to walk all type references than it is to walk all type
+/// declarations, so perhaps the conservative code here is fine.
+/// 2) At the same time, this check might miss some cases that are actually
```
john-h-kastner-aws, line 196: >  When converted to human syntax, the reference will resolve to the common type declaration in the current namespace instead.

Sounds like a bug?

cedar-policy-validator/src/schema/test_579.rs
cdisselkoen: I actually think it's #147.  See [this thread](https://github.com/cedar-policy/cedar/pull/1060#discussion_r1674552526)

cedar-policy-validator/src/schema.rs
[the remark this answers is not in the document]
cdisselkoen: Added tests for D and E.  (D tests required code changes -- the first version of this PR didn't apply the same resolution algorithm to entity parent declarations.)  F tests don't make sense because action attributes are not _typenames_, they are actual _attribute values_, so we can't use the same resolution algorithm.  And also, action attributes are not supported in the human syntax, only the JSON syntax.  I renamed G to F and am working on that.  But that one is subtle too -- G requires not that the _typenames_ conditionally resolve, but that _action EUIDs_ conditionally resolve.  Not based on whether the typename `NS1::Action` exists, but whether `NS1::Action::"foo"` was declared.  It's unclear whether #579 envisioned this.  Do we want `action Read in [Group]` to always unconditionally refer to `NS::Action::"Group"` where `NS` is the current namespace (current behavior)?  Or do we want it to conditionally resolve to `Action::"Group"` if `NS::Action::"Group"` doesn't exist?

cedar-policy-validator/src/human_schema/to_json_schema.rs
```diff
@@ -177,361 +152,228 @@ impl<'a> ConversionContext<'a> {
...
+        // Convert entity type decls
+        let entity_types = entity_types
+            .into_iter()
+            .map(convert_entity_decl)
+            .flatten()
+            .collect();
+
+        // Convert action decls, collecting all errors
+        let actions =
+            collect_all_errors(action.into_iter().map(convert_action_decl))?.collect::<Vec<_>>();
```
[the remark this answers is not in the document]
cdisselkoen: Removed this `collect` in the latest commit

cedar-policy-validator/src/schema.rs
```diff
@@ -674,16 +768,91 @@ impl TryInto<ValidatorSchema> for NamespaceDefinitionWithActionAttributes<RawNam
...
 }
 
+/// Get a `ValidatorSchemaFragment` describing the items that implicitly exist
+/// in the `__cedar` namespace.
+fn cedar_fragment(extensions: Extensions<'_>) -> ValidatorSchemaFragment<ConditionalName> {
+    // PANIC SAFETY: these are valid `Id`s
+    #[allow(clippy::unwrap_used)]
+    let mut common_types = HashMap::from_iter([
+        (
+            Id::from_str("Bool").unwrap(),
```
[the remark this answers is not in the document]
cdisselkoen: That constructor is currently private to Core.  Should we make it public from Core?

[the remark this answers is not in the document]
cdisselkoen: For hotness, this function is called exactly once per `ValidatorSchema` that is constructed.  Which is probably not hot enough to matter.  There are a lot of other things going on when constructing a `ValidatorSchema` that the performance here will probably be lost in the noise.  Maybe if we benchmarked specifically constructing an empty schema this would be a noticeable component of the runtime

cedar-policy-validator/src/schema_file_format.rs
cdisselkoen: You're correct. I'm not positive why this type is `Serialize` in the first place.  Only for DRT?  Does DRT serialize this format or the `ValidatorSchema` format?

cedar-policy-validator/src/human_schema/fmt.rs
cdisselkoen: While trying to reproduce I wondered, do we have a fuzz target that generates a JSON schema, converts to natural, and converts back?

john-h-kastner-aws: That should be https://github.com/cedar-policy/cedar-spec/blame/main/cedar-drt/fuzz/fuzz_targets/convert-schema-json-to-human.rs, but that's using arbitrary strings, so not the best coverage

cdisselkoen: Filed the bug as #1063.  It affects at minimum 3.2.1, I would guess all versions back to 3.1.0 which introduced the human/natural schema syntax.

cedar-policy-validator/src/schema/namespace_def.rs
```diff
@@ -612,11 +887,13 @@ impl TryInto<ValidatorNamespaceDef> for NamespaceDefinition<RawName> {
...
 
-/// Take an action identifier as a string and use it to construct an
-/// [`EntityUID`] for that action. The entity type of the action will always
-/// have the base type `Action`. The type will be qualified with any
-/// namespace provided in the `namespace` argument or with the namespace
-/// inside the [`ActionEntityUID`] if one is present.
+/// Take an action identifier and use it to construct a fully-qualified
+/// [`EntityUID`] for that action.
+///
+/// The entity type of the action will always have the base type `Action`.
```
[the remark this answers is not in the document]
cdisselkoen: As of the current commit of this PR, yes.  I assume that is a thing we want to allow though?

Quoting from [this comment](https://github.com/cedar-policy/cedar/pull/1060#discussion_r1676180619):
> [...] requires not that the typenames conditionally resolve, but that action EUIDs conditionally resolve. Not based on whether the typename NS1::Action exists, but whether NS1::Action::"foo" was declared. It's unclear whether https://github.com/cedar-policy/cedar/issues/579 envisioned this. Do we want `action Read in [Group]` to always unconditionally refer to `NS::Action::"Group"` where `NS` is the current namespace (current behavior)? Or do we want it to conditionally resolve to `Action::"Group"` if `NS::Action::"Group"` doesn't exist?

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -252,15 +257,18 @@ pub struct EntityType<N> {
...
+    pub fn conditionally_qualify_type_references(
+        self,
+        ns: Option<&Name>,
+    ) -> EntityType<ConditionalName> {
         EntityType {
             member_of_types: self
                 .member_of_types
                 .into_iter()
-                .map(|rname| rname.qualify_with(ns))
+                .map(|rname| rname.conditionally_qualify_with(ns, ReferenceType::Entity)) // REVIEW: Should we allow writing common-types in this position, as long as the common type resolves to an entity type?
```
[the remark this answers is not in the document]
cdisselkoen: I think it makes sense to punt this change to a future PR, particularly because there are concerns about a breaking change.  (I don't _think_ that anything in this PR is a breaking change -- that it only takes schemas that used to not parse because of attempted reference to an empty-namespace type, and allows them to parse.  There is a regression in error details discussed [here](https://github.com/cedar-policy/cedar/pull/1060#discussion_r1674552526), but I think we should accept that given that #579 is a higher-priority bug.)

cedar-policy-validator/src/schema_file_format.rs
john-h-kastner-aws: Does human-to-json conversion use this serialization?

cedar-policy-validator/src/schema_file_format.rs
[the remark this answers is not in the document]
john-h-kastner-aws: Sounds fine to me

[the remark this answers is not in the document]
cdisselkoen: Punted to #1064

cedar-policy-validator/src/schema/namespace_def.rs
```diff
@@ -157,41 +250,57 @@ impl ValidatorNamespaceDef {
...
 /// their corresponding [`SchemaType`]. The common type [`Name`]s (keys in the
-/// map) are fully qualified, and inside the [`SchemaType`]s (values in the
-/// map), all entity/common type references are also fully qualified.
+/// map) are fully qualified, but inside the [`SchemaType`]s (values in the
+/// map), entity/common type references may or may not be fully qualified yet,
+/// depending on `N`; see notes on [`SchemaType`].
 #[derive(Debug)]
-pub struct TypeDefs {
-    pub(super) type_defs: HashMap<Name, SchemaType<Name>>,
+pub struct TypeDefs<N> {
```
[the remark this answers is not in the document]
cdisselkoen: #1065

cedar-policy-validator/src/schema_file_format.rs
cdisselkoen: By the classic method of "remove it and see what breaks", I determined that the serialization of `SchemaType` is used in only the following places (at least in the `cedar` repo):
- `SchemaFragment::to_json_value()` and `SchemaFragment::as_json_string()` in the public API
- some tests that check we can roundtrip `SchemaFragment` into JSON and back

I didn't check whether it is also used by DRT.

cdisselkoen: This means that the issue you point out at the top of this thread is indeed a problem.  If:
* someone has a natural syntax schema
* calls `SchemaFragment::from_str_natural()` in the public API (producing a `SchemaFragment` that may use this variant)
* then calls `SchemaFragment::.to_json_value()` or `SchemaFragment::as_json_string()` in the public API

If they later convert that JSON back to a schema it might change meaning.

In fact, the CLI's `translate_schema_to_json()` makes exactly this call sequence...

cedar-policy-validator/src/schema.rs
```diff
@@ -674,16 +768,91 @@ impl TryInto<ValidatorSchema> for NamespaceDefinitionWithActionAttributes<RawNam
...
+}
+
+/// Get the names of all primitive types, as unqualified `Id`s
+fn primitive_types() -> impl Iterator<Item = Id> {
+    // PANIC SAFETY: these are valid `Id`s
+    #[allow(clippy::unwrap_used)]
+    [
+        Id::from_str("Bool").unwrap(),
+        Id::from_str("Long").unwrap(),
+        Id::from_str("String").unwrap(),
```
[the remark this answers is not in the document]
cdisselkoen: Done

cedar-policy-validator/src/err.rs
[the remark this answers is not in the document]
cdisselkoen: @khieta pointed me to it

cedar-policy-validator/src/err.rs
cdisselkoen: Renamed to `CommonTypeInvariantViolationError` and added appropriate help text

cedar-policy-validator/src/human_schema/to_json_schema.rs
```diff
@@ -46,12 +46,20 @@ use super::{
...
     schema: Schema,
     extensions: Extensions<'_>,
 ) -> Result<(SchemaFragment<RawName>, impl Iterator<Item = SchemaWarning>), ToJsonSchemaErrors> {
-    // First pass, figure out what each name is bound to
+    // combine all of the declarations in unqualified (empty) namespaces into a
+    // single unqualified namespace
+    //
+    // TODO: In the future, we could also combine namespaces with matching
+    // non-empty names, to allow reopening a namespace within the same (human)
+    // schema fragment.
```
khieta, line 62: The comment implies that we allow it in the JSON, but not human-readable syntax? That seems worth filing an issue for.

cedar-policy-validator/src/schema/test_579.rs
```diff
@@ -0,0 +1,2183 @@
+//! Tests described in https://github.com/cedar-policy/cedar/issues/579
```
khieta, line 1: Missing standard copyright header

cedar-policy-validator/src/schema/test_579.rs
```diff
@@ -0,0 +1,2183 @@
...
+            .help("`NS2::MyType` has not been declared as an entity type")
+            .build(),
+    );
+}
+
+// json versions
+#[test]
+fn A1a1_json() {
+    assert_parses_successfully_json(A1X1_json(a1_json()));
+}
```
khieta, line 1550: Might be nice to combine the human & json versions into a single test so we can easily see if there are cases where the two formats have different behavior.

khieta (commented):
Some more comments. Will finish reviewing the last few files later today.

cedar-policy-validator/src/human_schema/to_json_schema.rs
cdisselkoen: We can't allow reopening a namespace within the same JSON schema fragment, because the JSON schema structure is a map from namespace name to declarations, so it would be a duplicate key in the JSON map.  However, we could allow it in the human-readable syntax.

shaobo-he-aws (approved):

cedar-policy-validator/src/schema.rs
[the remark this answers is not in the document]
cdisselkoen: I'm going to proceed assuming the latter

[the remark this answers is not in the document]
khieta: I think that makes sense.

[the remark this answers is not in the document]
cdisselkoen: Pushed the F tests, and they pass.  This required some medium-weight code changes to support resolving action EUIDs, not just typenames.

cedar-policy-validator/src/schema/namespace_def.rs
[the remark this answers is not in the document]
cdisselkoen: The latest commit of this PR now supports referencing an unqualified action from inside a namespace.  Let's move further comments to the linked thread.

cedar-policy-validator/src/human_schema/to_json_schema.rs
john-h-kastner-aws: reopening a namespace between fragments should work.

cedar-policy-validator/src/err.rs
```diff
@@ -234,6 +242,37 @@ impl From<transitive_closure::TcError<EntityUID>> for SchemaError {
...
+        } else {
+            let (action_res_errors, other_errors): (Vec<_>, Vec<_>) =
+                other_errors.into_iter().partition_map(|e| match e {
+                    SchemaError::ActionResolution(e) => Either::Left(e),
+                    _ => Either::Right(e),
+                });
+            if let Some(errs) = NonEmpty::from_vec(action_res_errors) {
+                schema_errors::ActionResolutionError::join_nonempty(errs).into()
+            } else {
+                // PANIC SAFETY: `other_errors` was created by partitioning a `NonEmpty` into what we now know is an empty vector and this, so this cannot be empty
```
john-h-kastner-aws, line 268: nit: this `other_errors` was created by partition the shadowed `other_errors` which is a `Vec`, but shadowed `other_errors` wasn't empty by the same logic applied here.

cedar-policy-validator/src/err.rs
```diff
@@ -234,6 +242,37 @@ impl From<transitive_closure::TcError<EntityUID>> for SchemaError {
...
+                #[allow(clippy::expect_used)]
+                other_errors.into_iter().next().expect("cannot be nonempty")
```
john-h-kastner-aws, line 270:
```suggestion
                other_errors.into_iter().next().expect("cannot be empty")
```

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -441,11 +462,100 @@ impl<N: std::fmt::Display> std::fmt::Display for ActionEntityUID<N> {
...
+        return Err(ActionResolutionError(nonempty!(self)));
+    }
+
+    /// Get the possible fully-qualified [`ActionEntityUID<Name>`]s which this
+    /// [`ActionEntityUID<ConditionalName>`] might resolve to, in priority order
+    /// (highest-priority first).
+    pub(crate) fn possibilities(&self) -> impl Iterator<Item = ActionEntityUID<Name>> + '_ {
+        // PANIC SAFETY: by INVARIANT on self.ty
+        #[allow(clippy::expect_used)]
+        let ty = self.ty.as_ref().expect("by INVARIANT on self.ty");
```
john-h-kastner-aws, line 526: Perhaps encapsulate this with

```rust
impl ActionEntityUID<Name> {
  fn ty(&self) -> &Name {
    ...
  }
}
```

john-h-kastner-aws (approved):

cedar-policy-validator/src/schema/test_579.rs
cdisselkoen: Completed this reorg

cedar-policy-validator/src/schema.rs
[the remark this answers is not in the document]
cdisselkoen: Updated these TODO comments with an explanation and pointing to #174.  In my opinion we need to accept this regression; merging this PR and fixing #579 is more important.

cedar-policy-validator/src/human_schema/fmt.rs
```diff
@@ -183,6 +184,23 @@ impl NameCollisionsError {
 }
 
 /// Convert a [`SchemaFragment`] to a string containing human schema syntax
+///
+/// REVIEW: The existing code throws an error if any fully-qualified name in a
+/// non-empty namespace is a valid common type and also a valid entity type.
+/// 1) I think this check is more conservative than necessary? Schemas are
```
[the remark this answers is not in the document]
cdisselkoen: Updated this comment

cedar-policy-validator/src/schema/test_579.rs
cdisselkoen: Added

cedar-policy-validator/src/schema/namespace_def.rs
```diff
@@ -248,81 +409,159 @@ impl EntityTypesDef {
...
+                .shape
+                .into_inner()
+                .conditionally_qualify_type_references(schema_namespace),
+            parents: schema_file_type
                 .member_of_types
                 .into_iter()
-                .map(|raw_name| raw_name.qualify_with(schema_namespace).into())
+                .map(|raw_name| {
+                    raw_name.conditionally_qualify_with(schema_namespace, ReferenceType::Entity)
+                }) // REVIEW: here's another place to adjust if we decide we want to allow common types in this position
```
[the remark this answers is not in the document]
cdisselkoen: I think there is some value in defining a common type that points to an entity type.

Simple example (there may be more):
Suppose I have a namespace or fragment that uses `User` a lot, and I want to be able to flip between two different definitions of `User` for testing, or maybe because I’m in the process of migrating, or something.  Then I can change between `type User = User1;` and `type User = User2;` without changing `User` everywhere it appears.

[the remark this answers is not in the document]
cdisselkoen: At any rate, this was punted to #1064; we can continue this discussion there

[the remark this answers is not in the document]
cdisselkoen: Edited this comment to match the others that point to 1064

cedar-policy-validator/src/schema.rs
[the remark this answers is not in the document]
john-h-kastner-aws: I think that sounds fine. Although, this could be tracked separately from #174. The work described there covers capturing source information out of serde, but we only need to carry source info in the schema  (without capturing it tn the JSON case) to fix this regression.

cdisselkoen:
Just pushed a major overhaul involving merging in `main`.  This was nontrivial because of #969.  As discussed with maintainers, the proposal in this PR is to use `InternalName` (renamed from `UncheckedName`) in `ValidatorSchema` in order to allow `__cedar` names to be used internally.

cedar-policy-validator/src/schema.rs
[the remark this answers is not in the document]
cdisselkoen: Created #1084 for this.

cedar-policy-validator/src/schema_file_format.rs
cdisselkoen: Thinking some more about this.

- The human syntax allows references that might resolve to either entity or common types, unknown until we have all fragments.
- The JSON syntax requires knowing whether a reference resolves to an entity or a common type.
- Supporting the current public-API `SchemaFragment::to_json_value()` requires being able to serialize any `SchemaFragment` to JSON before we necessarily have all fragments.

I think these three bullets are incompatible; one of them has to give.  I don't think we want to change the first bullet.  So either we have to add support for entity-or-common references to the JSON syntax, or we have to change/remove the existing `SchemaFragment::to_json_value()` API (perhaps replace with just `Schema::to_json_value()`, which currently doesn't exist).

cedar-policy-validator/src/human_schema/to_json_schema.rs
```diff
@@ -46,12 +46,20 @@ use super::{
 /// This will let you subsequently decode that into the Validator AST for Schemas ([`crate::ValidatorSchema`]).
 /// On success, this function returns a tuple containing:
 ///     * The `SchemaFragment`
-///     * A vector of name collisions, that are essentially warnings
+///     * An iterator of warnings that were generated
+///
+/// TODO: In the future, these warnings should be generated later in the
+/// process, such that we apply the same checks to JSON and human schemas
```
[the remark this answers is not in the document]
cdisselkoen: #1085

cedar-policy-validator/src/schema_file_format.rs
khieta: Tbh I think it's weird that we don't support `.to_json()` or `.to_string()` on the `Schema` type, since that's the type users will interact with for validation. So I'm in favor of adding those.

john-h-kastner-aws: It is unfortunately not so straight forward unless we define a new json format which is json-list-of-json-schema-fragments

cdisselkoen: would it be possible to serialize as a single fragment?  By the time we're at `ValidatorSchema` we've lost all information of what came from what fragment (or how many fragments there were) anyway

cedar-policy-validator/src/human_schema/to_json_schema.rs
cdisselkoen: Opened #1086 for this, and pointed this TODO at that issue

cedar-policy-validator/src/human_schema/to_json_schema.rs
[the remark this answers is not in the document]
cdisselkoen: Added the issue number to this comment. Resolving.

cedar-policy-validator/src/schema_file_format.rs
john-h-kastner-aws: Good point. That should be possible. Still not trivial since IIRC, we don't have a function to convert squash into a single fragment, but that's possibly the nicer option.

cedar-policy-validator/src/schema.rs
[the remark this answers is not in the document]
cdisselkoen: Updated the comments to point to 1084 instead of 174.  Resolving this.

cedar-policy-validator/src/schema_file_format.rs
cdisselkoen: I'd like to figure out a way to punt this from this PR (which is already way too large without also writing a way to serialize `ValidatorSchema`).  Any ideas?

Alternately, we can make a separate PR to add serialization for `ValidatorSchema` and block this PR on that one.

cedar-policy-validator/src/err.rs
cdisselkoen: Updated this comment

cedar-policy-validator/src/err.rs
cdisselkoen: Fixed

cedar-policy-validator/src/schema_file_format.rs
cdisselkoen: Done, and improved all the invariant comments

cedar-policy-validator/src/schema_file_format.rs
khieta: I think it makes sense to make a separate PR to block this one.

cdisselkoen: I believe the mirror case is also a problem -- serializing a single fragment to the human syntax.  For symmetry, I'll remove both serialization forms from single fragment.

shaobo-he-aws (approved):

cedar-policy-validator/src/schema_file_format.rs
cdisselkoen: Following in-person discussion we reached these conclusions:

- Writing a serialization of `ValidatorSchema` is quite a bit of work. For instance, having to re-reverse the parents relation, which was reversed while constructing the `ValidatorSchema`. Further, this serialization will be error-prone, need thorough tests, etc.
- Writing the serialization as a list of `SchemaFragment` is probably not significantly easier. We'd still have to know for each reference whether it resolves to entity or common, so we'd have to run the full resolution algorithm, then abort it and turn things back into the `SchemaFragment` structures.
- The easier path is to just add a new syntax to the JSON format representing entity-or-common reference.  If we were designing the JSON format today, this is how we would do it, for closer correspondence with the human syntax.  This can be done backward-compatibly, representing the new type as `{ "type": "entityOrCommon", "name": "foo" }` meaning that `foo` can resolve to either entity type `foo` or common type `foo`.  This is a little ugly, but we care less about JSON readability now that we have the human syntax.
- No one objected to adding this syntax, and it will allow us to continue to roundtrip schema fragments between human and JSON, without resolving all typename references.

cdisselkoen: Pushed this change in the latest commit.

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -1412,8 +1540,9 @@ mod test {
             expect_err(
                 &src,
                 &miette::Report::new(e),
-                &ExpectedErrorMessageBuilder::error(r#"undeclared common type: Entity"#)
-                    .help("any common types used in entity or context attributes need to be declared in `commonTypes`")
+                &ExpectedErrorMessageBuilder::error(r#"failed to resolve type: Entity"#)
+                    // TODO: this help message could suggest that the user needs to add a `name` field
```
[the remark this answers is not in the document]
cdisselkoen: Made #1094 to track this, and linked it from this comment

cedar-policy-validator/src/schema_file_format.rs
```diff
@@ -940,6 +1319,31 @@ pub enum SchemaTypeVariant<N> {
...
+        /// `Entity`, nor between this and `SchemaType::Common`.
+        /// - To represent a must-be-entity-type reference in the JSON syntax,
+        ///     use `{ "type": "Entity", "name": "foo" }`. This ser/de as
+        ///     `SchemaType::Type(SchemaTypeVariant::Entity)`.
+        /// - To represent a must-be-common-type reference in the JSON syntax,
+        ///     use `{ "type": "foo" }`. This ser/de as
+        ///     `SchemaType::CommonTypeRef`.
+        /// - To represent an either-entity-or-common-type reference in the
+        ///     JSON syntax, use `{ "type": "EntityOrCommon", "name": "foo" }`.
+        ///     This ser/de as `SchemaType::Type(SchemaTypeVariant::EntityOrCommon`.
```
khieta, line 1339: Would a user ever want to write this syntax, or is it just used internally / to support round tripping?

khieta (approved):

cedar-policy-validator/src/schema_file_format.rs
cdisselkoen: Based on [this thread](https://github.com/cedar-policy/cedar/pull/1060#discussion_r1692049738), we support users writing this syntax, if they want to.

khieta: Sure, but there is no reason they would need to, right?

john-h-kastner-aws: Yeah, I don't expect users to write this. Presumably they know which it is, and I'd encourage users to be explicit.

cdisselkoen: There's no reason they would need to, but they might find it convenient, who knows.

I see no reason to try to artificially prevent users from writing this syntax.  It would cause problems for roundtrip properties too.

cedar-policy/CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ Cedar Language Version: 4.0
   to combine `Context`s, returning an error on duplicate keys (#1027,
   resolving #1013)
 - Additional functionality to the JSON FFI including parsing utilities (#1079)
+- Schema JSON syntax now accepts a type `EntityOrCommon` representing a typename
```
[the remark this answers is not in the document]
cdisselkoen: Done

shaobo-he-aws (approved):

PR #1062, opened 2024-07-11
Remove Unknown expression from EST for functions named "unknown"

### LINKED ISSUE #1061: EST parser and Cedar parser return different ASTs on policies with functions named "unknown"

### Bug Category

Cedar Parser

### Describe the bug

The following policy EST fails [this check](https://github.com/cedar-policy/cedar-spec/blob/main/cedar-drt/fuzz/src/parsing_utils.rs#L36-L39) in convert-policy-json-to-cedar.

```
{
    "effect": "permit",
    "principal": {  "op": "All"  },
    "action": {  "op": "All"   },
    "resource": {  "op": "All"  },
    "conditions": [
        {
            "kind": "when",
            "body": {
                "unknown": [
                    {
                        "Value": ""
                    }
                ]
            }
        }
    ]
}
```
This is due to the difference in AST expression kind built when parsing the EST (Unknown) vs parsing Cedar (ExtensionFunctionApp).

Also related to #891 

### Expected behavior

.

### Reproduction steps

.

### Code Snippet

_No response_

### Log output

_No response_

### Additional configuration

_No response_

### Operating  System

_No response_

_No response_

vasumv:
Fixes #1061 

## Description of changes
Removes `ExprNoExt::Unknown` from `est`. Conversion from `ast::Expr::Unknown` returns an extension function call with one arg literal of the unknown name. 

Test case included similar to input found by DRT.

## Issue #, if available
#1061 

### DIFF
--- a/cedar-policy-core/src/est/expr.rs
+++ b/cedar-policy-core/src/est/expr.rs
@@ -108,12 +108,6 @@ pub enum ExprNoExt {
     Var(ast::Var),
     /// Template slot
     Slot(#[cfg_attr(feature = "wasm", tsify(type = "string"))] ast::SlotId),
-    /// Unknown (for partial evaluation)
-    Unknown {
-        /// Name of the unknown
-        #[cfg_attr(feature = "wasm", tsify(type = "string"))]
-        name: SmolStr,
-    },
     /// `!`
     #[serde(rename = "!")]
     Not {
@@ -346,11 +340,6 @@ impl Expr {
         Expr::ExprNoExt(ExprNoExt::Slot(slot))
     }
 
-    /// Partial-evaluation unknowns
-    pub fn unknown(name: impl Into<SmolStr>) -> Self {
-        Expr::ExprNoExt(ExprNoExt::Unknown { name: name.into() })
-    }
-
     /// `!`
     pub fn not(e: Expr) -> Self {
         Expr::ExprNoExt(ExprNoExt::Not { arg: Arc::new(e) })
@@ -570,9 +559,6 @@ impl Expr {
                 .map_err(Into::into),
             Expr::ExprNoExt(ExprNoExt::Var(var)) => Ok(ast::Expr::var(var)),
             Expr::ExprNoExt(ExprNoExt::Slot(slot)) => Ok(ast::Expr::slot(slot)),
-            Expr::ExprNoExt(ExprNoExt::Unknown { name }) => {
-                Ok(ast::Expr::unknown(ast::Unknown::new_untyped(name)))
-            }
             Expr::ExprNoExt(ExprNoExt::Not { arg }) => {
                 Ok(ast::Expr::not((*arg).clone().try_into_ast(id)?))
             }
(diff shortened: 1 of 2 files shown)


cedar-policy-core/src/est/expr.rs
```diff
@@ -1611,7 +1602,6 @@ fn maybe_with_parens(f: &mut std::fmt::Formatter<'_>, expr: &Expr) -> std::fmt::
         Expr::ExprNoExt(ExprNoExt::Value(_)) |
         Expr::ExprNoExt(ExprNoExt::Var(_)) |
         Expr::ExprNoExt(ExprNoExt::Slot(_)) |
-        Expr::ExprNoExt(ExprNoExt::Unknown { .. }) => write!(f, "{expr}"),
```
john-h-kastner-aws, line 1614: shouldn't delete the `write!` part of this line

cedar-policy-core/src/est/expr.rs
```diff
@@ -735,7 +721,13 @@ impl From<ast::Expr> for Expr {
             ast::ExprKind::Lit(lit) => lit.into(),
             ast::ExprKind::Var(var) => var.into(),
             ast::ExprKind::Slot(slot) => slot.into(),
-            ast::ExprKind::Unknown(ast::Unknown { name, .. }) => Expr::unknown(name),
+            ast::ExprKind::Unknown(ast::Unknown { name, .. }) => {
```
john-h-kastner-aws, line 724: The natural follow-up question is if we need `ast::ExprKind::Unknown`. I'm not sure myself

cedar-policy-core/src/est/expr.rs
```diff
@@ -108,12 +108,6 @@ pub enum ExprNoExt {
     Var(ast::Var),
     /// Template slot
     Slot(#[cfg_attr(feature = "wasm", tsify(type = "string"))] ast::SlotId),
-    /// Unknown (for partial evaluation)
-    Unknown {
-        /// Name of the unknown
-        #[cfg_attr(feature = "wasm", tsify(type = "string"))]
-        name: SmolStr,
```
john-h-kastner-aws, line 115: Looking at this unknown variant closer, I see that it would accept `{ "unknown": "foo" }`, but the extension form accepts `{"unknown": [ {"value": "foo" }]}`. So, we are breaking one representation of unknown in the est. This is probably fine based on the justification that it's an experimental feature (and we're coming up on a major version anyways)

vasumv: Yes, good catch.

cedar-policy-core/src/est/expr.rs
vasumv: I think there are still many dependencies on `ast::ExprKind::Unknown`, don't think we should remove it for this PR

cedar-policy-core/src/est/expr.rs
```diff
@@ -346,11 +340,6 @@ impl Expr {
         Expr::ExprNoExt(ExprNoExt::Slot(slot))
     }
 
-    /// Partial-evaluation unknowns
-    pub fn unknown(name: impl Into<SmolStr>) -> Self {
-        Expr::ExprNoExt(ExprNoExt::Unknown { name: name.into() })
-    }
```
cdisselkoen, line 352: If it would be convenient, it would be possible to keep this constructor and just change its implementation to produce the new representation of `unknown`

cedar-policy-core/src/est/expr.rs
```diff
@@ -735,7 +721,13 @@ impl From<ast::Expr> for Expr {
...
             ast::ExprKind::Var(var) => var.into(),
             ast::ExprKind::Slot(slot) => slot.into(),
-            ast::ExprKind::Unknown(ast::Unknown { name, .. }) => Expr::unknown(name),
+            ast::ExprKind::Unknown(ast::Unknown { name, .. }) => {
+                // Create an extension call with one arg, which is the name of the unknown
+                Expr::ext_call(
+                    "unknown".to_string().into(),
+                    vec![Expr::lit(CedarValueJson::String(name))],
+                )
+            }
```
cdisselkoen, line 730: continuing my earlier comment, then we wouldn't need this change at all; this could still be a call to `Expr::unknown`

cdisselkoen (approved):

john-h-kastner-aws (approved):

PR #1059, opened 2024-07-11
rename `SchemaType::TypeDef`

cdisselkoen:
## Description of changes

I'm preparing a large PR for https://github.com/cedar-policy/cedar/issues/579, and to reduce the size I'm trying to split out some separable changes into smaller PRs. This renames `SchemaType::TypeDef` to `SchemaType::CommonTypeRef` because (1) we call these "common types" in the docs, so having "common" in the variant name improves clarity; and (2) `Def` -> `Ref` because this is not _defining_ a common-type, but rather _referencing_ a common-type.

This rename is non-breaking and doesn't change the JSON format, because of `serde(untagged)` on this enum.

## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [x] A change "invisible" to users (e.g., documentation, changes to "internal" crates like `cedar-policy-core`, `cedar-validator`, etc.)

I confirm that this PR (choose one, and delete the other options):

- [x] Does not update the CHANGELOG because my change does not significantly impact released code.

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [x] Requires updates, and I have made / will make these updates myself. (Please include in your description a timeline or link to the relevant PR in `cedar-spec`, and how you have tested that your updates are correct.)

PR #1065, opened 2024-07-12
rename some types, fields, functions, and variables related to common types

cdisselkoen:
## Description of changes

On the heels of #1059, and following [this comment](https://github.com/cedar-policy/cedar/pull/1060#discussion_r1674561634) on #1060, this PR renames a bunch of other types, fields, functions, and variables related to common types.

* Explicitly use "common type" instead of "typedef" where it makes sense, following our docs
* Distinguish between "type_defs" (definitions of common types) and "type_refs" (references to common types)
* Make a couple field names more concise (to change e.g. `self.type_defs.type_defs` into `self.common_types.defs`)

I believe that none of the changes here are breaking, or even visible outside the `cedar-policy-validator` crate.  The field renames are in structs that don't impl `Serialize` or `Deserialize`, and the types and functions are crate-private.  Happy to be corrected if I'm wrong.

### DIFF
--- a/cedar-policy-validator/src/schema.rs
+++ b/cedar-policy-validator/src/schema.rs
@@ -253,7 +253,7 @@ impl ValidatorSchema {
         fragments: impl IntoIterator<Item = ValidatorSchemaFragment>,
         extensions: Extensions<'_>,
     ) -> Result<ValidatorSchema> {
-        let mut type_defs = HashMap::new();
+        let mut common_types = HashMap::new();
         let mut entity_type_fragments: HashMap<EntityType, _> = HashMap::new();
         let mut action_fragments = HashMap::new();
 
@@ -263,8 +263,8 @@ impl ValidatorSchema {
             // already added by the `ValidatorNamespaceDef`, so the same base
             // type name may appear multiple times so long as the namespaces are
             // different.
-            for (name, ty) in ns_def.type_defs.type_defs {
-                match type_defs.entry(name) {
+            for (name, ty) in ns_def.common_types.defs {
+                match common_types.entry(name) {
                     Entry::Vacant(v) => v.insert(ty),
                     Entry::Occupied(o) => {
                         return Err(DuplicateCommonTypeError(o.key().clone()).into());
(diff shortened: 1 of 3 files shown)


cedar-policy-validator/src/schema.rs
```diff
@@ -253,7 +253,7 @@ impl ValidatorSchema {
         fragments: impl IntoIterator<Item = ValidatorSchemaFragment>,
         extensions: Extensions<'_>,
     ) -> Result<ValidatorSchema> {
-        let mut type_defs = HashMap::new();
+        let mut common_types = HashMap::new();
         let mut entity_type_fragments: HashMap<EntityType, _> = HashMap::new();
         let mut action_fragments = HashMap::new();
```
john-h-kastner-aws, line 259: it won't let me comment on the exact line, by "map of declared typedefs" in the comment a couple lines below

cedar-policy-validator/src/schema/namespace_def.rs
```diff
@@ -170,26 +170,26 @@ impl ValidatorNamespaceDef {
...
+    pub(super) defs: HashMap<Name, SchemaType<Name>>,
 }
 
-impl TypeDefs {
-    /// Construct a [`TypeDefs`] by converting the structures used by the schema
-    /// format to those used internally by the validator.
+impl CommonTypeDefs {
+    /// Construct a [`CommonTypeDefs`] by converting the structures used by the
+    /// schema format to those used internally by the validator.
     pub(crate) fn from_raw_typedefs(
```
john-h-kastner-aws, line 180: `from_raw_common_type_defs`, or maybe just `from_raw_type_defs` since "common type" is clear from context

john-h-kastner-aws (approved):

cedar-policy-validator/src/schema.rs
cdisselkoen: Good catch.  I had grepped `TypeDefs` and `type_defs` and `typ_defs` and some other things, but hadn't grepped `typedefs`

cedar-policy-validator/src/schema/namespace_def.rs
cdisselkoen: `from_raw_common_types` since it's taking `namespace_def.common_types`?

john-h-kastner-aws: sounds good

Issue #801, opened 2024-04-19
Provide Wasm (and Rust) function to get valid principal/action/resource for a policy


cdisselkoen:
### Category

User level API features/changes

### Describe the feature you'd like to request

[vscode-cedar](https://github.com/cedar-policy/vscode-cedar)'s autocomplete functionality needs a function in Wasm which takes a policy and schema, and returns all of the valid principal types for that policy, all of the valid actions for that policy, and all of the valid resource types for that policy. 

The Rust API doesn't currently have this functionality either, so while we're at it, we should add this to the public Rust API too.

### Describe alternatives you've considered

As a hack/workaround, the current code creates special intentionally-invalid policies and parses the errors produced by those.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

hakanson:
An example based on the vscode-cedar [testdata](https://github.com/cedar-policy/vscode-cedar/blob/prerelease/testdata/narrow/policies.cedar):

```
@id("E1,E2 a,a2 R2")
permit (principal, action, resource is NS::R2);
```

the constraints (policy export to JSON) are `All` for `principal` and `action`

```json
{
  "effect": "permit",
  "principal": {
    "op": "All"
  },
  "action": {
    "op": "All"
  },
  "resource": {
    "op": "is",
    "entity_type": "NS::R2"
  },
  "conditions": [],
  "annotations": {
    "id": "E1,E2 a,a2 R2"
  }
}
```

but looking at the namespace, they are:
- principal: `NS::E1`, `NS::E2`
- action: `NS::Action::"a"`, `NS::Action::"a2"`
- resource: `NS::R2`

```
namespace NS {
  entity E;
  entity R1 in [R] = {"p1": String};
  entity R;
  entity R2 in [R] = {"p1": Long};
  entity E1 in [E] = {"p1": String};
  entity E2 in [E] = {"p1": Long};
  action "as";
  action "a" in [Action::"as"] appliesTo {
    principal: [E1, E2],
    resource: [R1, R2],
    context: {"c1": Long}
  };
  action "a1" in [Action::"as"] appliesTo {
    principal: [E1],
    resource: [R1],
    context: {"c1": Long}
  };
  action "a2" in [Action::"as"] appliesTo {
    principal: [E2],
    resource: [R2],
    context: {"c1": Long}
  };
}
```

shaobo-he-aws:
[...]

PR #1074, opened 2024-07-17
Add APIs to get valid principals, actions, and resources

shaobo-he-aws:
## Issue #, if available
#801 


shaobo-he-aws:
I still need to add FFI functions but thought I should start the review process so you can take a look at the signature and implementation of the Rust function.

cedar-policy/src/api.rs
```diff
@@ -3617,3 +3618,36 @@ pub fn eval_expression(
         eval.interpret(&expr.0, &ast::SlotEnv::new())?,
     ))
 }
+
+/// Get a tuple of valid principal [`EntityTypeName`]s, action [`EntityUid`]s, and resource [`EntityTypeName`]s.
+pub fn get_valid_request_pars(
```
khieta, line 3623: I assume `pars` = principals, actions, resources? This might not be obvious to others, so maybe a name like `get_valid_request_types` would be better. Alternatively, we could provide separate functions: `get_valid_request_principal_types`, `get_valid_request_actions`, `get_valid_request_resource_types`

khieta: Also: maybe these could be methods for the `Policy` type?

shaobo-he-aws: The return values are not all entity names (actions are entity uids). So maybe `get_valid_request_envs`?

shaobo-he-aws: > Also: maybe these could be methods for the `Policy` type?

I don't have any preference. @hakanson What do you think?

john-h-kastner-aws: I agree that this should be a method on `Policy`

cedar-policy/src/api.rs
```diff
@@ -3617,3 +3618,36 @@ pub fn eval_expression(
         eval.interpret(&expr.0, &ast::SlotEnv::new())?,
     ))
 }
+
+/// Get a tuple of valid principal [`EntityTypeName`]s, action [`EntityUid`]s, and resource [`EntityTypeName`]s.
```
john-h-kastner-aws, line 3622: The doc comment should expand on what exactly `valid` means.

cedar-policy/src/api.rs
john-h-kastner-aws, line 1: I wonder how useful this will be given that it require a complete policy to be called in the first place. My imagined application for this feature is that you write `principal.` and then autocomplete suggests attributes based on the principal type, but the policy is quite unlikely to typecheck at that point in time.

shaobo-he-aws: I had similar concerns. The principle type is known statically at least constrained by an `is` operator or certain action.

shaobo-he-aws: @cdisselkoen @hakanson what use cases do you originally have in mind?

john-h-kastner-aws: I suppose the vscode plugin could extract the text of just the policy scope and then pass that to the Cedar sdk, so that's still useful.

Is there any application  for this function that would make it useful when given a complete policy?

john-h-kastner-aws: In general, a powerful IDE plugin will end up wanting a lot of utilities for working with partially written, syntactically invalid, and type incorrect policies that

hakanson: I add links to the vscode extension usage on #801 - both autocomplete and showing types on hover are uses cases.  

Not pretty, but I currently extract the head of the Cedar policy to find the entity types.

```typescript
const HEAD_REGEX = /(?:(permit|forbid)\s*\((.|\n)*?\))(?<!\s*(;|when|unless))/;

const match = cedarDoc.getText(policy.range).match(HEAD_REGEX);

const principalTypes = determineEntityTypes(schemaDoc, 'principal', head);
const resourceTypes = determineEntityTypes(schemaDoc, 'resource', head);
const actionIds = determineEntityTypes(schemaDoc, 'action', head);
```

types_on_hover

john-h-kastner-aws: so, it looks like with the proposed API you would still need to extract the policy scope the way that you do, but then you could pass that into the provided function (after appending an semicolon, if necessary). Then the hacky `determineEntityTypes` function could be deleted.

hakanson: @john-h-kastner-aws  What would valid values be in this policy?  Does the `when` clause change the results?  That might answer the question of just the scope vs. the complete policy.

```cedar
@id("(when)E1 a,a1,a2 (when)R2")
permit (principal, action, resource)
when { principal is NS::E1 && resource is NS::R2 };
```

john-h-kastner-aws: If you provide the whole policy, then it will take the clauses into account

cedar-policy/src/api.rs
shaobo-he-aws: Renamed and refactored to a method.

cedar-policy/src/api.rs
```diff
@@ -2068,6 +2069,61 @@ pub(crate) fn fold_partition<T, A, B, E>(
     Ok((lefts, rights))
 }
 
+/// The "type" of an [`Request`], i.e., the [`EntityTypeName`]s of principal
```
khieta, line 2072:
```suggestion
/// The "type" of a [`Request`], i.e., the [`EntityTypeName`]s of principal
```

cedar-policy/src/ffi/utils.rs
```diff
@@ -342,6 +345,24 @@ impl Template {
             .add_template(template)
             .wrap_err(format!("failed to add template{msg} to policy set"))
     }
+
+    /// Get valid principals, actions, and resources.
+    pub fn get_valid_request_envs(
+        self,
+        s: Schema,
+    ) -> Result<(BTreeSet<String>, BTreeSet<String>, BTreeSet<String>), miette::Report> {
```
khieta, line 353: It seems a little inconsistent to return a `HashSet<RequestEnv>` in the main API, but a `(BTreeSet<String>, BTreeSet<String>, BTreeSet<String>)` here. But maybe the interface here is the most useful when implementing an editor?

cedar-wasm/src/utils.rs
```diff
@@ -0,0 +1,35 @@
...
+/// struct that defines the result for the syntax validation function
+pub enum GetValidRequestEnvsResult {
+    /// represents successful syntax validation
+    Success {
+        principals: Vec<String>,
+        actions: Vec<String>,
+        resources: Vec<String>,
+    },
+    /// represents a syntax error and encloses a vector of the errors
+    Error { error: String },
```
khieta, line 19:
```suggestion
    /// represents a failed call to [`get_valid_request_envs()`]
    Error { 
      /// the encountered error
      error: String 
    },
```

cedar-wasm/src/utils.rs
```diff
@@ -0,0 +1,35 @@
+use cedar_policy::ffi::{Schema, Template};
+use serde::{Deserialize, Serialize};
+use tsify::Tsify;
+use wasm_bindgen::prelude::*;
+
+#[derive(Tsify, Debug, Serialize, Deserialize)]
+#[serde(tag = "type")]
+#[serde(rename_all = "camelCase")]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+/// struct that defines the result for the syntax validation function
```
khieta, line 10:
```suggestion
/// struct that defines the result of getting valid request environments
```

cedar-wasm/src/utils.rs
```diff
@@ -0,0 +1,35 @@
...
+pub enum GetValidRequestEnvsResult {
+    /// represents successful syntax validation
```
khieta, line 12:
```suggestion
    /// represents a successful call to [`get_valid_request_envs()`]
```

cedar-policy/src/ffi/utils.rs
shaobo-he-aws: No. It's to make return value order deterministic so that tests can pass 😂

shaobo-he-aws:
@khieta should I move my changes in `cedar-wasm` to the ffi module of `cedar-policy`?

khieta:
> should I move my changes in cedar-wasm to the ffi module of cedar-policy?

Fine either way. The reason I moved the other functionality was to support reusing code in the Java bindings. If we think we'll want this functionality in Java eventually, then it makes sense to move to `ffi`. Otherwise it can stay in `wasm`.

cedar-policy/src/ffi/utils.rs
khieta: I think it'd be more consistent to make the APIs match

cedar-policy/src/api.rs
khieta: Re-opening this thread: was the decision that complete policy analysis is sufficient? Or do you need functions that looks specifically at the scope?

shaobo-he-aws: I think the complete policy analysis is sufficient. We should be able to construct new policies with scopes of original policies if we want to look specifically at the scope.

cedar-policy/src/ffi/utils.rs
shaobo-he-aws: Done.

cedar-policy/src/api.rs
john-h-kastner-aws: My opinion is now that whole-policy is fine. Depending on the exact use case, some code may still need to manually extract the policy scope and then pass only that to the Cedar APIs, but that's fine.

cedar-policy/src/ffi/utils.rs
khieta: Doesn't seem like this is done in the current version?

To be more concrete about what I was suggesting: I think this ffi function should also return a `HashSet<RequestEnv>`.

cedar-wasm/src/utils.rs
```diff
@@ -0,0 +1,38 @@
...
+/// struct that defines the result of getting valid request environments
+pub enum GetValidRequestEnvsResult {
+    /// represents a successful call to [`get_valid_request_envs()`]
+    Success {
+        principals: Vec<String>,
+        actions: Vec<String>,
+        resources: Vec<String>,
+    },
+    /// represents a failed call to [`get_valid_request_envs()`]
+    Error {
```
khieta, line 19: For consistency with the other Wasm/FFI functions
```suggestion
    Failure {
```

cedar-policy/src/ffi/utils.rs
shaobo-he-aws: I changed the callee's signature to `BTreeSet` so that they are consistent. I think we should not be seeing a lot of `RequestEnv` so `BTreeSet` is fine.

khieta: Gotcha. I think an `impl Iterator` would be more consistent with the rest of our public API than a `BTreeSet` though.

shaobo-he-aws: Done.

khieta (approved):

cedar-policy/src/api.rs
```diff
@@ -2247,6 +2303,28 @@ impl Template {
...
+            .into_iter()
+            .filter_map(|(env, pc)| {
+                if matches!(pc, PolicyCheck::Success(_)) {
+                    Some(RequestEnv::from(env))
+                } else {
+                    None
+                }
+            })
+            .collect::<BTreeSet<_>>()
+            .into_iter()
```
cdisselkoen, line 2326: We should be able to avoid collecting into the set here?  Might need to add `+ '_` to the `impl Iterator` return type

cedar-policy/src/ffi/utils.rs
```diff
@@ -345,6 +345,35 @@ impl Template {
...
+            impl Iterator<Item = String>,
+            impl Iterator<Item = String>,
+        ),
+        miette::Report,
+    > {
+        let t = self.parse(None)?;
+        let (s, _) = s.parse()?;
+        let mut principals = Vec::new();
+        let mut actions = Vec::new();
+        let mut resources = Vec::new();
```
cdisselkoen, line 365: should we use `HashSet` to avoid duplicates?

cdisselkoen (approved):

cedar-policy/src/api.rs
```diff
@@ -2068,6 +2069,61 @@ pub(crate) fn fold_partition<T, A, B, E>(
...
+                principal,
+                action,
+                resource,
+                ..
+            } => Self {
+                principal: principal.clone().into(),
+                resource: resource.clone().into(),
+                action: action.clone().into(),
+            },
+            //PANIC SAFETY: partial validation should not be enabled and hence `RequestEnv::UndeclaredAction` should not show up
```
john-h-kastner-aws, line 2120: Since we only use this `From` impl once, I think I'd prefer in-lining it in `get_valid_request_envs` so that this invariant is immediately obvious. As it is, this function looks like it could panic, and I need to know that it's only called in that one place before I'm convinced it's safe.

john-h-kastner-aws (approved):

cedar-policy/src/api.rs
shaobo-he-aws: It seems that I can't avoid collecting because otherwise `tc` will be referenced, which is to a temporary variable.

cdisselkoen: Ok, no problem.

cedar-policy/src/api.rs
shaobo-he-aws: Done.

cedar-policy/src/api.rs
```diff
@@ -2068,6 +2069,61 @@ pub(crate) fn fold_partition<T, A, B, E>(
...
+impl From<cedar_policy_validator::types::RequestEnv<'_>> for RequestEnv {
+    fn from(value: cedar_policy_validator::types::RequestEnv<'_>) -> Self {
+        match value {
+            cedar_policy_validator::types::RequestEnv::DeclaredAction {
+                principal,
+                action,
+                resource,
+                ..
+            } => Self {
+                principal: principal.clone().into(),
```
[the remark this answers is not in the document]
shaobo-he-aws: `api::RequestEnv` is an owned struct so we have to clone anyways?

cedar-policy/src/ffi/utils.rs
shaobo-he-aws: Done. I used `BTreeSet` because the number of these types should not be large and we want to have an order.

cdisselkoen: Makes sense to me

cedar-policy/src/api.rs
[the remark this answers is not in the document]
john-h-kastner-aws: Yeah, I guess so.

john-h-kastner-aws (approved):

andrewmwells-amazon (approved):

PR #1075, opened 2024-07-17
fixes for the wasm build

khieta:
## Description of changes

While working on #854 I realized that the wasm build was not actually generating valid TypeScript (largely due to changes I made for #757). This PR fixes the issues & adds CI to avoid this mistake in the future.

Detailed summary:
* Fixes the bad TypeScript by adjusting various `tsify` annotations
* Adds CI to check that the generated TypeScript is valid
* Adds `#![cfg_attr(feature = "wasm", allow(non_snake_case))]` in a few files to suppress annoying VSCode warnings
* Removes the `cedar-wasm/src/{validator, authorizer}.rs` files since it's simpler to expose the FFI functions directly rather than introduce a trivial wrapper


khieta:
Still need to investigate the new `wasm-example` failure, but the PR is otherwise ready for review.

shaobo-he-aws (approved):

john-h-kastner-aws (approved):
Seems legit

Issue #415, opened 2023-11-09
[CLI] `check-parse` command (or another command) should check parsing of entity data and schemas, not just policies


cdisselkoen:
### Category

Other

### Describe the feature you'd like to request

Currently, the CLI `check-parse` command only checks the parsing of policies.

To check the parsing of entity data or schemas, you have to use other commands -- for instance `authorize` for entity data, or `validate` for schemas.  But this is inconvenient, as you also have to provide other inputs for those commands, like policies.  We don't have a dedicated way to check the parsing of entity data or schemas without also having to provide unnecessary inputs.

### Describe alternatives you've considered

We could have separate commands `check-policies-parse`, `check-entities-parse`, `check-schema-parse`; or (my preference) just have a single `check-parse` which checks that any or all of these parse.  `check-parse` could check the parsing of any or all of these in a single call -- e.g., if you pass just schema, it checks the parsing of the schema, or if you pass both entities and policies, it checks the parsing of the entities and policies; etc.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #1079, opened 2024-07-19
add parsing functions to ffi

khieta:
## Description of changes

Moves the parsing functionality from `cedar-wasm` to `cedar_policy::ffi` and deletes the validation `enabled` flag, which was intended to take the place of parsing. I still plan to do some more work on #854 (namely, adding support for formatting & conversion between Cedar and JSON formats), but that will come in a future PR.

Downstream breaks in the Cedar Wasm example are expected.


shaobo-he-aws (approved):

cedar-policy/src/ffi/validate.rs
```diff
@@ -155,28 +147,16 @@ impl ValidationCall {
...
 }
 
-impl Default for ValidationSettings {
-    fn default() -> Self {
-        Self {
-            enabled: true,
-            mode: ValidationMode::default(),
-        }
-    }
-}
```
cdisselkoen, line 178: Might make sense to keep this around?

cdisselkoen (approved):
Looks great.  As a side note, it would also be great to make all these functions available in the CLI -- https://github.com/cedar-policy/cedar/issues/415

cedar-policy/src/ffi/validate.rs
khieta: It's still there - it's just been moved to an automatic derivation of `Default`.

PR #1083, opened 2024-07-22
fix TryFrom<Name> for UnreservedId

cdisselkoen:
## Description of changes

I think #969 had the `TryFrom<Name> for UnreservedId` impl backwards, that is, `Ok` when it should be `Err` and vice versa.  This PR, reversing it, does not fail any tests, so I suspect this impl is unused.

Also uses `is_unqualified()` instead of manually checking `path`. This matches how the similar `TryFrom<UncheckedName> for Id` impl works.


shaobo-he-aws (approved):

john-h-kastner-aws (approved):

cdisselkoen:
Merging, the CI build-wasm failure looks unrelated.

PR #1078, opened 2024-07-19
Update build_and_test.yml

shaobo-he-aws:
## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [ ] A breaking change requiring a major version bump to `cedar-policy` (e.g., changes to the signature of an existing API).
- [ ] A backwards-compatible change requiring a minor version bump to `cedar-policy` (e.g., addition of a new API).
- [ ] A bug fix or other functionality change requiring a patch to `cedar-policy`.
- [ ] A change "invisible" to users (e.g., documentation, changes to "internal" crates like `cedar-policy-core`, `cedar-validator`, etc.)
- [ ] A change (breaking or otherwise) that only impacts unreleased or experimental code.

I confirm that this PR (choose one, and delete the other options):

- [ ] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).
- [ ] Does not update the CHANGELOG because my change does not significantly impact released code.

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [ ] Does not require updates because my change does not impact the Cedar formal model or DRT infrastructure.
- [ ] Requires updates, and I have made / will make these updates myself. (Please include in your description a timeline or link to the relevant PR in `cedar-spec`, and how you have tested that your updates are correct.)
- [ ] Requires updates, but I do not plan to make them in the near future. (Make sure that your changes are hidden behind a feature flag to mark them as experimental.)
- [ ] I'm not sure how my change impacts `cedar-spec`. (Post your PR anyways, and we'll discuss in the comments.)

PR #1087, opened 2024-07-23
add formatting and JSON conversion utilities to ffi

### LINKED ISSUE #854: Add parsing APIs to FFI

khieta:
## Description of changes

* Adds JSON APIs to apply the policy formatter & convert between the Cedar/JSON formats for policies/schemas
* Removes the corresponding definitions from `cedar-wasm/`
* Tweaks an error message in the formatter
* Fixes a typo in the wasm build script introduced by #1078. Apparently `&` is a special character for `sed` in some positions.

PR to fix downstream breaks in wasm example: https://github.com/cedar-policy/cedar-examples/pull/180

## Issue #, if available

Resolves #854


cedar-policy/Cargo.toml
```diff
@@ -14,6 +14,7 @@ repository = "https://github.com/cedar-policy/cedar"
 [dependencies]
 cedar-policy-core = { version = "=4.0.0", path = "../cedar-policy-core" }
 cedar-policy-validator = { version = "=4.0.0", path = "../cedar-policy-validator" }
+cedar-policy-formatter = { version = "=4.0.0", path = "../cedar-policy-formatter" }
```
cdisselkoen, line 17: Did we have a reason to not take this dependency in the past?

cedar-policy/src/ffi/convert.rs
```diff
@@ -0,0 +1,396 @@
...
+#[cfg_attr(feature = "wasm", tsify(into_wasm_abi, from_wasm_abi))]
+pub enum SchemaToTextAnswer {
+    /// Represents a successful call
+    Success {
+        /// Cedar format schema
+        text: String,
+        /// Warnings
+        warnings: Vec<DetailedError>,
+    },
+    /// Represents a failed call(e.g., because the input is ill-formed)
```
cdisselkoen, line 176:
```suggestion
    /// Represents a failed call (e.g., because the input is ill-formed)
```

cedar-policy/src/ffi/format.rs
```diff
@@ -0,0 +1,164 @@
...
+#[cfg_attr(feature = "wasm", tsify(into_wasm_abi, from_wasm_abi))]
+#[serde(rename_all = "camelCase")]
+#[serde(deny_unknown_fields)]
+pub struct FormattingCall {
+    /// Policy text. May define multiple policies or templates in the Cedar policy format.
+    policy_text: String,
+    /// Line width
+    line_width: usize,
+    /// Indentation width
+    indent_width: isize,
```
cdisselkoen, line 85: should we be opinionated and set defaults for these?  Does the CLI interface to the formatter have defaults for these?

cedar-policy/src/ffi/format.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// Represents a failure to call the formatter
+    Failure {
+        /// Policy parse errors
+        errors: Vec<DetailedError>,
+    },
+    /// Represents a successful formatting call
+    Success {
+        /// Formatted policy
+        formatted_policy: String,
+    },
```
cdisselkoen, line 104: Any reason we have `Failure` first in this enum, and `Success` first in the others?

cedar-policy/src/ffi/convert.rs
```diff
@@ -0,0 +1,396 @@
...
+#[cfg(feature = "wasm")]
+use wasm_bindgen::prelude::wasm_bindgen;
+
+#[cfg(feature = "wasm")]
+extern crate tsify;
+
+/// Return the Cedar (textual) representation of a policy or template.
+///
+/// Note that the [`Template`] type can represent both static policies and
+/// policy templates.
```
cdisselkoen, line 33: Good note, I had forgotten this was the case in the FFI.  It's not true of `cedar_policy::Template`.

cdisselkoen (approved):

cedar-policy/Cargo.toml
khieta: I don't believe so, @shaobo-he-aws do you remember any special reason?

cedar-policy/src/ffi/format.rs
khieta: The CLI defaults are 80 for line width & 2 for indentation. Will use the same defaults here.

cedar-policy/src/ffi/format.rs
khieta: No reason. Does it matter?

cdisselkoen: It does not

cedar-policy/src/ffi/format.rs
khieta: Done in the latest commit

cedar-policy/Cargo.toml
shaobo-he-aws: I think there are two reasons. 1. We only use the formatter in CLI so it's unnecessary there. 2. I was too lazy to add these APIs. So we're all good here.

cedar-policy/src/ffi/convert.rs
cdisselkoen: Having `Policy` in the FFI support both static policies and templates might be additionally confusing because `Policy` in `cedar_policy` represents both static policies and template-linked policies, but not templates

khieta: Yuck. I'll leave it as-is then.

PR #1090, opened 2024-07-24
Remove or refactor evaluation extension function errors

### LINKED ISSUE #193: Improvements to ExtensionsError variants representing internal invariant violations
### Category

User level API changes

Several variants of `ExtensionsError` -- `HasNoType`, `FuncMultiplyDefined`, and `MultipleConstructorsSameSignature` -- represent internal invariant violations more than they represent something the user did wrong.  In the current world where we control all the language extensions, these should never be thrown, and if they are thrown, it represents a bug in our code.

We should mark these as internal errors -- perhaps by making a struct `InternalError` (or some other name) and having that be a variant of `ExtensionsError`.  If possible, we should also reorganize `ExtensionsError` and/or `EvaluationError` so that these errors are (a) caught as early as possible, eg before `is_authorized()`, and (b) not part of `EvaluationError`, so that consumers of `EvaluationError` don't have to worry about matching on these.

We want to catch these as early as possible, but we also don't want to do unnecessary expensive checks on the hot path.  One option is to do it when constructing an `Extensions` object, if that is not on the hot path or can be moved off the hot path.

As above.

Removing these variants entirely in favor of some other mechanism or refactor that makes those errors impossible.  This might be very difficult without restructuring our extensions mechanism.

### Additional context

_No response_

john-h-kastner-aws:
## Description of changes

Update extension function errors in evaluation to fix #193. A few more changes for validation will come in a second PR.

Changes:

* Move `FuncMultiplyDefinedError` into a new error enum `ExtensionInitializationError` which is only returned when constructing extension and never in evaluation.
* Move `MultipleConstructorsSameSignature` into the `ExtensionInitializationError` enum 
* replace `HasNoType` with the existing variant `UnknownInsufficientTypeInfoError` a `None` return type on an extension function could only happen when that extension function was `unknown`, so these errors were effectively the same. Adjust some names an comments to make it more clear what `None` in the returns means.


cedar-policy-core/src/ast/extension.rs
```diff
@@ -301,7 +303,7 @@ impl ExtensionFunction {
     ///
     /// If any given argument type is not constant (function works with multiple
     /// `SchemaType`s) then this will be `None` for that argument.
```
cdisselkoen, line 305: Comment needs to be removed here as well

cedar-policy-core/src/entities/json/schema_types.rs
```diff
@@ -319,6 +312,40 @@ pub struct HeterogeneousSetError {
...
+pub struct UnknownInsufficientTypeInfoError {
+    /// `Unknown` which has insufficient type information
+    unknown: Option<Unknown>,
+}
+
+impl std::fmt::Display for UnknownInsufficientTypeInfoError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "cannot compute type because of insufficient type information for"
```
cdisselkoen, line 328: I think you need a trailing space
```suggestion
            "cannot compute type because of insufficient type information for "
```

cedar-policy-core/src/entities/json/schema_types.rs
```diff
@@ -319,6 +312,40 @@ pub struct HeterogeneousSetError {
...
+/// residual which is not just a single `Unknown`). For now, we do not
+/// attempt to compute the [`SchemaType`] in these cases, and just return
+/// this error.
+#[derive(Debug, Diagnostic, Error)]
+#[error("cannot compute type of nontrivial residual `{residual}`")]
+pub struct NontrivialResidualError {
+    /// Nontrivial residual which we were trying to compute the
+    /// [`SchemaType`] of
+    residual: Box<Expr>,
+}
```
cdisselkoen, line 347: bonus points for a `Diagnostic` implementation to take `labels()` and `source_code()` from the expr's `Loc`.  There are other errors you can copy-paste this implementation from (or make a macro)

cedar-policy-core/src/entities.rs
```diff
@@ -174,7 +174,7 @@ impl Entities {
             // We do this before adding the actions, because we trust the
             // actions were already validated as part of constructing the
             // `Schema`
-            let checker = EntitySchemaConformanceChecker::new(schema, extensions);
+            let checker = EntitySchemaConformanceChecker::new(schema, extensions.clone());
```
cdisselkoen, line 177: is it possible for `EntitySchemaConformanceChecker` to hold an `&Extensions` instead?  (I do recognize that `Extensions` is still pretty cheap to clone, so this is a minor nit)

cedar-policy-core/src/entities/json/context.rs
```diff
@@ -82,12 +82,12 @@ impl<'e, 's, S: ContextSchema> ContextJsonParser<'e, 's, S> {
         &self,
         json: serde_json::Value,
     ) -> Result<Context, ContextJsonDeserializationError> {
-        let vparser = ValueParser::new(self.extensions);
+        let vparser = ValueParser::new(self.extensions.clone());
```
cdisselkoen, line 85: is it possible for `ValueParser` to hold an `&Extensions` instead?  (I do recognize that `Extensions` is still pretty cheap to clone, so this is a minor nit)

cedar-policy-core/src/entities/json/context.rs
```diff
@@ -82,12 +82,12 @@ impl<'e, 's, S: ContextSchema> ContextJsonParser<'e, 's, S> {
...
         let expected_ty = self.schema.map(|s| s.context_type());
         let rexpr = vparser.val_into_restricted_expr(json, expected_ty.as_ref(), || {
             JsonDeserializationErrorContext::Context
         })?;
-        Context::from_expr(rexpr.as_borrowed(), self.extensions)
+        Context::from_expr(rexpr.as_borrowed(), self.extensions.clone())
```
cdisselkoen, line 90: Is it possible for `Context::from_expr()` to take an `&Extensions` instead?

cedar-policy-core/src/evaluator.rs
```diff
@@ -172,7 +172,7 @@ impl<'e> RestrictedEvaluator<'e> {
                 match split(args) {
                     Either::Left(values) => {
                         let values : Vec<_> = values.collect();
-                        let efunc = self.extensions.func(fn_name).map_err(|err| err.with_maybe_source_loc(expr.source_loc().cloned()))?;
+                        let efunc = self.extensions.func(fn_name)?;
```
cdisselkoen, line 175: why this change?

cedar-policy-core/src/evaluator.rs
```diff
@@ -541,10 +541,7 @@ impl<'e> Evaluator<'e> {
                 match split(args) {
                     Either::Left(vals) => {
                         let vals: Vec<_> = vals.collect();
-                        let efunc = self
-                            .extensions
-                            .func(fn_name)
-                            .map_err(|err| err.with_maybe_source_loc(loc.cloned()))?;
+                        let efunc = self.extensions.func(fn_name)?;
```
cdisselkoen, line 544: why this change?

cedar-policy-core/src/extensions.rs
```diff
@@ -23,49 +23,152 @@ pub mod ipaddr;
...
 ///
 /// Clone is cheap for this type.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone)]
 pub struct Extensions<'a> {
     /// the actual extensions
     extensions: &'a [Extension],
+    /// All extension functions, collected from every extension used to
+    /// construct this object.  Built ahead of time so that we know during
+    /// extension function lookup that at most one extension functions exists
```
cdisselkoen, line 77:
```suggestion
    /// extension function lookup that at most one extension function exists
```

cedar-policy-core/src/extensions.rs
```diff
@@ -124,105 +214,75 @@ impl<'a> Extensions<'a> {
     /// `Err` is returned in the case that multiple constructors have that signature.
```
cdisselkoen, line 214: Comment needs updating here too. (and the polymorphic comment no longer applies)

cedar-policy-core/src/extensions.rs
```diff
@@ -233,7 +293,7 @@ pub mod extension_function_lookup_errors {
     // Don't make fields `pub`, don't make breaking changes, and use caution
     // when adding public methods.
     #[derive(Debug, PartialEq, Eq, Clone, Error)]
-    #[error("extension function `{name}` does not exist")]
+    #[error("internal invariant violated. Extension function `{name}` does not exist")]
```
cdisselkoen, line 296: Might be worth commenting why this is an internal invariant violation, and why you won't get this error from a policy that calls `foo()` (in either JSON or Cedar format), or entity JSON data that calls `foo()`

cdisselkoen (approved):

cedar-policy-core/src/evaluator.rs
john-h-kastner-aws: `Name`s carry a source location, so I update the error to use that

cedar-policy-core/src/entities.rs
john-h-kastner-aws: Yes. I'm confident I could get rid of all the clones and just do refs everywhere, but it creates a bunch of changes, so I didn't go through with it

cdisselkoen: Sounds fine

cedar-policy-core/src/extensions.rs
john-h-kastner-aws: Looked at this closer, and the supposed invariant is violated when parsing entity data:

```json
    {
        "uid": {  "type":"User", "id":"bob" },
        "attrs": {
            "score": { "__extn": { "fn": "foo", "arg": "0.899" }}
        },
        "parents": []
    }
```
```
  × failed to parse entities from file test.json
  ├─▶ error during entity deserialization
  ╰─▶ failed to evaluate attribute `score` of `User::"bob"`: internal invariant violated. Extension function `foo` does not exist
```

Not a Cedar bug because we're getting an error exactly where we want one. Just means this error variant can occur legitimately without some more refactoring

john-h-kastner-aws: IMO not worth restructuring just so we can say this is an internal invariant violation.

cedar-policy-core/src/entities/json/schema_types.rs
john-h-kastner-aws: done

cedar-policy-core/src/extensions.rs
cdisselkoen: Makes sense to me

shaobo-he-aws (approved):

cedar-policy-core/src/entities.rs
john-h-kastner-aws: not so bad actually #1100

PR #1102, opened 2024-07-30
Add Entity Manifests to Cedar

oflatt:
## Description of changes

This PR implements [RFC #74](https://github.com/cedar-policy/rfcs/pull/74) behind a new feature flag "entity-manifest".
It adds:
- A new public function `compute_entity_manifest` (with experimental warning)
- New structs`EntityManifest, RootAccessTrie, AccessTrie, EntitySliceError`. (with experimental warning)
- JSON serialization for the entity manifest, along with tests that use it.

Note that there's a TODO in the `compute_entity_manifest` function about returning type errors.

## Follow up PRs

- #1105 adds an entity slicing function that produces an `Entities` store from an existing `Entities` store. This enables strong corpus testing.
- Generalization to all of Cedar
- Type annotated entity manifests
- Support for a SimpleEntityLoader as described in the RFC.
- Support for an easy to use flattened format for the entity manifest, and a human-readable format.


cedar-policy-validator/src/entity_slicing.rs
```diff
@@ -314,7 +314,7 @@ pub fn compute_entity_manifest(
                 // Also, should we run full validation instead of just
                 // typechecking? Validation does a little more right?
                 PolicyCheck::Fail(errors) => {
-                    // PANIC SAFETY policy check fail
+                    // PANIC SAFETY: policy check fail
```
john-h-kastner-aws, line 317: it doesn't like the line breaks

oflatt: thanks!

cedar-policy-validator/src/entity_slicing.rs
```diff
@@ -0,0 +1,1087 @@
+//! Entity Slicing
```
shaobo-he-aws, line 1: Rename it to `entity_manifest`?

shaobo-he-aws: And add copyright claim.

cedar-policy-validator/src/entity_slicing.rs
```diff
@@ -0,0 +1,1087 @@
...
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    ValidationError(#[from] ValidationError),
+    /// A entities error was encountered
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    EntitiesError(#[from] EntitiesError),
+
+    /// The request was partial
+    #[error("Entity slicing requires a fully concrete request. Got a partial request.")]
```
shaobo-he-aws, line 173: This is inconsistent with Cedar's error type convention. You want a struct or enum here even though it doesn't contain anything.

cedar-policy-validator/src/entity_slicing.rs
oflatt: :+1:

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+                    // compute the trie from the typechecked expr
+                    // using static analysis
+                    compute_root_trie(&typechecked_expr, policy.id())
+                }
+                PolicyCheck::Irrelevant(_) => {
+                    // this policy is ireelevant, so we need no data
+                    Ok(RootAccessTrie::new())
+                }
+
+                // TODO is returning the first error correct?
```
oflatt, line 330: Need advice on this TODO

cedar-policy-core/src/ast/request.rs
```diff
@@ -49,6 +49,17 @@ pub struct Request {
     pub(crate) context: Option<Context>,
 }
 
+/// Represents the principal, action, and resource types.
```
cdisselkoen, line 52:
```suggestion
/// Represents the principal type, resource type, and action UID.
```

cedar-policy-core/src/ast/request.rs
```diff
@@ -49,6 +49,17 @@ pub struct Request {
...
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Deserialize, Serialize)]
+pub struct RequestType {
+    /// Principal type
+    pub principal: EntityType,
+    /// Aciton type
```
cdisselkoen, line 57:
```suggestion
    /// Action type
```

cedar-policy-core/src/ast/request.rs
```diff
@@ -186,6 +197,19 @@ impl Request {
...
     }
+
+    /// Get the request types that correspond to this request.
+    /// This includes the types of the principal, action, and resource.
+    /// [`RequestType`] is used by the entity manifest.
+    /// The context type is implied by the action's type.
+    /// Returns `None` if the request is not fully concrete.
+    pub fn to_request_type(&self) -> Option<RequestType> {
+        Some(RequestType {
+            principal: self.principal.uid()?.clone().components().0,
```
cdisselkoen, line 208: nit: `.clone().components().0` will result in cloning both components, whereas `.entity_type().clone()` only clones the component it needs

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+};
+
+/// Data structure storing what data is needed
+/// based on the the [`RequestType`].
+/// For each request type, the [`EntityManifest`] stores
+/// a [`RootAccessTrie`] of data to retrieve.
+#[doc = include_str!("../../cedar-policy/experimental_warning.md")]
+#[serde_as]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct EntityManifest<T = ()>
```
cdisselkoen, line 48: can we add a comment in the doc string about what `T` represents?

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+    #[serde_as(as = "Vec<(_, _)>")]
+    #[serde(bound(deserialize = "T: Default"))]
+    pub per_action: HashMap<RequestType, RootAccessTrie<T>>,
+}
+
+/// A map of data fields to [`AccessTrie`]s.
+/// The keys to this map form the edges in the access trie,
+/// pointing to sub-tries.
+#[doc = include_str!("../../cedar-policy/experimental_warning.md")]
+pub type Fields<T> = HashMap<SmolStr, Box<AccessTrie<T>>>;
```
cdisselkoen, line 62: Do we need the `Box`?

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+
+/// The root of an entity slice.
```
cdisselkoen, line 64: entity manifest?

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+/// A [`RootAccessTrie`] is a trie describing
+/// data paths to retrieve. Each edge in the trie
+/// is either a record or entity dereference.
+///
+/// If an entity or record field does not exist in the backing store,
+/// it is safe to stop loading data at that point.
+#[doc = include_str!("../../cedar-policy/experimental_warning.md")]
+#[serde_as]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct RootAccessTrie<T = ()>
```
cdisselkoen, line 92: Another place where it would be nice to describe what the `T` parameter is for in the doc string.  Also on `AccessTrie`, maybe other types.  If the description is long, you can just put a note like "`T`: see note on OtherType"

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct AccessTrie<T = ()> {
+    /// Child data of this entity slice.
+    /// The keys are edges in the trie pointing to sub-trie values.
+    #[serde_as(as = "Vec<(_, _)>")]
+    pub children: Fields<T>,
+    /// For entity types, this boolean may be `true`
+    /// to signal that all the parents in the entity hierarchy
+    /// are required (transitively).
+    pub parents_required: bool,
```
cdisselkoen, line 114: For clarity, since this requires all transitive ancestors, should we rename to `ancestors_required`?

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+#[serde_as]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct RootAccessTrie<T = ()>
+where
+    T: Clone,
+{
+    #[serde_as(as = "Vec<(_, _)>")]
+    #[serde(bound(deserialize = "T: Default"))]
+    /// The data that needs to be loaded, organized by root.
+    pub trie: HashMap<EntityRoot, AccessTrie<T>>,
```
cdisselkoen, line 99: tiny style nit: I think we usually put the doc comment on top, before `serde` attributes

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+    /// The kind of the expression that was unexpected
+    pub expr_kind: ExprKind<Option<Type>>,
+}
+
+impl Diagnostic for FailedAnalysisError {
+    impl_diagnostic_from_source_loc_field!();
+
+    fn help<'a>(&'a self) -> Option<Box<dyn Display + 'a>> {
+        Some(Box::new(format!(
+            "Entity slicing failed to analyze expression: {} operators are not allowed before accessing record or entity attributes.",
```
cdisselkoen, line 167:
```suggestion
            "failed to compute entity manifest: {} operators are not allowed before accessing record or entity attributes",
```

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+    EntitiesError(#[from] EntitiesError),
+
+    /// The request was partial
+    #[error("Entity slicing requires a fully concrete request. Got a partial request.")]
+    PartialRequestError,
+    /// A policy was partial
+    #[error(
+        "Entity slicing requires fully concrete policies. Got a policy with an unknown expression."
+    )]
+    PartialExpressionError,
```
cdisselkoen, line 195: So that we can extend these in the future backwards-compatibly, let's make structs `PartialRequestError` and `PartialExpressionError` that are empty, and do `PartialRequestError(#[from] PartialRequestError)` here

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+        } else {
+            res.insert(key.clone(), value.clone());
+        }
+    }
+    res
+}
+
+impl AccessPath {
+    /// Convert a [`AccessPath`] into corresponding [`RootAccessTrie`].
+    fn to_primary_slice(&self) -> RootAccessTrie {
```
cdisselkoen, line 220:
```suggestion
    fn to_root_access_trie(&self) -> RootAccessTrie {
```

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+        self.to_primary_slice_with_leaf(AccessTrie {
+            parents_required: true,
+            children: Default::default(),
+            data: (),
+        })
+    }
+
+    /// Convert an [`AccessPath`] to a [`RootAccessTrie`], and also
+    /// add a full trie as the leaf at the end.
+    fn to_primary_slice_with_leaf(&self, leaf_trie: AccessTrie) -> RootAccessTrie {
```
cdisselkoen, line 230:
```suggestion
    fn to_root_access_trie_with_leaf(&self, leaf_trie: AccessTrie) -> RootAccessTrie {
```

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+    /// Union two [`RootAccessTrie`]s together.
+    /// The new trie requests the data from both of the original.
+    fn union(&self, other: &Self) -> Self {
+        let mut res = self.clone();
+        for (key, value) in &other.trie {
+            if let Some(existing) = res.trie.get(key) {
+                res.trie.insert(key.clone(), (*existing).union(value));
+            } else {
+                res.trie.insert(key.clone(), value.clone());
+            }
```
cdisselkoen, line 268: here and elsewhere: the `.entry()` API will avoid the double hashmap lookup.  (In this code, you're doing a hashmap lookup for `.get()` and then repeating the lookup for the same key for `.insert()`)

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+) -> Result<EntityManifest, EntityManifestError> {
+    let mut manifest: HashMap<RequestType, RootAccessTrie> = HashMap::new();
+
+    // now, for each policy we add the data it requires to the manifest
+    for policy in policies.policies() {
+        // typecheck the policy and get all the request environments
+        let typechecker = Typechecker::new(schema, ValidationMode::Strict, policy.id().clone());
+        let request_envs = typechecker.typecheck_by_request_env(policy.template());
+        for (request_env, policy_check) in request_envs {
+            // match on the typechecking answer
```
cdisselkoen, line 318: this comment is probably unnecessary

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+            let new_primary_slice = match policy_check {
+                PolicyCheck::Success(typechecked_expr) => {
+                    // compute the trie from the typechecked expr
+                    // using static analysis
+                    compute_root_trie(&typechecked_expr, policy.id())
+                }
+                PolicyCheck::Irrelevant(_) => {
+                    // this policy is ireelevant, so we need no data
```
cdisselkoen, line 326:
```suggestion
                    // this policy is irrelevant, so we need no data
```

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+
+            let request_type = request_env
+                .to_request_type()
+                .ok_or(EntityManifestError::PartialRequestError)?;
+            // Add to the manifest based on the request type.
+            if let Some(existing) = manifest.get_mut(&request_type) {
+                *existing = existing.union(&new_primary_slice);
+            } else {
+                manifest.insert(request_type, new_primary_slice);
+            }
```
cdisselkoen, line 352: another place for `.entry()`

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+    policy_id: &PolicyID,
+    should_load_all: bool,
+) -> Result<(), EntityManifestError> {
+    match expr.expr_kind() {
+        // Literals, variables, and unkonwns without any GetAttr operations
+        // on them are okay, since no fields need to be loaded.
+        ExprKind::Lit(_) => (),
+        ExprKind::Var(_) => (),
+        ExprKind::Slot(_) => (),
+        ExprKind::Unknown(_) => return Err(EntityManifestError::PartialExpressionError),
```
cdisselkoen, line 388:
```suggestion
        ExprKind::Lit(_) => Ok(()),
        ExprKind::Var(_) => Ok(()),
        ExprKind::Slot(_) => Ok(()),
        ExprKind::Unknown(_) => Err(EntityManifestError::PartialExpressionError),
```

and then you don't need the separate return of `Ok(())` at the bottom

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+            UnaryOp::Not => add_to_root_trie(root_trie, arg, policy_id, should_load_all)?,
+            UnaryOp::Neg => add_to_root_trie(root_trie, arg, policy_id, should_load_all)?,
+        },
+        ExprKind::BinaryApp { op, arg1, arg2 } => match op {
+            // Special case! Equality between records requires
+            // that we load all fields.
+            // This could be made more precise if we check the type.
+            BinaryOp::Eq => {
+                add_to_root_trie(root_trie, arg1, policy_id, true)?;
+                add_to_root_trie(root_trie, arg1, policy_id, true)?;
```
cdisselkoen, line 416: doing the same call twice?

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+                flat_slice.parents_required = true;
+                *root_trie = root_trie.union(&flat_slice.to_primary_slice());
+            }
+            BinaryOp::Contains | BinaryOp::ContainsAll | BinaryOp::ContainsAny => {
+                // Like equality, another special case for records.
+                add_to_root_trie(root_trie, arg1, policy_id, true)?;
+                add_to_root_trie(root_trie, arg2, policy_id, true)?;
+            }
+            BinaryOp::Less | BinaryOp::LessEq | BinaryOp::Add | BinaryOp::Sub | BinaryOp::Mul => {
+                // These operators work on values, so no special
```
cdisselkoen, line 433: nit: in the terminology used elsewhere in the Cedar implementation, `Value` includes records.  Might be better to say "literals" or just "integers"

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+        | Type::Primitive { .. }
+        | Type::Set { .. } => AccessTrie::new(),
+        Type::EntityOrRecord(record_type) => entity_or_record_to_access_trie(record_type),
+    }
+}
+
+/// Compute the full [`AccessTrie`] for the given entity or record type.
+fn entity_or_record_to_access_trie(ty: &EntityRecordKind) -> AccessTrie {
+    match ty {
+        EntityRecordKind::ActionEntity { name: _, attrs }
```
cdisselkoen, line 510: Some sugar for this

```suggestion
        EntityRecordKind::ActionEntity { attrs, .. }
```

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+    Ok(match expr.expr_kind() {
+        ExprKind::Slot(slot_id) => {
+            if slot_id.is_principal() {
+                AccessPath {
+                    root: EntityRoot::Var(Var::Principal),
+                    path: vec![],
+                    parents_required: false,
+                }
+            } else {
+                assert!(slot_id.is_resource());
```
cdisselkoen, line 554: instead of this if-else with an `assert!`, should we just `match` on the `slot_id`?

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1103 @@
...
+
+action Read appliesTo {
+  principal: [User],
+  resource: [Document]
+};
+        ",
+            Extensions::all_available(),
+        )
+        .unwrap()
+        .0;
```
cdisselkoen, line 622: Since this schema is used repeatedly, we can just make a `schema()` helper function in this module that returns it

cedar-policy/CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ Starting with version 4.0, changes marked with a star (*) are _language breaking
 Cedar Language Version: 4.0
 
 ### Added
+- A new experimental API (`compute_entity_manifest`)
+  that provides the Entity Manifest: a data
+  structure that describes what data is required to satisfy a 
+  Cedar request. Entity Manifests improve Cedar performance dramatically
+  in a safe way.
```
cdisselkoen, line 20: Probably not putting the performance claim in the changelog. Among other things it's not clear what we're comparing against for the "improve performance dramatically".  Vs sending all entities?  Are users actually doing that?  I suspect they aren't, and are instead rolling their own slicing, in which case this is more about improving safety than performance

cdisselkoen: I'd also link to the RFC in the changelog entry

cedar-policy/src/api.rs
```diff
@@ -23,16 +23,22 @@
...
 
 mod err;
 pub use err::*;
 
 pub use ast::Effect;
 pub use authorizer::Decision;
-use cedar_policy_core::ast;
 #[cfg(feature = "partial-eval")]
 use cedar_policy_core::ast::BorrowedRestrictedExpr;
+use cedar_policy_core::ast::{self};
```
cdisselkoen, line 41: I _think_ this change has no effect?

cedar-policy/src/api.rs
```diff
@@ -3617,3 +3623,17 @@ pub fn eval_expression(
         eval.interpret(&expr.0, &ast::SlotEnv::new())?,
     ))
 }
+
+/// Given a schema and policy set, compute an entity slice manifest.
```
cdisselkoen, line 3627:
```suggestion
/// Given a schema and policy set, compute an entity manifest.
```

cedar-policy/src/api.rs
```diff
@@ -3617,3 +3623,17 @@ pub fn eval_expression(
...
+/// The policies must validate against the schema in strict mode,
+/// otherwise an error is returned.
+/// The manifest describes the data required to answer requests
+/// for each action type.
```
cdisselkoen, line 3631:
```suggestion
/// for each action.
```

cdisselkoen (approved):
Definitely a reasonable implementation of the RFC as it currently stands.  Some low-level comments, but at a high level this looks great.

cedar-policy-validator/src/entity_slicing.rs
oflatt: Is there a benefit to breaking it out into a struct?

cedar-policy-validator/src/entity_manifest.rs
oflatt: Unfortunately yes, since `AccessTrie` contains a `Fields`

cedar-policy-validator/src/entity_slicing.rs
oflatt: I saw Craig's comment: it's for backwards compatibility. Will do!

cedar-policy-validator/src/entity_manifest.rs
oflatt: I think I can't access the inner enum from this crate

cedar-policy-validator/src/entity_manifest.rs
oflatt: yep

cedar-policy-validator/src/entity_manifest.rs
oflatt: yeah

cedar-policy-validator/src/entity_manifest.rs
oflatt: done

cedar-policy/CHANGELOG.md
oflatt: Good idea

cedar-policy-validator/src/entity_manifest.rs
oflatt: Test case caught another bug!

cedar-policy-core/src/ast/request.rs
```diff
@@ -49,6 +49,17 @@ pub struct Request {
     pub(crate) context: Option<Context>,
 }
 
+/// Represents the principal type, resource type, and action UID.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Deserialize, Serialize)]
+pub struct RequestType {
```
khieta, line 54: Why do you need to define this type here if we already have `RequestEnv` in the validator code?

cedar-policy-validator/src/lib.rs
```diff
@@ -35,6 +35,7 @@ use cedar_policy_core::ast::{Policy, PolicySet, Template};
 use serde::Serialize;
 use std::collections::HashSet;
 
+pub mod entity_manifest;
```
khieta, line 38: IIRC other experimental features (like partial evaluation & partial validation) are hidden behind flags even in our "private" crates like `cedar-policy-validator`. So you may want to hide this import here behind `"entity-manifest"`

cedar-policy/src/api.rs
```diff
@@ -23,6 +23,12 @@
 )]
 
 mod id;
+#[cfg(feature = "entity-manifest")]
+use cedar_policy_validator::entity_manifest;
+#[cfg(feature = "entity-manifest")]
+pub use cedar_policy_validator::entity_manifest::{
+    AccessTrie, EntityManifest, EntityManifestError, EntityRoot, Fields, RootAccessTrie,
```
khieta, line 30: Nit: we've been trying to keep `pub use`s of error types (like `EntityManifestError`) in the `api/err.rs` file.

Also, for publicly used types, we usually add a "CAUTION..." comment to the type definition as a reminder that it's part of the public interface. Not super important for an experimental feature, but adding it now helps us avoid forgetting it in the future.

cedar-policy-validator/src/entity_manifest.rs
khieta: Returning the first error is fine, but returning all the errors would be more consistent with the rest of the validation APIs.

Re your second question: yes, validation does some additional steps (e.g., handling for templates). If you look at `validate` and `validate_policies` [here](https://github.com/cedar-policy/cedar/blob/672074429b67cf6ec2727c423409b4ac5ff73582/cedar-policy-validator/src/lib.rs#L110) you'll see that there are several pre-processing steps before we get to `typecheck_policy` (which is what calls `typecheck_by_request_env`). I think it'd be ideal to add you changes to the main validation path (if possible).

oflatt: I'm not sure how to add it to the validation path without it getting ugly- it would have to return both the current result and an optional entity manifest.
I'll keep this as-is for now.

cedar-policy-core/src/ast/request.rs
oflatt: As far as I can tell, entity manifests need less information than `RequestEnv` because we validate the policies in strict mode. This helps prevent some runtime errors (hopefully).

cedar-policy-validator/src/lib.rs
oflatt: thanks! done

cedar-policy/src/api.rs
oflatt: :+1:

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1266 @@
...
+
+/// The root of a data path or [`RootAccessTrie`].
+// CAUTION: this type is publicly exported in `cedar-policy`.
+// Don't make fields `pub`, don't make breaking changes, and use caution
+// when adding public methods.
+#[doc = include_str!("../../cedar-policy/experimental_warning.md")]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
+pub enum EntityRoot {
+    /// Literal entity ids
+    Literal(EntityUID),
```
john-h-kastner-aws, line 81: I think this will end up exporting `cedar_policy_core::ast::EntityUID` from `cedar_policy`. Not a blocking issue for an experimental feature, but we'll need to introduce some more wrapper type in `cedar_policy` if we want callers to be able to use this directly as a `cedar_policy::EntityUID`. Same comment for `Var` variant.

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1266 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct AccessTrie<T = ()> {
+    /// Child data of this entity slice.
+    /// The keys are edges in the trie pointing to sub-trie values.
+    #[serde_as(as = "Vec<(_, _)>")]
+    pub children: Fields<T>,
+    /// For entity types, this boolean may be `true`
+    /// to signal that all the ancestors in the entity hierarchy
+    /// are required (transitively).
+    pub ancestors_required: bool,
```
john-h-kastner-aws, line 140: depending on the use case for this serialization we'll probably want to include `#[serde(rename_all = "camelCase")]` on this and other structs. IIRC, all our external JSON formats use camel case. Not sure if we're consistent for internal DRT JSON

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1266 @@
...
+/// <datapath-expr> = <datapath-expr>.<field>
+///                   <datapath-expr> has <field>
+///                   <variable>
+///                   <entity literal>
+/// ```
+/// The `get_expr_path` function handles `datapath-expr` expressions.
+/// This error message tells the user not to use certain operators
+/// before accessing record or entity attributes, breaking this grammar.
+#[derive(Debug, Clone, Error, Hash, Eq, PartialEq)]
+#[error("For policy `{policy_id}`, failed to analyze expression while computing entity manifest.`")]
```
john-h-kastner-aws, line 178: IIRC. our error message convention is to omit trailing punctuation and not capitalize the first letter.

john-h-kastner-aws (commented):
Looks good. Main point of concern in the assumption that validation errors will be non-empty.

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1266 @@
...
+                    // using static analysis
+                    compute_root_trie(&typechecked_expr, policy.id())
+                }
+                PolicyCheck::Irrelevant(_) => {
+                    // this policy is irrelevant, so we need no data
+                    Ok(RootAccessTrie::new())
+                }
+
+                PolicyCheck::Fail(errors) => {
+                    // PANIC SAFETY: policy check fail should be a non-empty vector.
```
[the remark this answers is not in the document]
john-h-kastner-aws: Returning the full list of errors would avoid any uncertainty here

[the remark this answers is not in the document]
oflatt: That's a good idea
How should I implement `Diagnostic` for multiple validation errors? Should it print out the first one? What about when there are no errors?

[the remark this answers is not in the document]
cdisselkoen: You can take a look at [ParseErrors](https://github.com/cedar-policy/cedar/blob/a633019e3a8a456ef6581eb0d302eb5dbb5c3cf9/cedar-policy-core/src/parser/err.rs#L743) in Core for precedent.  To the first question, at least in that case we chose to just do the first one.  To the second question, we used `NonEmpty` to structurally guarantee we can't have an empty vector.

cedar-policy-validator/src/entity_manifest.rs
oflatt: Do you mean that this should use the existing wrapper `EntityUid`?
That's a little unfortunate since I can't import it from the validator crate.

cedar-policy-validator/src/types/request_env.rs
```diff
@@ -14,15 +14,16 @@
...
+use serde::Serialize;
 
 use crate::ValidatorSchema;
 
 use super::Type;
 
 /// Represents a request type environment. In principle, this contains full
 /// types for the four variables (principal, action, resource, context).
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize)]
```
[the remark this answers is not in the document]
oflatt: oops

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1266 @@
...
+    /// before a [`ExprKind::GetAttr`]
+    /// See [`FailedAnalysisError`] for more details.
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    FailedAnalysis(#[from] FailedAnalysisError),
+}
+
+/// Union two tries by combining the fields.
+fn union_fields<T: Clone>(first: &Fields<T>, second: &Fields<T>) -> Fields<T> {
+    let mut res = first.clone();
```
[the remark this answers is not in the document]
oflatt: Did this in a (coming soon) follow up PR that generalizes the analysis to all of cedar.

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1266 @@
...
+            trie: Default::default(),
+        }
+    }
+}
+
+impl<T: Clone> RootAccessTrie<T> {
+    /// Union two [`RootAccessTrie`]s together.
+    /// The new trie requests the data from both of the original.
+    fn union(&self, other: &Self) -> Self {
+        let mut res = self.clone();
```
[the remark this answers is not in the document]
oflatt: Did this in a (coming soon) follow up PR that generalizes the analysis to all of cedar.

cedar-policy-validator/src/entity_manifest.rs
john-h-kastner-aws: Right, Callers should be able to use it as they would the existing `EntityUid` wrapper. It does get a annoying since you'll end up wanting a wrapper for most of your structs, but we don't want consumers of `cedar-policy` to ever import `cedar-policy-core`

john-h-kastner-aws: The other option would be using the opaque inner struct pattern we use for error messages if users don't need to look inside the `Literal`.

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1266 @@
...
+#[serde_as]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct EntityManifest<T = ()>
+where
+    T: Clone,
+{
+    /// A map from request types to [`RootAccessTrie`]s.
+    #[serde_as(as = "Vec<(_, _)>")]
+    #[serde(bound(deserialize = "T: Default"))]
+    pub per_action: HashMap<RequestType, RootAccessTrie<T>>,
```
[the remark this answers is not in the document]
oflatt: Good catch

cedar-policy-validator/src/entity_manifest.rs
oflatt: That makes sense. I think people should be able to get the literal, so for now I'll add a TODO to implement wrappers for everything in `api.rs`

cedar-policy-validator/src/entity_manifest.rs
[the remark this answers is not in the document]
oflatt: Thanks to the pointer to `ParseErrors`
I'm still having trouble finding the case where there are no errors reported- could you point me in the right direction? What are undefined action and entity types?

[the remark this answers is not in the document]
oflatt: I'm also looking at `ValidationResult`, which claims that if there are no errors validation has passed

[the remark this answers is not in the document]
oflatt: One option could be to 
1) Run full validation, getting a proper error message.
2) Run `typecheck_by_request_env` to get the type-annotated policies, using panic if we encounter a type error

Another option would be to embed entity manifest computation into the typechecker to make this more efficient. But this seems kind of ugly.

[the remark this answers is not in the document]
john-h-kastner-aws: >  I'm still having trouble finding the case where there are no errors reported- could you point me in the right direction? What are undefined action and entity types?

https://github.com/cedar-policy/cedar/blob/a633019e3a8a456ef6581eb0d302eb5dbb5c3cf9/cedar-policy-validator/src/typecheck.rs#L390-L398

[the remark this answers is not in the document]
oflatt: In my case, all of the request environments are from the schema cross product so we should be okay I think.
But either way, I've rewritten the code to do full validation first, then get the type-annotated policies.

cedar-policy/src/api/id.rs
```diff
@@ -1,6 +1,6 @@
 /*
  * Copyright Cedar Contributors
- *
+y *
```
john-h-kastner-aws, line 3: `y`

oflatt: oops, thought I fixed that good catch

john-h-kastner-aws (approved):

cedar-policy-validator/src/entity_manifest.rs
oflatt: Ended up running full validation first before this code path.

cedar-policy/src/api/err.rs
```diff
@@ -1161,3 +1170,53 @@ pub mod request_validation_errors {
     #[diagnostic(transparent)]
     pub struct TypeOfContextError(#[from] cedar_policy_core::entities::json::GetSchemaTypeError);
 }
+
+/// An error generated by entity slicing.
+/// See [`FailedAnalysisError`] for details on the fragment
+/// of Cedar handled by entity slicing.
+// CAUTION: this type is publicly exported in `cedar-policy`.
+// Don't make fields `pub`, don't make breaking changes, and use caution
+// when adding public methods.
```
khieta, line 1179: No need for the warning here since this definition is directly in `cedar-policy`. Of course it doesn't hurt though 🤷‍♀️

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1306 @@
+//! Entity Manifest definition and static analysis.
+
+/*
+ * Copyright Cedar Contributors
```
khieta, line 4: In our other files, the copyright is at the top and any top-level comments come later. Not sure if there is an important reason for this, but might as well be consistent 🙂

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1306 @@
...
+}
+
+/// Error when expressions are partial during entity
+/// manifest computation
+// CAUTION: this type is publicly exported in `cedar-policy`.
+// Don't make fields `pub`, don't make breaking changes, and use caution
+// when adding public methods.
+#[derive(Debug, Clone, Error, Hash, Eq, PartialEq)]
+#[error(
+    "Entity slicing requires fully concrete policies. Got a policy with an unknown expression."
```
khieta, line 213: Our convention is that all error messages start with a lowercase letter, and do not end with punctuation
```suggestion
    "entity slicing requires fully concrete policies. Got a policy with an unknown expression"
```

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1306 @@
...
+
+impl Diagnostic for PartialExpressionError {}
+
+/// Error when the request is partial during entity
+/// manifest computation
+// CAUTION: this type is publicly exported in `cedar-policy`.
+// Don't make fields `pub`, don't make breaking changes, and use caution
+// when adding public methods.
+#[derive(Debug, Clone, Error, Hash, Eq, PartialEq)]
+#[error("Entity slicing requires a fully concrete request. Got a partial request.")]
```
khieta, line 225:
```suggestion
#[error("entity slicing requires a fully concrete request. Got a partial request")]
```

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1306 @@
...
+pub struct PartialRequestError {}
+impl Diagnostic for PartialRequestError {}
+
+/// An error generated by entity slicing.
+/// See [`FailedAnalysisError`] for details on the fragment
+/// of Cedar handled by entity slicing.
+#[derive(Debug, Error)]
+pub enum EntityManifestError {
+    /// A validation error was encountered
+    // TODO impl Error for ValidationResult (it already is implemented for api::ValidationResult)
```
khieta, line 235: Same note about making an issue for TODOs

cedar-policy-validator/src/entity_manifest.rs
```diff
@@ -0,0 +1,1306 @@
...
+    #[error("A validation error occurred.")]
```
khieta, line 236:
```suggestion
    #[error("A validation error occurred")]
```

khieta (approved):
Looks great! I just left some small nit-picky comments

cedar-policy/src/api/err.rs
oflatt: I'll keep it since I keep moving it around (:

cedar-policy/src/api.rs
```diff
@@ -23,6 +23,13 @@
 )]
 
 mod id;
+#[cfg(feature = "entity-manifest")]
+use cedar_policy_validator::entity_manifest;
+// TODO implement wrappers for these structs before they become public
```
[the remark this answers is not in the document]
oflatt: thanks, good idea

cedar-policy-validator/src/entity_manifest.rs
oflatt: done (:

khieta (approved):

PR #1108, opened 2024-08-02
require templates in the `Template` constructors

### LINKED ISSUE #1095: Clarify `Template` vs. `Policy`

In Cedar, we have three types of policies in play: static policies, templates, and template-linked policies (see https://docs.cedarpolicy.com/policies/templates.html for more details). Currently, the Rust API and JSON FFI are inconsistent in how they distinguish between these types.

In the JSON FFI:
* `Policy` represents a static policy
* `Template` represents a template or static policy (aka template with zero slots)
* `TemplateLink` represents a template-linked policy

In the Rust API:
* `Policy` represents a static policy or template-linked policy
* `Template` represents a template (although still supports parsing static policies, see example below)

IMO the difference in representation between static policies and template-linked policies is harmless (and not easily avoidable in the current implementation), but the difference in handling of `Template`s is confusing. We should either:
1. Update the Rust API documentation to reflect that a `Template` may also be a static policy (but the policy will have no effect during authorization unless linked), or 
2. _(My preference)_ Update both the JSON FFI and Rust API to return an error when constructing a `Template` with no slots.

---
Example of treating a static policy like a template:
```rust
let id = PolicyId::new("id");
let text = "permit(principal, action, resource);"; // static policy that permits everything
let template = Template::parse(Some(id), text).unwrap(); // ok
let mut pset = PolicySet::new();
pset.add_template(template).unwrap(); // also ok

// ... some defns omitted
let response = authorizer.is_authorized(&request, &pset, &entities);
assert_eq!(response.decision(), Decision::Deny); // response will be DENY since the template isn't linked
```

_No response_

### Additional context

_No response_

- [X] ⚠️ This feature might incur a breaking change

khieta:
## Description of changes

* The public-facing `Template` constructors will now return an error if passed a static policy.
* The internal `Template` type will continue to support both templates and static policies. There are several places in the parser and validator where we want a single type to represent both options because we use the same logic for both. I added comments & did some renaming in places where I felt this was unclear.
* Added new error variants to `FromJsonError` (`PolicyToTemplate`) and `ParseError` (`ExpectedTemplate`) to return in the case where a static policy is being used as a template.
* Made corresponding changes to the FFI code and test cases.

## Issue #, if available

Resolves #1095 

### DIFF
--- a/cedar-policy-core/src/est/err.rs
+++ b/cedar-policy-core/src/est/err.rs
@@ -34,6 +34,9 @@ pub enum FromJsonError {
     #[error("tried to convert JSON representing a template to a static policy")]
     #[diagnostic(transparent)]
     TemplateToPolicy(#[from] ast::UnexpectedSlotError),
+    /// Tried to convert an EST representing a static policy to an AST representing a template
+    #[error("tried to convert JSON representing a static policy to a template")]
+    PolicyToTemplate,
     /// Slot name was not valid for the position it was used in. (Currently, principal slots must
     /// be named `?principal`, and resource slots must be named `?resource`.)
     #[error("invalid slot name or slot used in wrong position")]
--- a/cedar-policy-core/src/est/err.rs
+++ b/cedar-policy-core/src/est/err.rs
@@ -31,12 +31,13 @@ pub enum FromJsonError {
     #[diagnostic(transparent)]
     JsonDeserializationError(#[from] JsonDeserializationError),
     /// Tried to convert an EST representing a template to an AST representing a static policy
-    #[error("tried to convert JSON representing a template to a static policy")]
+    #[error(transparent)]
     #[diagnostic(transparent)]
-    TemplateToPolicy(#[from] ast::UnexpectedSlotError),
+    TemplateToPolicy(#[from] parse_errors::ExpectedStaticPolicy),
     /// Tried to convert an EST representing a static policy to an AST representing a template
-    #[error("tried to convert JSON representing a static policy to a template")]
-    PolicyToTemplate,
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    PolicyToTemplate(#[from] parse_errors::ExpectedTemplate),
     /// Slot name was not valid for the position it was used in. (Currently, principal slots must
     /// be named `?principal`, and resource slots must be named `?resource`.)
     #[error("invalid slot name or slot used in wrong position")]
@@ -105,4 +106,10 @@ pub enum LinkingError {
         /// Slot which didn't have a value provided for it
         slot: ast::SlotId,
     },
+}
+
+impl From<ast::UnexpectedSlotError> for FromJsonError {
+    fn from(err: ast::UnexpectedSlotError) -> Self {
+        Self::TemplateToPolicy(err.into())
+    }
 }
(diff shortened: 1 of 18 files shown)


cedar-policy-core/src/est/err.rs
```diff
@@ -34,6 +34,9 @@ pub enum FromJsonError {
     #[error("tried to convert JSON representing a template to a static policy")]
     #[diagnostic(transparent)]
     TemplateToPolicy(#[from] ast::UnexpectedSlotError),
+    /// Tried to convert an EST representing a static policy to an AST representing a template
+    #[error("tried to convert JSON representing a static policy to a template")]
+    PolicyToTemplate,
```
cdisselkoen, line 39: just for consistency, even though this isn't a public error type (I think?), should we make `struct PolicyToTemplateError {}` and have this be `PolicyToTemplate(#[from] PolicyToTemplateError)`?

cedar-policy-core/src/parser.rs
```diff
@@ -112,24 +113,40 @@ pub fn parse_policy_template(
     cst.to_policy_template(id)
 }
 
-/// Like `parse_policy_template()`, but also returns the (lossless) EST -- that
+/// Like `parse_policy_or_template()`, but also returns the (lossless) EST -- that
 /// is, the EST of the original template without any of the lossy transforms
```
cdisselkoen, line 117:
```suggestion
/// is, the EST of the original policy/template without any of the lossy transforms
```

cdisselkoen (approved):
Looks great

john-h-kastner-aws (approved):
DRT build fail should be fixed by https://github.com/cedar-policy/cedar-spec/pull/401

cedar-policy-core/src/est/err.rs
khieta: Done in the latest commit. Also updated the CST->AST path to use the same error structure for consistent messaging

khieta:
Need one re-approval here & approvals on the downstream fixes:
* https://github.com/cedar-policy/cedar-java/pull/182
* https://github.com/cedar-policy/cedar-spec/pull/407

cedar-policy-core/src/parser/err.rs
```diff
@@ -442,6 +449,29 @@ pub mod parse_errors {
...
+                ast::UnexpectedSlotError::FoundSlot(slot) => Self { slot },
+            }
+        }
+    }
+
+    /// Details about a `ExpectedTemplate` error.
+    #[derive(Debug, Clone, Diagnostic, Error, PartialEq, Eq)]
+    #[error("expected a template, got a static policy")]
+    #[diagnostic(help("a template should include slot(s) `?principal` or `?resource`"))]
+    pub struct ExpectedTemplate {}
```
cdisselkoen, line 473: can we make unit structs like this `non_exhaustive`?  Without `non_exhaustive`, since we have no private fields, users can still construct this struct with `ExpectedTemplate {}`, so adding fields is still breaking.

cdisselkoen (approved):

cedar-policy-core/src/parser/err.rs
khieta: Thanks for pointing that out - I didn't realize it was a problem. I've updated the type to match the approach taken [here](https://github.com/cedar-policy/cedar/blob/8b74d95cba4abe314783a152b34a6b4e9ca39265/cedar-policy/src/api/err.rs#L582) with a `_dummy` field.

cdisselkoen (approved):

PR #1111, opened 2024-08-02
rename `schema_file_format` to `json_schema`

cdisselkoen:
## Description of changes

A number of pedestrian changes in this PR:

* Rename the `schema_file_format` module to `json_schema`, to make clearer it describes the JSON format specifically. This mirrors the existing module `human_schema`.
* Rename the `json_schema::SchemaFragment` type to `json_schema::Fragment`.
* Rename the `json_schema::SchemaType` type to `json_schema::Type`.
    * Previously we had `Type` in the root namespace, `SchemaType` in the root namespace, and `human_schema::Type`.  Now we have `Type` in the root namespace, `json_schema::Type`, and `human_schema::Type`.
    * Also rename `try_schema_type_into_validator_type()` to `try_jsonschema_type_into_validator_type()`.
* Rename the `json_schema::SchemaTypeVariant` type to `json_schema::TypeVariant`.
* Don't `use` items from `json_schema` in the root namespace of `cedar_policy_validator`.  Also avoid `use`ing these items in our own internal submodules outside of the `json_schema` module.  This has the effect of clarifying where JSON structures are being used.  For some important examples:
    * In `human_schema/to_json_schema.rs`, one function signature is now `Schema -> json_schema::Fragment` instead of `Schema -> SchemaFragment`.
    * In `schema.rs`, instead of constructing a `ValidatorSchemaFragment` from a `SchemaFragment`, we now construct a `ValidatorSchemaFragment` from a `json_schema::Fragment`.
    * In `namespace_def.rs`, instead of constructing a `ValidatorNamespaceDef` from a `NamespaceDefinition`, we now construct a `ValidatorNamespaceDef` from a `json_schema::NamespaceDefinition`.
    * Also in `namespace_def.rs`, instead of converting `SchemaType`s to `Type`s, we convert `json_schema::Type`s to `Type`s.
    * Outside of the `json_schema` module itself, this further distinguishes `json_schema`'s `EntityType` from `core::ast::EntityType` which is frequently `use`d in `cedar_policy_validator`.
* Export `Path` in `human_schema` directly, as opposed to just living in `human_schema::ast`.
* In tests, prefer to use `json_schema::SchemaFragment::from_json_value()` and similar, over direct calls to `serde_json`.
* No changes to the public (`cedar_policy`) API.

### DIFF
--- a/cedar-policy-validator/src/err.rs
+++ b/cedar-policy-validator/src/err.rs
@@ -386,7 +386,7 @@ pub mod schema_errors {
     #[derive(Debug, Diagnostic, Error)]
     #[diagnostic(help("any actions appearing as parents need to be declared as actions"))]
     pub struct ActionResolutionError(
-        pub(crate) NonEmpty<crate::schema_file_format::ActionEntityUID<crate::ConditionalName>>,
+        pub(crate) NonEmpty<crate::json_schema::ActionEntityUID<crate::ConditionalName>>,
     );
 
     impl ActionResolutionError {
--- a/cedar-policy-validator/src/human_schema.rs
+++ b/cedar-policy-validator/src/human_schema.rs
@@ -17,6 +17,7 @@
 //! The Cedar human/natural syntax for schemas
 
 mod ast;
+pub use ast::Path;
 mod err;
 pub mod fmt;
 pub mod parser;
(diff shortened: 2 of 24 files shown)


khieta (approved):

shaobo-he-aws (approved):

john-h-kastner-aws (approved):
Looks good. We could also consider renaming the `ValidatorSchema` and similar structs to just `Schema`

PR #1124, opened 2024-08-08
enhance `impl_diagnostic` macros

cdisselkoen:
## Description of changes

Enhances our `impl_diagnostic` macros, allowing them to be used in a few more places.

Partially addresses #977 -- specifically, addresses #977 for `schema_warnings::ShadowsBuiltinWarning` and `schema_warnings::ShadowsEntityWarning`, and also for generic `cedar_policy_core::parser::Node`.

### DIFF
--- a/cedar-policy-core/src/evaluator/err.rs
+++ b/cedar-policy-core/src/evaluator/err.rs
@@ -330,7 +330,7 @@ pub mod evaluation_errors {
     // Or, we could have separate fields for source code and label instead of
     // combining them into `Loc`, which would work around the issue.
     impl Diagnostic for EntityDoesNotExistError {
-        impl_diagnostic_from_source_loc_field!();
+        impl_diagnostic_from_source_loc_opt_field!(source_loc);
 
         fn help<'a>(&'a self) -> Option<Box<dyn std::fmt::Display + 'a>> {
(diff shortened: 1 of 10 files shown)


cedar-policy-core/src/evaluator/err.rs
```diff
@@ -330,7 +330,7 @@ pub mod evaluation_errors {
     // Or, we could have separate fields for source code and label instead of
     // combining them into `Loc`, which would work around the issue.
     impl Diagnostic for EntityDoesNotExistError {
-        impl_diagnostic_from_source_loc_field!();
+        impl_diagnostic_from_source_loc_opt_field!(source_loc);
 
         fn help<'a>(&'a self) -> Option<Box<dyn std::fmt::Display + 'a>> {
             None
```
john-h-kastner-aws, line 336: While we're in here, I think we could delete these `None` functions

john-h-kastner-aws (approved):

cedar-policy-core/src/evaluator/err.rs
cdisselkoen: Sure, done in the latest commit

john-h-kastner-aws (approved):

Issue #1116, opened 2024-08-06
Make `Policy` and `Template` APIs consistent wrt annotations


cdisselkoen:
### Category

User level API features/changes

### Describe the feature you'd like to request

As noted in [this TODO in the code](https://github.com/cedar-policy/cedar/blob/97fd8cf53f6b483bb0b2ae0495340c2665317f5b/cedar-policy/src/api.rs#L1897)

`PolicySet` has two methods `annotation()` and `template_annotation()` which are very similar except that one expects a policy ID and one expects a template ID. However, the former returns `Option<&str>` and the latter returns `Option<String>`.  We should make this consistent; I think `Option<&str>` is better.

### Describe alternatives you've considered

We could converge on `Option<String>` instead.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [X] ⚠️ This feature might incur a breaking change

PR #1131, opened 2024-08-08
Unify the signature of `PolicySet::annotation` and `PolicySet::template_annotation`

shaobo-he-aws:
## Description of changes

This PR change the return type to both methods to `Option<&str>`. Due to the author's limited knowledge of Rust, I have to add a new function in core to get a reference to the `Template`. Otherwise we could let the return type be `Option<SmolStr>`.

## Issue #, if available
#1116 

### DIFF
--- a/cedar-policy-core/src/ast/policy_set.rs
+++ b/cedar-policy-core/src/ast/policy_set.rs
@@ -457,9 +457,14 @@ impl PolicySet {
         self.templates.is_empty() && self.links.is_empty()
     }
 
-    /// Lookup a template by policy id
+    /// Lookup a template by policy id, returns [`Option<Arc<Template>>`]
     pub fn get_template(&self, id: &PolicyID) -> Option<Arc<Template>> {
         self.templates.get(id).cloned()
+    }
+
+    /// Lookup a template by policy id, returns [`Option<&Template>`]
+    pub fn get_template_ref(&self, id: &PolicyID) -> Option<&Template> {
+        self.templates.get(id).map(AsRef::as_ref)
     }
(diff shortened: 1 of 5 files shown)


khieta (approved):
Looks good to me. I'll leave it to a Rust expert to comment on whether `get_template_ref` is needed.

cedar-policy-core/src/ast/policy_set.rs
```diff
@@ -457,11 +457,16 @@ impl PolicySet {
...
-    /// Lookup a template by policy id
+    /// Lookup a template by policy id, returns [`Option<Arc<Template>>`]
     pub fn get_template(&self, id: &PolicyID) -> Option<Arc<Template>> {
         self.templates.get(id).cloned()
     }
 
+    /// Lookup a template by policy id, returns [`Option<&Template>`]
+    pub fn get_template_ref(&self, id: &PolicyID) -> Option<&Template> {
+        self.templates.get(id).map(AsRef::as_ref)
+    }
```
cdisselkoen, line 468: I think in other places we call these `get_template_arc` and just `get_template`.  Should we change the other places to match this convention?

cdisselkoen (approved):
Looks good to me. I don't see a problem with the proposed Core interface

cedar-policy-core/src/ast/policy_set.rs
shaobo-he-aws: I think `get_template_arc` and `get_template` make more sense to me.

cdisselkoen (approved):

PR #1138, opened 2024-08-09
Context `merge` function (#1027)

khieta:
## Description of changes

Backport #1027, which adds new APIs for `Context`. It also modifies some error types, but these are not publicly exposed in 3.x.


cdisselkoen (approved):

khieta:
Note: DRT failure is due to a breaking change in `cedar-policy-core`, not the public API. It should be resolved by backporting https://github.com/cedar-policy/cedar-spec/pull/376 into `cedar-spec`.

shaobo-he-aws (approved):

PR #1140, opened 2024-08-09
require templates in the `Template` constructors (#1108)

khieta:
## Description of changes

Backport of #1108. Last of the currently planned changes for 3.3.0.

### DIFF
--- a/cedar-policy-core/src/ast/policy_set.rs
+++ b/cedar-policy-core/src/ast/policy_set.rs
@@ -514,7 +514,7 @@ mod test {
         let p1 = parser::parse_policy(Some("id".into()), "permit(principal,action,resource);")
             .expect("Failed to parse");
         pset.add_static(p1).expect("Failed to add!");
-        let template = parser::parse_policy_template(
+        let template = parser::parse_policy_or_template(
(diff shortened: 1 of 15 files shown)


john-h-kastner-aws (approved):

khieta:
The Java & SemVer CI failures are expected, but I'm a little confused about the `msrv` failure since I can't seem to reproduce it locally. Suggestions welcome.

cdisselkoen:
Seems to reproduce for me on this PR's branch with `cargo hack check --rust-version --workspace --all-targets --ignore-private`. Weirdly, the error is at `cedar-policy/src/tests.rs:3440` with 
```text
could not find `test_utils` in `cedar_policy_core`
```
But `cargo test --all-features` passes with Rust 1.80, so what about 1.76 is causing it to not find that module?

khieta:
Issue resolved by updating the `cedar-policy` `Cargo.toml` file to match the main branch. (The reason I couldn't reproduce before was that I was too lazy to run `cargo install cargo-hack` and was trying to reproduce manually... my bad.)

cdisselkoen (approved):

PR #1146, opened 2024-08-16
Level validation Rust implementation (RFC 76)

andrewmwells-amazon:
## Description of changes
Rust implementation of Level validation (RFC 76: https://github.com/cedar-policy/rfcs/pull/76)

@aaronjeline is working on Lean changes.

Example error messages:
Screenshot 2024-08-16 at 3 23 47 PM

## Issue #, if available
RFC 76 (https://github.com/cedar-policy/rfcs/pull/76)


cedar-policy-validator/src/diagnostics/validation_errors.rs
```diff
@@ -448,6 +449,114 @@ impl Diagnostic for HierarchyNotRespected {
...
+pub struct EntityDerefLevel {
+    /// `None` represents infinity.
+    pub level: Option<i64>,
+}
+
+impl Display for EntityDerefLevel {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::result::Result<(), std::fmt::Error> {
+        match self.level {
+            Some(l) => write!(f, "{l}"),
+            None => write!(f, "INF"),
```
aaronjeline, line 464: Is this a public display impl, if so we probably want a nicer display than `INF` (just `infinity`? `infinite`?)

cedar-policy-validator/src/diagnostics/validation_errors.rs
```diff
@@ -448,6 +449,114 @@ impl Diagnostic for HierarchyNotRespected {
     }
 }
 
+#[derive(Debug, Clone, Hash, Eq, PartialEq, Error, Copy)]
+/// Represents how many entity dereferences can be applied to a node.
+/// `None` represents infinity.
+pub struct EntityDerefLevel {
+    /// `None` represents infinity.
+    pub level: Option<i64>,
```
aaronjeline, line 457: Why signed?

cedar-policy-validator/src/lib.rs
```diff
@@ -206,6 +286,192 @@ impl Validator {
...
+            match policy_check {
+                PolicyCheck::Success(e) => {
+                    let res =
+                        self.check_entity_deref_level_helper(&e, max_allowed_level, policy_id);
+                    match res.1 {
+                        Some(e) => errs.push(ValidationError::EntityDerefLevelViolation(e)),
+                        None => (),
+                    }
+                }
+                PolicyCheck::Fail(e) => println!("Typechecking failed: {:?}", e),
```
aaronjeline, line 312: Why is this `println` here?

cedar-policy-validator/src/lib.rs
```diff
@@ -206,6 +286,192 @@ impl Validator {
...
+                PolicyCheck::Irrelevant(_) => (), //Don't report level violations if typechecking already failed
+            }
+        }
+        errs
+    }
+
+    #[cfg(feature = "level-validate")]
+    fn min(
```
aaronjeline, line 320: The minimum of an empty list is infinity?

aaronjeline: I think we can change this to take a non-empty vec and remove a potential footgun in the future

aaronjeline (changes requested):
This is slightly different the lean impl. In the lean impl `Level 0` is the level that can't be de-referenced. I think they should align on that. Otherwise, looks good. Very different algorithm to the lean, but I think you said that was so you can get better error messages.

cedar-policy-validator/src/lib.rs
```diff
@@ -206,6 +286,192 @@ impl Validator {
...
+                PolicyCheck::Success(e) => {
+                    let res =
+                        self.check_entity_deref_level_helper(&e, max_allowed_level, policy_id);
+                    match res.1 {
+                        Some(e) => errs.push(ValidationError::EntityDerefLevelViolation(e)),
+                        None => (),
+                    }
+                }
+                PolicyCheck::Fail(e) => println!("Typechecking failed: {:?}", e),
+                PolicyCheck::Irrelevant(_) => (), //Don't report level violations if typechecking already failed
```
john-h-kastner-aws, line 315: This might not be sound. IIRC, the irrelevant case includes policies with type `False`, but something like `principal.foo.bar.baz && false` has type `False` while still being a potential level violation.

cedar-policy-validator/src/lib.rs
```diff
@@ -206,6 +286,192 @@ impl Validator {
...
+    #[cfg(feature = "level-validate")]
+    fn min(
+        v: Vec<(EntityDerefLevel, Option<EntityDerefLevelViolation>)>,
+    ) -> (EntityDerefLevel, Option<EntityDerefLevelViolation>) {
+        match v.first() {
+            Some(e) => {
+                let mut p = e.clone();
+                for e in v {
+                    if e.0 < p.0 {
+                        p = e.clone();
```
john-h-kastner-aws, line 330: On a hunch, these `clone`s don't feel like they should be necessary

cedar-policy-validator/src/diagnostics/validation_errors.rs
andrewmwells-amazon: It's for error messages. Actual level is `allowed_level.add(actual_level.neg())`

cedar-policy-validator/src/lib.rs
```diff
@@ -206,6 +286,192 @@ impl Validator {
...
+                    }
+                }
+                p
+            }
+            None => (EntityDerefLevel { level: None }, None),
+        }
+    }
+
+    #[cfg(feature = "level-validate")]
+    ///Walk the type-annotated AST and compute the used level and possible violation
```
john-h-kastner-aws, line 340: Can you add a comment here to explain how I should interpret the returned `EntityDerefLevel`.

cedar-policy-validator/src/lib.rs
```diff
@@ -206,6 +286,192 @@ impl Validator {
...
+                Self::min(v)
+            }
+            ExprKind::Record(fields) => {
+                let v: Vec<(EntityDerefLevel, Option<_>)> = fields
+                    .iter()
+                    .map(|(_, l)| {
+                        self.check_entity_deref_level_helper(l, max_allowed_level, policy_id)
+                    })
+                    .collect();
+                Self::min(v)
```
john-h-kastner-aws, line 473: This case isn't obvious to me. What's the intended behavior for something like `{foo: principal, bar: <whatever>}.foo`?

cedar-policy-validator/src/lib.rs
andrewmwells-amazon: I started implementing that, but I think the current version is actually less error-prone. If we add more support for empty sets / extension functions with no args in the future, this code would handle that correctly, while `NonEmpty::from_vec(v).unwrap()` would panic. I don't want to add actual error handling for these cases that are currently impossible. Marking these expressions as non-dereferenceable should be future-proof (baring some possbile weird extension functions I haven't thought of).

cedar-policy-validator/src/lib.rs
```diff
@@ -511,3 +777,87 @@ mod test {
         );
     }
 }
+
+#[cfg(feature = "level-validate")]
+#[cfg(test)]
+mod levels_validation_tests {
```
john-h-kastner-aws, line 785: Needs some more test cases, particularly if-then-else, and records

cedar-policy-validator/src/diagnostics/validation_errors.rs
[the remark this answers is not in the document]
andrewmwells-amazon: Added a comment explaining what negative values represent.

`None` is no longer constructed, but I'd like to keep the code this way in case we add `level` back into the schema.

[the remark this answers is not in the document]
john-h-kastner-aws: I'd rather get rid of it and add it later if we need to, but I don't feel too strongly about it

cedar-policy-validator/src/lib.rs
```diff
@@ -206,6 +286,194 @@ impl Validator {
...
+                    (new_level, None)
+                }
+            }
+            ExprKind::BinaryApp { arg1, arg2, .. } => {
+                let es = [arg1, arg2];
+                let v: Vec<(EntityDerefLevel, Option<_>)> = es
+                    .iter()
+                    .map(|l| self.check_entity_deref_level_helper(l, max_allowed_level, policy_id))
+                    .collect();
+                Self::min(v)
```
john-h-kastner-aws, line 410: By returning the `min` here I think you get things like `principal == principal` has level `max_allowed_level` which feels wrong. Could that cause trouble anywhere? `(principal == principal).foo` won't ever be allowed due to typechecking, but maybe there's a another troublesome case.

cedar-policy-validator/src/lib.rs
```diff
@@ -206,6 +286,194 @@ impl Validator {
...
+                PolicyCheck::Fail(_) => unreachable!(),
+                PolicyCheck::Irrelevant(_) => (), //Don't report level violations if typechecking already failed
+            }
+        }
+        errs
+    }
+
+    #[cfg(feature = "level-validate")]
+    fn min(
+        v: Vec<(EntityDerefLevel, Option<EntityDerefLevelViolation>)>,
```
john-h-kastner-aws, line 323: We should be able to have this take an `impl IntoIterator<...>` to avoid some vec allocations at call sites

cedar-policy-validator/src/lib.rs
```diff
@@ -206,6 +286,194 @@ impl Validator {
...
+                self.check_entity_deref_level_helper(arg, max_allowed_level, policy_id)
+            }
+            // `In` operator decrements the LHS only
+            ExprKind::BinaryApp { op, arg1, arg2 } if op == &BinaryOp::In => {
+                let mut lhs =
+                    self.check_entity_deref_level_helper(arg1, max_allowed_level, policy_id);
+                let rhs = self.check_entity_deref_level_helper(arg2, max_allowed_level, policy_id);
+                lhs = (lhs.0.decrement(), lhs.1);
+                let new_level = Self::min(vec![lhs, rhs]).0;
+                if new_level.level < Some(0) {
```
john-h-kastner-aws, line 390: Is `rhs` relevant to this comparison?

cedar-policy-validator/src/lib.rs
andrewmwells-amazon: Fixed to only clone the return value.

cedar-policy-validator/src/lib.rs
john-h-kastner-aws: right, I'm wondering if it could ever cause `min` to be less than 0 when there wasn't already an error raised earlier

cedar-policy-validator/src/lib.rs
andrewmwells-amazon: My understanding of the RFC is that it should recurse into `<whatever>` in determining the maximum number of dereferences that could occur.

cedar-policy-validator/src/lib.rs
andrewmwells-amazon: That feels right to me. Each node's `level` is the maximum number of times it could be dereferenced. I don't see how `principal == principal` could be dereferenced at all, but in a hypothetical world where it could be, I'd exect it to be dereferenceable `max_allowed_level` times.

andrewmwells-amazon:
> This is slightly different the lean impl. In the lean impl Level 0 is the level that can't be de-referenced. I think they should align on that.

That's my understanding too. Each node's `level` gives the number of allowable dereferences. `0` means no dereferences are allowed. A negative number means we've already done more dereferences than allowed. `None` means infinite dereferences allowed. Is there a place where I'm not following this?

cedar-policy-validator/src/lib.rs
andrewmwells-amazon: Actually rereading the RFC, it isn't explicit but "Note that for records, individual attributes may have entities with different levels. E.g., { a: principal, b: principal.manager } would have type { a: User(2), b: User(1) }." sounds like it should only consider `foo` here.

cedar-policy-validator/src/lib.rs
john-h-kastner-aws: fair enough

cedar-policy-validator/src/lib.rs
john-h-kastner-aws: That also how I read it

andrewmwells-amazon: Latest revision considers the level of the specific attribute being accessed.

cedar-policy-validator/src/lib.rs
andrewmwells-amazon: Added in latest revision

cedar-policy-validator/src/typecheck.rs
```diff
@@ -155,7 +155,10 @@ impl<'a> Typechecker<'a> {
                 (false, true, None) => PolicyCheck::Fail(type_errors),
                 (false, true, Some(e)) => PolicyCheck::Success(e),
                 (false, false, _) => PolicyCheck::Fail(type_errors),
-                (true, _, _) => PolicyCheck::Irrelevant(type_errors),
+                (true, _, Some(e)) => PolicyCheck::Irrelevant(type_errors, e),
+                // PANIC SAFETY: `is_false` implies e is `Some(Lit(Bool(false)))`
+                #[allow(clippy::unreachable)]
+                (true, _, None) => unreachable!(),
```
john-h-kastner-aws, line 161: This and the one below should be safe, but the panic safety comment is too strong. `ty.typechecked() == true` implies `ty.into_typed_expr().is_some()`, but it might not be a lit bool

andrewmwells-amazon: Changed to "`is_false` implies `e` has a type implies `Some(e)`."

cedar-policy-validator/src/lib.rs
andrewmwells-amazon: Fixed

cedar-policy-validator/src/lib.rs
```diff
@@ -127,6 +134,33 @@ impl Validator {
         )
     }
 
+    #[cfg(feature = "level-validate")]
```
shaobo-he-aws, line 137: Super nit-pick: Can we move relevant code to a module (file)?

shaobo-he-aws (approved):

cedar-policy/src/api.rs
```diff
@@ -1258,22 +1258,21 @@ impl Validator {
...
-    /// was found. If a policy id included in the input policy set does not appear in the
-    /// output iterator, then that policy passed the validator. If the function
+    /// Strictly validate all policies in a policy set, collecting all validation errors
+    /// found into the returned `ValidationResult`. If strict validation passes, run level
+    /// validation (RFC 76). Each error is returned together with the policy id of the policy
+    /// where the error was found. If a policy id included in the input policy set does not
+    /// appear in the output iterator, then that policy passed the validator. If the function
     /// `validation_passed` returns true, then there were no validation errors found, so
     /// all policies in the policy set have passed the validator.
-    pub fn validate_with_level(
```
john-h-kastner-aws: What was the reason for this change? I don't see how the algorithm would rely on strict typechecking in particular.

I'd rather keep the old API with the `mode` param if that possible. Not a big deal otherwise

cedar-policy-validator/src/lib.rs
john-h-kastner-aws: agree. That might be nice

cedar-policy-validator/src/lib.rs
andrewmwells-amazon: I'm not sure I understand. Shouldn't the level of `principal.foo.bar.baz + resource.foo.bar.baz.biz` should consider the rhs? The most severe error is "propagated" up the expression tree using `min` so we can report the actual level.

john-h-kastner-aws (approved):

cedar-policy-validator/src/diagnostics/validation_errors.rs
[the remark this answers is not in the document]
andrewmwells-amazon: Done

cedar-policy/src/tests.rs
```diff
@@ -3923,6 +3923,220 @@ mod partial_schema {
...
+        let src = r#"permit(principal == User::"һenry", action, resource) when {principal in resource.foo};"#;
+        let p = Policy::parse(None, src).unwrap();
+        set.add(p).unwrap();
+
+        let result = validator.strict_validate_with_level(&set, 0);
+        assert!(!result.validation_passed());
+        assert_eq!(result.validation_errors().count(), 1);
+        assert_matches!(
+            result.validation_errors().next().unwrap(),
+            ValidationError::EntityDerefLevelViolation(_)
```
john-h-kastner-aws, line 4008: To check my understanding, `actual_level` in the struct is `-1` here and this then gets printed as `0 + - (-1)`, showing "Actual level is 1", correct?

john-h-kastner-aws (approved):

cedar-policy-validator/src/lib.rs
```diff
@@ -206,6 +267,205 @@ impl Validator {
...
+            ExprKind::UnaryApp { arg, .. } => {
+                self.check_entity_deref_level_helper(arg, max_allowed_level, policy_id)
+            }
+            // `In` operator decrements the LHS only
+            ExprKind::BinaryApp { op, arg1, arg2 } if op == &BinaryOp::In => {
+                let mut lhs =
+                    self.check_entity_deref_level_helper(arg1, max_allowed_level, policy_id);
+                let rhs = self.check_entity_deref_level_helper(arg2, max_allowed_level, policy_id);
+                lhs = (lhs.0.decrement(), lhs.1);
+                let new_level = Self::min(vec![lhs, rhs]).0;
```
aaronjeline, line 363: I'm not sure why we're looking for min here? 
> The rules for expression e1 in e2 (line (3) in Fig. 7)(a) in the paper) are extended to require that the entity type of e1 is E1(n) where n > 0. This requirement indicates that e1 must be an entity from which you can dereference at least one level (to access its ancestors).

Could you explain?

cedar-policy-validator/src/lib.rs
```diff
@@ -206,6 +267,205 @@ impl Validator {
...
+            }
+            ExprKind::UnaryApp { arg, .. } => {
+                self.check_entity_deref_level_helper(arg, max_allowed_level, policy_id)
+            }
+            // `In` operator decrements the LHS only
+            ExprKind::BinaryApp { op, arg1, arg2 } if op == &BinaryOp::In => {
+                let mut lhs =
+                    self.check_entity_deref_level_helper(arg1, max_allowed_level, policy_id);
+                let rhs = self.check_entity_deref_level_helper(arg2, max_allowed_level, policy_id);
+                lhs = (lhs.0.decrement(), lhs.1);
```
aaronjeline, line 362: Tiny nit, but this could lack `mut` and just bind a new `lhs`

cedar-policy-validator/src/lib.rs
aaronjeline: I _think_ this is correct, it's just phrased very different from the type system version. arg2 can be zero, because it checks for being less than zero, which only happens due to decrements.

aaronjeline (approved):

cedar-policy-validator/src/lib.rs
john-h-kastner-aws: ISTM the `min` is intended to handle situations where there may be a levels violation on the RHS of the `in`. Something like `principal in resource.foo.bar` validated at level 0 might get reported as a level violation at level 2 due to `resource.foo.bar` rather than at level 1 due to `principal in ...`. 

I asked about the same code here: https://github.com/cedar-policy/cedar/pull/1146#discussion_r1783398534

Probably worth another test demonstrating the case where the `min` value is determined b `arg2`, but `arg1` level is also less than 0.

cedar-policy/src/tests.rs
andrewmwells-amazon: Yes.

cedar-policy/src/api.rs
andrewmwells-amazon: I don't see why it wouldn't work with permissive validation, but I haven't thought about it enough (and more importantly haven't tested it) so I'd rather merge this and add permissive validation in the future.

cedar-policy-validator/src/lib.rs
andrewmwells-amazon: Done.

cedar-policy-validator/src/level_validate.rs
```diff
@@ -0,0 +1,340 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+//! Validator for Cedar policies
+#![forbid(unsafe_code)]
+#![deny(
```
john-h-kastner-aws, line 19: Should't need these outside the toplevel lib.rs

john-h-kastner-aws (approved):

cedar-policy-validator/src/level_validate.rs
andrewmwells-amazon: Interesting, it warned me that was the case for `#![warn(rust_2018_idioms)]` but not for these.

john-h-kastner-aws (approved):

PR #1169, opened 2024-09-03
add non_exhaustive to a couple more types

cdisselkoen:
## Description of changes

Adds `non_exhaustive` to a couple more error types related to parsing or validation.  Most of our public error types related to parsing or validation already had `non_exhaustive`.

- [x] updates the CHANGELOG appropriately

### DIFF
--- a/cedar-policy-validator/src/cedar_schema/err.rs
+++ b/cedar-policy-validator/src/cedar_schema/err.rs
@@ -580,6 +580,8 @@ impl Diagnostic for MultipleEAMapDeclarationsError {
 }
 
 /// Errors relating to embedded attribute maps (`EAMap`s)
+//
+// This is NOT a publicly exported error type.
 #[derive(Debug, Clone, Diagnostic, Error, PartialEq, Eq)]
 #[non_exhaustive]
 pub enum EAMapError {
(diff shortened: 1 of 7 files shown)


cedar-policy-validator/src/cedar_schema/err.rs
```diff
@@ -581,6 +581,7 @@ impl Diagnostic for MultipleEAMapDeclarationsError {
 
 /// Errors relating to embedded attribute maps (`EAMap`s)
 #[derive(Debug, Clone, Diagnostic, Error, PartialEq, Eq)]
+#[non_exhaustive]
```
khieta, line 584: Assuming you're adding `non_exhaustive` to this error type (and `CedarSchemaParseErrors`) because it is publicly exposed -- can you add a CAUTION comment noting this?

cdisselkoen: Actually, turns out it is not publicly exposed.  So it doesn't matter as much if we (do or don't) add `non_exhaustive`. Preference?

khieta: Ok, it's fine as-is then. Just wanted to make sure we hadn't forgot to mark something as public 🙂

khieta (approved):

cedar-policy-validator/src/cedar_schema/err.rs
cdisselkoen: `CedarSchemaParseErrors` is not publicly exposed either

cdisselkoen: Added comments to that effect, at a minimum

khieta (approved):

shaobo-he-aws (approved):

PR #1172, opened 2024-09-04
some refactoring in types.rs

cdisselkoen:
## Description of changes

Some pure refactors (no functional change) in `types.rs`.  I had these in service of RFC 68 code, but they are separable so I thought it would be useful to review them separately and thus reduce the size of the coming RFC 68 PR. 

Brief summary of the changes:
- Adding and improving comments
- Rename `.all_attrs()` to `.all_known_attrs()` for clarity
- Pulling some existing code into dedicated helper functions `AttributeType::is_subtype()` and `AttributeType::least_upper_bound()`
- Return a reference rather than an owned type in one place

### DIFF
--- a/cedar-policy-validator/src/types.rs
+++ b/cedar-policy-validator/src/types.rs
@@ -259,12 +259,14 @@ impl Type {
 
             // `None` as an element type represents the top type for the set
             // element, so every other set is a subtype of set<None>, making a
-            // least upper bound containing  set<None> and another set type
+            // least upper bound containing set<None> and another set type
             // equal to set<None>. This case should be impossible due to the
             // subtype checks in the first two match cases, but we handle it
             // explicitly as an alternative to panicking if it occurs.
-            (ty_lub @ Type::Set { element_type: None }, Type::Set { .. })
-            | (Type::Set { .. }, ty_lub @ Type::Set { element_type: None }) => Ok(ty_lub.clone()),
+            (Type::Set { element_type: None }, Type::Set { .. })
+            | (Type::Set { .. }, Type::Set { element_type: None }) => {
+                Ok(Type::Set { element_type: None })
+            }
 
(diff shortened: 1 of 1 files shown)


aaronjeline (approved):
I think these are great refactors

cedar-policy-validator/src/types.rs
```diff
@@ -689,10 +703,12 @@ impl Display for Type {
                 write!(f, "{{")?;
                 for (name, ty) in attrs.iter() {
                     write!(f, "{name}")?;
-                    if !ty.is_required {
+                    if !ty.is_required() {
                         write!(f, "?")?;
                     }
-                    write!(f, ": {},", ty.attr_type)?;
+                    write!(f, ": ")?;
+                    ty.display(f)?;
```
john-h-kastner-aws, line 710: Could rename this function to `display_type` or something so it's clear this doesn't re-print  the optional qualifier.

cedar-policy-validator/src/types.rs
```diff
@@ -1376,11 +1404,57 @@ impl AttributeType {
...
+    }
+
+    /// Is `ty0` a subtype of `ty1`?
+    fn is_subtype(
+        schema: &ValidatorSchema,
+        ty0: &AttributeType,
+        ty1: &AttributeType,
+        mode: ValidationMode,
+    ) -> bool {
+        (if mode.is_strict() {
```
john-h-kastner-aws, line 1452: While we're moving the code around I might also split this up with a local for the result of optional-qualifier subtyping

john-h-kastner-aws (approved):

cedar-policy-validator/src/types.rs
cdisselkoen: Done

john-h-kastner-aws (approved):

Issue #440, opened 2023-11-16
Consider a different implementation of `typecheck_restricted_expr_against_schematype()`


cdisselkoen:
### Category

Internal refactors/changes

### Describe the feature you'd like to request

This comes from a TODO in the code in cedar-policy-core/src/entities/conformance.rs (as of this writing).

Currently, Core typechecks a restricted expression against a Core `SchemaType` by first calculating the `SchemaType` of the restricted expression, and then determining if the two `SchemaType`s are "consistent".  It may be less confusing, more efficient, and/or more precise to just typecheck directly.

In [a comment on #430](https://github.com/cedar-policy/cedar/pull/430#discussion_r1395940271), @khieta says she prefers the code as-is because it prevents duplication and reuses `schematype_of_restricted_expr()`.  This concern should be balanced against the one above.

Either we should change the implementation, or we should just remove the TODO if we determine the current implementation is the most desirable.

### Describe alternatives you've considered

We can also refactor/adjust `schematype_of_restricted_expr()` and/or `SchemaType::is_consistent_with()` if that would make things clearer.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

andrewmwells-amazon:
https://github.com/cedar-policy/cedar/pull/1177 adds a dedicated method here (though it still could be optimized further).

john-h-kastner-aws:
Resolved by #1177, #1190, #1194, and #1197 which together remove `schematype_of_restricted_expr` entirely and re-implement this function to typecheck directly.

Issue #1176, opened 2024-09-05
Entity validation behaves incorrectly on record-typed attributes


cdisselkoen:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Schemas and Validation

### Describe the bug

The [documentation on `Entities::from_entities()`](https://docs.rs/cedar-policy/3.3.0/cedar_policy/struct.Entities.html#method.from_entities) says

> If a schema is present, this function will also ensure that the produced entities fully conform to the schema – for instance, it will error if attributes have the wrong types (e.g., string instead of integer), or if required attributes are missing or superfluous attributes are provided.

In actuality, this function will _not_ correctly error if inside a record-typed attribute, an (inner) superfluous attribute is provided.

Confirmed this behavior on `main` and 3.3.0 as of this writing.  Doesn't seem to affect `Entities::from_json_*()` because the JSON deserialization code does catch the superfluous inner attribute.  For the same reason, this bug is not reproducible in the CLI (as far as I know).  It only affects users of `Entities::from_entities()` who constructed entities programmatically.

### Expected behavior

Expected behavior to match the docs.  Expected `Entities::from_entities()` to reject an entity with a superfluous inner attribute.

### Reproduction steps

The following works in an otherwise-blank Cargo project with dependency on `cedar_policy` 3.3.0 (and `miette` 7.2.0).

```rust
use cedar_policy::{Entities, Entity, EntityUid, RestrictedExpression, Schema};
use std::collections::{HashMap, HashSet};
use std::str::FromStr;

fn main() {
    let (schema, _) = Schema::from_cedarschema_str(
        "
        entity E {
          rec: {
            foo: Long
          }
        };
        action Act appliesTo {
          principal: [E],
          resource: [E],
        };
    ",
    )
    .unwrap();
    let entity = Entity::new(
        EntityUid::from_str(r#"E::"abc""#).unwrap(),
        HashMap::from_iter([(
            "rec".into(),
            RestrictedExpression::new_record([
                ("foo".into(), RestrictedExpression::new_long(4567)),
                (
                    "extra".into(),
[...]

PR #1177, opened 2024-09-05
Fix 1176

andrewmwells-amazon:
## Description of changes
When constructing the `SchemaType` of a concrete entity, make attributes required.

## Issue #, if available
https://github.com/cedar-policy/cedar/issues/1176

### DIFF
--- a/cedar-policy-core/src/entities/conformance.rs
+++ b/cedar-policy-core/src/entities/conformance.rs
@@ -182,35 +182,29 @@ pub fn typecheck_value_against_schematype(
 /// Check whether the given `RestrictedExpr` is a valid instance of `SchemaType`
 pub fn does_restricted_expr_implement_schematype(
     expr: BorrowedRestrictedExpr<'_>,
-    expr_ty: &SchemaType,
     expected_ty: &SchemaType,
     extensions: &Extensions<'_>,
 ) -> bool {
     use SchemaType::*;
-    if expr_ty == expected_ty {
-        return true;
-    }
 
-    match (expr_ty, expected_ty) {
-        (Set { .. }, EmptySet) => true,
-        (EmptySet, Set { .. }) => true,
-        (
-            Set {
-                element_ty: expr_elm_ty,
-            },
-            Set { element_ty: elty },
-        ) => match expr.as_set_elements() {
-            Some(mut els) => els.all(|e| {
-                does_restricted_expr_implement_schematype(e, expr_elm_ty, elty, extensions)
-            }),
+    match expected_ty {
+        Bool => expr.as_bool().is_some(),
+        Long => expr.as_long().is_some(),
+        String => expr.as_string().is_some(),
+        EmptySet => expr.as_set_elements().is_some(),
+        Set { .. }
+            if expr.as_set_elements().is_some() && expr.as_set_elements().unwrap().count() == 0 =>
+        {
+            true
+        }
+
+        Set { element_ty: elty } => match expr.as_set_elements() {
+            Some(mut els) => {
+                els.all(|e| does_restricted_expr_implement_schematype(e, elty, extensions))
+            }
             None => false,
         },
-        (
-            Record {
-                attrs: expr_attrs, ..
-            },
-            Record { attrs, open_attrs },
-        ) => match expr.as_record_pairs() {
+        Record { attrs, open_attrs } => match expr.as_record_pairs() {
             Some(pairs) => {
                 let pairs_map: BTreeMap<&SmolStr, BorrowedRestrictedExpr<'_>> = pairs.collect();
                 let all_req_schema_attrs_in_record = attrs.iter().all(|(k, v)| {
@@ -218,7 +212,6 @@ pub fn does_restricted_expr_implement_schematype(
                         || match pairs_map.get(k) {
                             Some(inner_e) => does_restricted_expr_implement_schematype(
                                 *inner_e,
-                                &expr_attrs.get(k).unwrap().attr_type,
                                 &v.attr_type,
                                 extensions,
                             ),
@@ -229,7 +222,6 @@ pub fn does_restricted_expr_implement_schematype(
                     pairs_map.iter().all(|(k, inner_e)| match attrs.get(*k) {
                         Some(sch_ty) => does_restricted_expr_implement_schematype(
                             *inner_e,
-                            &expr_attrs.get(*k).unwrap().attr_type,
                             &sch_ty.attr_type,
                             extensions,
                         ),
@@ -239,7 +231,17 @@ pub fn does_restricted_expr_implement_schematype(
             }
             None => false,
         },
-        _ => false,
+        Extension { name } => match expr.as_extn_fn_call() {
+            Some((actual_name, _)) => match format!("{:?}", name.0.id).as_str() {
+                "Id(\"ipaddr\")" => format!("{:?}", actual_name.0.id).as_str() == "Id(\"ip\")",
+                _ => name == actual_name,
+            },
+            None => false,
+        },
+        Entity { ty } => match expr.as_euid() {
+            Some(actual_euid) => actual_euid.entity_type() == ty,
+            None => false,
+        },
     }
 }
 
@@ -257,8 +259,7 @@ pub fn typecheck_restricted_expr_against_schematype(
     // directly?
     match schematype_of_restricted_expr(expr, extensions) {
         Ok(actual_ty) => {
-            if does_restricted_expr_implement_schematype(expr, &actual_ty, expected_ty, extensions)
-            {
+            if does_restricted_expr_implement_schematype(expr, expected_ty, extensions) {
                 // typecheck passes
                 Ok(())
             } else {
--- a/cedar-policy-core/src/entities/conformance.rs
+++ b/cedar-policy-core/src/entities/conformance.rs
@@ -213,16 +213,18 @@ pub fn does_restricted_expr_implement_schematype(
         ) => match expr.as_record_pairs() {
             Some(pairs) => {
                 let pairs_map: BTreeMap<&SmolStr, BorrowedRestrictedExpr<'_>> = pairs.collect();
-                let all_req_schema_attrs_in_record =
-                    attrs.iter().all(|(k, v)| match pairs_map.get(k) {
-                        Some(inner_e) => does_restricted_expr_implement_schematype(
-                            *inner_e,
-                            &expr_attrs.get(k).unwrap().attr_type,
-                            &v.attr_type,
-                            extensions,
-                        ),
-                        None => !v.required,
-                    });
+                let all_req_schema_attrs_in_record = attrs.iter().all(|(k, v)| {
+                    !v.required
+                        || match pairs_map.get(k) {
+                            Some(inner_e) => does_restricted_expr_implement_schematype(
+                                *inner_e,
+                                &expr_attrs.get(k).unwrap().attr_type,
+                                &v.attr_type,
+                                extensions,
+                            ),
+                            None => false,
+                        }
+                });
                 let all_rec_attrs_in_schema =
                     pairs_map.iter().all(|(k, inner_e)| match attrs.get(*k) {
                         Some(sch_ty) => does_restricted_expr_implement_schematype(
--- a/cedar-policy-core/src/entities/conformance.rs
+++ b/cedar-policy-core/src/entities/conformance.rs
@@ -226,8 +226,8 @@ pub fn does_restricted_expr_implement_schematype(
             None => false,
         },
         Extension { name } => match expr.as_extn_fn_call() {
-            Some((actual_name, _)) => match format!("{:?}", name.0.id).as_str() {
-                "Id(\"ipaddr\")" => format!("{:?}", actual_name.0.id).as_str() == "Id(\"ip\")",
+            Some((actual_name, _)) => match name.0.id.as_ref() {
+                "ipaddr" => actual_name.0.id.as_ref() == "ip",
                 _ => name == actual_name,
             },
             None => false,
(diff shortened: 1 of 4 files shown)


shaobo-he-aws (approved):

cedar-policy-core/src/entities/json/schema_types.rs
```diff
@@ -353,12 +353,8 @@ impl Diagnostic for NontrivialResidualError {
 /// For records, we can't know whether the attributes in the given record are
 /// required or optional.
 /// This function, when given a record that has keys A, B, and C, will return a
-/// `SchemaType` where A, B, and C are all marked as optional attributes, but no
+/// `SchemaType` where A, B, and C are all marked as required attributes, but no
 /// other attributes are possible.
-/// That is, this assumes that all existing attributes are optional, but that no
-/// other optional attributes are possible.
-/// Compared to marking A, B, and C as required, this allows the returned
-/// `SchemaType` to `is_consistent_with()` more types.
```
cdisselkoen, line 361: Presumably, now with this change, `SchemaType` will be `is_consistent_with()` less types (unless this comment was wrong).  Are we reasonably sure this won't cause other problems?

cedar-policy/src/tests.rs
```diff
@@ -3375,6 +3375,45 @@ mod schema_based_parsing_tests {
...
+                .unwrap(),
+            )]),
+            HashSet::new(),
+        )
+        .unwrap();
+        assert_matches!(
+            Entities::from_entities([entity], Some(&schema)),
+            Err(EntitiesError::InvalidEntity(_))
+        );
+    }
```
cdisselkoen, line 3416: Nit: this test should probably belong in the `entity_validate_tests` and not in the `schema_based_parsing_tests` module, since schema-based parsing of JSON entity data is specifically not affected.

cedar-policy/src/tests.rs
```diff
@@ -3375,6 +3375,45 @@ mod schema_based_parsing_tests {
             Err(EntitiesError::TransitiveClosureError(_))
         ));
     }
+
+    /// Record inside entity doesn't conform to schema
```
cdisselkoen, line 3379: Perhaps in addition to this test (with an extra attribute on the inner record), we should also have a test where an attribute on the inner record is missing? maybe also both of those for a schema where `foo` is optional?

cedar-policy-core/src/entities/json/schema_types.rs
```diff
@@ -382,10 +378,10 @@ pub fn schematype_of_restricted_expr(
...
                         extensions,
                     )?;
-                    // we can't know if the attribute is required or optional,
-                    // but marking it optional is more flexible -- allows the
-                    // attribute type to `is_consistent_with()` more types
-                    Ok((k.clone(), AttributeType::optional(attr_type)))
+                    // We can't know if the attribute is required or optional.
+                    // Assume it's required so that calling `is_consistent_with` on this and a type from the Schema
+                    // will reject attributes not present in the schema.
+                    Ok((k.clone(), AttributeType::required(attr_type)))
```
john-h-kastner-aws, line 384: I think this is the correct choice. `required <: optional`, and we know the attribute is present, so there's no reason to leave it as optional.

john-h-kastner-aws (approved):

cedar-policy-core/src/entities/json/schema_types.rs
john-h-kastner-aws: If `is_consistent_with` is just a typing judgement, then this should be fine because `required <: optional`.

cdisselkoen: `is_consistent_with` doesn't use the validator's type system. Let's double check that it does the right thing.

cedar-policy-core/src/entities/json/schema_types.rs
andrewmwells-amazon: Actually this doesn't work because it marks more sets as heterogeneous. I'll add a comment in the issues.

cedar-policy-core/src/entities/json/schema_types.rs
andrewmwells-amazon: No longer modifying `is_consistent_with`.

cedar-policy-core/src/entities/conformance.rs
```diff
@@ -176,6 +179,68 @@ pub fn typecheck_value_against_schematype(
     }
 }
 
+/// Check whether the given `RestrictedExpr` is a valid instance of `SchemaType`
+pub fn does_restricted_expr_implement_schematype(
+    expr: BorrowedRestrictedExpr<'_>,
+    expr_ty: &SchemaType,
+    expected_ty: &SchemaType,
+    extensions: &Extensions<'_>,
+) -> bool {
```
cdisselkoen, line 188: having `expr_ty` as an input to this function is a little confusing, given that the comments say that we're checking `expr` itself against `expected_ty`

cedar-policy/src/tests.rs
```diff
@@ -1709,6 +1710,158 @@ mod entity_validate_tests {
             }
         }
     }
+
+    /// Record inside entity doesn't conform to schema
```
cdisselkoen, line 1714: Bonus points for tests where a (required/optional) inner record attribute is _missing_ in the entity data.  (as opposed to this test where there is an _extra_ attribute in the entity data.)  But these tests adequately address the #1176 report.

cedar-policy-core/src/entities/json/schema_types.rs
john-h-kastner-aws: I don't see why that would be the case. Do you mean `[{a: 1}, {b: 2}]` is treated as heterogeneous when the types are `{a: Long}` and `{b: Long}`, but not when it's `{a?: Long}` and `{b?: Long}`? That doesn't make sense to me.

cedar-policy-core/src/entities/conformance.rs
andrewmwells-amazon: We could compute it within the function but that's wasteful (and requires duplicate error-handling)

andrewmwells-amazon: Removed in latest revision.

cedar-policy-core/src/entities/conformance.rs
```diff
@@ -176,6 +179,66 @@ pub fn typecheck_value_against_schematype(
...
+
+    match expected_ty {
+        Bool => expr.as_bool().is_some(),
+        Long => expr.as_long().is_some(),
+        String => expr.as_string().is_some(),
+        EmptySet => expr.as_set_elements().is_some(),
+        // PANIC SAFETY: LHS checks that `is_some()`.
+        #[allow(clippy::unwrap_used)]
+        Set { .. }
+            if expr.as_set_elements().is_some() && expr.as_set_elements().unwrap().count() == 0 =>
```
cdisselkoen, line 197: might be able to avoid the unwrap with something like
```
Set { .. } if matches!(expr.as_set_elements(), Some(iter) if iter.count() == 0) =>
```

cedar-policy-core/src/entities/conformance.rs
```diff
@@ -176,6 +179,66 @@ pub fn typecheck_value_against_schematype(
...
+                        }
+                });
+                let all_rec_attrs_in_schema =
+                    pairs_map.iter().all(|(k, inner_e)| match attrs.get(*k) {
+                        Some(sch_ty) => {
+                            does_restricted_expr_implement_schematype(*inner_e, &sch_ty.attr_type)
+                        }
+                        None => false,
+                    });
+                (*open_attrs || all_rec_attrs_in_schema) && all_req_schema_attrs_in_record
```
cdisselkoen, line 224: Does this do the right thing in the case where we have `open_attrs == true`, there is an optional attribute, and `pairs_map` has a value for the optional attribute but it's not the type declared in the schema?

In my reading, `all_rec_attrs_in_schema` will be `false`, but we'll still pass overall due to the logic in line 224

cedar-policy-core/src/entities/conformance.rs
```diff
@@ -190,7 +253,7 @@ pub fn typecheck_restricted_expr_against_schematype(
     // directly?
```
cdisselkoen, line 253: comment can be updated/removed.  (And this issue can also be `resolves #440`)

cedar-policy/src/tests.rs
cdisselkoen: Latest rev tests look good

cedar-policy-core/src/entities/conformance.rs
andrewmwells-amazon: That gives the error:
```
cannot move out of `iter` in pattern guard
variables bound in patterns cannot be moved from until after the end of the pattern guard
```
and I didn't find a work around.

cedar-policy-core/src/entities/conformance.rs
andrewmwells-amazon: Good catch!

cedar-policy-core/src/entities/conformance.rs
andrewmwells-amazon: I'm hoping John comes up with a way to avoid calling `schematype_of_restricted_expr` without impacting error reporting (maybe only calling if if the types don't match). I wasn't sure if we want to call 440 resolved while we're still using `schematype_of_restricted_expr` on the main path.

andrewmwells-amazon: Done

andrewmwells-amazon:
Also fixes https://github.com/cedar-policy/cedar/issues/440

cedar-policy-core/src/entities/conformance.rs
```diff
@@ -176,6 +179,66 @@ pub fn typecheck_value_against_schematype(
...
+    expr: BorrowedRestrictedExpr<'_>,
+    expected_ty: &SchemaType,
+) -> bool {
+    use SchemaType::*;
+
+    match expected_ty {
+        Bool => expr.as_bool().is_some(),
+        Long => expr.as_long().is_some(),
+        String => expr.as_string().is_some(),
+        EmptySet => expr.as_set_elements().is_some(),
```
john-h-kastner-aws, line 193: Does this say that any set (even a non-empty set) implements the `EmptySet` type? I would expect that only empty sets implement `EmptySet.

cedar-policy-core/src/entities/conformance.rs
```diff
@@ -176,6 +179,66 @@ pub fn typecheck_value_against_schematype(
...
+                            does_restricted_expr_implement_schematype(*inner_e, &sch_ty.attr_type)
+                        }
+                        None => *open_attrs,
+                    });
+                all_rec_attrs_match_schema && all_req_schema_attrs_in_record
+            }
+            None => false,
+        },
+        Extension { name } => match expr.as_extn_fn_call() {
+            Some((actual_name, _)) => match format!("{:?}", name.0.id).as_str() {
```
john-h-kastner-aws, line 229: Could this match on `name.0.id.as_ref()` instead?

cedar-policy-core/src/entities/json/schema_types.rs
```diff
@@ -382,9 +379,8 @@ pub fn schematype_of_restricted_expr(
                         BorrowedRestrictedExpr::new_unchecked(v), // assuming the invariant holds for the record as a whole, it will also hold for each attribute value
                         extensions,
                     )?;
-                    // we can't know if the attribute is required or optional,
-                    // but marking it optional is more flexible -- allows the
-                    // attribute type to `is_consistent_with()` more types
+                    // We can't know if the attribute is required or optional.
+                    // Keep as optional to minimize heterogeneous sets.
```
john-h-kastner-aws, line 383: How relevant is this comment with the latest changes?

cedar-policy-core/src/entities/conformance.rs
andrewmwells-amazon: Added two new tests

cedar-policy-core/src/entities/json/schema_types.rs
andrewmwells-amazon: We investigated this and decided we'd prefer the behavior of `is_consistent_with` with this set to `required`. So I replaced what I view as a misleading comment with the reason we didn't proceed down that route.

cedar-policy-core/src/entities/conformance.rs
andrewmwells-amazon: That matches the behavior in `is_consistent_with`
https://github.com/cedar-policy/cedar/blob/c1b078296da73b812dff28e88ccc1f7266d74c21/cedar-policy-core/src/entities/json/schema_types.rs#L123

cedar-policy/src/tests.rs
andrewmwells-amazon: Done

cedar-policy-core/src/entities/conformance.rs
john-h-kastner-aws: It's going to be right for `is_consistent_with` since that's documented as checking "here exist some concrete values that have both types", but I don't think it's going to be correct here. 

Also, it doesn't look like this match case is reachable.

cedar-policy-core/src/entities/conformance.rs
```diff
@@ -176,6 +179,68 @@ pub fn typecheck_value_against_schematype(
...
+    expected_ty: &SchemaType,
+) -> bool {
+    use SchemaType::*;
+
+    match expected_ty {
+        Bool => expr.as_bool().is_some(),
+        Long => expr.as_long().is_some(),
+        String => expr.as_string().is_some(),
+        EmptySet => {
+            expr.as_set_elements().is_some() && expr.as_set_elements().unwrap().count() == 0
```
john-h-kastner-aws, line 194: Here and below we could use `is_some_and` to avoid unwrapping

john-h-kastner-aws (approved):
clippy is unhappy, but looks good otherwise

PR #1185, opened 2024-09-09
Apply clippy auto-fixes

shaobo-he-aws:
## Description of changes

Applied clippy 1.81

### DIFF
--- a/cedar-policy-validator/src/schema.rs
+++ b/cedar-policy-validator/src/schema.rs
@@ -474,7 +474,7 @@ impl ValidatorSchema {
             for parent in entity_type.parents.iter() {
                 entity_children
                     .entry(internal_name_to_entity_type(parent.clone())?)
-                    .or_insert_with(HashSet::new)
+                    .or_default()
                     .insert(name.clone());
             }
         }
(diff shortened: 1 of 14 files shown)


cedar-policy-validator/src/schema.rs
```diff
@@ -474,7 +474,7 @@ impl ValidatorSchema {
             for parent in entity_type.parents.iter() {
                 entity_children
                     .entry(internal_name_to_entity_type(parent.clone())?)
-                    .or_insert_with(HashSet::new)
+                    .or_default()
```
john-h-kastner-aws, line 477: Can you check if this is actually the same? I can imagine that `or_default` could reasonably not insert default value

john-h-kastner-aws (approved):

cedar-policy-validator/src/schema.rs
shaobo-he-aws: It seems `or_default` does the job: https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_default

khieta (approved):
Lgtm.

Issue #1184, opened 2024-09-09
Some usability improvements


aaronjeline:
### Category

User level API features/changes

## New APIS
* `Entity::new` returns a result, it would be nice to have a version that never fails
* `Entity` does not implement `Hash`, yet we require entity collections
  to not have duplicates! This means you can't use `HashSet` :(
* `Entities::from_entities` does not indicate in documentation that it
   will error on duplicates

### Describe alternatives you've considered

Don't do this

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #1186, opened 2024-09-09
Added some useful methods and docs

aaronjeline:
## Description of changes
- Adds `Entity::new_without_attrs` that can't error
- `Entity` is now `Hash`, which is very useful as we require that entity collections lack duplicates
- Added documentation to methods on `Entities` clarifying when errors are returned
 
## Issue #, if available
#1184 

### DIFF
--- a/cedar-policy-core/src/ast/entity.rs
+++ b/cedar-policy-core/src/ast/entity.rs
@@ -317,8 +317,17 @@ pub struct Entity {
     ancestors: HashSet<EntityUID>,
 }
 
+impl std::hash::Hash for Entity {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.uid.hash(state);
+    }
+}
+
 impl Entity {
     /// Create a new `Entity` with this UID, attributes, and ancestors
+    ///
+    /// # Errors
+    /// - Will error if any of the Restricted Expressions in `attrs` error when evaluated
     pub fn new(
         uid: EntityUID,
         attrs: HashMap<SmolStr, RestrictedExpr>,
@@ -344,6 +353,19 @@ impl Entity {
             attrs: evaluated_attrs,
             ancestors,
         })
+    }
+
+    /// Creates a new `Entity` with this UID, ancestors, and an empty set of attributes
+    /// Since it lacks any attributes, this method returns `Self` instead of `Result<Self>`
+    pub fn new_empty_attrs(
+        uid: EntityUID,
+        ancestors: HashSet<EntityUID>,
+        extensions: &Extensions<'_>,
+    ) -> Self {
+        // PANIC SAFETY
+        // Safe as the hashmap is empty
+        #[allow(clippy::unwrap_used)]
+        Self::new(uid, HashMap::new(), ancestors, extensions).unwrap()
     }
 
     /// Create a new `Entity` with this UID, attributes, and ancestors.
--- a/cedar-policy-core/src/ast/entity.rs
+++ b/cedar-policy-core/src/ast/entity.rs
@@ -327,7 +327,7 @@ impl Entity {
     /// Create a new `Entity` with this UID, attributes, and ancestors
     ///
     /// # Errors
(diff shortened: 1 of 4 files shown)


cedar-policy-core/src/ast/entity.rs
```diff
@@ -346,6 +355,19 @@ impl Entity {
         })
     }
 
+    /// Creates a new `Entity` with this UID, ancestors, and an empty set of attributes
```
john-h-kastner-aws, line 358: Consistency with `new` doc comment.
```suggestion
    /// Create a new `Entity` with this UID, ancestors, and an empty set of attributes
```

cedar-policy/src/api.rs
```diff
@@ -345,6 +356,10 @@ impl Entities {
     /// error if attributes have the wrong types (e.g., string instead of
     /// integer), or if required attributes are missing or superfluous
     /// attributes are provided.
+    /// # Errors
```
john-h-kastner-aws, line 359: We could consider enabling the [missing_error_doc lint](https://rust-lang.github.io/rust-clippy/stable/index.html#/missing_errors_doc) if we believe this sort of documentation is valuable.

john-h-kastner-aws (approved):

cedar-policy-core/src/ast/entity.rs
```diff
@@ -317,8 +317,17 @@ pub struct Entity {
...
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.uid.hash(state);
+    }
+}
+
 impl Entity {
     /// Create a new `Entity` with this UID, attributes, and ancestors
+    ///
+    /// # Errors
+    /// - Will error if any of the Restricted Expressions in `attrs` error when evaluated
```
cdisselkoen, line 330:
```suggestion
    /// - Will error if any of the [`RestrictedExpr]`s in `attrs` error when evaluated
```

cedar-policy-core/src/ast/entity.rs
```diff
@@ -346,6 +355,20 @@ impl Entity {
         })
     }
 
+    /// Create a new `Entity` with this UID, ancestors, and an empty set of attributes
+    /// Since it lacks any attributes, this method returns `Self` instead of `Result<Self>`
```
cdisselkoen, line 359:
```suggestion
    /// Create a new `Entity` with this UID, ancestors, and an empty set of attributes
    ///
    /// Since there are no attributes, this method does not error, and returns `Self` instead of `Result<Self>`
```

cedar-policy-core/src/ast/entity.rs
```diff
@@ -346,6 +355,20 @@ impl Entity {
...
+    pub fn new_empty_attrs(
+        uid: EntityUID,
+        ancestors: HashSet<EntityUID>,
+        extensions: &Extensions<'_>,
+    ) -> Self {
+        // Safe as the attributes hashmap is empty, and `Self::new`
+        // can only error when evaluating an attribute expression
+        // PANIC SAFETY (see above)
+        #[allow(clippy::unwrap_used)]
+        Self::new(uid, HashMap::new(), ancestors, extensions).unwrap()
```
cdisselkoen, line 369: Relying on this invariant for panic safety seems sketchy to me, I could imagine a future PR adding a way `Entities::new()` could error, but not updating this code or the referenced error docs.

I think we could avoid the unwrap by relying on Core's `Entity::new_with_attr_partial_value()` which does not error.  (You can pass an empty attr map to it.)

cedar-policy/src/api.rs
```diff
@@ -371,6 +386,10 @@ impl Entities {
     ///
     /// Re-computing the transitive closure can be expensive, so it is advised
     /// to not call this method in a loop.
+    /// # Errors
```
cdisselkoen, line 389: Here and elsewhere in this file, it might render in the docs html a little nicer (in my opinion/taste) with `## Errors` rather than `# Errors`

cdisselkoen (commented):
Looks good overall, one comment about panic safety

cdisselkoen (approved):

PR #1190, opened 2024-09-10
Avoid calling `schematype_of_restricted_expr` for parsing extension type entity attribute

john-h-kastner-aws:
This change assumes that there exists one constructor for each entity type, so it's enough to know the type we want to construct without also needing to infer the actual type of the arguments.

Multiple single argument constructors previously could exist for one extension type as long as the argument types were distinct. This would now be an error when building the `Extensions` struct. Our current extensions abide by this limitation.

First PR towards removing all uses of `schematype_of_restricted_expr`

### DIFF
--- a/cedar-policy-core/src/entities/json/err.rs
+++ b/cedar-policy-core/src/entities/json/err.rs
@@ -16,14 +16,13 @@
 
 use std::fmt::Display;
 
-use super::{HeterogeneousSetError, SchemaType};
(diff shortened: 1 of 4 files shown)


cdisselkoen (approved):
Seems reasonable to me, although I do envision that at some point in the future we will want to support multiple constructors for a single extension type.  (The datetime RFC doesn't propose this, does it?)

PR #1194, opened 2024-09-11
Delete schema-based parsing for action attribtues

john-h-kastner-aws:
## Description of changes

Schema based parsing currently supports parsing action attributes according to their types, but the Cedar schema format does not support defining actions with their attributes or declaring the type of their attributes, so this code is not reachable by consumers of the Cedar library.

By deleting this code we can remove a caller of `schematype_of_partialvalue` which is known to be buggy for set and records.

Note that action entities are compared against the action entities from the schema when constructing an `Entities` object, so we still report an error if an action entity contains unexpected attributes.

### DIFF
--- a/cedar-policy-core/src/entities/json/entities.rs
+++ b/cedar-policy-core/src/entities/json/entities.rs
@@ -23,7 +23,7 @@ use crate::ast::{BorrowedRestrictedExpr, Entity, EntityUID, PartialValue, Restri
 use crate::entities::conformance::EntitySchemaConformanceChecker;
 use crate::entities::{
     conformance::err::{EntitySchemaConformanceError, UnexpectedEntityTypeError},
-    schematype_of_partialvalue, Entities, EntitiesError, GetSchemaTypeError, TCComputation,
+    Entities, EntitiesError, TCComputation,
 };
 use crate::extensions::Extensions;
 use crate::jsonvalue::JsonValueWithNoDuplicateKeys;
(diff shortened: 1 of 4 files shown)


khieta (approved):

cedar-policy-core/src/entities/json/entities.rs
```diff
@@ -343,59 +339,6 @@ impl<'e, 's, S: Schema> EntityJsonParser<'e, 's, S> {
                     };
                     Ok((k.clone(), rexpr))
                 }
-                EntitySchemaInfo::Action(action) => {
-                    // We'll do schema-based parsing assuming optimistically that
-                    // the type in the JSON is the same as the type in the schema.
-                    // (As of this writing, the schema doesn't actually tell us
-                    // what type each action attribute is supposed to be)
```
cdisselkoen, line 350: I _think_ this is a behavior change -- suppose you have an action attribute of type `ip` in the schema, then with this code you're allowed to pass just `"192.168.0.1"` for that attribute in the entity data (ie, just a string), whereas after this code is removed, you'll need the `{ __extn: { ... } }` form.

however, this is probably inconsequential, because most users are probably not passing actions in their entity data if they are also passing the schema, since the schema contains the action info and we already automatically pull the proper action info from the schema when constructing entities.

cdisselkoen (approved):
Great to see some of this code being removed.

PR #1197, opened 2024-09-11
Delete `schematype_of_restricted_expr` function

john-h-kastner-aws:
Finish removing the buggy `schematype_of_restricted_expr` function.

This changes regresses on some error messages where we would previously report the "actual" and "expected" type for an expression we now only print the "expected" type and the expression itself. Getting a nice replacement for the "actual" part of the error message  wasn't trivial, so I'm put this PR up now for review now. We can discuss what to do with those errors.


cedar-policy-core/src/entities/conformance.rs
```diff
@@ -179,42 +170,78 @@ pub fn typecheck_value_against_schematype(
...
+    /// Returns `Ok(true)` only when all elements are `Ok(true)`. Return an error
+    /// result if any elements are `Err(_)`. Otherwise returns `Ok(false)`.
+    fn try_all<E>(i: impl Iterator<Item = Result<bool, E>>) -> Result<bool, E> {
+        Ok(i.collect::<Result<Vec<_>, _>>()?.iter().all(|b| *b))
+    }
+
+    // Check for `unknowns`.  In this case we just don't have the information to
+    // know whether the attribute value (an unknown) matches the expected type.
+    // For now we consider this as passing -- we can't really report a type
+    // error <https://github.com/cedar-policy/cedar/issues/418>.
```
cdisselkoen, line 192: This comment is confusing in light of the fact that the code below actually does check `u.type_annotation` if it is present.  Maybe the comment belongs in the `None` branch, ie when no type annotation is present?

cdisselkoen (commented):
I definitely approve of the direction of this PR.  Wondering if we can make the error message regression a little less painful

cedar-policy-core/src/ast/restricted_expr.rs
```diff
@@ -447,6 +448,16 @@ impl<'a> BorrowedRestrictedExpr<'a> {
             _ => None,
         }
     }
+
+    /// Try to compute the runtime type of this expression. See
+    /// [Expr::try_type_of] for exactly what this computes.
```
cdisselkoen, line 453:
```suggestion
    /// [`Expr::try_type_of`] for exactly what this computes.
```

cedar-policy-core/src/ast/restricted_expr.rs
```diff
@@ -447,6 +448,16 @@ impl<'a> BorrowedRestrictedExpr<'a> {
...
+    ///
+    /// On a restricted expression, there are fewer case where we might fail to
```
cdisselkoen, line 455:
```suggestion
    /// On a restricted expression, there are fewer cases where we might fail to
```

cedar-policy-core/src/entities/conformance.rs
```diff
@@ -172,109 +171,158 @@ pub fn typecheck_value_against_schematype(
...
-        Extension { name } => match expr.as_extn_fn_call() {
-            Some((actual_name, _)) => match name.0.id.as_ref() {
-                "ipaddr" => actual_name.0.id.as_ref() == "ip",
-                _ => name == actual_name,
-            },
-            None => false,
+            None => type_mismatch_err(),
         },
+        // Extension functions are handled by the first `match` in this function.
+        Extension { .. } => type_mismatch_err(),
```
cdisselkoen, line 321: nit: in the first `match`, I think we only handled the case where we expected an extension type.  What if we are expecting a builtin type, and the restrictedexpr is an extension function that returns that type?  (Pretty sure this case is not reachable from public API surface so it kinda doesn't matter, but)

cedar-policy-core/src/entities/json/err.rs
```diff
@@ -482,19 +482,87 @@ pub enum JsonDeserializationErrorContext {
...
+    /// attribute we didn't expect.
+    UnexpectedAttr(SmolStr),
+    /// We saw a record type expression as expected, but it did not contain an
+    /// attribute we expected.
+    MissingRequiredAtr(SmolStr),
+    /// No further detail available.
+    None,
+}
+
+impl std::fmt::Display for TypeMismatchReason {
```
cdisselkoen, line 512: This display impl could be a `derive(Error)` and then marking each case with `#[error("actually has type {ty}")]` etc

cedar-policy-core/src/entities/json/schema_types.rs
```diff
@@ -83,6 +83,20 @@ impl SchemaType {
...
+        match self {
+            SchemaType::Bool => Type::Bool,
+            SchemaType::Long => Type::Long,
+            SchemaType::String => Type::String,
+            SchemaType::Set { .. } => Type::Set,
+            SchemaType::EmptySet => Type::Set,
+            SchemaType::Record { .. } => Type::Record,
+            SchemaType::Entity { ty } => Type::Entity { ty },
+            SchemaType::Extension { name } => Type::Extension { name },
+        }
```
cdisselkoen, line 97: Might be sensible to make this an `impl From`

cdisselkoen (approved):

shaobo-he-aws (approved):

cedar-policy-core/src/entities/conformance.rs
john-h-kastner-aws: I think I actually fixed the behavior on extension functions with the changes in this PR. The first `match` compares the return type to the expected type instead of looking at the function name.

john-h-kastner-aws: also added some test cases hitting this

PR #1211, opened 2024-09-18
hide RFC 82 implementation behind entity-tags crate feature

cdisselkoen:
## Description of changes

Hides the existing RFC 82 implementation code (#1204) behind an experimental feature flag, `entity-tags`.

This PR chooses to leave the parser grammar and JSON structures the same, and only report an error at the point we construct a `ValidatorSchema`. This enables us to have a much nicer error message if someone tries to use tags without enabling the feature.  It does mean that a lot of the "experimental" parser code is not deactivated when the feature is deactivated, but I think this is acceptable.


john-h-kastner-aws (approved):

aaronjeline (approved):
I agree, leaving the parser code is reasonable given the better error message

Issue #1063, opened 2024-07-12
[3.x.x only] Converting a JSON-syntax schema to human syntax and back can change the target of an entity type reference


cdisselkoen:
### Before opening, please confirm:

- [X] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Schemas and Validation

### Describe the bug

For the following schema:

```
{
  "": {
    "entityTypes": {
      "User": {}
    },
    "actions": {}
  },
  "NS": {
    "commonTypes": {
      "User": { "type": "String" }
    },
    "entityTypes": {
      "Foo": {
        "shape": {
          "type": "Record",
          "attributes": {
            "owner": { "type": "Entity", "name": "User" }
          }
        }
      }
    },
    "actions": {}
  }
}
```

When converted to natural/human syntax, we get
```
namespace NS {
  type User = __cedar::String;
  entity Foo = {"owner": User};
}
entity User;
```
(I added whitespace for clarity.)

When converting that back to JSON syntax, we get
```
{
  "NS": {
    "commonTypes": {
      "User": { "type":"String" }
    },
    "entityTypes": {
      "Foo": {
        "shape": {
          "type": "Record",
          "attributes": {
            "owner": { "type": "User" }
          }
        }
      }
    },
    "actions": {}
  },
  "": {
    "entityTypes": {
      "User": {}
    },
    "actions": {}
  }
}
```
(again, I added whitespace for clarity.)

This is not the same as the original schema: In the original schema, `Foo.owner` had the type `User`, an entity type in the empty namespace.  In the final schema, `Foo.owner` has the type `NS::User`, a common type which resolves to `String`.

[...]

PR #1214, opened 2024-09-19
Update changelong on main for fix in #1212

john-h-kastner-aws:
Update for #1063 fix on 3.4.x in #1212


cdisselkoen (approved):
Looks good.  Might be worth a similar note in the CLI changelog about the `translate-schema` command.

khieta (approved):

PR #1215, opened 2024-09-19
refactor validator tests

cdisselkoen:
## Description of changes

The main purpose of this refactor is so that we can assert things about the returned errors other than equality to a specific `ValidationError`.  In particular, paving the way to use `expect_err()` to check more miette details of error messages.  I believe this refactor might also improve test-failure messages in some cases.

Also removed a few helper functions that were simply calling another function with one argument hardcoded.  In my opinion, it is clearer to inline these and just specify the argument at each callsite.


khieta (approved):
Looks fine, pending resolution of CI failures.

cedar-policy-validator/src/typecheck/test/test_utils.rs
```diff
@@ -384,15 +381,32 @@ pub(crate) fn assert_typechecks_empty_schema_permissive(expr: Expr, expected: Ty
...
+) -> HashSet<ValidationError> {
+    assert_typecheck_fails(empty_schema_file(), expr, None)
+}
+
+/// Assert that the given `HashSet` has exactly one error. Return it.
+/// If there are more than one, panic and display all the errors in pretty format.
+#[track_caller]
+pub(crate) fn assert_exactly_one_error<T: miette::Diagnostic + Send + Sync + 'static>(
+    set: HashSet<T>,
+) -> T {
```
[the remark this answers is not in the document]
cdisselkoen: I originally named it `assert_exactly_one_element`, but since it really only works for error types (and not general T), I changed it.  Suggestions?

cedar-policy-validator/src/typecheck/test/expr.rs
```diff
@@ -762,29 +798,26 @@ fn in_set_typechecks_strict() {
 #[test]
 fn in_typecheck_fails() {
     let src = "0 in true";
-    assert_typecheck_fails_empty_schema(
-        src.parse().unwrap(),
-        Type::primitive_boolean(),
+    let errors =
+        assert_typecheck_fails_empty_schema(src.parse().unwrap(), Type::primitive_boolean());
+    assert_eq!(errors.len(), 2);
+    assert!(errors.contains(&ValidationError::expected_type(
```
john-h-kastner-aws, line 804: I think assertion failures will be substantially worse here. The old assertion would have done an `assert_eq!` on the error sets, printing a pretty diff because of the `similar-asserts` dependency while this assert just fails.

cedar-policy-validator/src/typecheck/test/test_utils.rs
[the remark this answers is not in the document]
cdisselkoen: I guess `assert_exactly_one_diagnostic`?

cedar-policy-validator/src/typecheck/test/namespace.rs
```diff
@@ -140,51 +137,62 @@ fn namespaced_entity_get_attr() {
...
 fn namespaced_entity_wrong_namespace() {
-    assert_expr_typecheck_fails_namespace_schema(
+    let errors = assert_typecheck_fails(
+        namespaced_entity_type_schema(),
         Expr::from_str(r#"N::S::T::Foo::"alice""#).expect("Expr should parse."),
         None,
-        [],
     );
-    assert_expr_typecheck_fails_namespace_schema(
+    assert_eq!(errors.len(), 0);
```
john-h-kastner-aws, line 165: I'd rather `assert_eq!(errors, [])` so we get the actual contents of `errors` in any assertion failure.

cedar-policy-validator/src/typecheck/test/test_utils.rs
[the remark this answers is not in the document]
khieta: Yeah, that makes sense.

cedar-policy-validator/src/typecheck/test/expr.rs
cdisselkoen: Good point. (also I forgot about `similar-asserts`, thanks for the reminder)

cedar-policy-validator/src/typecheck/test/namespace.rs
cdisselkoen: Good call

cdisselkoen: Fixed in latest revision

cedar-policy-validator/src/typecheck/test/expr.rs
cdisselkoen: Fixed in latest revision

cedar-policy-validator/src/typecheck/test/test_utils.rs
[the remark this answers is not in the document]
cdisselkoen: Fixed in latest revision

john-h-kastner-aws (approved):

cdisselkoen:
Java CI failures are unrelated; merging this.

PR #1231, opened 2024-09-25
Add support for annotations without explicit value

john-h-kastner-aws:
## Description of changes

Adds support for annotations like `@foo` as sugar for `@foo("")`.

Will require a small DRT update to account for the change in data structures. Will also require a small change to the docs to note that this syntax is also allowed. 

### DIFF
--- a/cedar-policy-core/src/ast/policy.rs
+++ b/cedar-policy-core/src/ast/policy.rs
@@ -1170,43 +1170,11 @@ pub struct Annotation {
     /// Annotation value. `None` for annotations without a value, i.e., `@foo`.
     /// An annotation without a value should be treated as equivalent to the
     /// value being `""`. This interpretation is implemented by the `AsRef<str>`
-    /// impl an `val` method below.
-    val: Option<SmolStr>,
+    /// impl below.
+    pub val: Option<SmolStr>,
     /// Source location. Note this is the location of _the entire key-value
     /// pair_ for the annotation, not just `val` above
-    loc: Option<Loc>,
-}
-
-impl Annotation {
-    /// Construct a new annotation.
-    pub fn new(val: Option<SmolStr>, loc: Option<Loc>) -> Self {
-        Self { val, loc }
-    }
-
-    /// Get the annotation value, treating an annotation without a value as
-    /// being equivalent to the value `""`.
-    pub fn val(&self) -> &str {
-        self.as_ref()
-    }
-
-    /// Get the annotation value, returning `None` for annotations without a
-    /// value. We generally want to treat this as equivalent to `""`, but we
-    /// occasionally want a lossless representation.
-    pub fn raw_val(&self) -> Option<&SmolStr> {
-        self.val.as_ref()
-    }
-
-    /// Get the annotation value, returning `None` for annotations without a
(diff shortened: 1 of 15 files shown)


shaobo-he-aws (approved):
LGTM.

Though I'd like to propose a new terminology. Can we call what's after `@` the annotation name and what's inside `()` the annotation argument.

john-h-kastner-aws:
> Though I'd like to propose a new terminology. Can we call what's after `@` the annotation name and what's inside `()` the annotation argument.

I don't particularly want to pull a change in terminology into this PR.

cedar-policy-core/src/ast/policy.rs
```diff
@@ -1163,16 +1167,51 @@ impl From<BTreeMap<AnyId, Annotation>> for Annotations {
 /// Struct which holds the value of a particular annotation
 #[derive(Serialize, Deserialize, Clone, Hash, Eq, PartialEq, Debug, PartialOrd, Ord)]
 pub struct Annotation {
-    /// Annotation value
-    pub val: SmolStr,
+    /// Annotation value. `None` for annotations without a value, i.e., `@foo`.
+    /// An annotation without a value should be treated as equivalent to the
+    /// value being `""`. This interpretation is implemented by the `AsRef<str>`
+    /// impl an `val` method below.
+    raw_val: Option<SmolStr>,
```
cdisselkoen, line 1174: Why `Option` and explicitly saying that `None` and `Some("")` are equivalent?  This seems like a footgun for later.  Wouldn't it be better to just have a `SmolStr` here so that there's only one representation?

cdisselkoen (commented):
Leaving a high-level comment, will leave a detailed review in a bit

cedar-policy-core/src/ast/policy.rs
john-h-kastner-aws: The reason is that conversion to EST, which is currently written to go through this AST struct, even when converting directly from CST, should be lossless, so this struct needed to know if it was originally present.

cdisselkoen: Personally I would propose that we pick one or the other representation to be canonical and output that EST.  So either we decide that empty annotation values are always omitted, or always explicit `""`.  I believe "lossless" already drops extraneous parens, is this any different?

john-h-kastner-aws: A related decision: In the EST I updated annotations so that `null` is an allowed value, equivalent to `""`. Thoughts? I think we would eventually need to allow that in a future world where we differentiate no-value and empty-string, but it's not necessary for now.

cdisselkoen: I could see an argument either way there.  Some sugar in the EST can be helpful, but the feature request was probably primarily for sugar in the Cedar policy format, and it's totally reasonable to require explicitness in the EST.

john-h-kastner-aws: Made some of the cst-to-ast code generic so that est-to-ast can still reuse the same id and string validation checks without passing through the ast structs to avoid this issue

cdisselkoen (approved):
Looks great.

Issue #1120, opened 2024-08-06
API to substitute all occurences of an entity literal with another entity literal


andrewmwells-amazon:
### Category

User level API features/changes

### Describe the feature you'd like to request

An API to substitute all occurrences of an entity literal with another entity literal.

### Describe alternatives you've considered

Status quo

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

laraelangdon:
Should we cancel this now given latest clarified requirements?

cdisselkoen:
Resolved by #1233

PR #1233, opened 2024-09-25
Entity Literal Substitution

andrewmwells-amazon:
## Description of changes
API for Entity literal substitution.

## Issue #, if available
#1120 

### DIFF
--- a/cedar-policy-core/src/entities/json/value.rs
+++ b/cedar-policy-core/src/entities/json/value.rs
@@ -389,6 +389,54 @@ impl CedarValueJson {
             },
         }
     }
+
+    /// Substitute entity literals
+    pub fn sub_entity_literals(
+        self,
+        mapping: &BTreeMap<EntityUID, EntityUID>,
+    ) -> Result<Self, JsonDeserializationError> {
+        match self.clone() {
+            // TODO: Should we error for this (alegedly impossible) case?
+            CedarValueJson::ExprEscape { __expr } => Ok(self.clone()),
+            CedarValueJson::EntityEscape { __entity } => {
+                let euid = EntityUID::try_from(__entity);
+                match euid {
+                    Ok(euid) => match mapping.get(&euid) {
+                        Some(new_euid) => Ok(CedarValueJson::EntityEscape {
+                            __entity: new_euid.into(),
+                        }),
+                        None => Ok(self.clone()),
+                    },
+                    // TODO: What can cause this to error and how should we handle?
+                    Err(_) => Ok(self.clone()),
+                }
+            }
+            CedarValueJson::ExtnEscape { __extn } => Ok(CedarValueJson::ExtnEscape {
+                __extn: FnAndArg {
+                    ext_fn: __extn.ext_fn,
+                    arg: Box::new((*__extn.arg).clone().sub_entity_literals(mapping)?),
+                },
+            }),
+            CedarValueJson::Bool(_) => Ok(self.clone()),
+            CedarValueJson::Long(_) => Ok(self.clone()),
+            CedarValueJson::String(_) => Ok(self.clone()),
+            CedarValueJson::Set(v) => {
+                let mut new_v = vec![];
+                for j in v {
+                    new_v.push(j.sub_entity_literals(mapping)?)
+                }
+                Ok(CedarValueJson::Set(new_v))
+            }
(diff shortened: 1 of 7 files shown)


shaobo-he-aws:
There seem to be conflicts with main.

cedar-policy-core/src/entities/json/value.rs
```diff
@@ -389,6 +389,52 @@ impl CedarValueJson {
...
+                        }),
+                        None => Ok(self.clone()),
+                    },
+                    Err(_) => Ok(self.clone()),
+                }
+            }
+            CedarValueJson::ExtnEscape { __extn } => Ok(CedarValueJson::ExtnEscape {
+                __extn: FnAndArg {
+                    ext_fn: __extn.ext_fn,
+                    arg: Box::new((*__extn.arg).clone().sub_entity_literals(mapping)?),
```
cdisselkoen, line 415: I'm guessing you don't need this `.clone()` on a freshly created `Box`?

cedar-policy-core/src/entities/json/value.rs
```diff
@@ -389,6 +389,52 @@ impl CedarValueJson {
...
+            }),
+            CedarValueJson::Bool(_) => Ok(self.clone()),
+            CedarValueJson::Long(_) => Ok(self.clone()),
+            CedarValueJson::String(_) => Ok(self.clone()),
+            CedarValueJson::Set(v) => {
+                let mut new_v = vec![];
+                for j in v {
+                    new_v.push(j.sub_entity_literals(mapping)?)
+                }
+                Ok(CedarValueJson::Set(new_v))
```
cdisselkoen, line 426: alternately: something like `Ok(CedarValueJson::Set(v.into_iter().map(|j| j.sub_entity_literals(mapping)).collect())`

cedar-policy-core/src/est/expr.rs
```diff
@@ -15,8 +15,8 @@
  */
 
 use super::FromJsonError;
-use crate::ast;
 use crate::ast::InputInteger;
+use crate::ast::{self, EntityUID};
```
cdisselkoen, line 19: can merge the `InputInteger` import to this line while we're here

cedar-policy/src/api.rs
```diff
@@ -2942,6 +2942,38 @@ impl Policy {
...
+            .expect("Internal error, failed to construct est.")
+            .clone();
+
+        let mapping = mapping.into_iter().map(|(k, v)| (k.0, v.0)).collect();
+
+        // PANIC SAFETY: This can't fail for a policy that was already constructed
+        #[allow(clippy::expect_used)]
+        let est = cloned_est
+            .sub_entity_literals(&mapping)
+            .expect("Internal error, failed to sub entity literals.");
```
cdisselkoen, line 2962: help me understand why this `expect` is safe -- the panic-safety comment notes that the est must be valid; but what if the mapping isn't valid or something?

cdisselkoen (approved):
Looks good overall; tiny nits and one question about an `.expect()`

cedar-policy/src/api.rs
andrewmwells-amazon: I think this particular part will be panic-free because we're substituting `EntityUID`s for `EntityUID`s. I'm less confident about the construction of the AST. 

My original thought was that this method can't ensure a policy that previously passed validation still passes after substitution, so the caller will be responsible for checking that policies are valid. I'm not sure that's the best idea. There are some errors that we could catch e.g., mapping `Action::"view"` -> `User::"Alice"`. Maybe we should try to catch as many as possible and then the caller still needs to re-validate against their schema?

Another option would be to require that the mapping not change entity types.

cdisselkoen: If we wanted the mapping to not change entity types, it could be an Eid->Eid mapping instead of EntityUID->EntityUID.  If this fits the need, this might be a better solution.

andrewmwells-amazon: Maybe that's the right call for now? It would work well for the internal use case, but I'm hesitant because I think using the same type for human-readable and `UUID`s is an anti-pattern. It requires adding constraints like disallowing some prefix from the human-readable `Eid` and requiring it for the `UUID`s.

john-h-kastner-aws (approved):

PR #1258, opened 2024-10-02
Suggest obvious fix when action id includes type

### LINKED ISSUE #166: Suggested replacement for unrecognized action entity misses obvious fix

### Bug Category

Schemas and Validation

### Describe the bug

The `actions` section of the schema lists entity IDs, whereas the `entityTypes` section lists entity types. If a user mistakenly includes the `Action` entity type (and any namespace) as part of their action ID, the error messages are confusing. 

### Expected behavior

Ideally, the error message can identify the user's mistake more clearly, and do better at suggesting a fix. The suggestion would need to account for this particular mistake of including part of the type/namespace in the action ID.

### Reproduction steps

In [tinytodo's `tinytodo.cedarschema.json`](https://github.com/cedar-policy/cedar-examples/blob/main/tinytodo/tinytodo.cedarschema.json), modify the `actions` element `"createList"` to instead be `"Action::CreateList"`. When validating `polices.cedar` we get the error message
```
Validation error on policy policy0: Unrecognized action id Action::"CreateList", did you mean Action::"GetList"?
```
This is confusing! With this change, we are defining an action entity `Action::"Action::CreateList"`. This is not what the user intended, because they failed to appreciate the distinction between entity IDs and entity types.

### Code Snippet

_No response_

### Log output

_No response_

### Additional configuration

_No response_

### Operating  System

_No response_

_No response_

adpaco-aws:
## Description of changes

This PR extends the validation of action IDs so that we detect cases in which the user may have mistakenly included the entity type (i.e., `"Action::"`) in the action ID. In that case, we don't suggest another action ID like we used to, but instead draw attention to the fact that the action ID was defined with the entity type in its name. This avoids misleading the user when the fix is much simpler.

Resolves #166 

### DIFF
--- a/cedar-policy-validator/src/diagnostics/validation_errors.rs
+++ b/cedar-policy-validator/src/diagnostics/validation_errors.rs
@@ -133,7 +133,10 @@ pub fn unrecognized_action_id_help(
         ))
     } else {
         // Otherwise, suggest using another id
-        let euids_strs = schema.known_action_ids().map(ToString::to_string).collect::<Vec<_>>();
+        let euids_strs = schema
+            .known_action_ids()
+            .map(ToString::to_string)
+            .collect::<Vec<_>>();
         fuzzy_search(euid.eid().as_ref(), &euids_strs)
             .map(UnrecognizedActionIdHelp::SuggestAlternative)
     }
--- a/cedar-policy-validator/src/diagnostics/validation_errors.rs
+++ b/cedar-policy-validator/src/diagnostics/validation_errors.rs
@@ -70,18 +70,17 @@ pub struct UnrecognizedActionId {
     pub source_loc: Option<Loc>,
     /// Policy ID where the error occurred
     pub policy_id: PolicyID,
-    /// Action Id seen in the policy.
+    /// Action Id seen in the policy
     pub actual_action_id: String,
-    /// An action id from the schema that the user might reasonably have
-    /// intended to write.
-    pub unrecognized_action_id_help: Option<UnrecognizedActionIdHelp>,
+    /// Hint for resolving the error
+    pub hint: Option<UnrecognizedActionIdHelp>,
 }
 
 impl Diagnostic for UnrecognizedActionId {
     impl_diagnostic_from_source_loc_opt_field!(source_loc);
 
     fn help<'a>(&'a self) -> Option<Box<dyn std::fmt::Display + 'a>> {
-        self.unrecognized_action_id_help
+        self.hint
             .as_ref()
             .map(|help| Box::new(help) as Box<dyn std::fmt::Display>)
     }
@@ -108,8 +107,8 @@ pub fn unrecognized_action_id_help(
     let eid_with_type = format!("Action::{}", eid_str);
     let eid_with_type_and_quotes = format!("Action::\"{}\"", eid_str);
     let maybe_id_with_type = schema.known_action_ids().find(|euid| {
-        let eid_string = <Eid as AsRef<str>>::as_ref(euid.eid()).to_string();
-        eid_string.contains(&eid_with_type) || eid_string.contains(&eid_with_type_and_quotes)
+        let eid = <Eid as AsRef<str>>::as_ref(euid.eid());
+        eid.contains(&eid_with_type) || eid.contains(&eid_with_type_and_quotes)
     });
     if let Some(id) = maybe_id_with_type {
         // In that case, let the user know about it
--- a/cedar-policy-validator/src/diagnostics/validation_errors.rs
+++ b/cedar-policy-validator/src/diagnostics/validation_errors.rs
@@ -31,6 +31,7 @@ use cedar_policy_core::parser::join_with_conjunction;
 
 use crate::fuzzy_match::fuzzy_search;
 use crate::types::{EntityLUB, EntityRecordKind, RequestEnv, Type};
+use crate::ValidatorSchema;
 use itertools::Itertools;
 use smol_str::SmolStr;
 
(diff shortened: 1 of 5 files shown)


cedar-policy-validator/src/rbac.rs
```diff
@@ -79,7 +78,6 @@ impl Validator {
         let known_action_ids = self
             .schema
             .known_action_ids()
-            .map(ToString::to_string)
             .collect::<Vec<_>>();
```
john-h-kastner-aws, line 81: Might be able to get rid of this `collect` by using `&Schema` as the parameter type for `unrecognized_action_id_help`

cedar-policy-validator/src/diagnostics/validation_errors.rs
```diff
@@ -72,20 +73,55 @@ pub struct UnrecognizedActionId {
...
+/// Help for resolving an unrecognized action id error
+#[derive(Debug, Clone, Error, Hash, Eq, PartialEq)]
+pub enum UnrecognizedActionIdHelp {
+    /// Draw attention to action id including action type (e.g., `Action::"Action::view"`)
+    AvoidActionTypeInActionId(String),
+    /// Suggest an alternative action
+    SuggestAlternative(String),
+}
+
+impl std::fmt::Display for UnrecognizedActionIdHelp {
```
john-h-kastner-aws, line 100: Do we need this Display impl? Alternatively, so we just use it in the `help` function?

cedar-policy-validator/src/rbac.rs
```diff
@@ -727,6 +725,38 @@ mod test {
         assert_eq!(notes.len(), 1, "{:?}", notes);
     }
 
+    #[test]
+    fn validate_action_id_with_action_type() {
```
john-h-kastner-aws, line 729: Probably worth adding a test case with a non-empty namespace

john-h-kastner-aws (approved):
Looks pretty good.

john-h-kastner-aws (commented):
IMO worth adding an line in the changelog. We weren't consistent with mentioning error message changes like this leading up to the 4.0 release since we had made a ton of error message tweaks, but I like including these small improvements.

cedar-policy-validator/src/diagnostics/validation_errors.rs
```diff
@@ -72,20 +73,55 @@ pub struct UnrecognizedActionId {
     pub actual_action_id: String,
     /// An action id from the schema that the user might reasonably have
     /// intended to write.
-    pub suggested_action_id: Option<String>,
+    pub unrecognized_action_id_help: Option<UnrecognizedActionIdHelp>,
 }
```
cdisselkoen, line 77: Probably worth updating the doc comment.

cedar-policy-validator/src/diagnostics/validation_errors.rs
cdisselkoen: Alternatively, our pattern elsewhere is to not manually impl `Display` for types that `derive(Error)` like this one -- instead, you can use `#[error("blah blah blah")]` on each enum variant to put the `Display` impl inline in the definition.  One example of this is the declaration of `LubHelp` in this same file.

cedar-policy-validator/src/diagnostics/validation_errors.rs
```diff
@@ -72,20 +73,55 @@ pub struct UnrecognizedActionId {
...
+            Self::SuggestAlternative(s) => write!(f, "the action id {s} may be an alternative"),
+        }
+    }
+}
+
+/// Determine the help to offer in the presence of an unrecognized action id error.
+pub fn unrecognized_action_id_help(euid: &EntityUID, euids: &[&EntityUID]) -> Option<UnrecognizedActionIdHelp> {
+    // Check if the user has included the type (i.e., `Action::`) in the action id
+    let eid_with_type = format!("Action::{}", <Eid as AsRef<str>>::as_ref(euid.eid()));
+    let maybe_id_with_type = euids.iter().find(|euid| <Eid as AsRef<str>>::as_ref(euid.eid()).to_string().contains(&eid_with_type));
```
cdisselkoen, line 113: nit: do we need the `.to_string()`? does `&str` not support `.contains()`?

cdisselkoen (approved):
Looks good overall.

cedar-policy-validator/src/rbac.rs
adpaco-aws: Ok, we're now passing `&Schema`

cedar-policy-validator/src/rbac.rs
adpaco-aws: Added one.

cedar-policy-validator/src/diagnostics/validation_errors.rs
adpaco-aws: You're right, but for some reason I thought it wasn't. Removed it.

cedar-policy-validator/src/diagnostics/validation_errors.rs
adpaco-aws: Updated and renamed this field to `hint` as we do in other types.

cedar-policy-validator/src/diagnostics/validation_errors.rs
```diff
@@ -72,20 +73,55 @@ pub struct UnrecognizedActionId {
...
+            Self::AvoidActionTypeInActionId(s) => write!(f, "the action id {s} is similar but includes the type"),
+            Self::SuggestAlternative(s) => write!(f, "the action id {s} may be an alternative"),
+        }
+    }
+}
+
+/// Determine the help to offer in the presence of an unrecognized action id error.
+pub fn unrecognized_action_id_help(euid: &EntityUID, euids: &[&EntityUID]) -> Option<UnrecognizedActionIdHelp> {
+    // Check if the user has included the type (i.e., `Action::`) in the action id
+    let eid_with_type = format!("Action::{}", <Eid as AsRef<str>>::as_ref(euid.eid()));
```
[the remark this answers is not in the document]
adpaco-aws: Yes, there's no reason not to, but I'll admit I don't know how one goes about creating such action ID. Regardless, we're now checking if the entity IDs contain either `Action::<ID>` or `Action::\"<ID>\"`.

cedar-policy/CHANGELOG.md
```diff
@@ -23,6 +23,11 @@ Cedar Language Version: TBD
   Cedar request. To use this API you must enable the `entity-manifest` feature flag.
 - Added public APIs to get language and SDK version numbers (#1219).
 
+### Changed
+
+- The validator provides a more specific hint when an action ID cannot be found
+  and the same action ID with `Action::` has been defined.
```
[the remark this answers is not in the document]
adpaco-aws: Added locally, thanks!

john-h-kastner-aws (approved):

PR #1262, opened 2024-10-04
Reorganize `to_expr_or_special` for `cst::Member`

john-h-kastner-aws:
## Description of changes

This method has always struck me as unnecessarily complex. This PR tries to simplify it by observing that most of the cases can only occur on the first iteration of the loop. On the first iteration, we might have an access on a name (`foo.bar`),  variable (`principal.bar`), string-literal (`"foo".bar`) or expression (`(principal.foo).bar`), but on all subsequent iterations we can only have the expression case. The string literal case was identical to the expression case, so these can share code.

### DIFF
--- a/cedar-policy-core/src/parser/cst_to_ast.rs
+++ b/cedar-policy-core/src/parser/cst_to_ast.rs
@@ -1128,6 +1128,46 @@ impl Node<Option<cst::Member>> {
         }
     }
 
+    /// Construct an attribute access or method call on an expression. This also
+    /// handles function calls, but a function call of an arbitrary expression
+    /// is always an error.
+    fn build_expr_accessor<'a>(
+        &self,
+        head: ast::Expr,
+        next: &mut AstAccessor,
+        tail: &'a mut [AstAccessor],
+    ) -> Result<(ast::Expr, &'a mut [AstAccessor])> {
+        use AstAccessor::*;
+        match (next, tail) {
+            // trying to "call" an expression as a function like `(1 + 1)("foo")`. Always an error.
+            (Call(_), _) => Err(self.to_ast_err(ToASTErrorKind::ExpressionCall).into()),
+
+            // method call on arbitrary expression like `[].contains(1)`
+            (Field(id), [Call(args), rest @ ..]) => {
+                // move the expr and args out of the slice
+                let args = std::mem::take(args);
+                // move the id out of the slice as well, to avoid cloning the internal string
+                let id = mem::replace(id, ast::UnreservedId::empty());
+                Ok((id.to_meth(head, args, &self.loc)?, rest))
+            }
+
+            // field of arbitrary expr like `(principal.foo).bar`
+            (Field(id), rest) => {
+                let id = mem::replace(id, ast::UnreservedId::empty());
+                Ok((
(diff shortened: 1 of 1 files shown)


shaobo-he-aws (approved):
Looks great.

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -1128,6 +1128,46 @@ impl Node<Option<cst::Member>> {
...
+    ) -> Result<(ast::Expr, &'a mut [AstAccessor])> {
+        use AstAccessor::*;
+        match (next, tail) {
+            // trying to "call" an expression as a function like `(1 + 1)("foo")`. Always an error.
+            (Call(_), _) => Err(self.to_ast_err(ToASTErrorKind::ExpressionCall).into()),
+
+            // method call on arbitrary expression like `[].contains(1)`
+            (Field(id), [Call(args), rest @ ..]) => {
+                // move the expr and args out of the slice
+                let args = std::mem::take(args);
```
cdisselkoen, line 1148: is this necessary? can we just use `args` directly in the below?

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -1137,198 +1177,115 @@ impl Node<Option<cst::Member>> {
...
-                    let nn = mem::replace(
-                        name,
-                        ast::Name::unqualified_name(ast::UnreservedId::empty()),
-                    );
-                    head = nn.into_func(args, self.loc.clone()).map(|expr| Expr {
-                        expr,
-                        loc: self.loc.clone(),
-                    })?;
-                    tail = rest;
+                    let args = std::mem::take(args);
```
cdisselkoen, line 1210: same here, do we need the `std::mem::take`

cdisselkoen (approved):

cedar-policy-core/src/parser/cst_to_ast.rs
john-h-kastner-aws: Should be avoidable, but `tail` is a slice, and `to_meth` wants to take ownership, so not trivially. I think it would mean rewriting this to work of an `impl Iterator<Item = AstAccessor>`, but then doing the lookahead is kind of annoying.

cdisselkoen: Ah I see, I missed the types involved here.

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -1128,6 +1128,46 @@ impl Node<Option<cst::Member>> {
         }
     }
 
+    /// Construct an attribute access or method call on an expression. This also
+    /// handles function calls, but a function call of an arbitrary expression
+    /// is always an error.
```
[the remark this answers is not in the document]
john-h-kastner-aws: Maybe there's a better phrasing, but the "method call on expression" is just any normal method call like `principal.foo.contains("thing")`, while "function call on expression" would be (principal.foo)(2)`. I put examples of these in the actual code.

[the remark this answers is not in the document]
cdisselkoen: Makes sense

cdisselkoen (approved):

PR #1296, opened 2024-10-29
Speed Improvement: Make incoming entities Arc<Entities> like schema entities

tomlikestorock:
## Description of changes
This provides a 3x speedup based on local benchmarking with a large schema that has big action entity hierarchy relationships. Today Entities.entities is a HashMap of `<EntityUID, Entity>`. When from_entities brings in the entities and actions from a given schema, it currently relies on `Arc::unwrap_or_clone`, which frequently ends up being clone. For a large set of very interrelated Actions, this repeated clone becomes expensive.

Changing Entities.entities to be a HashMap of `<EntityUID, Arc<Entity>>` allows us to keep the clone cheap, and Arc'ing the incoming Entities becomes cheap as well.

Changing the interface of the HashMap means altering what create_entities_map returns, adding an implementation of TCNode for `Arc<Entity>`, and udpating how IntoIterator works to return values from the Arc's.

## Issue #, if available
This contributes to 1285.

### DIFF
--- a/cedar-policy-core/src/entities.rs
+++ b/cedar-policy-core/src/entities.rs
@@ -58,7 +58,7 @@ pub struct Entities {
     /// Important internal invariant: for any `Entities` object that exists, the
     /// the `ancestor` relation is transitively closed.
     #[serde_as(as = "Vec<(_, _)>")]
-    entities: HashMap<EntityUID, Entity>,
+    entities: HashMap<EntityUID, Arc<Entity>>,
 
     /// The mode flag determines whether this store functions as a partial store or
     /// as a fully concrete store.
@@ -109,7 +109,7 @@ impl Entities {
 
     /// Iterate over the `Entity`s in the `Entities`
     pub fn iter(&self) -> impl Iterator<Item = &Entity> {
-        self.entities.values()
+        self.entities.values().map(|e| e.as_ref())
     }
 
     /// Adds the [`crate::ast::Entity`]s in the iterator to this [`Entities`].
@@ -140,7 +140,7 @@ impl Entities {
                     return Err(EntitiesError::duplicate(entity.uid().clone()))
                 }
                 hash_map::Entry::Vacant(vacant_entry) => {
-                    vacant_entry.insert(entity);
+                    vacant_entry.insert(Arc::new(entity));
                 }
             }
         }
@@ -213,7 +213,7 @@ impl Entities {
                 schema
                     .action_entities()
                     .into_iter()
-                    .map(|e| (e.uid().clone(), Arc::unwrap_or_clone(e))),
+                    .map(|e: Arc<Entity>| (e.uid().clone(), e)),
             );
         }
         Ok(Self {
@@ -252,6 +252,7 @@ impl Entities {
     fn to_ejsons(&self) -> Result<Vec<EntityJson>> {
         self.entities
             .values()
+            .map(Arc::as_ref)
             .map(EntityJson::from_entity)
             .collect::<std::result::Result<_, JsonSerializationError>>()
             .map_err(Into::into)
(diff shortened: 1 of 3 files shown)


tpaulus (approved):

cedar-policy-core/src/entities.rs
```diff
@@ -322,13 +323,13 @@ impl Entities {
 }
 
 /// Create a map from EntityUids to Entities, erroring if there are any duplicates
-fn create_entity_map(es: impl Iterator<Item = Entity>) -> Result<HashMap<EntityUID, Entity>> {
+fn create_entity_map(es: impl Iterator<Item = Entity>) -> Result<HashMap<EntityUID, Arc<Entity>>> {
```
cdisselkoen, line 326: should we also have the input `es` here be an iterator of `Arc<Entity>`, and force callers to map `Arc::new` if needed?

cedar-policy-core/src/entities.rs
```diff
@@ -338,10 +339,14 @@ fn create_entity_map(es: impl Iterator<Item = Entity>) -> Result<HashMap<EntityU
...
 
-    type IntoIter = hash_map::IntoValues<EntityUID, Entity>;
+    type IntoIter = std::vec::IntoIter<Entity>;
 
     fn into_iter(self) -> Self::IntoIter {
-        self.entities.into_values()
+        self.entities
+            .into_values()
+            .map(Arc::unwrap_or_clone)
+            .collect::<Vec<Entity>>()
```
cdisselkoen, line 348: Having to `.collect()` here seems like a potential performance problem as well ... is it possible to avoid it with a different choice of `type IntoIter` above, so that the implementation here can just be `self.entities.into_values().map(Arc::unwrap_or_clone)`?  or alternately, should we change the `type Item` to be `Arc<Entity>`?

cdisselkoen (approved):

cedar-policy-core/src/entities.rs
```diff
@@ -213,7 +213,7 @@ impl Entities {
                 schema
                     .action_entities()
                     .into_iter()
-                    .map(|e| (e.uid().clone(), Arc::unwrap_or_clone(e))),
```
john-h-kastner-aws, line 216: It took me a second to puzzle out why this actually gives a speed up, so I'll leave my reasoning here. Maybe worth elaborating in a comment.

We remove this `unwrap_or_clone`, but add a `make_mut` (which I assume must clone if there's another reference to the data) call inside TC computation, so at first glance it feels like that should be a wash. But, the TC was already computed for theentity map (which we just constructed, so one ref to each `Arc`), and we separately computed TC for the actions inside the schema constructor (while it had unique ownership). Put together these should mean we so never trigger a clone on the `make_mut` during TC compunction.

john-h-kastner-aws (approved):
Looks good. Craig's comments are probably worth addressing.

cedar-policy-core/src/entities.rs
tomlikestorock: Ah, I wasn't sure how interested you'd be in propagating this Arc expectation (at least internally). I can make change that reflect this for sure.

cedar-policy-core/src/entities.rs
tomlikestorock: Definitely, I'll check that out.

tpaulus (approved):

cedar-policy-core/src/entities.rs
tomlikestorock: Updated to this request in 1da976e9668283ae545d252e652b1003f7a682fb, let me know what you think.

cedar-policy-core/src/entities.rs
tomlikestorock: Updated in 11e70bd. I changed the IntoIter type to reflect the .map call as the return type. This gave another speed boost as is.

cdisselkoen (approved):

cdisselkoen:
Thanks for this work!

john-h-kastner-aws (approved):
Looks the build error is coming from the protobuf feature

tomlikestorock:
Protobuf tests updated to use Arc wrapped entities in 857509b.

john-h-kastner-aws (approved):

PR #1308, opened 2024-11-08
improve performance of deserializing invalid json policies

### LINKED ISSUE #1284: Optimize parsing EST from deeply nested JSON objects
### Category

Other

Large, deeply nested JSON objects can take a long time to parse (or fail to parse) as Cedar policy ESTs. For example: https://gist.github.com/john-h-kastner-aws/f4ec6ca7e245b1565888763698fb89b5#file-slow-est-json takes almost 10 seconds to run when passed to the `convert-policy-json-to-cedar` fuzz target `cargo fuzz run convert-policy-json-to-cedar slow-unit-41767207ce4a2bed039c0df5364ecf2e507258e6  -s none`, and this is a somewhat reduced example taken from an even slower test case.

We have encountered a similar issue before and fixed it by a change to test infrastructure (https://github.com/cedar-policy/cedar-spec/pull/419), but, I think a proper fix would involve a fairly invasive change to our JSON parsing.

As a starting point, I suspect that our use of `untagged` enums may be responsible. See [this](https://stackoverflow.com/questions/73458178/why-serde-json-rust-so-slow-when-deserializing-in-untagged-enums), [this](https://old.reddit.com/r/rust/comments/w3q1oq/things_i_wish_i_had_known_about_serde_json/), and [this](https://users.rust-lang.org/t/how-to-deserialize-untagged-enums-fast/28331)

.

### Additional context

_No response_

cdisselkoen:
## Description of changes

Replaces the derived `Deserialize` implementation for `est::Expr` with a custom one, resulting in greatly improved performance on the #1284 benchmark (see #1307).  Should be a performance improvement for a wide variety of invalid `est::Expr`s.  Hopefully not a slowdown for the happy path, that is, valid `est::Expr`s.

On my machine, the new numbers for the #1284 benchmark are:
```
Policy::from_json       time:   [107.54 µs 107.61 µs 107.70 µs]
                        change: [-99.997% -99.997% -99.997%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low mild
  3 (3.00%) high mild
  2 (2.00%) high severe
```

That is, performance has improved from about 3.4s to execute once, to about 108 µs.

## Issue #, if available

Fixes #1284.

### DIFF
--- a/cedar-policy-core/src/est/expr.rs
+++ b/cedar-policy-core/src/est/expr.rs
@@ -21,6 +21,7 @@ use crate::entities::json::{
     CedarValueJson, FnAndArg, TypeAndId,
 };
 use crate::extensions::Extensions;
+use crate::jsonvalue::JsonValueWithNoDuplicateKeys;
 use crate::parser::cst::{self, Ident};
 use crate::parser::err::{ParseErrors, ToASTError, ToASTErrorKind};
 use crate::parser::unescape::to_unescaped_string;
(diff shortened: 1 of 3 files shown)


cedar-policy-core/src/est/expr.rs
```diff
@@ -21,34 +21,100 @@ use crate::entities::json::{
...
+                        return Err(serde::de::Error::custom(format!("JSON object representing an Expr should have only one key, but found two keys: `{k}` and `{k2}`")));
+                    }
+                };
+                match k.as_str() {
+                    // All the valid keys for `ExprNoExt`.
+                    // If a variant or alias is added to `ExprNoExt`, it must also be added here.
+                    "Value" | "Var" | "Slot" | "!" | "neg" | "==" | "!=" | "in" | "<" | "<="
+                    | ">" | ">=" | "&&" | "||" | "+" | "-" | "*" | "contains" | "containsAll"
+                    | "containsAny" | "getTag" | "hasTag" | "." | "has" | "like" | "is"
+                    | "if-then-else" | "Set" | "Record" => {
```
adpaco-aws, line 97: I can see this coming back to bite us in the future. Is there a good reason not to have an `enum` with all the valid keys and use it here?

cedar-policy-core/src/est/expr.rs
```diff
@@ -21,34 +21,100 @@ use crate::entities::json::{
...
+                            "empty map is not a valid expression",
+                        ))
+                    }
+                    Some((k, v)) => (k, v),
+                };
+                match map.next_key()? {
+                    None => (),
+                    Some(k2) => {
+                        let k2: SmolStr = k2;
+                        return Err(serde::de::Error::custom(format!("JSON object representing an Expr should have only one key, but found two keys: `{k}` and `{k2}`")));
```
adpaco-aws, line 88:
```suggestion
                        return Err(serde::de::Error::custom(format!("JSON object representing an `Expr` should have only one key, but found two keys: `{k}` and `{k2}`")));
```

cedar-policy-core/src/est/expr.rs
cdisselkoen: 1) I think the overall risk here is low.  If someone adds a case to `ExprNoExt` and not here, it will be caught by literally any EST deserialization test involving the new variant.  And no one will add a variant to `ExprNoExt` without also adding a test for it.
2) Putting these in an `enum` doesn't solve the problem of having to maintain the correspondence between the keys here and the keys of `ExprNoExt`.  But if you think having another `enum` would improve readability, what specifically is the proposal?  That we would try to deserialize `k` into that enum, instead of having the logic inline?

cdisselkoen: We could also invert the logic -- have a list of known extension function names, and in those cases deserialize into the `ExtFuncCall` variant, in all other cases deserialize into the `ExprNoExt` variant.  Then we'd have to update on any addition of an extension function, rather than any addition to `ExprNoExt`.  My feeling was that additions of extension functions are likely to be more frequent.

cdisselkoen: Actually, we might be able to use the `Extensions` object to dynamically look up whether a string is a valid extension function name, and route the deserialization accordingly.  This would rely on a single source of truth for defined extensions, and we wouldn't have to make updates in both places.  Would that be better?

adpaco-aws: Yes, I'm fine with any alternative here.

It's true that adding an `enum` doesn't solve the problem, but adding just one `impl`  where we `match` on each variant of that `enum` should force updates to it when adding new ones. However, I haven't checked how `ExprNoExt` works in details so it's possible that wouldn't be so useful in the case of new keys.

john-h-kastner-aws: I like that solution. Though, we would ideally want to parametrize deserailzation on an `Extensions`, and I'm not sure how that would want.

cedar-policy-core/src/est/expr.rs
```diff
@@ -21,34 +21,100 @@ use crate::entities::json::{
...
+                            "empty map is not a valid expression",
+                        ))
+                    }
+                    Some((k, v)) => (k, v),
+                };
+                match map.next_key()? {
+                    None => (),
+                    Some(k2) => {
+                        let k2: SmolStr = k2;
+                        return Err(serde::de::Error::custom(format!("JSON object representing an `Expr` should have only one key, but found two keys: `{k}` and `{k2}`")));
```
john-h-kastner-aws, line 88: do we ever test this error case?

cedar-policy-core/src/est/expr.rs
john-h-kastner-aws: make sense. I guess then you should make sure we have a test case for parsing a deactivated but nonetheless existent extension function.

cdisselkoen: Hmm actually, I think the current behavior is that a policy using a deactivated-but-nonetheless-existent extension function does parse (all the way to AST) and you only get an evaluation error

cedar-policy-core/src/est/expr.rs
cdisselkoen: Added a test case

shaobo-he-aws (approved):

cedar-policy-core/src/est/expr.rs
cdisselkoen: Hmm this version with `is_known_extension_func_name()` is 20% slower on the 1284 benchmark than the original version (with the hardcoded match on list of ExprNoExt keys).  (now 129 µs on my machine.)  Not sure if we care since this is an edge case anyway.  More interested in if the happy-path performance changed

john-h-kastner-aws (approved):
Looks good.

It might be worth opening a follow up issue to investigate other uses of `untaqged`.

cedar-policy-core/src/est/expr.rs
john-h-kastner-aws: ah, that's unfortunate. Maybe due to calling `ast::Name::parse_unqualified_name` a lot?
I think that still sounds like a (very) good improvement over current performance, so I'm still happy to see this merged.

cdisselkoen: Yeah, it probably is due to that.  I wonder if there's a faster way to check this.

We could take advantage of the fact that currently no extension functions are namespaced, and have a function `is_unknown_extension_func_name()` that works on `Id` instead of `Name`.  That would avoid parsing.  We could even justify using `Id::new_unchecked`, because even if the name was reserved, it wouldn't match an extension func name

PR #1316, opened 2024-11-11
RFC 48 implementation

shaobo-he-aws:
## Description of changes

This PR implements RFC 48 (schema annotations).


cedar-policy-validator/src/cedar_schema/grammar.lalrpop
```diff
@@ -82,10 +82,18 @@ match {
 
     // other tokens
     ",", ";", ":", "::", "{", "}", "[", "]",
-    "<", ">", "=", "?",
+    "<", ">", "=", "?", "@", "(", ")",
 
 }
 
+// Annotations := {'@' Ident '(' String ')'}
+Annotation: Node<()> = {
```
john-h-kastner-aws, line 90: I wonder if we can start sharing some code between this grammar and the policy grammar

cedar-policy-validator/src/cedar_schema/err.rs
```diff
@@ -401,6 +402,10 @@ pub enum ToJsonSchemaError {
     #[error(transparent)]
     #[diagnostic(transparent)]
     ReservedSchemaKeyword(#[from] ReservedSchemaKeyword),
+    /// Duplicate annotations
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    DuplicateAnnotations(#[from] DuplicateAnnotations),
```
john-h-kastner-aws, line 408: Breaking change. This is a exhaustive public error type. Semver checks is passing, but I think it's not looking into `pub use`'ed items

cedar-policy-validator/src/cedar_schema/ast.rs
```diff
@@ -29,11 +29,47 @@ use smol_str::ToSmolStr;
...
+use super::err::DuplicateAnnotations;
+
 pub const BUILTIN_TYPES: [&str; 3] = ["Long", "String", "Bool"];
 
 pub(super) const CEDAR_NAMESPACE: &str = "__cedar";
 
-pub type Schema = Vec<Node<Namespace>>;
+pub type Schema = Vec<Annotated<Namespace>>;
+
+pub type Annotation = (Node<AnyId>, Node<SmolStr>);
```
john-h-kastner-aws, line 40: Will this handle this `@foo` annotation syntax?

cedar-policy-validator/src/cedar_schema/grammar.lalrpop
```diff
@@ -82,21 +85,35 @@ match {
...
 
+#[inline]
+AnyIdent: Node<AnyId> = {
+    <l:@L> <id:IDENTIFIER> <r:@R>
+        => Node::with_source_loc(AnyId::new_unchecked(id), Loc::new(l..r, Arc::clone(src))),
+}
+
+// Annotations := {'@' IDENTIFIER '(' String ')'}
+Annotation: Node<Annotation> = {
+    <l:@L> "@" <key:AnyIdent> <value: ("(" <STR> ")")> <r:@R> => Node::with_source_loc((key, value), Loc::new(l..r, Arc::clone(src)))
```
john-h-kastner-aws, line 100: Doesn't look like this handles `@foo` empty annotations. Ideally we could use the same grammar production to ensure they're exactly the same

cedar-policy-validator/src/schema/namespace_def.rs
```diff
@@ -132,12 +132,31 @@ impl ValidatorNamespaceDef<ConditionalName, ConditionalName> {
 
         // Convert the common types, actions and entity types from the schema
         // file into the representation used by the validator.
-        let common_types =
-            CommonTypeDefs::from_raw_common_types(namespace_def.common_types, namespace.as_ref())?;
-        let actions =
-            ActionsDef::from_raw_actions(namespace_def.actions, namespace.as_ref(), extensions)?;
-        let entity_types =
-            EntityTypesDef::from_raw_entity_types(namespace_def.entity_types, namespace.as_ref())?;
+        let common_types = CommonTypeDefs::from_raw_common_types(
```
john-h-kastner-aws, line 135: Could we update these `from_raw_*` functions to take an `impl IntoIterator<..>` to avoid rebuilding the map?

cedar-policy-validator/src/json_schema.rs
```diff
@@ -1314,6 +1326,163 @@ impl<N> From<TypeVariant<N>> for Type<N> {
     }
 }
 
+/// The fields for a `Type`. Used for implementing deserialization.
```
john-h-kastner-aws, line 1329: Why duplicate this for annotated types? Do we still need to be able to parse types without annotations?

cedar-policy-validator/src/cedar_schema/to_json_schema.rs
```diff
@@ -165,25 +195,63 @@ impl TryFrom<Namespace> for json_schema::NamespaceDefinition<RawName> {
         let (entity_types, action, common_types) = into_partition_decls(n.decls);
 
         // Convert entity type decls, collecting all errors
-        let entity_types = collect_all_errors(entity_types.into_iter().map(convert_entity_decl))?
-            .flatten()
-            .collect();
+        let entity_types = collect_all_errors(entity_types.into_iter().map(|et| {
+            let et = annotations::Annotated::try_from(et)
+                .map_err(|err| ToJsonSchemaErrors::from(ToJsonSchemaError::from(err)))?;
+            convert_entity_decl(et.data).map(|i| {
```
john-h-kastner-aws, line 201: IMO it might be nicer if `convert_entity_decl` took the annotated entity type as it argument and returned the converted decl with annotations

cedar-policy-validator/src/json_schema.rs
```diff
@@ -2933,3 +3134,149 @@ mod test_duplicates_error {
...
+                    }
+                }
+            }
+        }
+                    },
+                    "actions": {},
+                    "commonTypes": {}
+                }});
+        let schema: Result<Fragment<RawName>, _> = serde_json::from_value(src);
+        assert_matches!(schema, Err(_));
```
john-h-kastner-aws, line 3240: Shouldn't have any test case just asserting `assert_matches!(schema, Err(_));`

cedar-policy-validator/src/cedar_schema/grammar.lalrpop
```diff
@@ -82,21 +85,35 @@ match {
...
     <ns: Namespace*> => ns,
 }
 
 // Namespace := 'namespace' Path '{' {Decl} '}'
-Namespace: Node<Namespace> = {
-    <l:@L> NAMESPACE <p: Path> "{" <decls: Decl*> "}" <r:@R>
-        => Node::with_source_loc(Namespace { name: Some(Node::with_source_loc(p, Loc::new(l..r, Arc::clone(src)))), decls}, Loc::new(l..r, Arc::clone(src))),
-     <l:@L> <decl: Decl> <r:@R> => Node::with_source_loc(Namespace {name: None, decls: vec![decl]}, Loc::new(l..r, Arc::clone(src))),
+Namespace: Annotated<Namespace> = {
+    <annotations: Annotation*> NAMESPACE <p: Path> "{" <decls: Annotated<Decl>*> "}"
```
john-h-kastner-aws, line 114: What's the `Annotated<E>` production doing if you still need to write out ` <annotations: Annotation*>` here?

john-h-kastner-aws (changes requested):
To major issues:
* Doesn't handle `@foo` empty annotation syntax
* Adds a new error variant to a public and exhaustive enum

shaobo-he-aws:
> * Doesn't handle `@foo` empty annotation syntax

I did do it but realized that it wasn't in the RFC.

cedar-policy-validator/src/cedar_schema/grammar.lalrpop
shaobo-he-aws: lalrpop errors if I use `Annotated<E>` here, i.e., the grammar is problematic.

cedar-policy-validator/src/cedar_schema/err.rs
shaobo-he-aws: Got it. I guess that we need to move it to a user error thrown by lalrpop, which actually makes the code much simpler.

john-h-kastner-aws:
> > * Doesn't handle `@foo` empty annotation syntax
> 
> I did do it but realized that it wasn't in the RFC.

fair, but empty annotation came about after the RFC was written, so I'd assume the intent of the RFC was to support annotations in schema as they exist for policies

cedar-policy-validator/src/cedar_schema/err.rs
john-h-kastner-aws: We should track an issue for this breaking change so we remember to apply it for an eventual API version 5.0

cedar-policy-validator/src/annotations.rs
```diff
@@ -0,0 +1,48 @@
...
+//! Annotations used in Cedar/JSON schemas
+
+use std::collections::BTreeMap;
+
+use cedar_policy_core::ast::AnyId;
+use serde::{Deserialize, Serialize};
+use smol_str::SmolStr;
+
+/// Annotations
+pub type Annotations = BTreeMap<AnyId, SmolStr>;
```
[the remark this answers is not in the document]
shaobo-he-aws: On second thought, I think maybe it's not proper that we reuse code here. The reason is that the validator adds serde annotations to the struct. These annotations should be specific to the validator and not propagate to the general core library.

cedar-policy-validator/src/cedar_schema/err.rs
shaobo-he-aws: Implement a non-breaking version.

cedar-policy-validator/src/json_schema.rs
shaobo-he-aws: I believe so. `Type<N>` is used as a value of `shape`. We wouldn't want something like,
```JSON
"shape": {
  "annotations": {"a": ""},
  "type": "Foo"
}
```

cedar-policy-validator/src/cedar_schema/grammar.lalrpop
shaobo-he-aws: Done.

cedar-policy-validator/src/json_schema.rs
shaobo-he-aws: Agreed. I tried to eliminate the duplication like the following,

```rust
    fn visit_map<A>(self, mut map: A) -> std::result::Result<Self::Value, A::Error>
    where
        A: MapAccess<'de>
    {
        let mut pairs: Vec<(String, serde_json::Value)> = Vec::new();
        let mut annotations: Option<Annotations> = None;

        while let Some(key) = map.next_key::<String>()? {
            if key == "annotations" {
                annotations = Some(map.next_value()?);
            } else {
                pairs.push((key, map.next_value()?));
            }
        }
        let rest: MapDeserializer<'_, std::vec::IntoIter<(String, serde_json::Value)>, A::Error> = MapDeserializer::new(pairs.into_iter());
        let ty_visitor = TypeVisitor {
            _phantom: PhantomData,
        };
        Ok(Self::Value {
            data: rest.deserialize_any(ty_visitor)?,
            annotations: annotations.unwrap_or_default(),
        })
    }
```

But I couldn't get type checker to pass.

john-h-kastner-aws: If it means we can simplify the code could we consider revising the RFC allowing? I doesn't seem too bad to me

```json
"shape": {
  "annotations": {"a": ""},
  "type": "Foo"
}
```

i guess in the Cedar format it would be

```
entity foo 
@a("")
{
...
}
```

cedar-policy-validator/src/annotations.rs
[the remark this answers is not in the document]
john-h-kastner-aws: Yeah i see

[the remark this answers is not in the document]
shaobo-he-aws: In the end, I decided to use core's. I realized we can always change core without worrying semver. So why not use the clean version.

cedar-policy-validator/src/schema/namespace_def.rs
shaobo-he-aws: Maybe another PR? This one is already huge.

cedar-policy-validator/src/json_schema.rs
shaobo-he-aws: let's be faithful to the RFC.

cedar-policy-core/src/ast/annotation.rs
```diff
@@ -0,0 +1,167 @@
...
+        self.0.iter()
+    }
+
+    /// Tell if it's empty
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+}
+
+/// Wraps the [`BTreeMap`]` into an opaque type so we can change it later if need be
```
cdisselkoen, line 67:
```suggestion
/// Wraps the [`BTreeMap`] into an opaque type so we can change it later if need be
```

cedar-policy-core/src/ast/id.rs
```diff
@@ -274,7 +276,7 @@ impl AnyId {
     ///
     /// This method was created for the `From<cst::Ident> for AnyId` impl to use.
     /// See notes on `Id::new_unchecked()`.
-    pub(crate) fn new_unchecked(s: impl Into<SmolStr>) -> AnyId {
+    pub fn new_unchecked(s: impl Into<SmolStr>) -> AnyId {
```
cdisselkoen, line 279: Just curious why we need this to be `pub` instead of using `.parse()` as suggested in the comment.  Should we update the comment?

cedar-policy-core/src/est/annotation.rs
```diff
@@ -0,0 +1,86 @@
...
+extern crate tsify;
+
+/// Similar to [`ast::Annotations`] but allow annotation value to be `null`
+#[derive(Serialize, Deserialize, Clone, Hash, Eq, PartialEq, PartialOrd, Ord, Debug)]
+#[cfg_attr(feature = "wasm", derive(tsify::Tsify))]
+pub struct Annotations(
+    #[serde(default)]
+    #[serde(skip_serializing_if = "BTreeMap::is_empty")]
+    #[serde(with = "::serde_with::rust::maps_duplicate_key_is_error")]
+    #[cfg_attr(feature = "wasm", tsify(type = "Record<string, Annotation>"))]
```
cdisselkoen, line 32: maybe my comment above about `loc`s in the EST should have been here.  I didn't realize there were two `Annotations` types now

cedar-policy-validator/src/cedar_schema/ast.rs
```diff
@@ -29,11 +29,53 @@ use smol_str::ToSmolStr;
...
+        }
+    }
+    Ok(Annotated {
+        data,
+        annotations: unique_annotations
+            .into_iter()
+            .map(|(key, value)| {
+                (
+                    key.node,
+                    Annotation::with_optional_value(value.map(|n| n.node), None),
```
cdisselkoen, line 73: is `None` correct here?  does `value` contain a source location we could propagate?

cedar-policy-validator/src/schema/namespace_def.rs
john-h-kastner-aws: that'd be fine. I excepted this change to be trivial, but if it's not we can do it later

cedar-policy-validator/src/cedar_schema/test.rs
```diff
@@ -1984,6 +1988,266 @@ mod translator_tests {
...
+            namespace N {
+              @ae1("🌕")
+              @ae2("moon")
+              entity Moon;
+            }
+            @a("🌌")
+            action "🚀","🛸" appliesTo {
+                principal: [Astronaut, ET],
+                resource: Earth,
+            };
```
cdisselkoen, line 2142: love it

cedar-policy-validator/src/cedar_schema/to_json_schema.rs
```diff
@@ -172,38 +187,46 @@ impl TryFrom<Namespace> for json_schema::NamespaceDefinition<RawName> {
         // Convert action decls, collecting all errors
         let actions = collect_all_errors(action.into_iter().map(convert_action_decl))?
             .flatten()
+            .map(|(key, value)| (key, value))
```
cdisselkoen, line 190: what does this do?

cedar-policy-validator/src/json_schema.rs
```diff
@@ -1655,7 +1749,9 @@ impl TypeOfAttribute<RawName> {
     fn into_n<N: From<RawName>>(self) -> TypeOfAttribute<N> {
         TypeOfAttribute {
             ty: self.ty.into_n(),
+
```
cdisselkoen, line 1752: maybe an accidental blank line?  (also on lines 1765, 1785, 1797)

cedar-policy-validator/src/schema/namespace_def.rs
cdisselkoen: This is indeed a large PR.  It probably could have been split into (1) the refactoring changes in Core and then (2) the changes in cedar-policy-validator.  But at this point I don't think it's worth making the split, just a comment for the future that splitting large PRs like this would be helpful

cdisselkoen (approved):
Looks great, thanks for all the great work

cedar-policy-validator/src/cedar_schema/fmt.rs
```diff
@@ -31,8 +31,8 @@ impl<N: Display> Display for json_schema::Fragment<N> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         for (ns, def) in &self.0 {
             match ns {
-                None => write!(f, "{def}")?,
```
john-h-kastner-aws, line 34: nit: this line shouldn't need to change

john-h-kastner-aws (approved):
looking good

cedar-policy-core/src/ast/annotation.rs
```diff
@@ -0,0 +1,167 @@
...
+        Self(value)
+    }
+}
+
+/// Struct which holds the value of a particular annotation
+#[derive(Educe, Clone, Debug, Serialize, Deserialize, Default)]
+#[educe(Hash, PartialEq, Eq, PartialOrd, Ord)]
+#[serde(transparent)]
+#[cfg_attr(feature = "wasm", derive(tsify::Tsify))]
+#[cfg_attr(feature = "wasm", tsify(into_wasm_abi, from_wasm_abi))]
```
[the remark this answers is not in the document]
shaobo-he-aws: It seems that `Annotation` is still a type alias to `SmolStr` in the generated TS files.

cedar-policy-validator/src/cedar_schema/to_json_schema.rs
john-h-kastner-aws: noticed the same thing in json_schema.rs lines 330 and 334

cedar-policy-core/src/ast/id.rs
shaobo-he-aws: Reverted. Use `from_str` (i.e., `.parse()`) instead.

cedar-policy-validator/src/cedar_schema/ast.rs
shaobo-he-aws: Fixed.

cedar-policy-validator/src/cedar_schema/to_json_schema.rs
shaobo-he-aws: Good catch. Fixed.

cedar-policy-validator/src/json_schema.rs
shaobo-he-aws: Yes. Fixed.

cedar-policy-validator/src/schema/namespace_def.rs
shaobo-he-aws: The changes are indeed trivial. Made them.

john-h-kastner-aws (approved):

PR #1321, opened 2024-11-13
tell dependabot not to update url dependency

cdisselkoen:
## Description of changes

I think this is the right syntax.

For reasoning why this change, see the notes in cedar-wasm/Cargo.toml: 
> Lock `url` (dependencies of cargo-lock) to 2.5.2 because they may introduce a dependency on a crate licensed under the Unicode 3.0 license in a future minor version, and we do not have explicit legal aproval to use that license.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,3 +18,6 @@ updates:
       rust-dependencies:
         patterns:
           - "*"
+    # Ignore (do not update) some dependencies
+    ignore:
+      - dependency-name: "url"  # see notes in cedar-wasm/Cargo.toml


john-h-kastner-aws:
I just told it to `@dependabot ignore url patch version` on it's PR, so let's see what that look like after it rebases

john-h-kastner-aws (approved):
Based on https://github.com/cedar-policy/cedar/pull/1322#issuecomment-2474904790 I think we still want this to ignore any future updates

cdisselkoen:
Closing in favor of #1323

PR #1324, opened 2024-11-14
derives `Clone` on a bunch more of our types

cdisselkoen:
Many of these maybe _shouldn't_ have to be cloned, but there's no reason not to derive `Clone`, and maybe it enables some more threading usecases for our users or something

### DIFF
--- a/cedar-policy-core/src/authorizer.rs
+++ b/cedar-policy-core/src/authorizer.rs
@@ -40,7 +40,7 @@ pub use partial_response::ErrorState;
 pub use partial_response::PartialResponse;
 
 /// Authorizer
-#[derive(Clone)]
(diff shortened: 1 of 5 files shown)


cedar-policy-core/src/authorizer.rs
```diff
@@ -40,6 +40,7 @@ pub use partial_response::ErrorState;
 pub use partial_response::PartialResponse;
 
 /// Authorizer
+#[derive(Clone)]
```
john-h-kastner-aws, line 43: May as well derive `Debug` while we're in here

```suggestion
#[derive(Debug, Clone)]
```

john-h-kastner-aws (approved):

adpaco-aws (approved):

PR #1343, opened 2024-11-27
remove Loc from protobufs in AST.proto

cdisselkoen:
## Description of changes

Removes source locations from all the protobuf formats in AST.proto.  They are not used by the Lean code which consumes them.  And, since each `Loc` includes a full copy of the original source string, they blow up the encoded size of the protobufs significantly.  

Breaking change, but `protobufs` is an experimental feature and this is exactly the kind of change folks should expect while the feature is experimental.


john-h-kastner-aws (approved):
Also needs the update in `cedar-policy-cli`

cdisselkoen:
Thanks for the reminder, done

shaobo-he-aws (approved):

PR #1345, opened 2024-12-02
Expose protobufs in public API

cdisselkoen:
## Description of changes

To date, the protobuf serialization/deserialization experimental feature is only available to direct consumers of cedar-policy-core and cedar-policy-validator, and callers are required to manually call `from()` / `into()` to convert into the appropriate struct in `cedar_policy_core::ast::proto` or `cedar_policy_validator::proto`. 

This PR introduces a new `Protobuf` trait, exposed in our public API (guarded by the `protobufs` feature flag, of course), which allows consumers of the public API to serialize/deserialize various public-API types like `Policy` and `Schema` using protobufs.  All necessary conversions are performed under-the-hood, and the types in the `proto` modules are not exposed publicly.

### DIFF
--- a/cedar-policy-core/src/ast/policy.rs
+++ b/cedar-policy-core/src/ast/policy.rs
@@ -619,6 +619,15 @@ impl From<&proto::LiteralPolicy> for LiteralPolicy {
 }
 
 #[cfg(feature = "protobufs")]
+impl TryFrom<&proto::LiteralPolicy> for Policy {
+    type Error = ReificationError;
+    fn try_from(policy: &proto::LiteralPolicy) -> Result<Self, Self::Error> {
+        // TODO: do we need to provide a nonempty `templates` argument to `.reify()`
+        LiteralPolicy::from(policy).reify(&HashMap::new())
+    }
+}
+
+#[cfg(feature = "protobufs")]
 impl From<&LiteralPolicy> for proto::LiteralPolicy {
     fn from(v: &LiteralPolicy) -> Self {
         let template_id_str: &str = v.template_id.as_ref();
--- a/cedar-policy-validator/protobuf_schema/Validator.proto
+++ b/cedar-policy-validator/protobuf_schema/Validator.proto
@@ -102,8 +102,8 @@ message EntityRecordKind {
 }
 
 enum OpenTag {
-        OpenAttributes = 0;
-        ClosedAttributes = 1;
+    OpenAttributes = 0;
+    ClosedAttributes = 1;
 }
 
 message Attributes {
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -2019,6 +2019,7 @@ impl PolicySet {
     }
 
     /// Build the [`PolicySet`] from just the AST information
+    #[cfg_attr(not(feature = "protobufs"), allow(dead_code))]
     fn from_ast(ast: ast::PolicySet) -> Result<Self, PolicySetError> {
         Self::from_policies(ast.into_policies().map(Policy::from_ast))
     }
@@ -2687,6 +2688,7 @@ impl Template {
         })
     }
 
+    #[cfg_attr(not(feature = "protobufs"), allow(dead_code))]
     fn from_ast(ast: ast::Template) -> Self {
         Self {
             lossless: LosslessPolicy::Est(ast.clone().into()),
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -83,6 +83,55 @@ pub(crate) mod version {
     }
 }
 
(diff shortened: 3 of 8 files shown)


john-h-kastner-aws (approved):

cedar-policy/src/api.rs
```diff
@@ -1927,6 +2016,11 @@ impl PolicySet {
         })
     }
 
+    /// Build the [`PolicySet`] from just the AST information
+    pub fn from_ast(ast: ast::PolicySet) -> Result<Self, PolicySetError> {
```
john-h-kastner-aws, line 2020: Should this be `pub`?

cedar-policy-validator/protobuf_schema/Validator.proto
```diff
@@ -102,8 +102,8 @@ message EntityRecordKind {
 }
 
 enum OpenTag {
-        OpenAttributes = 0;
-        ClosedAttributes = 1;
+    OpenAttributes = 0;
```
john-h-kastner-aws, line 105: A general note about the protobuf impl. Does this mean that `OpenAttributes` on records in the schema would become a publicly accessible feature when we stabilize protobuf? What other internal details might be exposed this way?

cedar-policy-core/src/ast/policy.rs
```diff
@@ -618,6 +618,15 @@ impl From<&proto::LiteralPolicy> for LiteralPolicy {
     }
 }
 
+#[cfg(feature = "protobufs")]
+impl TryFrom<&proto::LiteralPolicy> for Policy {
+    type Error = ReificationError;
+    fn try_from(policy: &proto::LiteralPolicy) -> Result<Self, Self::Error> {
+        // TODO: do we need to provide a nonempty `templates` argument to `.reify()`
+        LiteralPolicy::from(policy).reify(&HashMap::new())
```
adpaco-aws, line 626: Do we need an issue to track this work?

adpaco-aws (approved):

cedar-policy/src/api.rs
cdisselkoen: Correct, it shouldn't be

cedar-policy-validator/protobuf_schema/Validator.proto
cdisselkoen: Interesting question.

Protobuf is not self-describing.  In this PR, public-API users only see a blob of bytes, and if one of those bits indicates open-attributes vs closed-attributes, they would never know.  But if anyone actually relies on our `.proto` files to encode/decode themselves, then yes, they see these attributes.

I believe the Lean protobuf parser errors if it sees OpenAttributes.

cedar-policy-core/src/ast/policy.rs
cdisselkoen: There's a bunch of other code todos behind the `protobufs` feature flag, maybe we should have an issue to track all of them in general?  For instance a number of potential panics in this code

adpaco-aws: Yes, we could have a single issue with all the "tasks" related to the feature.

cdisselkoen: #1348

cedar-policy-validator/protobuf_schema/Validator.proto
cdisselkoen: Arguably if we want to stabilize protobufs before OpenAttributes, we should remove OpenAttributes from the protobuf format.

john-h-kastner-aws: Yeah, I'd agree with that. And also

* [attributes on actions](https://github.com/cedar-policy/cedar/blob/770b9f0bff6cf3e4c0074c3153b9983adf598e78/cedar-policy-validator/protobuf_schema/Validator.proto#L53)
* [never](https://github.com/cedar-policy/cedar/blob/770b9f0bff6cf3e4c0074c3153b9983adf598e78/cedar-policy-validator/protobuf_schema/Validator.proto#L70) and [empty set](https://github.com/cedar-policy/cedar/blob/770b9f0bff6cf3e4c0074c3153b9983adf598e78/cedar-policy-validator/protobuf_schema/Validator.proto#L73) types
* [Action Entity](https://github.com/cedar-policy/cedar/blob/770b9f0bff6cf3e4c0074c3153b9983adf598e78/cedar-policy-validator/protobuf_schema/Validator.proto#L85) type
* and [permissive validation mode](https://github.com/cedar-policy/cedar/blob/770b9f0bff6cf3e4c0074c3153b9983adf598e78/cedar-policy-validator/protobuf_schema/Validator.proto#L124) (behind it's own feature)

john-h-kastner-aws (approved):

PR #1350, opened 2024-12-02
followup performance optimization to 1308

cdisselkoen:
## Description of changes

Inspired by [this thread](https://github.com/cedar-policy/cedar/pull/1308#discussion_r1834981214) in #1308.  Avoids repeatedly calling `ast::Name::parse_unqualified_name()` during EST parsing.  On my machine (a different machine than the one I reported numbers on in #1308), the performance of our `deeply_nested_est` benchmark goes from 71.6 µs before this PR, to 63.3 µs after this PR, an 11% improvement.

### DIFF
--- a/cedar-policy-core/src/parser/cst_to_ast.rs
+++ b/cedar-policy-core/src/parser/cst_to_ast.rs
@@ -91,6 +91,7 @@ fn load_styles() -> ExtStyles<'static> {
                 }
             }
             CallStyle::MethodStyle => {
+                debug_assert!(func.name().is_unqualified());
                 methods.insert(func.name().basename());
             }
         };
--- a/cedar-policy-core/src/parser/cst_to_ast.rs
+++ b/cedar-policy-core/src/parser/cst_to_ast.rs
@@ -67,10 +67,10 @@ type Result<T> = std::result::Result<T, ParseErrors>;
 struct ExtStyles<'a> {
     /// All extension function names (just functions, not methods), as `Name`s
     functions: HashSet<&'a ast::Name>,
-    /// All unqualified extension function names (just functions, not methods), as `Id`s
-    unqualified_functions: HashSet<&'a ast::Id>,
     /// All extension function methods. `UnreservedId` is appropriate because methods cannot be namespaced.
     methods: HashSet<ast::UnreservedId>,
+    /// All extension function and method names (both qualified and unqualified), in their string (`Display`) form
+    functions_and_methods_as_str: HashSet<SmolStr>,
 }
 
 // Store extension function call styles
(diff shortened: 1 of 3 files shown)


cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -75,14 +79,27 @@ lazy_static::lazy_static! {
...
-            CallStyle::MethodStyle => methods.insert(func.name().basename()),
+            CallStyle::FunctionStyle => {
+                let name = func.name();
+                functions.insert(name);
+                if name.is_unqualified() {
+                    unqualified_functions.insert(name.basename_as_ref());
+                }
+            }
+            CallStyle::MethodStyle => {
+                methods.insert(func.name().basename());
```
john-h-kastner-aws, line 94: Might be nice to add an assertion here that `func.name().is_unqualified()`

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -1605,6 +1622,35 @@ impl Node<Option<cst::Name>> {
...
+    }
+}
+
+/// If this [`ast::Id`] is a known extension function/method name or not
+pub(crate) fn is_known_extension_func_id(id: &ast::Id) -> bool {
+    EXTENSION_STYLES.unqualified_functions.contains(id) || is_known_extension_method_id(id.clone())
+}
+
+/// If this [`SmolStr`] is a known extension function/method name or not
+pub(crate) fn is_known_extension_func_str(s: SmolStr) -> bool {
```
john-h-kastner-aws, line 1645: I don't particularly like this function since it's not clear if it will treat `s` as an `Id` or a `Name`. It might be better to force callers to do the conversion to `ast::Id`, or maybe it's enough to make it explicit in the name

john-h-kastner-aws (approved):

cedar-policy-core/src/parser/cst_to_ast.rs
cdisselkoen: I considered forcing the caller to do the conversion, but the use of `ast::Id::unchecked` is justified based on internals in this file, so I thought it might be better to put the use in this file as well.  However, I'm happy to move it back to the caller, with an appropriate comment.

cedar-policy-core/src/parser/cst_to_ast.rs
cdisselkoen: Added

cedar-policy-core/src/parser/cst_to_ast.rs
cdisselkoen: Actually a better idea: let's just keep a hashset of the string versions of all extension function/method names in `EXTENSION_STYLES`

cdisselkoen:
Updated performance numbers after my latest push: now a 14% improvement on the `deeply_nested_est` benchmark from before this PR to after this PR

shaobo-he-aws (approved):

cedar-policy-core/src/parser/cst_to_ast.rs
john-h-kastner-aws: looks good, thanks

PR #1357, opened 2024-12-05
Improve error messages for request validation errors

### LINKED ISSUE #998: Improve error messages for `RequestValidationError`

#996 added some tests for the error messages produced by request validation, and they could use some polishing.

Suggestions from that review thread:
* The error message when passing an incorrect principal/action/resource is something like ```resource type `Principal` is not valid for `Action::"action"` ```. It should be easy to add help text that recommends a valid principal/action/resource.
* When passing an invalid context, the error message is something like ```context `<first-class record with 1 fields>` is not valid for `Action::"action"` ```. It would be better to print the actual context instead of "<first-class record with 1 fields>". It would be even better to reuse the error types/error text from the entity conformance checks (see `JsonDeserializationError`).

_No response_

### Additional context

_No response_

cdisselkoen:
## Description of changes

Makes the improvements suggested in #998. 

## Issue #, if available

Fixes #998.

### DIFF
--- a/cedar-policy-core/src/ast/value.rs
+++ b/cedar-policy-core/src/ast/value.rs
@@ -158,6 +158,19 @@ impl Value {
     pub fn eq_and_same_source_loc(&self, other: &Self) -> bool {
         self == other && self.source_loc() == other.source_loc()
     }
+
+    /// Alternate `Display` impl, that truncates large sets/records (including recursively).
+    ///
+    /// `n`: the maximum number of set elements, or record key-value pairs, that
+    /// will be shown before eliding the rest with `..`.
+    /// `None` means no bound.
+    pub fn bounded_display(
+        &self,
+        f: &mut std::fmt::Formatter<'_>,
+        n: Option<usize>,
+    ) -> std::fmt::Result {
+        self.value.bounded_display(f, n)
+    }
 }
 
 impl ValueKind {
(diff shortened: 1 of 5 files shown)


cedar-policy-core/src/ast/value.rs
```diff
@@ -484,38 +485,49 @@ impl std::fmt::Display for ValueKind {
...
                     }
-                    n => write!(f, "<set with {} elements>", n),
                 }
+                write!(f, "]")?;
+                Ok(())
             }
             Self::Record(record) => {
-                write!(f, "<first-class record with {} fields>", record.len())
+                write!(f, "{{")?;
+                for (i, (k, v)) in record.as_ref().iter().enumerate() {
```
adpaco-aws, line 514: Nit: This looks nicer in the general case, but it's shown as `` `{}` `` in the case with 0 fields. In that case, I'm not sure the help message is so useful (might take a while to realize it's an empty context).

adpaco-aws (approved):

cedar-policy-core/src/ast/value.rs
```diff
@@ -484,38 +485,49 @@ impl std::fmt::Display for ValueKind {
...
+                    // in a deterministic order already
+                    for (i, item) in authoritative.as_ref().iter().enumerate() {
+                        write!(f, "{item}")?;
+                        if i < authoritative.len() - 1 {
+                            write!(f, ", ")?;
                         }
-                        write!(f, "]")?;
-                        Ok(())
                     }
-                    n => write!(f, "<set with {} elements>", n),
```
john-h-kastner-aws, line 514: Should we keep this case for the same reason we applied #887? Similarly should consider adding a case like this for the record branch.

cdisselkoen: A goal for the policy pretty-printer is to produce valid, parseable syntax.  Is a similar goal reasonable here, that the value pretty-printer should produce syntax that is parseable as a Cedar expression?

john-h-kastner-aws: Fair point. 

But, for request validation errors in particular, I don't think it's a reasonable goal. It isn't ideal to have `InvalidContextError` send back the entire context when the context is invalid.

cdisselkoen: Updated the request-validation error message to truncate after 5 key-value pairs in the `Context` record.  Left the `Display` impl for `Value` as printing valid Cedar syntax.  Is that a good compromise between the competing goals here?

cdisselkoen: Refactored to new `bounded_display()` method on `Value` and `ValueKind`, which shares code with their `Display` impls.  Addresses bounding the display of large sets and records, including recursively; doesn't address large literals or extension values (?)

john-h-kastner-aws: Seems good enough for now. To me it feels more likely that the context record will be very large in the normal operation of an application than very large individual strings. If this turns about to be wrong we can truncate these as well.

cedar-policy-validator/src/coreschema.rs
```diff
@@ -389,11 +445,16 @@ pub mod request_validation_errors {
...
+        /// Resource types which actually are valid for that action
+        pub fn valid_resource_tys(&self) -> impl Iterator<Item = &ast::EntityType> {
+            self.valid_resource_tys.iter()
+        }
     }
 
     /// Context does not comply with the shape specified for the request action
     #[derive(Debug, Error, Diagnostic)]
-    #[error("context `{context}` is not valid for `{action}`")]
+    #[error("context `{}` is not valid for `{action}`", pretty_print(&.context))]
```
john-h-kastner-aws, line 457: this might be a good place to use the new thiserror syntax `#[error(fmt = path::to::myfmt)]` to avoid allocating a string here

john-h-kastner-aws (approved):

cedar-policy-validator/src/coreschema.rs
cdisselkoen: Performance on the error path is not critical, but seems reasonable to make a followup issue for this.  #1371

PR #1326, opened 2024-11-19
Add serialization to JSON/protobuf to CLI

andrewmwells-amazon:
## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [ ] A backwards-compatible change requiring a minor version bump to `cedar-policy` (e.g., addition of a new API).

I confirm that this PR (choose one, and delete the other options):

- [ ] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [ ] Does not require updates because my change does not impact the Cedar formal model or DRT infrastructure.

I confirm that [`docs.cedarpolicy.com`](https://docs.cedarpolicy.com/) (choose one, and delete the other options):

- [ ] Does not require updates because my change does not impact the Cedar language specification.

PR #1359, opened 2024-12-06
bump crate version number to 4.3.0

cdisselkoen:
## Description of changes

Bumps the version number to 4.3.0 (here on `main`).  The `main` branch already contains features that are targeted for 4.3.0, such as the `protobufs` experimental feature.

### DIFF
--- a/cedar-policy-cli/Cargo.toml
+++ b/cedar-policy-cli/Cargo.toml
@@ -13,8 +13,8 @@ repository.workspace = true
 [dependencies]
 cedar-policy = { version = "=4.3.0", path = "../cedar-policy" }
 cedar-policy-formatter = { version = "=4.3.0", path = "../cedar-policy-formatter" }
-cedar-policy-core = { package= "cedar-policy-core", git = "https://github.com/cedar-policy/cedar.git", branch = "main", version = "4.*"}
-cedar-policy-validator = { package= "cedar-policy-validator", git = "https://github.com/cedar-policy/cedar.git", branch = "main", version = "4.*"}
+cedar-policy-core = { version = "=4.3.0", path = "../cedar-policy-core" }
+cedar-policy-validator = { version = "=4.3.0", path = "../cedar-policy-validator" }
 clap = { version = "4", features = ["derive", "env"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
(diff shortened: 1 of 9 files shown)


cedar-policy-cli/Cargo.toml
```diff
@@ -11,10 +11,10 @@ homepage.workspace = true
 repository.workspace = true
 
 [dependencies]
-cedar-policy = { version = "=4.1.0", path = "../cedar-policy" }
-cedar-policy-formatter = { version = "=4.1.0", path = "../cedar-policy-formatter" }
-cedar-policy-core = { package= "cedar-policy-core", git = "https://github.com/cedar-policy/cedar.git", branch = "main", version = "=4.1.0"}
-cedar-policy-validator = { package= "cedar-policy-validator", git = "https://github.com/cedar-policy/cedar.git", branch = "main", version = "=4.1.0"}
+cedar-policy = { version = "=4.3.0", path = "../cedar-policy" }
+cedar-policy-formatter = { version = "=4.3.0", path = "../cedar-policy-formatter" }
+cedar-policy-core = { package= "cedar-policy-core", git = "https://github.com/cedar-policy/cedar.git", branch = "main", version = "4.*"}
```
john-h-kastner-aws, line 16: Why does this change to `4.*`? (Also, why does the cli need to depend on core at all)

john-h-kastner-aws (approved):

cedar-policy-cli/Cargo.toml
cdisselkoen: It can't say 4.3.0 because current git `main` is not 4.3.0.  An alternate plan would be to move this from a git to a path dependency, like the dependency on `cedar-policy`.

CLI dependency on core was added in #1326 and is only necessary for the protobufs-related features.  Now that we have #1345, I suspect this is no longer needed, and I'm working on removing it.

john-h-kastner-aws: Perhaps related that Cargo.lock has two entries for `cedar-policy-core` at different commits.

cdisselkoen: Oof. Looking closer, #1326 also added some JSON-related commands that are not guarded by the protobufs experimental flag, and rely on the undocumented JSON serialization of internal structures (which is what Lean expects as of this writing).  Be nice to remove those when Lean switches to accepting protobuf instead, but technically that's a breaking change to functionality not marked as experimental

adpaco-aws: Same goes for `cedar-policy-validator` (one line below). Can we use path dependencies for both?

john-h-kastner-aws: > Be nice to remove those when Lean switches to accepting protobuf instead, but technically that's a breaking change to functionality not marked as experimental

All unrelased code, so please break

cdisselkoen: Done

adpaco-aws (approved):

john-h-kastner-aws (approved):

Issue #1348, opened 2024-12-02
Code cleanups prior to stabilizing protobufs feature


cdisselkoen:
### Describe the improvement you'd like to request

The code for the `protobufs` experimental needs several small cleanups before stabilization.  At a minimum,
- resolving all `PANIC SAFETY` comments that today just explain "it's an experimental feature".  Panics might be acceptable in experimental features but are not for stabilized features.
- resolving the TODO introduced by #1345 (see https://github.com/cedar-policy/cedar/pull/1345#discussion_r1866223460)

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this internal improvement
- [ ] ⚠️ This feature might incur a breaking change

john-h-kastner-aws:
we also need to carefully consider what implementation details the protobuf format effectively makes public. E.g.: https://github.com/cedar-policy/cedar/pull/1345#discussion_r1866364882

The protobuf format is based on our internal structures rather than our external data formats. This is fine for diff testing, but not for public consumption.

john-h-kastner-aws:
Linking #1286 and #1352

john-h-kastner-aws:
Another note https://github.com/cedar-policy/cedar/pull/1366#discussion_r1876723750

john-h-kastner-aws:
Linking another comment https://github.com/cedar-policy/cedar/pull/1506#discussion_r1972378468

current entities protobuf assume TC is already computed. We probably want to include a flag to allow the parser to compute TC

john-h-kastner-aws:
Linking with optimizations suggested in #1815

liamwh:
Hi all,

Great to see this issue, I think this would be a very valuable upgrade for Cedar.

Has there been much progress on this issue lately? This is a feature we're interested in depending on but are hesitant whilst it's still experimental. Is anyone able to give any indication of when this might be stabilised? Is there work actively being done on it?

Thanks and regards,

Liam

cdisselkoen:
This is not at the top of our list currently, but we're happy to accept contributions on the issues mentioned in this thread

john-h-kastner-aws:
Another point for consideration when stablizing:

Do we want the `TryFrom` impls in the public API? E.g., https://docs.rs/cedar-policy/latest/cedar_policy/struct.PolicySet.html#impl-TryFrom%3C%26PolicySet%3E-for-PolicySet

These could be useful for some callers, but exposing them gives us less room to change implementation details later on.

[...]

PR #1366, opened 2024-12-09
Remove unreleased JSON serialization code from the CLI

cdisselkoen:
## Description of changes

Following comment by @john-h-kastner-aws [here](https://github.com/cedar-policy/cedar/pull/1359#discussion_r1873592002).

#1326 added some JSON serialization commands to the CLI that were not guarded by the `protobufs` experimental feature.  These are due to be replaced by protobuf versions.  The code added by #1326 has not yet been released, so we're free to make changes before it is released without worrying about breaking things.  The immediate steps taken in this PR are to:
1. move all of the related code behind the `protobufs` experimental feature, and also to flatten the module while we're here.
2. Remove the JSON-serialization command, leaving just the protobuf-serialization commands.

### DIFF
--- a/cedar-policy-cli/src/lib.rs
+++ b/cedar-policy-cli/src/lib.rs
@@ -118,17 +118,14 @@ pub enum Commands {
     New(NewArgs),
     /// Partially evaluate an authorization request
     PartiallyAuthorize(PartiallyAuthorizeArgs),
-    /// Ouput a JSON file for consumption by Lean
-    #[command(subcommand)]
-    WriteDRTJson(serialization::AnalysisCommands),
     /// Output a protobuf binary file for consumption by Lean
     #[cfg(feature = "protobufs")]
     #[command(subcommand)]
-    WriteDRTProto(serialization::AnalysisCommands),
+    WriteDRTProto(protobufs::AnalysisCommands),
     /// Output a protobuf binary file for consumption by Lean
     #[cfg(feature = "protobufs")]
     #[command(subcommand)]
-    WriteDRTProtoFromJSON(serialization::AnalyzeCommandsFromJson),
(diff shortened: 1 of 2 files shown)


cedar-policy-cli/src/lib.rs
```diff
@@ -118,17 +118,14 @@ pub enum Commands {
...
     /// Partially evaluate an authorization request
     PartiallyAuthorize(PartiallyAuthorizeArgs),
-    /// Ouput a JSON file for consumption by Lean
-    #[command(subcommand)]
-    WriteDRTJson(serialization::AnalysisCommands),
     /// Output a protobuf binary file for consumption by Lean
     #[cfg(feature = "protobufs")]
     #[command(subcommand)]
-    WriteDRTProto(serialization::AnalysisCommands),
+    WriteDRTProto(protobufs::AnalysisCommands),
```
john-h-kastner-aws, line 124: I guess I don't really know what protofbuf functionality was added to the CLI in the frist place, but how does the variant name (`WriteDRTProto`) relate to inner struct (`AnalysisCommands`)? Also, is this protobuf format still DRT specific, or should we rename this to something more general before release?

john-h-kastner-aws (approved):

cedar-policy-cli/src/lib.rs
cdisselkoen: The naming was done by #1326, I think the relationship is that `AnalysisCommands` is a list of the protobuf formats supported by `WriteDRTProto` (today, just `Equivalence`), all of which relate to analysis (at least today).  I agree this should be changed prior to release.  Let's track that in #1348; I see you've already linked this comment there

adpaco-aws (approved):

PR #1373, opened 2024-12-12
introduce `BoundedDisplay` trait and properly bound error message

cdisselkoen:
## Description of changes

In #1357, we introduced a `bounded_display()` method on `Value` and `ValueKind`, but we failed to actually use it in the request-validation error message that motivated it.  This PR finishes that job.  It creates a `BoundedDisplay` trait rather than method, due to the number of types that need a bounded-display operation.  And, it uses `BoundedDisplay` in the request-validation error message that motivated this work.

Also changes one `HashMap` to `BTreeMap` internally, to aid in deterministic error messages.

Adds some tests for bounded-display functionality for ESTs.

### DIFF
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -772,8 +772,8 @@ impl<T: Clone> std::fmt::Display for Expr<T> {
 
 impl<T: Clone> BoundedDisplay for Expr<T> {
     fn fmt(&self, f: &mut impl std::fmt::Write, n: Option<usize>) -> std::fmt::Result {
-        // Like for the `std::fmt::Display` impl, we convert to EST and use the
-        // EST pretty-printer. Note that converting AST->EST is lossless and infallible.
+        // Like the `std::fmt::Display` impl, we convert to EST and use the EST
+        // pretty-printer. Note that converting AST->EST is lossless and infallible.
         BoundedDisplay::fmt(&crate::est::Expr::from(self.clone()), f, n)
     }
 }
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -770,6 +770,14 @@ impl<T: Clone> std::fmt::Display for Expr<T> {
     }
 }
 
+impl<T: Clone> BoundedDisplay for Expr<T> {
+    fn fmt(&self, f: &mut impl std::fmt::Write, n: Option<usize>) -> std::fmt::Result {
+        // Like for the `std::fmt::Display` impl, we convert to EST and use the
+        // EST pretty-printer. Note that converting AST->EST is lossless and infallible.
+        BoundedDisplay::fmt(&crate::est::Expr::from(self.clone()), f, n)
+    }
+}
+
 impl std::str::FromStr for Expr {
     type Err = ParseErrors;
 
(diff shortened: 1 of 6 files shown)


john-h-kastner-aws (approved):

cedar-policy-core/src/ast/expr.rs
```diff
@@ -770,6 +770,14 @@ impl<T: Clone> std::fmt::Display for Expr<T> {
     }
 }
 
+impl<T: Clone> BoundedDisplay for Expr<T> {
+    fn fmt(&self, f: &mut impl std::fmt::Write, n: Option<usize>) -> std::fmt::Result {
+        // Like for the `std::fmt::Display` impl, we convert to EST and use the
```
shaobo-he-aws, line 775: `Like for` -> `Like`

cedar-policy-validator/src/coreschema.rs
```diff
@@ -473,43 +473,6 @@ pub mod request_validation_errors {
             &self.action
         }
     }
-
-    const MAX_KEYS_TO_PRETTY_PRINT: usize = 5;
```
shaobo-he-aws, line 477: Maybe keep this constant so that we don't have a magic number here.

cedar-policy-core/src/est/expr.rs
```diff
@@ -1705,40 +1718,71 @@ fn display_cedarvaluejson(f: &mut std::fmt::Formatter<'_>, v: &CedarValueJson) -
...
                 }
             }
         }
         CedarValueJson::Set(v) => {
-            write!(f, "[")?;
-            for (i, val) in v.iter().enumerate() {
-                display_cedarvaluejson(f, val)?;
-                if i < (v.len() - 1) {
-                    write!(f, ", ")?;
+            match n {
```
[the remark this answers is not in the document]
cdisselkoen: Tried this, but we need the unwrapped value of `n` inside the `if` (for `.take(n)`), so I think this match works better

cedar-policy-core/src/ast/value.rs
```diff
@@ -556,6 +542,48 @@ impl StaticallyTyped for ValueKind {
...
+///
+/// `n`: the maximum number of set elements, or record key-value pairs, that
+/// will be shown before eliding the rest with `..`.
+/// `None` means no bound.
+///
+/// Intended for error messages, to avoid very large/long error messages.
+pub trait BoundedDisplay {
+    /// Write `self` to the writer `f`, truncating set elements or key-value
+    /// pairs if necessary based on `n`
+    fn fmt(&self, f: &mut impl std::fmt::Write, n: Option<usize>) -> std::fmt::Result;
```
shaobo-he-aws, line 556: Nit: we can add two methods with default implementations `fmt_bounded` and `fmt_unbounded` so that users don't have to add `Some` or `None` as an argument.

cedar-policy-core/src/ast/value.rs
```diff
@@ -556,6 +542,48 @@ impl StaticallyTyped for ValueKind {
...
+///
+/// `n`: the maximum number of set elements, or record key-value pairs, that
+/// will be shown before eliding the rest with `..`.
+/// `None` means no bound.
+///
+/// Intended for error messages, to avoid very large/long error messages.
+pub trait BoundedToString {
+    /// Convert `self` to a `String`, truncating set elements or key-value pairs
+    /// if necessary based on `n`
+    fn to_string(&self, n: Option<usize>) -> String;
```
shaobo-he-aws, line 570: Same as above.

cedar-policy-core/src/ast/value.rs
cdisselkoen: Sure done.

cedar-policy-core/src/ast/value.rs
cdisselkoen: Sure, done.

shaobo-he-aws (approved):

PR #1376, opened 2024-12-16
Apply clippy auto-fixes

shaobo-he-aws:
## Description of changes

Apply clippy fixes to test code. Also disabled `cognitive_complexity` warnings.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,8 +55,6 @@ needless_doctest_main = "allow"
 # see #878
 result_large_err = "allow"
 large_enum_variant = "allow"
-cognitive_complexity = "allow"
-
 
 redundant_clone = "deny"
 # Error on potential panics
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,8 @@ needless_doctest_main = "allow"
 # see #878
 result_large_err = "allow"
 large_enum_variant = "allow"
+cognitive_complexity = "allow"
+
 
 redundant_clone = "deny"
 # Error on potential panics
--- a/cedar-policy-cli/tests/sample.rs
+++ b/cedar-policy-cli/tests/sample.rs
@@ -792,9 +792,7 @@ fn test_link_samples() {
         &linked_file_name,
         "AccessVacation",
         "AliceAccess",
-        [(SlotId::principal(), "User::\"alice\"".to_string())]
-            .into_iter()
-            .collect(),
+        std::iter::once((SlotId::principal(), "User::\"alice\"".to_string())).collect(),
         CedarExitCode::Failure,
     );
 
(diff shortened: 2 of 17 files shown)


Cargo.toml
```diff
@@ -55,6 +55,8 @@ needless_doctest_main = "allow"
 # see #878
 result_large_err = "allow"
 large_enum_variant = "allow"
+cognitive_complexity = "allow"
```
john-h-kastner-aws, line 58: how often was this showing up? If it's only one or two place I might  prefer allowing it an the individual functions since IMO it's nice that clippy calls out complex functions even if we decide against refactoring.

cedar-policy-cli/tests/sample.rs
```diff
@@ -792,9 +792,7 @@ fn test_link_samples() {
         &linked_file_name,
         "AccessVacation",
         "AliceAccess",
-        [(SlotId::principal(), "User::\"alice\"".to_string())]
-            .into_iter()
-            .collect(),
+        std::iter::once((SlotId::principal(), "User::\"alice\"".to_string())).collect(),
```
john-h-kastner-aws, line 795: Here and else where: I think these could be `HashMap::from([...])` instead going through an iterator. Or maybe `run_link_test` could just take an `impl Into<HashMap<_,_>>`

cedar-policy-core/src/entities.rs
```diff
@@ -1750,9 +1750,8 @@ mod json_parsing_tests {
             roundtrip(&empty_entities).expect("should roundtrip without errors")
         );
 
-        let (e0, e1, e2, e3) = test_entities();
         let entities = Entities::from_entities(
-            [e0, e1, e2, e3],
+            <[Entity; 4]>::from(test_entities()),
```
john-h-kastner-aws, line 1754: could `test_entities` return `[Entity; 4]` directly?

john-h-kastner-aws (approved):

Cargo.toml
shaobo-he-aws: There are a few places now. I will annotate them individually.

cedar-policy-core/src/entities.rs
shaobo-he-aws: Done.

Cargo.toml
shaobo-he-aws: Done.

cedar-policy/src/tests.rs
```diff
@@ -17,6 +17,7 @@
 #![cfg(test)]
 // PANIC SAFETY unit tests
 #![allow(clippy::panic)]
+#![allow(clippy::cognitive_complexity, clippy::too_many_lines)]
```
john-h-kastner-aws, line 20: yeah, these lints are definitely bogus in test code.

When I run clippy,  I never see lints in `#![cfg(test)]` code (though running with `--fix` does fix lints in it). Is there something different about how you're running it?

john-h-kastner-aws (approved):

cedar-policy/src/tests.rs
shaobo-he-aws: I always enable `--fix`.

cdisselkoen (approved):

PR #1387, opened 2024-12-20
Improve substitution

B-Lorentz:
## Description of changes

Adds a new function `reauthorize_with_iter` to `PartialResponse`, that takes an `IntoIterator` instead of the `HashMap<SmolString, RestrictedExpression>`. This has two advantages:
- The SmolStr, which is an implementation detail of the cedar-policy-core crate is not exposed to the callers, instead &str is used. This in in line with the apparent policy of the api in cedar-policy.
- `reauthorize` immediately consumes the hashmap, the new api removes the need to allocate it if the remappings are  already available in some other format.

I'm considering to add one that takes a Fn(&str) -> Option<Value> too, that one would remove the need to iterate them all.
In fact the whole substitution infrastructure in -core could be using such a function, but the current reauthorize API is not (easily) implementable in those terms, because of the fallibility of conversion from RestrictedExpression to Value.

However, I believe the most common case will be when the caller provides literal types in the substitutions: If we make a specialized method that takes only valid values, the need to evaluate the 'restricted expression' disappears.
But I suppose ast::Value shoudn't be exposed in the public API, instead EvalResult or a similar wrapper would be needed?

### DIFF
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -1057,20 +1057,20 @@ impl PartialResponse {
     }
 
     /// Attempt to re-authorize this response given a mapping from unknowns to values.
-    /// Consider ``reauthorize_with_iter`` if mappings are not naturally in a hashmap.
     #[allow(clippy::needless_pass_by_value)]
+    #[deprecated = "use reauthorize_with_bindings"]
     pub fn reauthorize(
         &self,
         mapping: HashMap<SmolStr, RestrictedExpression>,
         auth: &Authorizer,
         es: &Entities,
     ) -> Result<Self, ReauthorizationError> {
-        self.reauthorize_with_iter(mapping.iter().map(|(k, v)| (k.as_str(), v)), auth, es)
+        self.reauthorize_with_bindings(mapping.iter().map(|(k, v)| (k.as_str(), v)), auth, es)
     }
 
     /// Attempt to re-authorize this response given a mapping from unknowns to values, provided as an iterator.
     /// Exhausts the iterator, returning any evaluation errors in the restricted expressions, regardless whether there is a matching unknown.
-    pub fn reauthorize_with_iter<'m>(
+    pub fn reauthorize_with_bindings<'m>(
         &self,
         mapping: impl IntoIterator<Item = (&'m str, &'m RestrictedExpression)>,
         auth: &Authorizer,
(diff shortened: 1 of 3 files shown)


shaobo-he-aws:
LGTM. Nit: I prefer the new API to be named like `reauthorize_with_bindings`.

shaobo-he-aws (approved):

B-Lorentz:
> LGTM. Nit: I prefer the new API to be named like `reauthorize_with_bindings`.
I can do that. `with_iter` seems to suggest some 'deeper' reliance on iteration than it being just the form the bindings are supplied. But I think `with_bindings` is also suboptimal, since the old `reauthorize` also uses bindings just as much, just takes them as a map. Can we deprecate the old method?

shaobo-he-aws:
> Can we deprecate the old method?

Yes, we theoretically can. You can mark it and other team members can weigh in when they review it.

cedar-policy/src/api.rs
```diff
@@ -1056,12 +1056,25 @@ impl PartialResponse {
...
         mapping: HashMap<SmolStr, RestrictedExpression>,
         auth: &Authorizer,
         es: &Entities,
+    ) -> Result<Self, ReauthorizationError> {
+        self.reauthorize_with_bindings(mapping.iter().map(|(k, v)| (k.as_str(), v)), auth, es)
+    }
+
+    /// Attempt to re-authorize this response given a mapping from unknowns to values, provided as an iterator.
+    /// Exhausts the iterator, returning any evaluation errors in the restricted expressions, regardless whether there is a matching unknown.
+    pub fn reauthorize_with_bindings<'m>(
```
john-h-kastner-aws, line 1073: Does it need this explicit lifetime?

john-h-kastner-aws (approved):
Looks good. One comment.

cedar-policy/src/api.rs
B-Lorentz: Without it, I get an error:
```
anonymous lifetimes in `impl Trait` are unstable
expected named lifetime parameterrustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic message [2]?2#file:///home/lorinc/github.com/B-Lorentz/cedar/cedar-policy/src/api.rs)
api.rs(1075, 51): consider introducing a named lifetime parameter: `'a `, `<'a>`
```

PR #1389, opened 2024-12-20
Add APIs to get schema annotations

shaobo-he-aws:

### DIFF
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -1371,7 +1371,9 @@ fn annotations_to_pairs(annotations: &est::Annotations) -> impl Iterator<Item = 
 
 impl SchemaFragment {
     /// Get annotations of a non-empty namespace.
-    /// We do not allow namespace-level annotations on the empty namespace
+    ///
+    /// We do not allow namespace-level annotations on the empty namespace.
+    ///
     /// Returns `None` if `namespace` is not found in the [`SchemaFragment`]
     pub fn namespace_annotations(
         &self,
@@ -1384,6 +1386,7 @@ impl SchemaFragment {
     }
 
     /// Get annotations of a common type declaration
+    ///
     /// Returns `None` if `namespace` is not found in the [`SchemaFragment`] or
     /// `ty` is not a valid common type ID or `ty` is not found in the
     /// corresponding namespace definition
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -1421,7 +1421,7 @@ impl SchemaFragment {
     /// Get annotations of an action declaration
     /// Returns `None` if `namespace` is not found in the [`SchemaFragment`] or
     /// `id` is not found in the corresponding namespace definition
-    pub fn action_type_annotations(
+    pub fn action_annotations(
         &self,
         namespace: Option<EntityNamespace>,
         id: EntityId,
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -30,6 +30,7 @@ use cedar_policy_validator::entity_manifest;
 pub use cedar_policy_validator::entity_manifest::{
     AccessTrie, EntityManifest, EntityRoot, Fields, RootAccessTrie,
 };
+use cedar_policy_validator::json_schema;
 use cedar_policy_validator::typecheck::{PolicyCheck, Typechecker};
 pub use id::*;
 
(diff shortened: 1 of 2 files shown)


cedar-policy/src/api.rs
```diff
@@ -1361,7 +1362,77 @@ pub struct SchemaFragment {
...
 
+fn annotations_to_pairs(annotations: &est::Annotations) -> impl Iterator<Item = (&str, &str)> {
+    annotations
+        .0
+        .iter()
+        .map(|(key, value)| (key.as_ref(), value.as_ref().map_or("", |a| a.as_ref())))
+}
+
 impl SchemaFragment {
+    /// Get annotations of a non-empty namespace
```
john-h-kastner-aws, line 1373: Need periods here and elsewhere so docs render properly

```suggestion
    /// Get annotations of a non-empty namespace.
```

cedar-policy/src/api.rs
```diff
@@ -1361,7 +1362,77 @@ pub struct SchemaFragment {
...
+        self.lossless
+            .0
+            .get(&Some(namespace.0))
+            .map(|ns_def| annotations_to_pairs(&ns_def.annotations))
+    }
+
+    /// Get annotations of a common type declaration
+    /// Returns `None` if `namespace` is not found in the [`SchemaFragment`] or
+    /// `ty` is not a valid common type ID or `ty` is not found in the
+    /// corresponding namespace definition
```
john-h-kastner-aws, line 1389: overloading the `None` case here isn't great. Maybe we could publicly expose an `Id` data type so that error can be reported sooner? Though, using that in other places in the API would probably be a breaking change, so perhaps it's better to stick with this signature for now

cedar-policy/src/api.rs
```diff
@@ -1361,7 +1362,77 @@ pub struct SchemaFragment {
...
+        ns_def
+            .entity_types
+            .get(&ty)
+            .map(|ty| annotations_to_pairs(&ty.annotations))
+    }
+
+    /// Get annotations of an action declaration
+    /// Returns `None` if `namespace` is not found in the [`SchemaFragment`] or
+    /// `id` is not found in the corresponding namespace definition
+    pub fn action_type_annotations(
```
john-h-kastner-aws, line 1424: `action_annotations` would IMO be a better name

john-h-kastner-aws (commented):
Also, this doesn't add an API for getting annotations on entity/context attributes. That's probably a more complicated API due to the recursive structure of types, but can you open a issue so we don't forget about it if  you don't plan on doing it in this PR?

cedar-policy/src/api.rs
shaobo-he-aws: Done.

shaobo-he-aws:
> Also, this doesn't add an API for getting annotations on entity/context attributes. That's probably a more complicated API due to the recursive structure of types, but can you open a issue so we don't forget about it if you don't plan on doing it in this PR?

Good idea. Issue: https://github.com/cedar-policy/cedar/issues/1404

cedar-policy/src/api.rs
```diff
@@ -1361,7 +1362,77 @@ pub struct SchemaFragment {
...
+    annotations
+        .0
+        .iter()
+        .map(|(key, value)| (key.as_ref(), value.as_ref().map_or("", |a| a.as_ref())))
+}
+
 impl SchemaFragment {
+    /// Get annotations of a non-empty namespace.
+    /// We do not allow namespace-level annotations on the empty namespace
+    /// Returns `None` if `namespace` is not found in the [`SchemaFragment`]
```
cdisselkoen, line 1375: for better Markdown rendering
```suggestion
    /// Get annotations of a non-empty namespace.
    ///
    /// We do not allow namespace-level annotations on the empty namespace.
    ///
    /// Returns `None` if `namespace` is not found in the [`SchemaFragment`]
```

cedar-policy/src/api.rs
```diff
@@ -1361,7 +1362,77 @@ pub struct SchemaFragment {
...
+        namespace: EntityNamespace,
+    ) -> Option<impl Iterator<Item = (&str, &str)>> {
+        self.lossless
+            .0
+            .get(&Some(namespace.0))
+            .map(|ns_def| annotations_to_pairs(&ns_def.annotations))
+    }
+
+    /// Get annotations of a common type declaration
+    /// Returns `None` if `namespace` is not found in the [`SchemaFragment`] or
```
cdisselkoen, line 1387: for better Markdown rendering
```suggestion
    /// Get annotations of a common type declaration
    ///
    /// Returns `None` if `namespace` is not found in the [`SchemaFragment`] or
```

cedar-policy/src/api.rs
```diff
@@ -1361,7 +1362,77 @@ pub struct SchemaFragment {
...
+}
+
 impl SchemaFragment {
+    /// Get annotations of a non-empty namespace.
+    /// We do not allow namespace-level annotations on the empty namespace
+    /// Returns `None` if `namespace` is not found in the [`SchemaFragment`]
+    pub fn namespace_annotations(
+        &self,
+        namespace: EntityNamespace,
+    ) -> Option<impl Iterator<Item = (&str, &str)>> {
```
cdisselkoen, line 1379: Having all of these APIs return `Iterator<Item = (&str, &str)>` is good, but I wonder if we need an additional API to get the value of an annotation with a user-specified key.  For instance if I want to specifically look up the value of the `foo` annotation on a particular namespace.  As is, I'd be forced to iterate through the entire iterator looking for the key `foo`.  We could solve this either by adding a separate API `namespace_annotation(&self, namespace, key)` or perhaps by returning a map structure (`HashMap` or `BTreeMap`) instead of an iterator in this API, so that users could do their own lookups.

Same for the other new annotation-getting APIs.

shaobo-he-aws: Good point. I think that having an iterator is a generic. That is, users can collect them into a `HashMap` or `BTreeMap` given their preferences (e.g., performance vs order). I also think it may be a good idea to let user collect the iterator and get the value by key, instead of having APIs. Maybe I'm too paranoid about breaking changes.

cdisselkoen: I guess I'm concerned that having the user collect the iterator in order to get the value by key is

1) a little complicated, compared to having a direct method for it -- so, creates a suboptimal user experience
2) potentially poor performance, as getting any value by key at all requires collecting the entire iterator

If we're worried about breaking changes, we don't have to return a map structure, we can just implement our own getter, something like:

```rust
pub fn namespace_annotation(
    &self,
    namespace: EntityNamespace,
    annotation_key: &str,
) -> Option<&str> { ... }
```
that gets a particular annotation value given a key.  I don't think adding a method like this is overly committing to any particular implementation details

cedar-policy/src/api.rs
```diff
@@ -1361,7 +1362,80 @@ pub struct SchemaFragment {
...
+        let ty = json_schema::CommonTypeId::new(ast::UnreservedId::from_normalized_str(ty).ok()?)
+            .ok()?;
+        ns_def
+            .common_types
+            .get(&ty)
+            .map(|ty| annotations_to_pairs(&ty.annotations))
+    }
+
+    /// Get annotations of an entity type declaration
+    /// Returns `None` if `namespace` is not found in the [`SchemaFragment`] or
```
john-h-kastner-aws, line 1408:
```suggestion
    /// Get annotations of an entity type declaration
    ///
    /// Returns `None` if `namespace` is not found in the [`SchemaFragment`] or
```

cedar-policy/src/api.rs
```diff
@@ -1361,7 +1362,80 @@ pub struct SchemaFragment {
...
+        let ty = ast::UnreservedId::from_normalized_str(ty).ok()?;
+        ns_def
+            .entity_types
+            .get(&ty)
+            .map(|ty| annotations_to_pairs(&ty.annotations))
+    }
+
+    /// Get annotations of an action declaration
+    /// Returns `None` if `namespace` is not found in the [`SchemaFragment`] or
+    /// `id` is not found in the corresponding namespace definition
```
john-h-kastner-aws, line 1426:
```suggestion
    /// Get annotations of an action declaration
    ///
    /// Returns `None` if `namespace` is not found in the [`SchemaFragment`] or
    /// `id` is not found in the corresponding namespace definition
```

john-h-kastner-aws (approved):
Approving since the code looks good. I agree with craig that we should provide direct accessor methods also/instead of the iterators

cedar-policy/src/api.rs
shaobo-he-aws: Sure. I can do it.

shaobo-he-aws:
> Approving since the code looks good to merge as is. I agree with craig that we should provide direct accessor methods also/instead of the iterators

Thanks for the review. I'll add these APIs.

cdisselkoen (approved):

PR #1391, opened 2024-12-23
Allow typed unknown resource and principal

B-Lorentz:
## Description of changes

Allows the caller to create a request with an unknown principal or resource, that is still known to belong to a certain type.

I had to modify the evaluator a little bit, to take the type annotation into account for 'is' expressions. I know this is enroaching on the open rfc of https://github.com/cedar-policy/rfcs/pull/83/files
but I found it to be very useful for a 'query construction' usecase, similar to the one outlined here:
https://cedarland.blog/usage/partial-evaluation/content.html#what-can-alice-access

Without this, the only way to remove from the residual policies which are clearly excluded by the scope is to construct a full dummy entity ID of the right type, and then insert into the entity store an entry with that id.
But this requires all the fields of the entity to be explicitly set to 'unknown', and makes the evaluator assume the entity has no parents (since the parents of a concrete entity can't be unknown atm)

## Issue #, if available

https://github.com/cedar-policy/cedar/issues/1393

(I guess partial-eval is not part of the formal spec, right)

### DIFF
--- a/cedar-policy-core/src/evaluator.rs
+++ b/cedar-policy-core/src/evaluator.rs
@@ -6399,10 +6399,7 @@ pub(crate) mod test {
         assert_eq!(r, Either::Left(Value::from(true)));
 
         // Two differently typed unknowns should not be equal
-        let e = Expr::is_eq(
-            Expr::var(Var::Principal),
-            Expr::var(Var::Resource),
-        );
+        let e = Expr::is_eq(Expr::var(Var::Principal), Expr::var(Var::Resource));
         let r = eval.partial_eval_expr(&e).unwrap();
         assert_eq!(r, Either::Left(Value::from(false)));
     }
--- a/cedar-policy-core/src/evaluator.rs
+++ b/cedar-policy-core/src/evaluator.rs
@@ -673,6 +673,16 @@ impl<'e> Evaluator<'e> {
                         Ok((v.get_as_entity()?.entity_type() == entity_type).into())
                     }
                     PartialValue::Residual(r) => {
+                        if let ExprKind::Unknown(Unknown {
+                            type_annotation:
+                                Some(Type::Entity {
+                                    ty: type_of_unknown,
+                                }),
+                            ..
+                        }) = r.expr_kind()
+                        {
+                            return Ok((type_of_unknown == entity_type).into());
+                        }
                         Ok(Expr::is_entity_type(r, entity_type.clone()).into())
                     }
                 }
@@ -6285,5 +6295,20 @@ pub(crate) mod test {
             "#).unwrap()), Err(EvaluationError::RecordAttrDoesNotExist(err)) => {
             assert_eq!(err.attr, "d");
         });
+    }
+
+    #[test]
+    fn typed_unknown_entity_id() {
+        let mut q = basic_request();
+        let entities = basic_entities();
+        q.principal = EntityUIDEntry::UnknownOfType {
+            ty: EntityType::from_str("different_test_type").expect("must parse"),
+            loc: None,
+        };
+        let eval = Evaluator::new(q, &entities, Extensions::none());
+
+        let e = Expr::is_entity_type(Expr::var(Var::Principal), EntityUID::test_entity_type());
+        let r = eval.partial_eval_expr(&e).unwrap();
+        assert_eq!(r, Either::Left(Value::from(false)));
     }
 }
--- a/cedar-policy-core/src/evaluator.rs
+++ b/cedar-policy-core/src/evaluator.rs
@@ -399,7 +399,10 @@ impl<'e> Evaluator<'e> {
                         return Ok(PartialValue::Residual(Expr::binary_app(*op, e1, v2.into())));
                     }
                     (PartialValue::Residual(e1), PartialValue::Residual(e2)) => {
-                        return Ok(PartialValue::Residual(Expr::binary_app(*op, e1, e2)))
+                        if let Some(val) = self.short_circuit_two_typed_residuals(&e1, &e2, *op) {
+                            return Ok(val);
+                        }
+                        return Ok(PartialValue::Residual(Expr::binary_app(*op, e1, e2)));
                     }
                 };
                 match op {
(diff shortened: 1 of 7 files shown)


cedar-policy-core/src/evaluator.rs
```diff
@@ -386,10 +386,16 @@ impl<'e> Evaluator<'e> {
                 ) {
                     (PartialValue::Value(v1), PartialValue::Value(v2)) => (v1, v2),
                     (PartialValue::Value(v1), PartialValue::Residual(e2)) => {
-                        return Ok(PartialValue::Residual(Expr::binary_app(*op, v1.into(), e2)))
+                        if let Some(val) = self.short_circuit_typed_residual(&v1, &e2, *op) {
```
cdisselkoen, line 389: Could update the comments on lines 380-382 since this does now partially evaluate (some) binary operators and take advantage of the type information in unknowns

cedar-policy/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ Cedar Language Version: TBD
 - Added protobuf and JSON generation code to `cedar-policy-cli`.
 - Added a new get helper method to Context that allows easy extraction of generic values from the context by key. This method simplifies the common use case of retrieving values from Context objects.
 - Implemented [RFC 62 (extended `has` operator)](https://github.com/cedar-policy/rfcs/blob/main/text/0062-extended-has.md)  (#1327, resolving #1329)
+- Added unknown_*_with_type methods to the RequestBuilder in partial-eval, allowing an unknown pricipal, resource or action to be constrained to a certain entity type.
```
cdisselkoen, line 24:
```suggestion
- Added unknown_*_with_type methods to the RequestBuilder in partial-eval, allowing an unknown principal, resource or action to be constrained to a certain entity type.
```

cdisselkoen (approved):
Thanks for this!  Very nice code.  Appreciate the helpful incidental refactors too.

cedar-policy-core/src/evaluator.rs
B-Lorentz: Thanks, makes sense

cedar-policy-core/src/evaluator.rs
```diff
@@ -379,17 +379,24 @@ impl<'e> Evaluator<'e> {
...
+                //   Operators are not partially evaluated, except in a few 'simple' cases when comparing a concrete value with an unknown of known type
+                //   implemented in short_circuit_typed_residual
                 let (arg1, arg2) = match (
                     self.partial_interpret(arg1, slots)?,
                     self.partial_interpret(arg2, slots)?,
                 ) {
                     (PartialValue::Value(v1), PartialValue::Value(v2)) => (v1, v2),
                     (PartialValue::Value(v1), PartialValue::Residual(e2)) => {
-                        return Ok(PartialValue::Residual(Expr::binary_app(*op, v1.into(), e2)))
+                        if let Some(val) = self.short_circuit_typed_residual(&v1, &e2, *op) {
```
shaobo-he-aws, line 390: Nit: IMO, `short_circuit_typed_residual` should be "inlined" here, e.g., by more precise pattern matching.

cedar-policy-core/src/evaluator.rs
```diff
@@ -901,6 +918,35 @@ impl<'e> Evaluator<'e> {
             PartialValue::Residual(r) => Ok(Either::Right(r)),
         }
     }
+
+    fn short_circuit_typed_residual(
```
shaobo-he-aws, line 922: Maybe we can do better here: Two unknowns with different types are not equal.

cedar-policy/src/api.rs
```diff
@@ -3649,6 +3660,17 @@ impl<S> RequestBuilder<S> {
         }
     }
 
+    /// Set the action to be unknown, but known to belong to a certain entity type.
+    ///
+    /// This information is taken into account when evaluating 'is', '==' and '!=' expressions.
+    #[must_use]
+    pub fn unknown_action_with_type(self, action_type: EntityTypeName) -> Self {
```
shaobo-he-aws, line 3667: Should we be more conservative here: not all `EntityTypeName` are allowed; only those with basename `Action`.

shaobo-he-aws:
It looks reasonable to me. That being said, this PR kinda motivates #209. Specifically, changes of PE should not impact concrete evaluation.

cedar-policy/src/api.rs
B-Lorentz: Well, setting a concrete action just above also doesn't have such a check.

cedar-policy-core/src/evaluator.rs
B-Lorentz: Makes sense, that would be consistent

cedar-policy-core/src/evaluator.rs
B-Lorentz: I could do that.. but it would get quite a but long

B-Lorentz:
> It looks reasonable to me. That being said, this PR kinda motivates #209. Specifically, changes of PE should not impact concrete evaluation.

Should we just hold off this PR, and then rebase it on top of yours?

shaobo-he-aws:
> Should we just hold off this PR, and then rebase it on top of yours?

There's no need to hold off this PR as mine is postponed due to technical issues. I'll discuss this PR with other team members.

shaobo-he-aws:
After discussion, this PR looks reasonable to us and we will accept it once review feedback is addressed.

cedar-policy-core/src/evaluator.rs
B-Lorentz: I tried and promptly rediscovered why I made it a separate method: because expr_kind in Expr is private, and can only be accessed trough the expr_kind() method.
Therefore I can't easily match it precisely in a single match expression (I could call expr_kind in a match guard, but then I would need to extract it twice, once with matches! in the guard, and once in the match body (since you can't have an if-let match guard).

Unless you want me to make the expr_kind field public (or public(crate)), I think we are better of with this solution.

cedar-policy/src/api.rs
B-Lorentz: It feels to me this is more of an useful convention than a fundamental property that will be forever true.
But if you say so, I can implement the check, or even remove this method, as it's dramatically less useful than unknown typed resource and principal

shaobo-he-aws:
> Thank you. However I will need some guidance on this one: [#1391 (comment)](https://github.com/cedar-policy/cedar/pull/1391#discussion_r1904608194) I can't see a clean way to inline it into the top-level match due to needing to look at some private fields that are only accessible by a helper method.

I'm fine with not inlining it.

> Also, it seems my semver-breaking the AST broke the DRT build. Is that okay? I will just make a followup PR to the repo to fix it?

Yes, it's Ok. Please submit a PR to cedar-spec to fix it.

shaobo-he-aws (approved):

cedar-policy-core/src/evaluator.rs
```diff
@@ -379,20 +379,30 @@ impl<'e> Evaluator<'e> {
...
                     (PartialValue::Value(v1), PartialValue::Residual(e2)) => {
-                        return Ok(PartialValue::Residual(Expr::binary_app(*op, v1.into(), e2)))
+                        if let Some(val) = self.short_circuit_typed_residual(&v1, &e2, *op) {
+                            return Ok(val);
+                        }
+                        return Ok(PartialValue::Residual(Expr::binary_app(*op, v1.into(), e2)));
                     }
                     (PartialValue::Residual(e1), PartialValue::Value(v2)) => {
-                        return Ok(PartialValue::Residual(Expr::binary_app(*op, e1, v2.into())))
+                        if let Some(val) = self.short_circuit_typed_residual(&v2, &e1, *op) {
```
john-h-kastner-aws, line 396: Swapping the order of `e1` and `v2` in this call is suspicious. Feels likely to introduce bugs if we ever extend partial evaluation to handle non commutative operators. I'd at least like a comment on the definition of `short_circuit_typed_residual` that it may be called this way

B-Lorentz: You are right, as currently written, short_circuit_typed_residual can only handle commutative operations.
I will add a comment

john-h-kastner-aws (approved):

PR #1392, opened 2024-12-23
Decouple concrete evaluator from partial evaluator

shaobo-he-aws:
## Issue #, if available
#209 


cdisselkoen (approved):
This does introduce some duplication between the partial-evaluation and concrete-evaluation code, and I'm vaguely worried that the two could get out of sync if we update one and forget to update the other, but nonetheless I'll approve this, it seems like a reasonable code organization

shaobo-he-aws:
> This does introduce some duplication between the partial-evaluation and concrete-evaluation code, and I'm vaguely worried that the two could get out of sync if we update one and forget to update the other

But isn't it the goal of this PR? That being said, I admit the duplication could be reduced. The major obstacle is that the two evaluators still share the same request and entity store types.

cdisselkoen:
Yes, this PR is doing exactly as #209 requests.  My nervousness is indeed about #209 itself.  Before we had a PR, it was hard to make a compelling argument about duplication or maintainability.  Now that we have a PR (thanks!) we can much more concretely (ha) evaluate whether the duplication is worth the benefits of #209.

shaobo-he-aws:
Closed this PR for now pending prominent decisions to make.

PR #1402, opened 2024-12-31
Add `Entity::new_with_tags()` and `Entity::tag()`

### LINKED ISSUE #1374: Add ability to create an Entity with tags in `Entity::new()`
### Category

The entity creation function has no way to create an entity with tags. I'd like to see tags added as an optional parameter to the `Entity::new()` function, and have them be used in authorization calculations. Additionally, a `tag` function that behaves similarly to the `attr` function, might be warranted.

Another option would be to add an `update_tags` function, however that seems to go against the pattern of immutability I see in the current API for Entity.

### Additional context

The only way, currently, to include tags on the entity is to parse it from JSON.

cdisselkoen:
## Description of changes

Adds public methods `Entity::new_with_tags()` and `Entity::tag()` as discussed in #1374.

Discussion points:
- I made `new_with_tags()` more generic than the existing `Entity::new()`; we can take any `IntoIterator` for the `attrs` and `parents` parameters, there's no reason we need to force callers to have `HashMap` and `HashSet` specifically.  Theoretically I believe we could adjust the `Entity::new()` signature in the same way without breaking semver if we wanted to keep these aligned.
- There's also no real reason we need owned `String` for attribute and tag names; the implementation immediately converts to `SmolStr` anyway, so we could take any `ToSmolStr` for instance.  I haven't made that change in the current version of this PR because I'm not sure how much we want to keep `SmolStr` and related types/functions as an implementation detail.

## Issue #, if available

Fixes #1374

### DIFF
--- a/cedar-policy/CHANGELOG.md
+++ b/cedar-policy/CHANGELOG.md
@@ -17,6 +17,7 @@ Cedar Language Version: TBD
 
 - Implemented [RFC 48 (schema annotations)](https://github.com/cedar-policy/rfcs/blob/main/text/0048-schema-annotations.md)  (#1316)
 - New `.isEmpty()` operator on sets (#1358, resolving #1356)
+- New `Entity::new_with_tags()` and `Entity::tag()` functions (#1402, resolving #1374)
 - Added protobuf schemas and (de)serialization code using on `prost` crate behind the experimental `protobufs` flag.
 - Added protobuf and JSON generation code to `cedar-policy-cli`.
 - Added a new get helper method to Context that allows easy extraction of generic values from the context by key. This method simplifies the common use case of retrieving values from Context objects.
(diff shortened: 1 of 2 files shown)


john-h-kastner-aws (approved):

john-h-kastner-aws:
> I made new_with_tags() more generic than the existing Entity::new(); we can take any IntoIterator for the attrs and parents parameters, there's no reason we need to force callers to have HashMap and HashSet specifically. Theoretically I believe we could adjust the Entity::new() signature in the same way without breaking semver if we wanted to keep these aligned.

I do prefer the new signature. Changing the old signature would probably break type inference in some edges cases is we want to take a strict definition of semver.
 
> There's also no real reason we need owned String for attribute and tag names; the implementation immediately converts to SmolStr anyway, so we could take any ToSmolStr for instance. I haven't made that change in the current version of this PR because I'm not sure how much we want to keep SmolStr and related types/functions as an implementation detail.

No strong opinion of whether SmolStr should be part of our API. A quick search finds one non-experimental function that already exposes it: https://github.com/cedar-policy/cedar/blob/d625809aef63d248e4a2830c581b641e482c0643/cedar-policy/src/api/id.rs#L64

shaobo-he-aws (approved):

cdisselkoen:
I think I'll leave this API using `String` to keep things a little more consistent between `new()` and `new_with_tags()`

PR #1405, opened 2025-01-02
Propagate some source locations from cedar syntax schema parser to JSON syntax structures

### LINKED ISSUE #1084: Carry source info through schema structures
### Category

Internal refactors/changes -- consider "Internal improvement" instead of "Feature request"

#174 requests source-location information for errors in JSON schemas.  This issue has a smaller scope -- it asks that we simply carry source-location information inside the schema JSON structures, and propagate it from the human-syntax as the human-syntax is converted to the JSON structures in `to_json_schema.rs`.  This should be sufficient to get source locations in many human-syntax errors, even without providing source locations for errors in JSON syntax (that is still #174).

This is an optional stepping stone to #174 with reduced scope.

### Additional context

_No response_

cdisselkoen:
## Description of changes

Propagates source locations on the Cedar schema format's `EntityDecl`, `ActionDecl`, and `TypeDecl` to the corresponding structures in `json_schema`.  Partially addresses #1084, although we probably need to propagate source locations on even more types in order to fully resolve #1084. 

## Issue #, if available

Partial resolution of #1084

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -386,6 +386,7 @@ dependencies = [
  "arbitrary",
  "cedar-policy-core",
  "cool_asserts",
+ "educe",
  "itertools 0.13.0",
  "lalrpop",
  "lalrpop-util",
--- a/cedar-policy-validator/Cargo.toml
+++ b/cedar-policy-validator/Cargo.toml
@@ -26,6 +26,7 @@ arbitrary = { version = "1", features = ["derive"], optional = true }
 lalrpop-util = { version = "0.22.0", features = ["lexer", "unicode"] }
 lazy_static = "1.4.0"
 nonempty = "0.10.0"
+educe = "0.6.0"
 
 # wasm dependencies
 serde-wasm-bindgen = { version = "0.6", optional = true }
--- a/cedar-policy-validator/src/cedar_schema/test.rs
+++ b/cedar-policy-validator/src/cedar_schema/test.rs
@@ -343,6 +343,7 @@ mod demo_tests {
             applies_to: None,
             member_of: None,
             annotations: Annotations::new(),
+            loc: None,
         };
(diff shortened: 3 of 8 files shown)


shaobo-he-aws:
A high level question: Should we have a `Node` struct instead? It seems that we will have to add a `loc` field wherever and whenever we want to propagate source locations.

john-h-kastner-aws (approved):
Looks good to me. Maybe add a test case demonstrating the new source locations?

cdisselkoen:
> Should we have a `Node` struct instead?

This is a great idea.  It would mean that we have to change the `serde` representation of `Node` though, in order to not break the existing schema JSON format.  Specifically we'd have to `serde(skip)` the `loc` field and `serde(flatten)` the `node` field.  `Node` is already `Deserialize` and `Serialize`, I wonder where that's currently used

cdisselkoen:
Well the answer is, it's not used anywhere in `cedar` repo, at least `cargo test --all-features` builds and passes even with `Serialize` and `Deserialize` removed from `Node`. So probably breaking the serialized representation of `Node` is OK

cdisselkoen:
Ah, quickly discovered the problem: `Node` has a non-optional `Loc`.  We need an `Option<Loc>` here.  We could create another flavor of `Node` that carries `Option<Loc>`, but is that worth it?

shaobo-he-aws:
> Ah, quickly discovered the problem: `Node` has a non-optional `Loc`. We need an `Option<Loc>` here. We could create another flavor of `Node` that carries `Option<Loc>`, but is that worth it?

I'm fine as it is. We can always refactor the code when it's deemed necessary.

shaobo-he-aws (approved):

john-h-kastner-aws (approved):

PR #1408, opened 2025-01-03
add source locations to a bunch more errors

### LINKED ISSUE #1084: Carry source info through schema structures

cdisselkoen:
## Description of changes

Finishes #1084 by actually adapting the `Diagnostic` implementations of a bunch of errors to take advantage of the now-propagated source locs.

As a side effect of this PR, a number of JSON schema errors now have source locations, although fairly unhelpful, as it's just based on parsing a single string in the JSON.  For instance, an undefined-entity-type-name error for `Foo` now has a source location where the source is just `Foo` and the span is the entire thing.  I view this as neither helpful nor harmful; this PR is still a step towards actual JSON source locations (#174), we just need to populate the source-locs in the types differently.

## Issue #, if available

Resolves #1084

### DIFF
--- a/cedar-policy-validator/src/schema/err.rs
+++ b/cedar-policy-validator/src/schema/err.rs
@@ -265,7 +265,10 @@ impl From<transitive_closure::TcError<EntityUID>> for SchemaError {
                 SchemaError::ActionTransitiveClosure(Box::new(e).into())
             }
             transitive_closure::TcError::HasCycle(err) => {
-                schema_errors::CycleInActionHierarchyError(err.vertex_with_loop().clone()).into()
+                schema_errors::CycleInActionHierarchyError {
+                    uid: err.vertex_with_loop().clone(),
+                }
+                .into()
             }
         }
     }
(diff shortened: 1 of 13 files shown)


cedar-policy-validator/src/schema/err.rs
```diff
@@ -363,40 +369,63 @@ pub mod schema_errors {
     // CAUTION: this type is publicly exported in `cedar-policy`.
     // Don't make fields `pub`, don't make breaking changes, and use caution
     // when adding public methods.
-    #[derive(Debug, Diagnostic, Error)]
-    #[diagnostic(help(
-        "any entity types appearing anywhere in a schema need to be declared in `entityTypes`"
-    ))]
-    pub struct UndeclaredEntityTypesError(pub(crate) BTreeSet<EntityType>);
```
john-h-kastner-aws, line 370: Will the order still be deterministic here?

john-h-kastner-aws (approved):

shaobo-he-aws (approved):

cedar-policy-validator/src/schema/err.rs
cdisselkoen: Added a sort in the `Display` trait, which will ensure the message is deterministic

shaobo-he-aws (approved):

PR #1415, opened 2025-01-07
Make datetime extension an experimental feature

shaobo-he-aws:


cdisselkoen (approved):

john-h-kastner-aws (commented):
Why did you undo the change to remove `datetime` as a default feature in the core and validator crates? I would think that's what we want

shaobo-he-aws:
> Why did you undo the change to remove `datetime` as a default feature in the core and validator crates? I would think that's what we want

I thought that we an experimental feature is meant to apply to cedar-policy since it contains public APIs. Do you think that we should provide the same guarantee in core and validator (i.e., default features should be DRTed) as well?

john-h-kastner-aws:
> Do you think that we should provide the same guarantee in core and validator (i.e., default features should be DRTed) as well?

This is more a question of how Cargo will handle this feature. If `datetime` is default in `cedar-policy-core`, and `cedar-policy` depends on it without `no-default-features`, then I would expect that the `datetime` extension would be enabled when parsing Cedar policies even when it's not explicitly enabled by the consumer of `cedar-policy`

shaobo-he-aws:
> This is more a question about how Cargo will handle this feature. If `datetime` is default in `cedar-policy-core`, and `cedar-policy` depends on it without `no-default-features`, then I would expect that the `datetime` extension would be enabled when parsing Cedar policies even when it's not explicitly enabled by the consumer of `cedar-policy`

I see. It makes sense. I'll revert the latest commit.

cdisselkoen (approved):

cedar-policy-core/src/evaluator.rs
```diff
@@ -422,15 +422,44 @@ impl<'e> Evaluator<'e> {
...
+                                    self.extensions
+                                        .types_with_operator_overloading()
+                                        .map(|n| Type::Extension { name: n.clone() })
+                                        .collect::<BTreeSet<_>>(),
+                                );
+                                Err(EvaluationError::type_error_with_advice(
+                                    expected_types.clone(),
+                                    &arg1,
+                                    format!(
+                                        "Only type {} support comparison",
```
john-h-kastner-aws, line 459:
```suggestion
                                        "Only types {} support comparison",
```

also, could re-use https://github.com/cedar-policy/cedar/blob/4c9c8d892982df5e540fdfe6c569249551525159/cedar-policy-core/src/parser/fmt.rs#L458

cedar-policy-core/src/evaluator.rs
```diff
@@ -422,15 +422,44 @@ impl<'e> Evaluator<'e> {
...
                             _ => {
-                                let mut expected_types = Extensions::types_with_operator_overloading().map(|name| Type::Extension { name });
-                                expected_types.push(Type::Long);
-                                Err(EvaluationError::type_error_with_advice(expected_types, &arg1, "Only `Long` and extension types `datetime`, `duration` support comparison".to_string()))
+                                let mut expected_types = nonempty::NonEmpty::singleton(Type::Long);
+                                expected_types.extend(
+                                    self.extensions
+                                        .types_with_operator_overloading()
+                                        .map(|n| Type::Extension { name: n.clone() })
+                                        .collect::<BTreeSet<_>>(),
```
john-h-kastner-aws, line 453: Could this be a call to `valid_comparison_op_types`? Probably needs to build the btree set for ordering, but it'd be a bit simpler. Maybe you could just `.sort()` the nonempty

john-h-kastner-aws (approved):

cedar-policy-core/src/evaluator.rs
shaobo-he-aws: Good point. Addressed.

john-h-kastner-aws (approved):

PR #1424, opened 2025-01-13
fix 1421, usage of `policy_id` in Typechecker

### LINKED ISSUE #1421: Improve interal interface for `Typechecker` struct

Our internal `Typchecker` struct stores a `PolicyId` for use in error reporting here https://github.com/cedar-policy/cedar/blob/7bd42a33d6051916b47dc4b232fb670882a060a5/cedar-policy-validator/src/typecheck.rs#L68-L73 but it also accepts policies to typecheck here https://github.com/cedar-policy/cedar/blob/7bd42a33d6051916b47dc4b232fb670882a060a5/cedar-policy-validator/src/typecheck.rs#L101-L106 and here https://github.com/cedar-policy/cedar/blob/7bd42a33d6051916b47dc4b232fb670882a060a5/cedar-policy-validator/src/typecheck.rs#L202-L205

When the Id's of those policies don't match the stored id, any reported type errors will incorrectly reference the stored policy id. For `typecheck_policy` this means callers need to be careful to only use the typecheker on the policy with a matching id. For `multi_typecheck_by_request_env`, it is not possible to correctly call the function on a slice containing more than one policy.

We should update these functions to automatically use the correct policy id taken from their arguments. We should also be able to remove the `PolicyId` argument to `Typechecker::new`.  

_No response_

### Additional context

_No response_

cdisselkoen:
## Description of changes

Fixes the remaining issues identified in #1421 (the ones that weren't already addressed in #1422). 
- Ensures policy_id is always set properly for the purposes of validation error reporting, without relying on callers to pass the correct id
- Removes `policy_id` argument from `Typechecker::new()`
- Documents that users of some of the `Typechecker` public interfaces will not get `impossible_policy` validation warnings

Also improves verbosity by hiding the `request_env` argument in `self` in quite a number of callers/callees, similar to how we already hide `schema` and `mode` and etc. 

## Issue #, if available

Resolves #1421 

### DIFF
--- a/cedar-policy-validator/src/typecheck.rs
+++ b/cedar-policy-validator/src/typecheck.rs
@@ -74,23 +74,15 @@ pub struct Typechecker<'a> {
     schema: &'a ValidatorSchema,
     extensions: &'static ExtensionSchemas<'static>,
     mode: ValidationMode,
-    policy_id: PolicyID,
 }
 
 impl<'a> Typechecker<'a> {
-    /// Construct a new typechecker.
-    pub fn new(
-        schema: &'a ValidatorSchema,
-        mode: ValidationMode,
-        policy_id: PolicyID,
-    ) -> Typechecker<'a> {
-        // Set the extensions using `all_available_extension_schemas`.
-        let extensions = ExtensionSchemas::all_available();
+    /// Construct a new typechecker. All extensions are enabled by default.
+    pub fn new(schema: &'a ValidatorSchema, mode: ValidationMode) -> Typechecker<'a> {
         Self {
             schema,
-            extensions,
+            extensions: ExtensionSchemas::all_available(),
             mode,
-            policy_id,
         }
     }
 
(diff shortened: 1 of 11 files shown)


cedar-policy-validator/src/typecheck.rs
```diff
@@ -357,16 +363,28 @@ impl<'a> Typechecker<'a> {
...
+    /// This method handles the majority of the work. Given an expression, and
+    /// the prior capability, return the result of typechecking the expression
+    /// in the single env this typechecker was constructed for, and add any
+    /// errors encountered into the `type_errors` list.
     fn typecheck<'b>(
         &self,
-        request_env: &RequestEnv<'_>,
         prior_capability: &CapabilitySet<'b>,
         e: &'b Expr,
         type_errors: &mut Vec<ValidationError>,
```
john-h-kastner-aws, line 390: Might be worth also hiding `type_errors`? We would need to make the whole `&mut self` mutable, but it'd make the recursive calls even nicer

john-h-kastner-aws (approved):

shaobo-he-aws (approved):

cedar-policy-validator/src/typecheck.rs
cdisselkoen: I'd kinda prefer a different refactor, that we return validation errors as part of the return value instead of by adding them to a `&mut Vec`, but maybe that's not feasible.

For now I think let's leave it like this and not require `&mut self`

PR #1437, opened 2025-01-22
Adds JSON serialized PolicySet from bindings to cedar_policy::PolicySet conversion

muditchaudhary:
## Description of changes
- Adds function to convert JSON serialized `PolicySet` object from bindings (E.g., CedarJava) to a `cedar_policy::PolicySet` instance
- The resulting instance can be used to expose `cedar_policy::PolicySet`'s methods (E.g., to_json) in bindings

### DIFF
--- a/cedar-policy/src/ffi/convert.rs
+++ b/cedar-policy/src/ffi/convert.rs
@@ -225,7 +225,7 @@ pub enum SchemaToJsonAnswer {
     },
 }
 
-/// Convert cedar_policy::ffi::PolicySet JSON to cedar_policy::PolicySet Object
+/// Convert `cedar_policy::ffi::PolicySet` JSON to `cedar_policy::PolicySet` Object
 pub fn json_to_policyset(
     json_string: &str,
 ) -> Result<crate::PolicySet, Box<dyn std::error::Error>> {
--- a/cedar-policy/src/ffi/convert.rs
+++ b/cedar-policy/src/ffi/convert.rs
@@ -18,7 +18,7 @@
 //! Cedar Wasm conversion functions are generated from the functions in this
 //! file.
 
-use super::utils::JsonValueWithNoDuplicateKeys;
+use super::utils::{JsonValueWithNoDuplicateKeys, PolicySet};
 use super::{DetailedError, Policy, Schema, Template};
 use serde::{Deserialize, Serialize};
 #[cfg(feature = "wasm")]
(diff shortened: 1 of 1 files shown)


cedar-policy/src/ffi/convert.rs
```diff
@@ -225,6 +225,23 @@ pub enum SchemaToJsonAnswer {
     },
 }
 
+/// Convert `cedar_policy::ffi::PolicySet` JSON to `cedar_policy::PolicySet` Object
+pub fn json_to_policyset(
+    json_string: &str,
+) -> Result<crate::PolicySet, Box<dyn std::error::Error>> {
```
cdisselkoen, line 231: Returning `Box<dyn Error>` instead of `DetailedError` or something that `impl`s `miette::Diagnostic` diverges from what a lot of the rest of the FFI functions do, and loses structured error information.  Is there a reason this API should have less structured error information (e.g., no help text, no source location) than other APIs?

cedar-policy/src/ffi/convert.rs
```diff
@@ -225,6 +225,23 @@ pub enum SchemaToJsonAnswer {
     },
 }
 
+/// Convert `cedar_policy::ffi::PolicySet` JSON to `cedar_policy::PolicySet` Object
+pub fn json_to_policyset(
```
cdisselkoen, line 229: Elsewhere (e.g., in `ffi/is_authorized.rs`), function names with just `json` take a `serde_json::Value`, while functions like this that take a string have names with `json_str`

cdisselkoen (commented):
Could you say a little bit more about why this function is required?  As opposed to just using `serde` to deserialize the string into a struct that includes `PolicySet` as a field, and then using `.parse()` on that `PolicySet`

cedar-policy/src/ffi/convert.rs
muditchaudhary: Will change this to match other FFI functions

muditchaudhary: > Could you say a little bit more about why this function is required? As opposed to just using `serde` to deserialize the string into a struct that includes `PolicySet` as a field, and then using `.parse()` on that `PolicySet`

`.parse` for `PolicySet` has visibility `pub (super)` so it is not visible to `CedarJavaFFI`. If changing its visibility to public is acceptable, I can use that instead.

cdisselkoen: I think changing the visibility is preferable to introducing a new function in this case.

PR #1452, opened 2025-02-06
move protobuf code to cedar-policy

cdisselkoen:
## Description of changes

Moves all protobuf-related code out of `cedar_policy_core` and `cedar_policy_validator` and into a new (public) submodule of `cedar_policy`.  This should resolve the build issues which resulted from packages having to depend on `.proto` files in other crates.

This necessitated various changes to the internal APIs of `cedar_policy_core` and `cedar_policy_validator`.  If someone feels strongly, I can split out those changes into a separate PR.

No changes to the public API, other than that the actual Rust structures compiled from protobuf are now public in `cedar_policy`, behind the `protobufs` experimental flag.


cdisselkoen:
I'm confused by the current build failure (which happens while trying to compile code that was generated by `prost`).  Rust-analyzer seems to find and compile the files correctly, but `cargo check` does not.

cdisselkoen:
Build is now passing. CI just reporting that DRT build fails, I will make a separate PR to deal with that (but not merge this until the DRT PR is ready).  This PR is ready for review.

cedar-policy/src/proto/validator.rs
```diff
@@ -0,0 +1,436 @@
...
+                tags,
+                enum_choices: vec![],
+            },
+            cedar_policy_validator::ValidatorEntityTypeKind::Enum(enum_choices) => Self {
+                name,
+                descendants,
+                attributes,
+                open_attributes,
+                tags,
+                enum_choices: enum_choices.into_iter().map(ToString::to_string).collect(),
```
adpaco-aws, line 178: Can't you use `new_enum` and `new_standard` here for some reason?

cedar-policy-core/src/entities.rs
```diff
@@ -90,6 +90,12 @@ impl Entities {
         }
     }
 
+    /// Is this a partial store (created with `.partial()`)
+    #[cfg(feature = "partial-eval")]
+    pub fn is_partial(&self) -> bool {
+        self.mode == Mode::Partial
+    }
```
adpaco-aws, line 97: I don't understand this. Is it needed?

adpaco-aws: Ok, found where it's used but I'm still confused as to why the `#[cfg(feature = "partial-eval")]` is not in the body.

cedar-policy/src/proto/validator.rs
cdisselkoen: Those are methods for constructing `cedar_policy_validator::ValidatorEntityType`.  This code is constructing `models::ValidatorEntityType`.

cedar-policy-core/src/entities.rs
cdisselkoen: Easy enough to make this method available without the feature, and just return `false` when the feature is disabled.

Should we also make `.partial()` available when the feature is disabled?

adpaco-aws: That's why I'm confused: Why does this code compile when `partial-eval` is not enabled? I'd assume `if cfg!(feature = "partial-eval") && v.is_partial()` to be an illegal expression since `is_partial` would be undefined in that case.

cdisselkoen: I don't think CI tests the `protobufs` feature without the `partial-eval` feature.  I think we only test with default features, no features, and all features.

adpaco-aws: Ok, then I'm happy with the change to make it always available.

cdisselkoen:
Ready for review -- https://github.com/cedar-policy/cedar-spec/pull/532 will address the downstream CI failure for DRT.

PR #1453, opened 2025-02-06
Entity removal

chaluli:
## Description of changes

Adds the ability to remove `Entity`s from  an `Entities` struct.


cedar-policy-core/src/ast/entity.rs
```diff
@@ -453,12 +462,22 @@ impl Entity {
...
+        self.parents.contains(e)
     }
 
     /// Iterate over this entity's ancestors
     pub fn ancestors(&self) -> impl Iterator<Item = &EntityUID> {
-        self.ancestors.iter()
+        self.parents.iter().chain(self.ancestors.iter())
+    }
+
+    /// Iterate over this entity's parents
```
cdisselkoen, line 478:
```suggestion
    /// Iterate over this entity's direct parents
```

cedar-policy-core/src/ast/entity.rs
```diff
@@ -497,15 +516,19 @@ impl Entity {
...
             ancestors: HashSet::new(),
+            parents: HashSet::new(),
             tags: BTreeMap::new(),
         }
     }
 
     /// Test if two `Entity` objects are deep/structurally equal.
     /// That is, not only do they have the same UID, but also the same
-    /// attributes, attribute values, and ancestors.
+    /// attributes, attribute values, and ancestors/parents.
```
cdisselkoen, line 526:
```suggestion
    /// attributes, attribute values, and ancestors/parents.
    ///
    /// Does not test that they have the same _direct_ parents, only that they have the same overall ancestor set.
```

cedar-policy-core/src/ast/entity.rs
```diff
@@ -545,28 +568,51 @@ impl Entity {
 
     /// Mark the given `UID` as an ancestor of this `Entity`
```
cdisselkoen, line 569:
```suggestion
    /// Mark the given `UID` as an (indirect) ancestor of this `Entity`
```

cedar-policy-core/src/ast/entity.rs
```diff
@@ -545,28 +568,51 @@ impl Entity {
...
     pub fn add_ancestor(&mut self, uid: EntityUID) {
-        self.ancestors.insert(uid);
+        if !self.parents.contains(&uid) {
+            self.ancestors.insert(uid);
+        }
+    }
+
+    /// Mark the given `UID` as a parent of this `Entity`
+    /// remove the UID from ancestors (indirect parents)
+    /// if it was previously added as an ancestor
```
cdisselkoen, line 578:
```suggestion
    /// Mark the given `UID` as a (direct) parent of this `Entity`, and
    /// remove the UID from (indirect) ancestors
    /// if it was previously added as an ancestor
```

cedar-policy-core/src/ast/entity.rs
```diff
@@ -545,28 +568,51 @@ impl Entity {
...
+    pub fn add_parent(&mut self, uid: EntityUID) {
+        self.ancestors.remove(&uid);
+        self.parents.insert(uid);
     }
 
-    /// Consume the entity and return the entity's owned Uid, attributes, parents, and tags.
+    /// Remove the given `UID` as an ancestory of this `Entity`
```
cdisselkoen, line 584:
```suggestion
    /// Remove the given `UID` as an ancestor of this `Entity`.
    /// 
    /// No effect if the `UID` was a direct parent.
```

cedar-policy-core/src/ast/entity.rs
```diff
@@ -545,28 +568,51 @@ impl Entity {
...
+    pub fn remove_ancestor(&mut self, uid: &EntityUID) {
+        self.ancestors.remove(uid);
+    }
+
+    /// Remove the given `UID` as a parent of this `Entity`
```
cdisselkoen, line 589:
```suggestion
    /// Remove the given `UID` as a (direct) parent of this `Entity`.
    ///
    /// No effect on the `Entity`'s (indirect) ancestors.
```

cedar-policy-core/src/ast/entity.rs
```diff
@@ -545,28 +568,51 @@ impl Entity {
...
+    pub fn remove_parent(&mut self, uid: &EntityUID) {
+        self.parents.remove(uid);
```
cdisselkoen, line 591: Should we also update the `ancestors` here?  What if we had A --> B --> C and we remove parent B from A, shouldn't we also remove ancestor C from A?

cedar-policy-core/src/entities.rs
```diff
@@ -152,6 +152,42 @@ impl Entities {
...
+    ) -> Result<Self> {
+        for uid_to_remove in collection.into_iter() {
+            match self.entities.remove(&uid_to_remove) {
+                None => (),
+                Some(entity_to_remove) => {
+                    for entity in self.entities.values_mut() {
+                        if entity.is_descendant_of(&uid_to_remove) {
+                            // remove any direct or indirect link between `entity` and `entity_to_remove`
+                            Arc::make_mut(entity).remove_ancestor(&uid_to_remove);
+                            Arc::make_mut(entity).remove_parent(&uid_to_remove);
```
cdisselkoen, line 173: Based on [the docs for `Arc::make_mut()`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.make_mut), I'm wondering if this is doing the right thing -- if it makes a copy in order to get a mutable reference, are we actually updating what we think we are, or only the copy?  What happens to the copy, is it dropped or saved?

cdisselkoen: I guess this must be doing the right thing since the tests pass

cedar-policy-core/src/entities.rs
```diff
@@ -2239,7 +2338,7 @@ mod schema_based_parsing_tests {
         }
 
         fn allowed_parent_types(&self) -> Arc<HashSet<EntityType>> {
-            Arc::new(HashSet::new())
+            Arc::new([].into_iter().collect())
```
cdisselkoen, line 2341: Curious why this change is necessary

cedar-policy-validator/src/schema.rs
```diff
@@ -876,6 +876,7 @@ impl ValidatorSchema {
             Entity::new_with_attr_partial_value_serialized_as_expr(
                 action_id.clone(),
                 action.attributes.clone(),
+                [].into_iter().collect(),
```
cdisselkoen, line 879: would `HashSet::new()` be clearer here?

cedar-policy/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Cedar Language Version: TBD
 
 ### Added
 
+- Added ability to remove `Entity`s from an `Entities` struct (resolving #701)
```
cdisselkoen, line 18:
```suggestion
- Added `Entities::remove_entities()`, allowing you to remove `Entity`s from an `Entities` struct (resolving #701)
```

cedar-policy/src/api.rs
```diff
@@ -503,6 +506,22 @@ impl Entities {
         ))
     }
 
+    /// Removes each of the [`EntityUID`]s in the interator
```
cdisselkoen, line 509:
```suggestion
    /// Removes each of the [`EntityUID`]s in the iterator
```

cedar-policy-core/src/ast/entity.rs
chaluli: I view this and remove_ancestor as a "low level" interface that allows a user to modify individual entity items. The end user is responsible for ensuring the hierarchy/TC invariants are maintained if they use this function directly.

If they want to have the invariants automatically maintained they should make use of Entities::remove_entities to remove the desired entities and TC edges they induce.

cedar-policy-core/src/entities.rs
chaluli: I was a little fuzzy on how `Arc::make_mut()` works, but it does seem to work to modify the original entity object. Perhaps it works because it makes a clone of the pointer and not of the entity?

cedar-policy-core/src/entities.rs
chaluli: Not necessary. Reverting back to original for readability purposes.

cedar-policy-core/src/entities.rs
cdisselkoen: Reading again, I think it rewrites the internal pointer contained in the `Arc` to point to a newly-allocated copy of `entity`

cedar-policy-core/src/ast/entity.rs
cdisselkoen: Cool.  Maybe add a comment on this function that the caller is responsible for ensuring hierarchy/TC is maintained?

cedar-policy-core/src/entities.rs
chaluli: It seems that it either (1) returns a mutable reference to the original entity object if no other `Arc`s point to that object or (2) if there are other `Arc`s pointing to the original entity it clones the entity and returns a mutable pointer to the clone of the original entity.

From my understanding, in our code-base we would always be in case (1). This is because we do not expose the internal `Arc` representation of the `Entities` struct (i.e., there should be only one `Arc` pointing to any given `entity` object).

cdisselkoen: I don't think we're always in case (1) -- e.g., someone could have made a copy of the `Entities` struct.  In fact, I believe that's the reason for the `Arc` in the first place, so that making a copy of the `Entities` struct is not a full deep copy but can just increment a bunch of reference counts in `Arc`s and copy pointers.

cdisselkoen: Nonetheless, this code is definitely doing the right thing

cedar-policy-core/src/ast/entity.rs
```diff
@@ -339,10 +339,13 @@ pub struct Entity {
     /// `RestrictedExpr`s, for mostly historical reasons.
     attrs: BTreeMap<SmolStr, PartialValueSerializedAsExpr>,
 
-    /// Set of ancestors of this `Entity` (i.e., all direct and transitive
-    /// parents), as UIDs
+    /// Set of (indirect) ancestors of this `Entity` (e.g., grandparents,
+    /// etc.), as UIDs
```
john-h-kastner-aws, line 343: Could you add a note clarifying whether or not these are disjoint? At first I thought they should be, but an entity could be a direct parent and also an indirect ancestor via another parent, so I'm entirely not sure what to expect here.

cedar-policy-core/src/ast/entity.rs
```diff
@@ -453,12 +462,22 @@ impl Entity {
...
+
+    /// Is this `Entity` a (direct) child of `e` in the entity hierarchy?
+    pub fn is_child_of(&self, e: &EntityUID) -> bool {
+        self.parents.contains(e)
     }
 
     /// Iterate over this entity's ancestors
     pub fn ancestors(&self) -> impl Iterator<Item = &EntityUID> {
-        self.ancestors.iter()
+        self.parents.iter().chain(self.ancestors.iter())
```
john-h-kastner-aws, line 475: This is a little confusing. Internal to this struct direct `parents` aren't (necessarily?) included in ancestors, but when calling this function in the rust of our code `ancestors` means all ancestors. Could ancestors in this struct be called `indirect_ancestors` or something else explicit like that?

cedar-policy-core/src/ast/entity.rs
```diff
@@ -543,30 +568,65 @@ impl Entity {
...
+        self.ancestors.remove(&uid);
+        self.parents.insert(uid);
+    }
+
+    /// Remove the given `UID` as an ancestor of this `Entity`.
+    ///
+    /// No effect if the `UID` was a direct parent.
+    /// The caller of this code is responsible for maintaining
+    /// transitive closure of hierarchy.
+    pub fn remove_ancestor(&mut self, uid: &EntityUID) {
```
john-h-kastner-aws, line 595: Following on from my comment on `pub fn ancestors(&self)`. If that function returns all ancestors, then it's very confusing that this ignore direct parents.

john-h-kastner-aws (commented):
Impl looks good, but I think the names on internal APIs are confusing

cedar-policy/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Cedar Language Version: TBD
 
 ### Added
 
+- Added `Entities::remove_entities()`, allowing you to remove `Entity`s from an `Entities` struct (resolving #701)
```
adpaco-aws, line 18: We can avoid the "allowing you" and make it more concise:
```suggestion
- Added `Entities::remove_entities()` to remove `Entity`s from an `Entities` struct (resolving #701)
```

cedar-policy-core/src/ast/entity.rs
chaluli: Yes, disjoint.

chaluli: I'll rename the internal field to indirect ancestors and update the function names according to if they affect only indirect ancestors or combo of indirect ancestors and parents.

adpaco-aws (approved):

PR #1457, opened 2025-02-07
clarify docs and changelog for 1446

cdisselkoen:


cedar-policy/src/api.rs
```diff
@@ -872,9 +872,12 @@ impl Authorizer {
     ///
     /// The authorizer uses the `stacker` crate to manage stack size and tries to use a sane default.
     /// If the default is not right for you, you can try wrapping the authorizer or individual calls
-    /// to `is_authorized` in `stacker::grow`. Note that `stacker` does not provide support in some
-    /// platforms like `wasm32`. In such cases, the authorizer will assume that the stack size is
-    /// sufficient.
+    /// to `is_authorized` in `stacker::grow`.
+    /// Note that on platforms not supported by `stacker` (e.g., `wasm32`, Android),
```
adpaco-aws, line 876: I'd say "Wasm" here like you did in the changelog.
```suggestion
    /// Note that on platforms not supported by `stacker` (e.g., `wasm32`, Android),
```

cedar-policy/src/api.rs
```diff
@@ -872,9 +872,12 @@ impl Authorizer {
...
+    /// the authorizer will simply assume that the stack size is sufficient. As a result, large inputs
+    /// may result in stack overflows and crashing the process.
+    /// But on all platforms supported by `stacker` (Linux, macOS, Windows, iOS, ...), Cedar will
```
adpaco-aws, line 879: I don't think we need to mention that many platforms, and we can't be sure those are all actually supported unless we're running them.
```suggestion
    /// But on all platforms supported by `stacker` (Linux, macOS, etc.), Cedar will
```

john-h-kastner-aws (approved):

adpaco-aws (approved):

Issue #1465, opened 2025-02-12
reauthorize only practically works with one unknown


luxas:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [x] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Policy Evaluation

### Describe the bug

Hi! Thanks for open sourcing Cedar, I'm really excited to use it in a Kubernetes (API) context. (And yes, I've been in contact with Micah already some time :D)

One of the gotchas I ran into the other day when experimenting with partial evaluation for multi-stage authorizations, when gradually supplying more data as needed, was that practically (usually) only one unknown is supported, as the partial evaluation stops whenever it encounters a residual on the LHS of an && (#1445)

So for a policy that refers to two unknowns (`resource.first_unknown` and `resource.second_unknown`) in the initial partial_evaluate stage, the following happens:

1. `partial_evaluate` without the two unknowns specified simplifies the policy the best it can, until it sees `unknown("first_unknown") && resource.second_unknown && ... more concrete stuff`, and feeds this into a partial response. If evaluation would have proceeded fully, or until an error occurs (#1445), then `resource.second_unknown` would have evaluated to `unknown("second_unknown")`, but it did not.
2. `reauthorize` with the two unknowns now specified first traverses the expression and maps unknowns to concrete values, but as partial_evaluate returned early, it is not known at this time that `resource.second_unknown` will evaluate to `unknown("second_unknown")`, so second_unknown is not populated from the reauthorize mappings, only `unknown("first_unknown")` is mapped.
3. `partial_evaluate` is invoked on the mapped expression, and is able to concretize the first unknown (let's assume, to true), thus proceeding to the second unknown, first now realizing that it is unknown. However, the reauthorize mappings are not part of this stage, thus must the PE terminate at the residual `unknown("second_unknown") && ... more concrete stuff`, although `unknown("second_unknown")` indeed was concretely known at reauthorize time.

I played with some quick ways to fix this, for my own learning and exploration of the Cedar codebase. In simplicity order:

1. A full evaluation option #1445 during the first `partial_evaluate` pass would make PE realize both resource fields are indeed unknowns. However, this has some limiations, especially around typing and errors.
[...]

Issue #1445, opened 2025-01-27
Full Evaluation Option for Policy Expressions.


amzn-mdamine:
### Category

User level API features/changes

### Describe the feature you'd like to request

# Feature Request: Full Evaluation Option for Policy Expressions

## Current Behavior

In Cedar, when evaluating policy expressions containing `and`, `or`, or `if-else` statements, the evaluation process stops after partially evaluating the left side of the expression. This partial evaluation can lead to missed opportunities to fully resolve the expression or  resolve unknowns on the right side.

## Proposed Enhancement

I propose adding a boolean flag to enable full evaluation of policy expressions. This enhancement would allow users to choose between the current partial evaluation behavior and a new full evaluation option.

### Benefits

1. More comprehensive results: Evaluating both sides of an expression can provide a complete picture of the policy's outcome.
2. Improved resolution of unknowns: Full evaluation help resolve unknowns on the right side of expressions.
3. Flexibility: Users can choose between partial and full evaluation based on their specific needs.

### Current Code (Example for `And` expression):

```rust
ExprKind::And { left, right } => {
    match self.partial_interpret(left, slots)? {
        PartialValue::Residual(e) => {
            Ok(PartialValue::Residual(Expr::and(e, right.as_ref().clone())))
        }
        PartialValue::Value(v) => {
            if v.get_as_bool()? {
                match self.partial_interpret(right, slots)? {
                    PartialValue::Residual(right) => {
                        Ok(PartialValue::Residual(Expr::and(Expr::val(true), right)))
                    }
                    PartialValue::Value(v) => Ok(v.get_as_bool()?.into()),
                }
            } else {
                Ok(false.into())
            }
        }
    }
}
```
### Proposed pseudo code:

```rust
ExprKind::And { left, right } => {
    match self.partial_interpret(left, slots)? {
        PartialValue::Residual(left_residual) => {
            if self.full_evaluation {
                match self.partial_interpret(right, slots)? {
                    PartialValue::Residual(right_residual) => {
                        Ok(PartialValue::Residual(Expr::and(left_residual, right_residual)))
                    }
                    PartialValue::Value(rhs) => {
                        if rhs.get_as_bool()? {
                            Ok(PartialValue::Residual(Expr::and(left_residual, Expr::val(true))))
                        } else {
                            Ok(false.into())
                        }
                    }
                }
            } else {
                Ok(PartialValue::Residual(Expr::and(left_residual, right.as_ref().clone())))
            }
        }
[...]

PR #1466, opened 2025-02-12
Allow reauthorization with also initially undiscovered unknowns in the Expr

luxas:
## Description of changes

Fixes: #1465 
Please read the issue description first for context.

There might be "undiscovered" unknowns in a partially evaluated `Expr` residual. In my example in #1465, that would be `resource.newlabels` that is during the first partial evaluation call omitted and thus unknown, but not yet evaluated by Cedar into a `unknown("resource.newlabels")`, as the LHS of the `&&` didn't proceed to evaluate the RHS (#1445).

In the issue I described three quick potential solutions, of which this one probably is the most realistic to actually implement.

This PR solves the described issue by moving the reauthorization substitution logic from `expr.substitute` to `partial_interpret`, where unknowns can be looked up any number of times needed, also while proceeding to the RHS of an `&&` (of course, given the LHS evaluated to true). The code isn't super neat, having to pass the `mappings` around everywhere, I'm definitely open to discuss neater solutions.

Note: `Expr.substitute` is now almost unused; the only place I saw it used is in `Context.substitute`, which indeed is used.

Longer-term, indeed [RFC 95](https://github.com/cedar-policy/rfcs/pull/95) is the better solution, addressing this bug and probably a whole class of others in one go. However, I wanted to open this PR to show the approach, as at the end of the day, the fix wasn't very complicated (and hopefully I didn't break anything else).

If this becomes an accepted approach to go forward in the short term, I'll make tests, etc. so this PR becomes ready to merge.

- [x] A change/bug fix "invisible" to users (e.g., documentation, changes to "internal" crates like `cedar-policy-core`, `cedar-validator`, etc.). Also this only touches experimental code.

- [x] Does not require updates because my change does not impact the Cedar formal model or DRT infrastructure. (AFAIK there is no formal model yet for partial evaluation)

### DIFF
--- a/cedar-policy-core/src/authorizer.rs
+++ b/cedar-policy-core/src/authorizer.rs
@@ -92,6 +92,17 @@ impl Authorizer {
         entities: &Entities,
     ) -> PartialResponse {
         let eval = Evaluator::new(q.clone(), entities, self.extensions);
+        self.is_authorized_core_internal(eval, q, pset)
+    }
+
+    /// The same as is_authorized_core, but for any Evaluator.
+    /// A PartialResponse caller constructs its own evaluator, with an unknown mapper function.
+    pub(crate) fn is_authorized_core_internal(
+        &self,
+        eval: Evaluator<'_>,
+        q: Request,
+        pset: &PolicySet,
+    ) -> PartialResponse {
         let mut true_permits = vec![];
         let mut true_forbids = vec![];
         let mut false_permits = vec![];
--- a/cedar-policy-core/src/authorizer/partial_response.rs
+++ b/cedar-policy-core/src/authorizer/partial_response.rs
@@ -17,15 +17,25 @@
 use std::collections::HashMap;
 
 use either::Either;
-use smol_str::SmolStr;
 use std::sync::Arc;
 
 use super::{
-    err::{ConcretizationError, ReauthorizationError},
-    Annotations, AuthorizationError, Authorizer, Context, Decision, Effect, EntityUIDEntry, Expr,
-    Policy, PolicySet, PolicySetError, Request, Response, Value,
+    Annotations, AuthorizationError, Decision, Effect, EntityUIDEntry, Expr,
+    Policy, Request, Response, 
 };
-use crate::{ast::PolicyID, entities::Entities, evaluator::{EvaluationError, Evaluator}};
+use crate::{ast::PolicyID, evaluator::EvaluationError};
+
+#[cfg(feature = "partial-eval")]
+use smol_str::SmolStr;
+
+#[cfg(feature = "partial-eval")]
+use crate::{evaluator::Evaluator, entities::Entities};
+
+#[cfg(feature = "partial-eval")]
+use super::{
+    err::{ConcretizationError, ReauthorizationError},
+    Context, PolicySet, PolicySetError, Authorizer, Value,
+};
 
 type PolicyComponents<'a> = (Effect, &'a PolicyID, &'a Arc<Expr>, &'a Arc<Annotations>);
 
@@ -65,6 +75,7 @@ pub struct PartialResponse {
     /// The trivial `false` expression, used for materializing a residual for non-satisfied policies
     false_expr: Arc<Expr>,
     /// The request associated with the partial response
+    #[cfg(feature = "partial-eval")]
     request: Arc<Request>,
 }
 
@@ -79,7 +90,7 @@ impl PartialResponse {
         false_forbids: impl IntoIterator<Item = (PolicyID, (ErrorState, Arc<Annotations>))>,
         residual_forbids: impl IntoIterator<Item = (PolicyID, (Arc<Expr>, Arc<Annotations>))>,
         errors: impl IntoIterator<Item = AuthorizationError>,
-        request: Arc<Request>,
+        _request: Arc<Request>,
     ) -> Self {
         Self {
             satisfied_permits: true_permits.into_iter().collect(),
@@ -91,7 +102,8 @@ impl PartialResponse {
             errors: errors.into_iter().collect(),
             true_expr: Arc::new(Expr::val(true)),
             false_expr: Arc::new(Expr::val(false)),
-            request,
+            #[cfg(feature = "partial-eval")]
+            request: _request,
         }
     }
 
@@ -311,6 +323,7 @@ impl PartialResponse {
     }
 
     /// Attempt to re-authorize this response given a mapping from unknowns to values
+    #[cfg(feature = "partial-eval")]
     pub fn reauthorize(
         &self,
         mapping: &HashMap<SmolStr, Value>,
@@ -329,6 +342,7 @@ impl PartialResponse {
         Ok(auth.is_authorized_core_internal(eval, new_request, &policyset))
     }
 
+    #[cfg(feature = "partial-eval")]
     fn all_residual_policies(&self) -> Result<PolicySet, PolicySetError> {
         PolicySet::try_from_iter(
             self.all_permit_residuals()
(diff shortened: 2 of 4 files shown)


cedar-policy-core/src/authorizer.rs
```diff
@@ -90,6 +92,15 @@ impl Authorizer {
...
         entities: &Entities,
+    ) -> PartialResponse {
+        self.is_authorized_core_with_mappings(q, pset, entities, None)
+    }
+    pub(crate) fn is_authorized_core_with_mappings(
+        &self,
+        q: Request,
+        pset: &PolicySet,
+        entities: &Entities,
+        mappings: Option<&HashMap<SmolStr, Value>>,
```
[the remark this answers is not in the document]
luxas: Absolutely, happy to do that 👍

luxas:
Any other comments than? https://github.com/cedar-policy/cedar/pull/1466#discussion_r1957065695

If this is accepted, then I'll change to use a function, and implement some unit tests.

cedar-policy-core/src/evaluator.rs
```diff
@@ -235,7 +234,11 @@ impl<'e> Evaluator<'e> {
     ///    it doesn't consider whether we're processing a `Permit` policy or a
     ///    `Forbid` policy.
     pub fn partial_evaluate(&self, p: &Policy) -> Result<Either<bool, Expr>> {
-        match self.partial_interpret(&p.condition(), p.env())? {
+        self.partial_evaluate_with_mappings(p, None)
+    }
+
+    pub(crate) fn partial_evaluate_with_mappings(&self, p: &Policy, mappings: Option<&HashMap<SmolStr, Value>>) -> Result<Either<bool, Expr>> {
```
cdisselkoen, line 240: Doc comment here (describing that `mappings` is for unknowns, and not, say, template slots) would be nice

cedar-policy-core/src/evaluator.rs
```diff
@@ -259,10 +262,14 @@ impl<'e> Evaluator<'e> {
     /// May return an error, for instance if the `Expr` tries to access an
     /// attribute that doesn't exist.
     pub fn partial_interpret(&self, expr: &Expr, slots: &SlotEnv) -> Result<PartialValue> {
-        stack_size_check()?;
+        self.partial_interpret_with_mappings(expr, slots, None)
+    }
 
-        let res = self.partial_interpret_internal(expr, slots);
+    fn partial_interpret_with_mappings(&self, expr: &Expr, slots: &SlotEnv, mappings: Option<&HashMap<SmolStr, Value>>) -> Result<PartialValue> {
```
cdisselkoen, line 268: Similarly doc comment here

cedar-policy-core/src/authorizer.rs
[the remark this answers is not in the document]
cdisselkoen: If we did `Fn(&str) -> Option<Value>` we probably wouldn't need the outer `Option`

luxas:
Awesome! 💯 I'll try to find some time in coming week or two to make this ready to merge 👍

shaobo-he-aws (commented):
Although we have DRT as a guardrail, I'm still not very confident to take the risk of changing the core evaluation logic in order to get a PE performance improvement.

luxas:
> Although we have DRT as a guardrail, I'm still not very confident to take the risk of changing the core evaluation logic in order to get a PE performance improvement.

I did not submit this PR as a performance improvement, but being able to evaluate more than two unknown objects. See #1465 for details. Or did I misunderstand the comment?

KubeCon kept me busy until now, but I'll take a stab at rebasing now 👍

luxas:
PR rebased.

Instead of propagating a HashMap, now a generic mapping function is passed around.

Instead of passing the mapping function around in functions, I moved it to the evaluator to make the code cleaner.

Any specific requests for what unit tests to write? I'll make a couple in the coming days.

cedar-policy-core/src/evaluator.rs
```diff
@@ -208,6 +210,20 @@ impl<'e> Evaluator<'e> {
             },
             entities,
             extensions,
+            unknowns_mapper: Box::new(|_: &str| -> Option<Value> {None}),
+        }
+    }
+
+    // Constructs an Evaluator for a given unknowns mapper function.
+    pub(crate) fn with_unknowns_mapper(self, unknowns_mapper: Box<dyn Fn(&str) -> Option<Value> + 'e>) -> Self {
```
luxas, line 218: Do you prefer to make a new, more generic constructor instead of this?

luxas:
One question: Should I implement this same logic for the restricted evaluator?

luxas:
I added a comment on `Expr.substitute` now that most likely, it will only substitute one unknown, if the Expr was built by an evaluator, that stopped evaluating when encountering that one first unknown.

Also, I understand your concern @shaobo-he-aws about concrete and partial evaluation already being intertwined, and this logic possibly complicating that more. To make it safer, I put the logic behind the feature flag, to make sure the logic stays the same for concrete evaluation for now, while fixing the issue for current partial evaluation. Let me know what you think.

cedar-policy-core/src/authorizer/partial_response.rs
```diff
@@ -317,17 +317,31 @@ impl PartialResponse {
         auth: &Authorizer,
```
luxas, line 317: It seems like `reauthorize` (and all of `PartialResponse`) is not behind a feature flag, which would otherwise be convenient to require, to be able to rely on feature-flagged functions (in this case `with_unknowns_mapper` if feature-gated) downstream

cedar-policy-core/src/evaluator.rs
```diff
@@ -67,6 +67,8 @@ pub struct Evaluator<'e> {
     entities: &'e Entities,
     /// Extensions which are active for this evaluation
     extensions: &'e Extensions<'e>,
+    /// Mapper of unknown values into concrete ones, if recognized
+    unknowns_mapper: Box<dyn Fn(&str) -> Option<Value> + 'e>,
```
luxas, line 71: As discussed above, it'd be convenient to feature-gate this field too; I can do that if you like, as otherwise Rust identifies it as unused (not read) when the feature gate is not set.

john-h-kastner-aws:
Hi! We'd like to merge this change but it looks like there's a build error. Could you fix this up for us? 

```
error: field `unknowns_mapper` is never read
  --> cedar-policy-core/src/evaluator.rs:71:5
   |
53 | pub struct Evaluator<'e> {
   |            --------- field in this struct
...
71 |     unknowns_mapper: Box<dyn Fn(&str) -> Option<Value> + 'e>,
   |     ^^^^^^^^^^^^^^^
   |
   = note: `-D dead-code` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(dead_code)]`

error: could not compile `cedar-policy-core` (lib) due to 1 previous error
```

luxas:
> build error, field `unknowns_mapper` is never read

Yes this was my question [here](https://github.com/cedar-policy/cedar/pull/1466#discussion_r2036850652), if you had any preference in how far I'll propagate the build flags "up the stack". I'll go further in separating the compiled code for with the feature gate as without. Will fix this later today, and add the high-level integration test that I was using while testing this code out, as there was no other specific ones asked for.

cedar-policy-core/src/evaluator.rs
shaobo-he-aws: It looks reasonable to me. Please proceed with the change.

cedar-policy-core/src/evaluator.rs
luxas: Fixed with https://github.com/cedar-policy/cedar/pull/1466#issuecomment-2822587582

cdisselkoen (approved):
LGTM

cdisselkoen (approved):

PR #1467, opened 2025-02-12
add conversions between api types and proto types

cdisselkoen:
## Description of changes

* Adds conversions between `api` types and the newly public types in `proto`.  This is in addition to the public `Protobuf` trait that allows directly serializing/deserializing to/from binary data.  These new conversions are potentially useful for consumers that want to not just ser/de a single struct, but define their own protobuf formats that include these types.
* Moves more existing protobuf code into the new public `proto` module that was created in #1452.  This is a breaking change for experimental code -- in particular, the public `Protobuf` trait is now `proto::traits::Protobuf` instead of available at the crate root.

### DIFF
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -86,59 +86,10 @@ pub(crate) mod version {
     }
 }
 
-/// Private functions to support implementing the `Protobuf` trait on various types
-#[cfg(feature = "protobufs")]
-mod proto {
-    use std::default::Default;
-
-    /// Encode `thing` into `buf` using the protobuf format `M`
-    ///
-    /// `Err` is only returned if `buf` has insufficient space.
-    #[allow(dead_code)] // experimental feature, we might have use for this one in the future
-    pub(super) fn encode<M: prost::Message>(
-        thing: impl Into<M>,
-        buf: &mut impl prost::bytes::BufMut,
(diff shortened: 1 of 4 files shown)


cdisselkoen:
the SemVer failure is expected: as noted above, we have moved the (experimental) `Protobuf` trait from the crate root to `proto::traits::Protobuf`

cedar-policy/src/api.rs
```diff
@@ -2228,7 +2139,7 @@ impl PolicySet {
 
     /// Build the [`PolicySet`] from just the AST information
     #[cfg_attr(not(feature = "protobufs"), allow(dead_code))]
-    fn from_ast(ast: ast::PolicySet) -> Result<Self, PolicySetError> {
+    pub(crate) fn from_ast(ast: ast::PolicySet) -> Result<Self, PolicySetError> {
```
adpaco-aws, line 2142: Just checking: Do we want to keep the `cfg_attr(not(feature = "protobufs")` here?

cdisselkoen: This says "if protobufs is not enabled, allow(dead_code)".  We do still need that I believe.

adpaco-aws (approved):

shaobo-he-aws (approved):

PR #1470, opened 2025-02-12
 Allow error expressions in AST and handle empty When error

eidelmanjonathan:
## Description of changes

This change allows us to parse AST's with expressions that are Errors. This will be useful in cases where we might want to parse an incomplete or invalid Cedar policy without generating an error. An example of where this might be useful would be in implementing auto-complete as part of an LSP or IDE plugin.

This particular change allows us to parse the following invalid policy 

`permit(principal, action, resource) when {};`

by calling the following:

```
 text_to_cst::parse_policy(text)
        .expect("failed parser")
        .to_policy_with_errors(ast::PolicyID::from_string("id"))
        .unwrap_or_else(|errs| {
            panic!("failed conversion to AST:\n{:?}", miette::Report::new(errs))
        })
```

- [ x] A change "invisible" to users (e.g., documentation, changes to "internal" crates like `cedar-policy-core`, `cedar-validator`, etc.)

- [ x ] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).
- [] Does not update the CHANGELOG because my change does not significantly impact released code.

- [x ] Does not require updates because my change does not impact the Cedar formal model or DRT infrastructure.

- [x ] Does not require updates because my change does not impact the Cedar language specification.

### DIFF
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+#[cfg(feature = "error-ast")]
+use super::expr_allows_errors::AstExprErrorKind;
 use crate::{
     ast::*,
     expr_builder::{self, ExprBuilder as _},
@@ -32,8 +34,6 @@ use std::{
     sync::Arc,
 };
 use thiserror::Error;
-
-use super::expr_allows_errors::AstExprErrorKind;
 
 #[cfg(feature = "wasm")]
 extern crate tsify;
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -33,6 +33,8 @@ use std::{
 };
 use thiserror::Error;
 
+use super::expr_allows_errors::AstExprErrorKind;
+
 #[cfg(feature = "wasm")]
 extern crate tsify;
 
@@ -155,6 +157,12 @@ pub enum ExprKind<T = ()> {
     Set(Arc<Vec<Expr<T>>>),
     /// Anonymous record (whose elements may be arbitrary expressions)
     Record(Arc<BTreeMap<SmolStr, Expr<T>>>),
+
+    // Error expression - allows us to continue parsing even when we have errors
+    Error {
+        error_kind: AstExprErrorKind, 
+        sub_expression: Option<Arc<Expr<T>>>
+    },
 }
 
 impl From<Value> for Expr {
@@ -191,7 +199,7 @@ impl From<PartialValue> for Expr {
 }
 
 impl<T> Expr<T> {
-    fn new(expr_kind: ExprKind<T>, source_loc: Option<Loc>, data: T) -> Self {
+    pub(crate) fn new(expr_kind: ExprKind<T>, source_loc: Option<Loc>, data: T) -> Self {
         Self {
             expr_kind,
             source_loc,
@@ -385,6 +393,7 @@ impl<T> Expr<T> {
             ExprKind::Is { .. } => Some(Type::Bool),
             ExprKind::Set(_) => Some(Type::Set),
             ExprKind::Record(_) => Some(Type::Record),
+            ExprKind::Error { .. } => None
         }
     }
 }
@@ -717,6 +726,7 @@ impl Expr {
                 expr.substitute_general::<T>(definitions)?,
                 entity_type.clone(),
             )),
+            ExprKind::Error { .. } => Ok(self.clone()),
         }
     }
 }
@@ -856,6 +866,8 @@ impl<T: Default + Clone> expr_builder::ExprBuilder for ExprBuilder<T> {
     type Expr = Expr<T>;
 
     type Data = T;
+
+    type ErrorType = ParseErrors;
 
     fn loc(&self) -> Option<&Loc> {
         self.source_loc.as_ref()
@@ -1167,6 +1179,10 @@ impl<T: Default + Clone> expr_builder::ExprBuilder for ExprBuilder<T> {
             expr: Arc::new(expr),
             entity_type,
         })
+    }
+        
+    fn error(self, parse_errors: ParseErrors, sub_expression: Option<Arc<Self::Expr>>) -> Result<Self::Expr, Self::ErrorType> {
+        Err(parse_errors)
     }
 }
 
@@ -1482,6 +1498,7 @@ impl<T> Expr<T> {
                 expr.hash_shape(state);
                 entity_type.hash(state);
             }
+            ExprKind::Error { error_kind, sub_expression } => error_kind.hash(state),
         }
     }
 }
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -162,7 +162,7 @@ pub enum ExprKind<T = ()> {
     Error {
         /// Type of error that led to the failure
         error_kind: AstExprErrorKind,
-        /// Any subexpressions that were successfully parsed - useful for nested errors 
+        /// Any subexpressions that were successfully parsed - useful for nested errors
         sub_expression: Option<Arc<Expr<T>>>,
     },
 }
@@ -1505,10 +1505,7 @@ impl<T> Expr<T> {
                 expr.hash_shape(state);
                 entity_type.hash(state);
             }
-            ExprKind::Error {
-                error_kind,
-                ..
-            } => error_kind.hash(state),
+            ExprKind::Error { error_kind, .. } => error_kind.hash(state),
         }
     }
 }
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -1184,6 +1184,7 @@ impl<T: Default + Clone> expr_builder::ExprBuilder for ExprBuilder<T> {
     }
 
     /// Don't support AST Error nodes - return the error right back
+    #[cfg(feature = "error-ast")]
     fn error(self, parse_errors: ParseErrors) -> Result<Self::Expr, Self::ErrorType> {
         Err(parse_errors)
     }
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -158,10 +158,12 @@ pub enum ExprKind<T = ()> {
     /// Anonymous record (whose elements may be arbitrary expressions)
     Record(Arc<BTreeMap<SmolStr, Expr<T>>>),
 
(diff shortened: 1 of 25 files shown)


cdisselkoen (commented):
Some places where the comments refer to `InlinePolicy`, which doesn't exist, and the code uses `StaticPolicy` instead -- some of these were pre-existing before your PR, but I noticed them while reviewing this PR

cedar-policy-core/src/evaluator.rs
```diff
@@ -760,6 +760,11 @@ impl<'e> Evaluator<'e> {
                     }
                 }
             }
+            // PANIC SAFETY: We should never be evaluating an error node - Will never happen because we do not use error-enabled parsing for evaluation
+            #[allow(clippy::panic)]
+            ExprKind::Error { .. } => {
+                panic!("We cannot evaluate an error node - this should never happen")
+            }
```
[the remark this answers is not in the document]
eidelmanjonathan: Yes you are correct that there is no safeguard for that. If someone were to call the `to_policy_with_errors` they could reach this case.

I guess my issue here is that it seems like a reasonable behaviour to me that we would panic if we tried to evaluate an expression that contained an error. That said, I can instead introduce a new `EvaluationError`. Maybe that is a better approach

cedar-policy-core/src/est/expr.rs
```diff
@@ -1113,6 +1118,11 @@ impl<T: Clone> From<ast::Expr<T>> for Expr {
                         .map(|(k, v)| (k, v.into())),
                 )
                 .unwrap(),
+            // PANIC SAFETY: There is no reason to support having deliberate Errors in EST - This should never happen
+            #[allow(clippy::panic)]
+            ast::ExprKind::Error { .. } => {
+                panic!("We do not support converting an AST Error node into an EST");
+            }
```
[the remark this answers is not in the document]
eidelmanjonathan: Your concern definitely makes sense. I am relying here on the fact that in production, we should never be using the parsing that allows errors.

I will modify the the EST to support the errors

cedar-policy-validator/src/level_validate.rs
```diff
@@ -238,6 +238,11 @@ impl Validator {
                     .collect();
                 Self::min(v)
             }
+            // PANIC SAFETY: We do not allow validation of AST's with error nodes
+            #[allow(clippy::panic)]
+            ExprKind::Error { .. } => {
+                panic!(" We do not allow validation of AST's with error nodes - this should never happen");
+            }
```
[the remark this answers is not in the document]
eidelmanjonathan: Yes will do

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -178,8 +203,12 @@ impl Node<Option<cst::Policy>> {
     }
 
     /// Convert `cst::Policy` to an AST `InlinePolicy`. (Will fail if the CST is for a template)
-    pub fn to_policy(&self, id: ast::PolicyID) -> Result<ast::StaticPolicy> {
-        let maybe_template = self.to_policy_template(id);
+    pub fn to_policy_helper(
```
adpaco-aws, line 206: Is this part of the public API for ASTs? If the answer is yes, then I'm not sure about renaming the function. Consumers are going to break anyway due to the extra `allow_errors` argument. We could even go as far as to create new `to_policy_with_errors` instead of changing this, but that'd lead to duplicated code for `to_policy_template` which I don't think we want.

cedar-policy-validator/src/typecheck.rs
```diff
@@ -1174,6 +1174,11 @@ impl<'a> SingleEnvTypechecker<'a> {
                     },
                 )
             }
+            // PANIC SAFETY: We should never be trying to type check or evaluate code that allows errors - This is avoided by using the default cst-to-ast parsing which fails on parse errors
```
adpaco-aws, line 1177: Split the comment into at least two lines.

adpaco-aws (commented):
I share the same concerns @cdisselkoen brought up: We need another way to handle error expressions in validation and evaluation.

cedar-policy-validator/src/typecheck.rs
eidelmanjonathan: Will do

cedar-policy-core/src/parser/cst_to_ast.rs
eidelmanjonathan: The helper functions should not be public, I'll fix that.

I am also planning to add a feature flag so that none of this error functionality will work unless it is explicitly enabled

cedar-policy-validator/src/level_validate.rs
```diff
@@ -77,8 +79,8 @@ impl Validator {
     }
 
     fn min(
-        v: impl IntoIterator<Item = (EntityDerefLevel, Option<EntityDerefLevelViolation>)>,
-    ) -> (EntityDerefLevel, Option<EntityDerefLevelViolation>) {
+        v: impl IntoIterator<Item = (EntityDerefLevel, Option<ValidationError>)>,
+    ) -> (EntityDerefLevel, Option<ValidationError>) {
```
adpaco-aws, line 83: I don't understand why the changes in this file are needed.

adpaco-aws (commented):
Looking good! I'm wondering if there's a better name to refer to this feature. The "error" word is quite overloaded in Cedar because they can happen in different ways (parsing, evaluation, etc.). A quick search shows me terms like "incomplete" or "tolerant". What do you think?

cedar-policy-validator/src/level_validate.rs
eidelmanjonathan: So the issue here is that we need to handle the case of an error node. It doesn't entirely make sense to return an `EntityDerefLevelViolation` in the case of an error. I discussed with John, and he suggested that this might be a better approach so that we could use `ValidationError::InternalInvariantViolation` here 

Definitely open to other suggestions!

cedar-policy-core/src/entities/json/err.rs
```diff
@@ -253,6 +258,14 @@ pub struct MissingRequiredRecordAttr {
     record_attr: SmolStr,
 }
 
+#[derive(Debug, Error, Diagnostic)]
+#[error("AST Error node")]
+/// Error type for a record missing a required attr
+pub struct ErrorNode {
```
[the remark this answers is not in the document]
eidelmanjonathan: Good catch, yes will fix

cedar-policy-core/src/ast/expr.rs
```diff
@@ -857,6 +869,8 @@ impl<T: Default + Clone> expr_builder::ExprBuilder for ExprBuilder<T> {
 
     type Data = T;
 
+    type ErrorType = ParseErrors;
+
```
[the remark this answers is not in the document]
eidelmanjonathan: Good idea, can add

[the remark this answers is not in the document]
eidelmanjonathan: Also per your above comment `tolerant-ast` works for me, I will change to that!

cedar-policy-core/src/entities/json/err.rs
```diff
@@ -120,6 +120,11 @@ pub enum JsonDeserializationError {
     #[error(transparent)]
     #[diagnostic(transparent)]
     TypeMismatch(TypeMismatch),
+    /// When trying to deserialize an AST with an error in it - this should fail
+    #[cfg(feature = "tolerant-ast")]
+    #[error("Unable to deserialize an AST Error node")]
+    #[diagnostic(help("AST error node indicates that the expression has failed to parse"))]
+    ASTErrorNode,
```
cdisselkoen, line 127: Was this variant intended to carry an `ErrorNode`?

cedar-policy-core/src/evaluator/err.rs
```diff
@@ -89,6 +89,12 @@ pub enum EvaluationError {
     #[diagnostic(transparent)]
     NonValue(#[from] evaluation_errors::NonValueError),
 
+    /// This is an expression AST node that gets generated when parsing fails
```
cdisselkoen, line 92:
```suggestion
    /// Trying to evaluate an expression AST node that gets generated when parsing fails
```

cedar-policy-validator/src/typecheck/typecheck_answer.rs
```diff
@@ -39,6 +39,10 @@ pub(crate) enum TypecheckAnswer<'a> {
 
     /// Recursion limit reached
     RecursionLimit,
+
+    /// Trying to typescheck an error node
```
cdisselkoen, line 43:
```suggestion
    /// Trying to typecheck an error node
```

cedar-policy/Cargo.toml
```diff
@@ -52,14 +52,15 @@ corpus-timing = []
...
-experimental = ["partial-eval", "permissive-validate", "partial-validate", "level-validate", "entity-manifest", "protobufs", "datetime"]
+experimental = ["partial-eval", "permissive-validate", "partial-validate", "level-validate", "entity-manifest", "protobufs", "datetime", "tolerant-ast"]
 entity-manifest = ["cedar-policy-validator/entity-manifest"]
 partial-eval = ["cedar-policy-core/partial-eval", "cedar-policy-validator/partial-eval"]
 permissive-validate = []
 partial-validate = ["cedar-policy-validator/partial-validate"]
 protobufs = ["dep:prost", "dep:prost-build"]
 level-validate = ["cedar-policy-validator/level-validate"]
 wasm = ["serde-wasm-bindgen", "tsify", "wasm-bindgen"]
+tolerant-ast = [ "cedar-policy-core/tolerant-ast",  "cedar-policy-validator/tolerant-ast"]
```
cdisselkoen, line 63:
```suggestion
tolerant-ast = ["cedar-policy-core/tolerant-ast",  "cedar-policy-validator/tolerant-ast"]
```

cdisselkoen (approved):
Looks great. Thanks for the work.

cedar-policy-core/src/entities/json/err.rs
eidelmanjonathan: Yup, will fix

eidelmanjonathan: Actually, on second thought the `ErrorNode` doesn't really add anything because the context is always just going to end up being unknown anyway. I'll remove it entirely

adpaco-aws (approved):
Thanks!

Can you please add this change to [the changelog](https://github.com/cedar-policy/cedar/blob/main/cedar-policy/CHANGELOG.md)? See the changes in 4.3.0 for examples adding/updating experimental features.

PR #1473, opened 2025-02-14
Allow identical duplicate entities

### LINKED ISSUE #613: Allow identical duplicates when adding to `Entities`
### Category

Currently when adding entities to an `Entities` instance an error is returned if any of the added entities have the same `EntityUID` that is already present in the `Entities`. This is good to ensure that there are no accidental errors that would be missed by silently ignoring duplicates. However, there are times when both the principal and the resource may share some parent entities, in which case it may be difficult to know if there are going to be any duplicate entries. To prevent the duplicates from causing errors the list of entities to be added would have to be pre-filtered by the user to contain no duplicates, which means extra work in checking for duplicates and extra work in building a filtered list of entities before addition.

It would be good if there would be a way to add entities to an `Entities` instance which would allow duplicates, but only if the added entity is exactly the same as the entity already present. This may be made the default mode of adding entities, or it may be a separate API. By doing this internally inside the library, any performance penalty of extra duplicate checking and filtering is eliminated, and only the penalty of comparing entities for equivalence remains – a check which is a beneficial sanity check with relatively low cost.

Alternative for the user is to not keep any `Entities` instances alive except momentarily. Any precomputed entity lists for principal and resource may be kept in data structures that are not handled by this library, deduplicated in those data structures and then simply converted to `Entities` just before the authorization check.

### Additional context

_No response_

- [X] 👋 I may be able to implement this feature request

katherine-hough:
## Description of changes
* Changed Entities::add_entities and Entities::from_entities to ignore identical entities with the same EUID
* Refactored common code from Entities::add_entities and Entities::create_entity_map into private function
* Updated documentation to describe the new behavior
* Modified tests to reflect the new behavior
* Added tests for the new behavior

## Issue #, if available
Closes #613 

### DIFF
--- a/cedar-policy-core/src/entities.rs
+++ b/cedar-policy-core/src/entities.rs
@@ -384,11 +384,12 @@ fn create_entity_map(
 /// if a structurally equal entity is found, the state of the map is unchanged.
 fn update_entity_map(map: &mut HashMap<EntityUID, Arc<Entity>>, entity: Arc<Entity>) -> Result<()> {
     match map.entry(entity.uid().clone()) {
-        hash_map::Entry::Occupied(entry) => {
+        hash_map::Entry::Occupied(occupied_entry) => {
             // Check whether the occupying entity is structurally equal to the
             // entity being processed
-            if !entity.deep_eq(entry.get()) {
-                return Err(EntitiesError::duplicate(entity.uid().clone()));
+            if !entity.deep_eq(occupied_entry.get()) {
+                let entry = occupied_entry.remove_entry();
+                return Err(EntitiesError::duplicate(entry.0));
             }
         }
         hash_map::Entry::Vacant(v) => {
--- a/cedar-policy-core/src/entities.rs
+++ b/cedar-policy-core/src/entities.rs
@@ -123,8 +123,10 @@ impl Entities {
     }
 
     /// Adds the [`crate::ast::Entity`]s in the iterator to this [`Entities`].
-    /// Fails if the passed iterator contains any duplicate entities with this structure,
-    /// or if any error is encountered in the transitive closure computation.
+    /// Fails if there is a pair of non-identical entities in the passed iterator
+    /// with the same Entity UID, or there is an entity in the passed iterator
+    /// with the same Entity UID as a non-identical entity in this structure.
+    /// Fails if any error is encountered in the transitive closure computation.
     ///
     /// If `schema` is present, then the added entities will be validated
     /// against the `schema`, returning an error if they do not conform to the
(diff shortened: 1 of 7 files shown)


cedar-policy-core/src/entities.rs
```diff
@@ -368,22 +365,39 @@ impl Entities {
...
+/// with the same EntityUID as the specified entity. If such an entity is found and is
+/// not structurally equal to the specified entity produces an error. Otherwise,
+/// if a structurally equal entity is found, the state of the map is unchanged.
+fn update_entity_map(map: &mut HashMap<EntityUID, Arc<Entity>>, entity: Arc<Entity>) -> Result<()> {
+    match map.entry(entity.uid().clone()) {
+        hash_map::Entry::Occupied(entry) => {
+            // Check whether the occupying entity is structurally equal to the
+            // entity being processed
+            if !entity.deep_eq(entry.get()) {
+                return Err(EntitiesError::duplicate(entity.uid().clone()));
```
john-h-kastner-aws, line 391: nit: maybe we could avoid this `clone` by re-taking ownership of the key with `entry.remove_entry()`

cedar-policy-core/src/entities.rs
```diff
@@ -368,22 +365,39 @@ impl Entities {
...
+        update_entity_map(&mut map, e)?;
     }
     Ok(map)
 }
 
+/// Adds an entry to the specified map associating the EntityUID of the specified entity
+/// to the specified entity. Checks whether there is an entity already in the map
+/// with the same EntityUID as the specified entity. If such an entity is found and is
+/// not structurally equal to the specified entity produces an error. Otherwise,
+/// if a structurally equal entity is found, the state of the map is unchanged.
```
adpaco-aws, line 384: Is this still an unexpected scenario? I'm wondering if we need to emit warning in that case.

cedar-policy-core/src/entities.rs
```diff
@@ -817,18 +831,58 @@ mod json_parsing_tests {
...
+
+    #[test]
+    fn add_inconsistent_duplicates_in_iterator() {
+        let parser: EntityJsonParser<'_, '_> =
+            EntityJsonParser::new(None, Extensions::all_available(), TCComputation::ComputeNow);
+        // Create the entities to be added
+        let new = serde_json::json!([
+            {"uid":{ "type" : "Test", "id" : "ruby" }, "attrs" : {"location": "France"}, "parents" : []},
+            {"uid":{ "type" : "Test", "id" : "jeff" }, "attrs" : {"location": "France"}, "parents" : []},
+            {"uid":{ "type" : "Test", "id" : "jeff" }, "attrs" : {}, "parents" : []}]);
```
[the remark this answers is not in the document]
katherine-hough: The no inconsistent duplicates invariant is preserved by EntityJsonParser::from_json_str since that will call Entities::from_entities to construct the entities instance which will check for inconsistent duplicate EUIDs. The test cedar_policy_core::entities::json::entities::test::reject_inconsistent_duplicates checks this. Is there another  way of constructing `Entities` from JSON that I missed?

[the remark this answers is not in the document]
adpaco-aws: I don't think so. Just checked the [docs for `Entities`](https://docs.rs/cedar-policy/latest/cedar_policy/struct.Entities.html) and, as long as we're checking in `add_entities*`, `from_entities` and `from_json*`, we should be good.

adpaco-aws (approved):

cedar-policy-core/src/entities.rs
adpaco-aws: Sounds like there isn't, no need to do anything here.

cedar-policy-core/src/entities.rs
katherine-hough: Good call. Thanks!

PR #1481, opened 2025-02-18
adds upsert_entities

a0js:
## Description of changes

Adds `upsert_entities` to the Entities implementation. Internally, it removes any entities that match the uid and performs an add.

### DIFF
--- a/cedar-policy-core/src/est.rs
+++ b/cedar-policy-core/src/est.rs
@@ -124,6 +124,14 @@ impl Policy {
             annotations: self.annotations,
         })
     }
+
+    /// Returns true if this policy is a template, i.e., it has at least one slot.
+    pub fn is_template(&self) -> bool {
+        self.principal.has_slot()
+            || self.action.has_slot()
+            || self.resource.has_slot()
+            || self.conditions.iter().any(|c| c.has_slot())
+    }
 }
 
 impl Clause {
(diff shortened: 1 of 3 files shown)


a0js:
This still requires iterating through all entities per updated entity in remove_entities which may not be the best alternative. I'm just not sure where to add an update capability.

cdisselkoen (approved):
Looks reasonable to me.

chaluli (changes requested):
This looks reasonable. Could you (@a0js) resolve the failing DCO status by signing off on your commit.

a0js:
Looking at this again, I don't really like the implementation here as we are iterating over the entire entities object just to update an entity. As one who is implementing my own entity store, this creates a lot of processing time. I'd like to take another stab at this potentially at the cedar-policy-core level.

@cdisselkoen @chaluli Could you help me understand why the add_entities function in the cedar-policy-core returns an error when an entity already exists, rather than updating it? (see https://docs.rs/cedar-policy-core/4.3.3/src/cedar_policy_core/entities.rs.html#150)

Is there a strict immutability requirement that I'm not aware of? Could I contribute a change in cedar-policy-core that adds an upsert functionality at that level, directly interacting with the the entities hashmap, allowing duplicates to override previous entries and then update this PR to use that functionality?

a0js:
Replaced by #1531

PR #1489, opened 2025-02-24
Add support for missing operands, invalid method names and invalid variables to 'tolerant-ast' experimental feature

eidelmanjonathan:
## Description of changes

This change expands the 'tolerant-ast' experimental feature to allow for several more error cases to be parsed into partial AST's. This allows us to succeed even if the policy has errors in it.

This change allows us to support the following new cases which will be helpful in creating a better LSP Autocomplete feature in the future:

1. Missing operands, eg: `permit(principal ==, action, resource);`
2. Invalid method names, eg: `permit ( principal,action,resource) when { ip(principal.ip).i() };`
3. Invalid variable, eg: `permit ( principal,action,resource) when { pri }`

- [ x ] A change (breaking or otherwise) that only impacts unreleased or experimental code.

- [ x ] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).

- [x  ] Does not require updates because my change does not impact the Cedar formal model or DRT infrastructure.

- [ x ] Does not require updates because my change does not impact the Cedar language specification.

### DIFF
--- a/cedar-policy-core/src/ast/entity.rs
+++ b/cedar-policy-core/src/ast/entity.rs
@@ -25,8 +25,7 @@ use crate::FromNormalizedStr;
 use educe::Educe;
 use itertools::Itertools;
 use miette::Diagnostic;
-use ref_cast::RefCast;
-use serde::{Deserialize, Serialize};
+use serde::{de::Deserializer, ser::Serializer, Deserialize, Serialize};
 use serde_with::{serde_as, TryFromInto};
 use smol_str::SmolStr;
 use std::collections::{BTreeMap, HashMap, HashSet};
@@ -34,15 +33,56 @@ use std::str::FromStr;
 use std::sync::Arc;
 use thiserror::Error;
 
+#[cfg(feature = "tolerant-ast")]
+static ERROR_NAME: std::sync::LazyLock<Name> =
+    std::sync::LazyLock::new(|| Name(InternalName::from(Id::new_unchecked("EntityTypeError"))));
+
+#[cfg(feature = "tolerant-ast")]
+static ERROR_EID_SMOL_STR: std::sync::LazyLock<SmolStr> =
+    std::sync::LazyLock::new(|| SmolStr::from("Eid::ErrorEid"));
+
 /// The entity type that Actions must have
 pub static ACTION_ENTITY_TYPE: &str = "Action";
 
-/// Entity type names are just [`Name`]s, but we have some operations on them specific to entity types.
-#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone, Hash, PartialOrd, Ord, RefCast)]
+// #[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone, Hash, PartialOrd, Ord, RefCast)]
+// #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+// #[serde(transparent)]
+// #[repr(transparent)]
+// pub struct EntityTypeImpl(Name);
+
+#[derive(PartialEq, Eq, Debug, Clone, Hash, PartialOrd, Ord)]
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
-#[serde(transparent)]
-#[repr(transparent)]
-pub struct EntityType(Name);
+/// Entity type - can be an error type when 'tolerant-ast' feature is enabled
+pub enum EntityType {
+    /// Entity type names are just [`Name`]s, but we have some operations on them specific to entity types.
+    EntityTypeImpl(Name),
+    #[cfg(feature = "tolerant-ast")]
+    /// Represents an error node of an entity that failed to parse
+    ErrorEntityType,
+}
+
+impl<'de> Deserialize<'de> for EntityType {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let name = Name::deserialize(deserializer)?;
+        Ok(EntityType::EntityTypeImpl(name))
+    }
+}
+
+impl Serialize for EntityType {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match self {
+            EntityType::EntityTypeImpl(name) => name.serialize(serializer),
+            #[cfg(feature = "tolerant-ast")]
+            EntityType::ErrorEntityType => serializer.serialize_str("EntityType::Error"),
+        }
+    }
+}
 
 impl EntityType {
     /// Is this an Action entity type?
@@ -50,22 +90,42 @@ impl EntityType {
     /// base name for the type, so this will return true for any entity type named
     /// `Action` regardless of namespaces.
     pub fn is_action(&self) -> bool {
-        self.0.as_ref().basename() == &Id::new_unchecked(ACTION_ENTITY_TYPE)
+        match self {
+            EntityType::EntityTypeImpl(name) => {
+                name.as_ref().basename() == &Id::new_unchecked(ACTION_ENTITY_TYPE)
+            }
+            #[cfg(feature = "tolerant-ast")]
+            EntityType::ErrorEntityType => false,
+        }
     }
 
     /// The name of this entity type
     pub fn name(&self) -> &Name {
-        &self.0
+        match self {
+            EntityType::EntityTypeImpl(name) => &name,
+            #[cfg(feature = "tolerant-ast")]
+            EntityType::ErrorEntityType => &ERROR_NAME,
+        }
     }
 
     /// The source location of this entity type
     pub fn loc(&self) -> Option<&Loc> {
-        self.0.as_ref().loc()
+        match self {
+            EntityType::EntityTypeImpl(name) => name.as_ref().loc(),
+            #[cfg(feature = "tolerant-ast")]
+            EntityType::ErrorEntityType => None,
+        }
     }
 
     /// Calls [`Name::qualify_with_name`] on the underlying [`Name`]
     pub fn qualify_with(&self, namespace: Option<&Name>) -> Self {
-        Self(self.0.qualify_with_name(namespace))
+        match self {
+            EntityType::EntityTypeImpl(name) => {
+                Self::EntityTypeImpl(name.qualify_with_name(namespace))
+            }
+            #[cfg(feature = "tolerant-ast")]
+            EntityType::ErrorEntityType => Self::ErrorEntityType,
+        }
     }
 
     /// Wraps [`Name::from_normalized_str`]
@@ -76,19 +136,27 @@ impl EntityType {
 
 impl From<Name> for EntityType {
     fn from(n: Name) -> Self {
-        Self(n)
+        Self::EntityTypeImpl(n)
     }
 }
 
 impl From<EntityType> for Name {
     fn from(ty: EntityType) -> Name {
-        ty.0
+        match ty {
+            EntityType::EntityTypeImpl(name) => name,
+            #[cfg(feature = "tolerant-ast")]
+            EntityType::ErrorEntityType => ERROR_NAME.clone(),
+        }
     }
 }
 
 impl AsRef<Name> for EntityType {
     fn as_ref(&self) -> &Name {
-        &self.0
+        match self {
+            EntityType::EntityTypeImpl(name) => name,
+            #[cfg(feature = "tolerant-ast")]
+            EntityType::ErrorEntityType => &ERROR_NAME,
+        }
     }
 }
 
@@ -96,13 +164,17 @@ impl FromStr for EntityType {
     type Err = ParseErrors;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        s.parse().map(Self)
+        s.parse().map(Self::EntityTypeImpl)
     }
 }
 
 impl std::fmt::Display for EntityType {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.0)
+        match self {
+            EntityType::EntityTypeImpl(name) => write!(f, "{}", name),
+            #[cfg(feature = "tolerant-ast")]
+            EntityType::ErrorEntityType => write!(f, "EntityType::Error"),
+        }
     }
 }
 
@@ -126,12 +198,11 @@ pub struct EntityUIDImpl {
 #[derive(Educe, Serialize, Deserialize, Debug, Clone)]
 #[educe(PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub enum EntityUID {
+    /// Unique ID for an entity. These represent entities in the AST
     EntityUID(EntityUIDImpl),
     #[cfg(feature = "tolerant-ast")]
-    Error{
-        eid: Eid,
-        ty: EntityType,
-    }
+    /// Represents the ID of an error that failed to parse
+    Error,
 }
 
 impl StaticallyTyped for EntityUID {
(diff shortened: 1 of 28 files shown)


cedar-policy-core/src/ast/entity.rs
```diff
@@ -25,47 +25,101 @@ use crate::FromNormalizedStr;
...
 use std::sync::Arc;
 use thiserror::Error;
 
+#[cfg(feature = "tolerant-ast")]
+static ERROR_NAME: std::sync::LazyLock<Name> =
+    std::sync::LazyLock::new(|| Name(InternalName::from(Id::new_unchecked("EntityTypeError"))));
+
+#[cfg(feature = "tolerant-ast")]
+static ERROR_EID_SMOL_STR: std::sync::LazyLock<SmolStr> =
+    std::sync::LazyLock::new(|| SmolStr::from("Eid::ErrorEid"));
```
cdisselkoen, line 42: nit, but I wonder if a `std::sync::LazyLock` is worth it when all we're saving is the construction of one `SmolStr`.  (the alternative being `pub static ERROR_EID: &str` and just do `SmolStr::from` every time the `SmolStr` is needed)

cedar-policy-core/src/ast/entity.rs
```diff
@@ -25,47 +25,101 @@ use crate::FromNormalizedStr;
...
-#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone, Hash, PartialOrd, Ord, RefCast)]
+#[derive(PartialEq, Eq, Debug, Clone, Hash, PartialOrd, Ord)]
 #[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
-#[serde(transparent)]
-#[repr(transparent)]
-pub struct EntityType(Name);
+/// Entity type - can be an error type when 'tolerant-ast' feature is enabled
+pub enum EntityType {
+    /// Entity type names are just [`Name`]s, but we have some operations on them specific to entity types.
+    EntityTypeImpl(Name),
```
cdisselkoen, line 52:
```suggestion
    EntityType(Name),
```

cedar-policy-core/src/ast/entity.rs
```diff
@@ -203,7 +326,16 @@ impl EntityUID {
...
-        write!(f, "{}::\"{}\"", self.entity_type(), self.eid.escaped())
+        match self {
+            EntityUID::EntityUID(entity_uid) => write!(
+                f,
+                "{}::\"{}\"",
+                self.entity_type(),
+                entity_uid.eid.escaped()
+            ),
+            #[cfg(feature = "tolerant-ast")]
+            EntityUID::Error => write!(f, "Expr::ExprError"),
```
cdisselkoen, line 337: This isn't a valid EntityUID, maybe we should still use `self.entity_type()` like in the other case, and then have something inside quotes so it's a valid EntityUID

cedar-policy-core/src/ast/entity.rs
```diff
@@ -242,38 +374,79 @@ impl<'a> arbitrary::Arbitrary<'a> for EntityUID {
 /// (see [#884](https://github.com/cedar-policy/cedar/issues/884)).
 /// To get an escaped representation, use `.escaped()`.
 /// To get an unescaped representation, use `.as_ref()`.
-#[derive(Serialize, Deserialize, PartialEq, Eq, Debug, Clone, Hash, PartialOrd, Ord)]
-pub struct Eid(SmolStr);
+#[derive(PartialEq, Eq, Debug, Clone, Hash, PartialOrd, Ord)]
+pub enum Eid {
+    /// Actual Eid
+    EidImpl(SmolStr),
```
cdisselkoen, line 380:
```suggestion
    Eid(SmolStr),
```

cedar-policy-core/src/parser/cst.rs
```diff
@@ -60,6 +60,16 @@ pub struct Policy {
     pub conds: Vec<Node<Cond>>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// Policy statement - can be an Error node when 'tolerant-ast' feature is switched on
+pub enum Policy {
+    /// Policy that has been parsed successfully
+    PolicyImpl(PolicyImpl),
```
cdisselkoen, line 67:
```suggestion
    Policy(PolicyImpl),
```

cedar-policy-core/src/parser/cst.rs
```diff
@@ -60,6 +60,16 @@ pub struct Policy {
     pub conds: Vec<Node<Cond>>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// Policy statement - can be an Error node when 'tolerant-ast' feature is switched on
+pub enum Policy {
+    /// Policy that has been parsed successfully
```
cdisselkoen, line 66: Nit, `Policy` can be created through other means than parsing

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -169,7 +169,7 @@ impl Node<Option<cst::Policy>> {
         self.to_policy_template(id)
     }
 
-    /// Convert `cst::Policy` to an AST `StaticPolicy` or `Template`
+    /// Convert `cst::Policy` to an AST `InlinePolicy` or `Template`
```
cdisselkoen, line 172: `InlinePolicy` does not exist. The correct terminology is `StaticPolicy`

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -823,15 +914,71 @@ impl Node<Option<cst::VariableDef>> {
...
-                        .into()
+            let action_constraint_res = action_constraint.contains_only_action_types();
+
+            action_constraint_res.map_err(|non_action_euids| {
+                rel_expr
+                    .to_ast_err(parse_errors::InvalidActionType {
+                        euids: non_action_euids,
+                    })
+                    .into()
+            })
```
cdisselkoen, line 925: this refactor can probably be reverted? the original code was more concise and I'm not sure that giving a name to `action_constraint_res` helps the readability

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -5330,22 +5602,6 @@ mod tests {
...
-                .build(),
-        );
-        expect_some_error_matches(
-            src,
-            &errs,
-            &ExpectedErrorMessageBuilder::error("invalid variable: u")
-                .help("the valid Cedar variables are `principal`, `action`, `resource`, and `context`; did you mean to enclose `u` in quotes to make a string?")
-                .exactly_one_underline("u")
-                .build(),
-        );
```
cdisselkoen, line 5348: can we leave these tests in?  (gated by `#[cfg(not(feature = "tolerant-ast"))]` if necessary)

cedar-policy-core/src/parser/text_to_cst.rs
```diff
@@ -69,6 +69,40 @@ fn parse_collect_errors<'a, P, T>(
     }
 }
 
+/// This helper function calls a generated parser. If the given string is unparsable, it will return the relevant errors
+/// If a string is parsable but has errors, it will still return the parse expression
+/// NOTE: This should only be used to construct an AST that includes error nodes and NOT for evaluation
+/// [`err::ParseErrors`].
```
cdisselkoen, line 75: Last line was probably mistakenly copied from somewhere else?

cedar-policy-core/src/parser/text_to_cst.rs
```diff
@@ -1003,6 +1061,22 @@ mod tests {
...
+            policies
+                .clone()
+                .0
+                .into_iter()
+                .filter_map(|p| p.node)
+                .count(),
+            3
+        );
+        #[cfg(feature = "tolerant-ast")]
+        return ();
```
cdisselkoen, line 1077: instead of this, we could guard the last assert (below) with `cfg(not(feature = "tolerant-ast"))`

cdisselkoen (approved):
Looks good

cedar-policy-core/src/parser/cst_to_ast.rs
eidelmanjonathan: Oops, this got snuck in during rebasing. Will fix

cedar-policy-core/src/parser/cst_to_ast.rs
eidelmanjonathan: Yup, this made sense in an earlier version of how I wanted to do this and forgot to change it back. Will fix

cedar-policy-core/src/parser/cst_to_ast.rs
eidelmanjonathan: Will fix

cedar-policy-core/src/parser/text_to_cst.rs
eidelmanjonathan: Sure, will do

cedar-policy-core/src/parser/text_to_cst.rs
eidelmanjonathan: Yes, will fix

cedar-policy-core/src/ast/entity.rs
eidelmanjonathan: Ah never mind, actually the issue here is that we can't return a reference to a local variable. I think the simplest way is what is already here, I'll leave it

cedar-policy-core/src/parser/cst.rs
eidelmanjonathan: Good point, will fix

cedar-policy-core/src/parser/cst_to_ast.rs
eidelmanjonathan: Ah sorry, looked closer at this. This whole test is guarded ` #[cfg(feature = "tolerant-ast")]` and uses the tolerant version of the parser which is why it has fewer error messages - these invalid expressions are correctly being translated into error nodes in the tolerant AST

There is another, separate test `show_policy1` which uses the normal parsing methods and it has these test cases

cdisselkoen: Cool 👍🏻

cedar-policy-core/src/ast/entity.rs
```diff
@@ -203,7 +324,16 @@ impl EntityUID {
...
-        write!(f, "{}::\"{}\"", self.entity_type(), self.eid.escaped())
+        match self {
+            EntityUID::EntityUID(entity_uid) => write!(
+                f,
+                "{}::\"{}\"",
+                self.entity_type(),
+                entity_uid.eid.escaped()
+            ),
+            #[cfg(feature = "tolerant-ast")]
+            EntityUID::Error => write!(f, "{}::\"ExprError\"", self.entity_type()),
```
adpaco-aws, line 335: You don't this specialization, right? We could just use the `entity_uid.eid()` instead of `ExprError`.

cedar-policy-core/src/ast/entity.rs
```diff
@@ -242,38 +372,79 @@ impl<'a> arbitrary::Arbitrary<'a> for EntityUID {
...
+        Eid::Eid(eid.into())
     }
 
     /// Get the contents of the `Eid` as an escaped string
     pub fn escaped(&self) -> SmolStr {
-        self.0.escape_debug().collect()
+        match self {
+            Eid::Eid(smol_str) => smol_str.escape_debug().collect(),
+            #[cfg(feature = "tolerant-ast")]
+            Eid::ErrorEid => "Eid::Error".into(),
```
adpaco-aws, line 418: We should put the string in a constant and use it everywhere (`serialize`, `as_ref`, etc.)

cedar-policy-core/src/est/scope_constraints.rs
```diff
@@ -461,6 +467,8 @@ impl std::fmt::Display for ActionConstraint {
                 std::fmt::Display::fmt(aic, f)?;
                 Ok(())
             }
+            #[cfg(feature = "tolerant-ast")]
+            Self::ErrorConstraint => write!(f, "action"),
```
adpaco-aws, line 471: This is not the same you were doing for other `Display` implementations.

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -906,9 +1059,10 @@ fn build_ast_error_node_if_possible<Build: ExprBuilder>(error: ParseErrors) -> R
 /// Since ExprBuilder ErrorType can be Infallible or ParseErrors, if we get an error from building the node pass the ParseErrors along
 fn convert_expr_error_to_parse_error<Build: ExprBuilder>(
     error: ParseErrors,
+    #[cfg(feature = "tolerant-ast")] loc: Option<&Loc>,
```
adpaco-aws, line 1062: Is this a good idea? Especially when the argument is an option type?

cedar-policy-core/src/ast/entity.rs
```diff
@@ -25,47 +25,99 @@ use crate::FromNormalizedStr;
...
+
+impl Serialize for EntityType {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        match self {
+            EntityType::EntityType(name) => name.serialize(serializer),
+            #[cfg(feature = "tolerant-ast")]
+            EntityType::ErrorEntityType => serializer.serialize_str("EntityType::Error"),
```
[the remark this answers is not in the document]
eidelmanjonathan: We don't really have an `ERROR_NAME` to speak of here - basically these error types just get created when part of an expression fails to parse, so we don't really have more detail than that. This is just a placeholder node so we can create an EntityUID without breaking public methods which might return a type

What I will do is make this into a string constant

cedar-policy-core/src/ast/entity.rs
```diff
@@ -76,40 +128,53 @@ impl EntityType {
...
     }
 }
 
 impl std::fmt::Display for EntityType {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.0)
+        match self {
+            EntityType::EntityType(name) => write!(f, "{}", name),
+            #[cfg(feature = "tolerant-ast")]
+            EntityType::ErrorEntityType => write!(f, "Error"),
```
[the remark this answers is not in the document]
eidelmanjonathan: Will do

cedar-policy-core/src/ast/entity.rs
eidelmanjonathan: Woops, good point

cedar-policy-core/src/ast/entity.rs
eidelmanjonathan: Will do

cedar-policy-core/src/est/scope_constraints.rs
eidelmanjonathan: Good catch, will fix

cedar-policy-core/src/parser/cst_to_ast.rs
eidelmanjonathan: So the reason for this is to avoid an unused variable when not using the `tolerant-ast`. I don't see an issue with it but if you prefer I can just suppress the clippy error

adpaco-aws: I'd rather suppress the warning then. Note that you can also suppress it conditionally with `#[cfg(not(feature = "tolerant-ast")), ...]` so it's easier to tell why it's there.

cedar-policy-core/src/ast/entity.rs
```diff
@@ -203,7 +330,16 @@ impl EntityUID {
...
+        match self {
+            EntityUID::EntityUID(entity_uid) => write!(
+                f,
+                "{}::\"{}\"",
+                self.entity_type(),
+                entity_uid.eid.escaped()
+            ),
+            #[cfg(feature = "tolerant-ast")]
+            EntityUID::Error => write!(f, "{}::\"{}\"", self.entity_type(), self.eid().escaped()),
+        }
```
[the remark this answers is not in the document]
eidelmanjonathan: Hah! Right you are :) Will do

cedar-policy-core/src/parser/cst_to_ast.rs
eidelmanjonathan: Ah nice, I can suppress conditionally sure

adpaco-aws (approved):

Issue #1201, opened 2024-09-12
Run nightly CI on release branches


khieta:
### Describe the improvement you'd like to request

We should be running nightly CI on our supported release branches to make sure everything builds as expected. Not doing this caused us to miss that the CI on the 2.4.x branch was failing (due to changes in the Rust compiler) until PR #1182. This is annoying for PR authors and has the potential to mask other issues.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this internal improvement
- [ ] ⚠️ This feature might incur a breaking change

PR #1494, opened 2025-02-24
Added Coverage Reports to CI

katherine-hough:
## Description of changes
Changed the "Nightly build" workflow to collect code coverage information and upload code coverage reports as workflow artifacts. These reports are only retained for 2 days to prevent storage build up. These reports are used to include information comparing the coverage of the HEAD of the PR to the coverage of the BASE of the PR.

Changed the CI "Cargo Build & Test" workflow that is run for pull requests to collect code coverage information and upload code coverage reports as workflow artifacts. These reports will be retained for the repository default of 90 days. These reports are processed as part of the "Cargo Build & Test" workflow to create a markdown comment summarizing the reports. Finally, a check is performed that fails the build if the observed coverage drops below a threshold value (currently set at 80%). Both the overall coverage for the repository and the coverage of added code is assessed for this check.

The produced markdown comment is posted to the PR that triggered the "Cargo Build & Test" workflow. This change will not be reflect until this change is merged into the default branch because the newly added "Comment on the pull request" workflow is triggered by a "workflow_run" event type and, therefore, must be present on the default branch to take effect. This separation and choice of event type is necessary for security reasons.

In order to demonstrate these changes, I created two pull requests within my fork of cedar:

In [Example PR 1](https://github.com/katherine-hough/cedar/pull/8), we see a PR that has added a new function, but not added any tests. The build is failed because less than 80% of the added code is covered by the test suite. 
[A comment is added the PR](https://github.com/katherine-hough/cedar/pull/8#issuecomment-2679773335) summarizing the coverage results. This comments shows a failure under the section "Coverage of Added or Modified Lines of Rust Code". If the "details" text is clicked, it shows that out of the ten lines of code added to the file "cedar-policy-core/src/entities.rs" zero of them were covered. It also listed which lines specifically were not covered. In the section "Coverage of All Lines of Rust Code", we can see the that overall coverage for the repository has not dropped below the 80% threshold. However, when the "details" text is clicked it reveals that coverage for the package "cedar-policy-core" has dropped from 91.50% to 91.47%.

In [Example PR 2](https://github.com/katherine-hough/cedar/pull/9), we see a PR that has added a new function with an accompanying test case. In this case the build succeeds, because at least 80% of the added code is covered by the test suite. A [report comment](https://github.com/katherine-hough/cedar/pull/9#issuecomment-2679774263) similar to the one for Example 1 is created summarizing the coverage results for this PR.

### DIFF
--- a/.github/actions/create_coverage_comment/action.yml
+++ b/.github/actions/create_coverage_comment/action.yml
@@ -3,8 +3,6 @@ description: >
   Build a markdown comment to be posted to a PR by the "comment_pr" workflow.
   If threshold is set, fails the build if the reported coverage does not meet
   met expected criteria.
-  Downloads the cobertura coverage report for the current workflow run.
-  TODO
   Uploads the created markdown comment and the number of the PR that triggered this action as workflow artifacts.
 inputs:
   threshold:
--- a/.github/actions/create_coverage_comment/action.yml
+++ b/.github/actions/create_coverage_comment/action.yml
@@ -1,0 +1,98 @@
+name: "Create Coverage PR Comment"
+description: >
+  Build a markdown comment to be posted to a PR by the "comment_pr" workflow.
+  If threshold is set, fails the build if the reported coverage does not meet
+  met expected criteria.
+  Downloads the cobertura coverage report for the current workflow run.
+  TODO
+  Uploads the created markdown comment and the number of the PR that triggered this action as workflow artifacts.
+inputs:
+  threshold:
+    description: >
+      Minimum proportion of modified lines that need to be covered for acceptance or -1 if there is no minimum.
+    required: false
+    default: "-1"
+  report-artifact-prefix:
+    description: "Prefix of the name of the artifact to be downloaded"
+    required: false
+    default: "coverage_report"
+  comment-artifact-name:
+    description: "Name of the artifact to be uploaded"
+    required: false
+    default: "coverage_comment"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Download report artifacts
+      uses: actions/download-artifact@v4
+      with:
+        name: ${{ inputs.report-artifact-prefix }}_cobertura
+    - name: Compute changed lines
+      id: changed_lines
+      uses: hestonhoffman/changed-lines@v1
+      with:
+        file_filter: ".rs"
+    - name: Set up python
+      uses: actions/setup-python@v5
+      with:
+        python-version: 3.13
+    - name: Find last nightly run for BASE
+      shell: bash
+      run: |
+        gh run list \
+          --commit $BASE_SHA \
+          --status "success" \
+          --workflow  "Nightly Build" \
+          --limit 1 \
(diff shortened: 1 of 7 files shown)


john-h-kastner-aws (commented):
Looks awesome based on the example PRs.

One initial comment. I looked at the coverage report HTML and it's counting test code as part of the total lines. Can we exclude all code after a `#[cfg(test)]` from the total?

katherine-hough:
@john-h-kastner-aws  Unfortunately, as far as I can tell neither [grcov](https://github.com/mozilla/grcov/issues/380) nor [rustc](https://github.com/rust-lang/rust/issues/68703) have explicit support for ignoring tests code. There is support in grcov for ignoring directories and files, so I will add "tests" directories to the ignore list and also build scripts (build.rs). Grcov also allows users to specify regexes to delimit regions to be excluded. So, I could exclude everything between ```^\#\[cfg\(test\)\]``` and ```^}```. This makes assumptions about the format. Do you think that is okay?

cdisselkoen:
That seems ok to me.  If the exclusions end up being wrong, we'll presumably notice at some point during PR reviews and can make tweaks in the future.

john-h-kastner-aws:
> So, I could exclude everything between `^\#\[cfg\(test\)\]` and `^}`.

Sounds fine to me if the results looks reasonable.

> This makes assumptions about the format. Do you think that is okay?

Maybe we could include something like `^// STOP COVERAGE` and `^// START COVERAGE` so we can tweak the behavior when these assumption fail

cdisselkoen (approved):

katherine-hough:
Matching on ```// STOP COVERAGE``` may cause some confusion because it will still match on ```}```. For example, you could have something like:
```
// STOP COVERAGE
#[test]
fn my_test() {
   // ...
}

#[test]
fn my_test2() {
   // ...
}
// START COVERAGE
```
It will match the first outer level ```}``` as the end of the exclusion not the final ```// START COVERAGE``` which is unintuitive. Should I still add it?

cdisselkoen:
My two cents: don't add it until it's needed.  Right now we don't have a motivating case where it would be needed, so we're designing against a hypothetical situation

.github/actions/create_coverage_comment/action.yml
```diff
@@ -0,0 +1,95 @@
...
+    - name: Compute changed lines
+      id: changed_lines
+      uses: hestonhoffman/changed-lines@v1
+      with:
+        file_filter: ".rs"
+    - name: Set up python
+      uses: actions/setup-python@v5
+      with:
+        python-version: 3.13
+    - name: Find last nightly run for BASE
```
john-h-kastner-aws, line 38: I think using the nightly run for the `BASE` won't work properly for any PRs to our release branches since we only do a nightly build of `main` (see #1201). I'm not particularly worried about this, but maybe this could check if it's running on `main`?

john-h-kastner-aws (approved):
One comment but happy to merge this as is

PR #1495, opened 2025-02-25
more simplification of protobuf schema format (followups to 1488)

cdisselkoen:
## Description of changes

More simplifications of the Protobuf format for `Schema`, some suggested in #1488 and some based on conversations with @john-h-kastner-aws offline.

- resolving the TODO: `Schema` is just a list of `EntityDecl` and a list of `ActionDecl`, not a map, as this is simpler and sufficient
- removing the ability to have open attributes, as this is not a stable feature (in both entities and records)
- the `context` for an Action must be a record, so instead of `Type`, we require it to be a `map<string, AttributeType>`
- rename `Type.Ty` to `Type.Prim` which is more clear
- remove the `Never`, `True`, `False`, `EmptySet`, and `AnyEntity` types, as these should never appear in schemas
- remove the `ActionEntity` type as action entities can just be encoded as `Entity` type (we do not support action attributes in Protobuf, as that is not a stable feature)
- remove `EntityRecordKind` in favor of inlining it into `Type`
- a few field renumberings, which is a breaking change we can make now but not after stabilization

### DIFF
--- a/cedar-policy-validator/src/types.rs
+++ b/cedar-policy-validator/src/types.rs
@@ -1075,13 +1075,14 @@ impl IntoIterator for Attributes {
 
 /// Used to tag record types to indicate if their attributes record is open or
 /// closed.
-#[derive(Hash, Ord, PartialOrd, Eq, PartialEq, Debug, Copy, Clone, Serialize)]
+#[derive(Hash, Ord, PartialOrd, Eq, PartialEq, Debug, Copy, Clone, Serialize, Default)]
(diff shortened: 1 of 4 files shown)


adpaco-aws (approved):

john-h-kastner-aws (approved):
looks like our new coverage check is correctly complaining about the lack of tests for validator protobuf. should ofc add tests before stabilizing, but we can bypass for now

shaobo-he-aws (approved):

PR #1506, opened 2025-02-26
more followups on Protobuf formats

cdisselkoen:
## Description of changes

- Remove the `Mode` for Entities, as we're not supported partial evaluation in protobuf currently.
- One more inlining, this time in `Expr`

### DIFF
--- a/cedar-policy/src/proto/entities.rs
+++ b/cedar-policy/src/proto/entities.rs
@@ -18,53 +18,35 @@
 
 use super::models;
 use cedar_policy_core::{ast, entities, extensions};
-use std::sync::Arc;
 
 impl From<&models::Entities> for entities::Entities {
     // PANIC SAFETY: experimental feature
     #[allow(clippy::expect_used)]
     fn from(v: &models::Entities) -> Self {
-        let entities: Vec<Arc<ast::Entity>> = v
-            .entities
-            .iter()
-            .map(|e| Arc::new(ast::Entity::from(e)))
-            .collect();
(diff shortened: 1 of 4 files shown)


cedar-policy/src/proto/entities.rs
```diff
@@ -18,53 +18,35 @@
...
-
-        result
-            .add_entities(
-                entities,
-                None::<&entities::NoEntitiesSchema>,
-                entities::TCComputation::AssumeAlreadyComputed,
-                extensions::Extensions::none(),
-            )
-            .expect("Should be able to add entities")
+        // REVIEW (before stabilization): does `AssumeAlreadyComputed` make
```
john-h-kastner-aws, line 28: Seems like a reasonable choice to me.

IMO it'd be worse if the protobuf parser recomputed TC every time, but it could also include a flag to control this, right?

john-h-kastner-aws (approved):

cedar-policy/src/proto/entities.rs
cdisselkoen: Yeah, maybe ultimately we'll need a flag here.  I propose we should leave that out of scope for this PR, since it would require changing our `Protobuf` trait probably.  This PR preserves the previous TC behavior

john-h-kastner-aws: sounds good

cdisselkoen:
^ improved tests in one affected file so that our coverage bar now passes :) (wasn't this PR's fault, but now our tests are better)

Issue #612, opened 2024-01-30
Incremental rebuilding of the transitive closure for `Entities`


nakedible-p:
### Describe the improvement you'd like to request

Currently whenever any entities are added to an `Entities` instance, the transitive closure is rebuilt from scratch. If a schema is used, this also means that all the action entities are part of the set and also are included in the rebuild. Implementing incremental rebuilding of the transitive closure that only traverses the new objects to be added should be possible, which would reduce the performance penalty of adding a few entities to an already existing large set of entities. This is especially useful in cases where a principal is shared for many requests (the entities for the principal can be cached) while the resource is often absent but sometimes present (the entities for the resource should be added to the existing set of entities).

### Describe alternatives you've considered

An alternative would be to be able to specify two (or more) sets of entities for a single authorization check. In that case, there wouldn't be a need to copy all the `Entities` to form a union of the entities for the principal and the resource, but instead both could be cached independently. However, there would likely be a performance penalty in entity traversal to do this, and the code may end up complex. Also, adding just two sets of entities doesn't solve the general issue of possibly needing entities referenced by the Context into the entity set as well.

### Additional context

May require formal verification of the incremental algorithm.

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this internal improvement
- [ ] ⚠️ This feature might incur a breaking change

PR #1509, opened 2025-02-27
Updates Transitive closure algorithm.

chaluli:
## Description of changes

Updates Cedar's transitive closure computation. The algorithm is only sound when computing the transitive closure of DAGs and graphs with simple cycles. However, the updated transitive closure algorithm when combined with `enforce_dag_from_tc` is sufficient for detecting cycles, even complex cycles. The implementation also uses a slightly more efficient algorithm (that does only one scan of the input graph) if one does not care to enforce that the graph is a DAG (while sound for computing the exact transitive closure of a DAG, this variant is not sufficient for detecting even simple cycles).

This pull request also adds the `remove_tc` that is capable of repairing an existing graph's transitive closure when only a sub-set of the nodes within the graph need recalculated. This is used within `add_entities` and `remove_entities` to avoid re-computing the entire graph's transitive closure. This change addresses the feature request in issu #612

### DIFF
--- a/cedar-policy-core/src/entities.rs
+++ b/cedar-policy-core/src/entities.rs
@@ -18,8 +18,8 @@
 
 use crate::ast::*;
 use crate::extensions::Extensions;
-use crate::transitive_closure::{compute_tc, enforce_tc_and_dag};
-use std::collections::{hash_map, HashMap};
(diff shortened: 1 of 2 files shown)


victornicolet:
Closing this PR in favor of #2317 which is rebased on main.

PR #1522, opened 2025-03-05
Allow an invalid entity uid ref in principal and resource scope in the 'tolerant-ast' experimental feature

eidelmanjonathan:
## Description of changes

This change extends the `tolerant-ast` feature to support invalid entity refs in the resource and principal constraints of a policy. Currently, these result in errors when converting from CST to AST. This is a common case that comes up when doing autocomplete in the context of an LSP or IDE extension.

We will now support the following cases:

```
#[cfg(feature = "tolerant-ast")]
    #[test]
    fn parsing_with_errors_succeeds_with_invalid_uid_resource_constraint() {
        let src = r#"
            permit (
                principal,
                action,
                resource in H
            )
            when { true };
        "#;
        assert_parse_policy_allows_errors(src);
    }

    #[cfg(feature = "tolerant-ast")]
    #[test]
    fn parsing_with_errors_succeeds_with_invalid_uid_principal_constraint() {
        let src = r#"
            permit (
                principal in J,
                action,
                resource
            )
            when { true };
        "#;
        assert_parse_policy_allows_errors(src);
    }
``` 

- [  ] A change "invisible" to users (e.g., documentation, changes to "internal" crates like `cedar-policy-core`, `cedar-validator`, etc.)
- [ x ] A change (breaking or otherwise) that only impacts unreleased or experimental code.

- [ x ] Does not update the CHANGELOG because my change does not significantly impact released code.

- [ x ] Does not require updates because my change does not impact the Cedar formal model or DRT infrastructure.

- [ x ] Does not require updates because my change does not impact the Cedar language specification.

### DIFF
--- a/cedar-policy-core/src/parser/cst_to_ast.rs
+++ b/cedar-policy-core/src/parser/cst_to_ast.rs
@@ -5638,7 +5638,7 @@ mod tests {
 
     #[cfg(feature = "tolerant-ast")]
     #[test]
-    fn parsing_with_errors_succeeds_with_invalid_uid_for_in() {
+    fn parsing_with_errors_succeeds_with_invalid_uid_resource_constraint() {
         let src = r#"
             permit (
                 principal,
@@ -5647,8 +5647,21 @@ mod tests {
             )
             when { true };
         "#;
-        let p = assert_parse_policy_allows_errors(src);
-        println!("{:?}", p);
+        assert_parse_policy_allows_errors(src);
+    }
+
+    #[cfg(feature = "tolerant-ast")]
+    #[test]
+    fn parsing_with_errors_succeeds_with_invalid_uid_principal_constraint() {
+        let src = r#"
+            permit (
+                principal in J,
+                action,
+                resource
+            )
+            when { true };
+        "#;
+        assert_parse_policy_allows_errors(src);
     }
 
     #[cfg(feature = "tolerant-ast")]
--- a/cedar-policy-core/src/parser/cst_to_ast.rs
+++ b/cedar-policy-core/src/parser/cst_to_ast.rs
@@ -1117,7 +1117,7 @@ impl Node<Option<cst::VariableDef>> {
                             return Err(self.to_ast_err(ToASTErrorKind::IsInActionScope).into());
                         }
                     }
-                    match rel_expr.to_refs(ast::Var::Action)? {
+                    match rel_expr.to_refs_tolerant_ast(ast::Var::Action)? {
                         OneOrMultipleRefs::Single(single_ref) => {
                             Ok(ActionConstraint::is_in([single_ref]))
                         }
@@ -1125,7 +1125,7 @@ impl Node<Option<cst::VariableDef>> {
                     }
                 }
                 cst::RelOp::Eq => {
-                    let single_ref = rel_expr.to_ref(ast::Var::Action)?;
+                    let single_ref = rel_expr.to_ref_tolerant_ast(ast::Var::Action)?;
                     Ok(ActionConstraint::is_eq(single_ref))
                 }
                 cst::RelOp::InvalidSingleEq => {
@@ -5669,6 +5669,21 @@ mod tests {
             when { true };
         "#;
         assert_parse_policy_allows_errors(src);
+    }
+
+    #[cfg(feature = "tolerant-ast")]
+    #[test]
+    fn repro_coles_bug() {
+        let src = r#"
+            permit (
+                principal,
+                action in A,
+                resource
+            )
+            when { true };
+        "#;
+        let p = assert_parse_policy_allows_errors(src);
+        println!("{:?}", p);
     }
 
     #[cfg(feature = "tolerant-ast")]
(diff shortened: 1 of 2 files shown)


cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -924,6 +957,72 @@ impl Node<Option<cst::VariableDef>> {
         }
     }
 
+    #[cfg(feature = "tolerant-ast")]
+    fn to_principal_or_resource_constraint_tolerant_ast(
+        &self,
+        expected: ast::Var,
+    ) -> Result<PrincipalOrResource> {
```
adpaco-aws, line 964: I don't think this amount of code duplication is acceptable. Either we refactor common parts and use them both here and in the "regular" implementation, or use conditional compilation on the latter and we don't have this function. The goal is to make it easier to spot the differences between the two implementations.

cedar-policy-core/src/parser/cst_to_ast/to_ref_or_refs.rs
```diff
@@ -253,6 +313,69 @@ impl Node<Option<cst::Primary>> {
                 .into()),
         }
     }
+
+    #[cfg(feature = "tolerant-ast")]
+    fn to_ref_or_refs_tolerant_ast<T: RefKind>(&self, var: ast::Var) -> Result<T> {
```
adpaco-aws, line 318: Same comment as earlier. I can't easily spot the difference between this implementation and the other one.

adpaco-aws (changes requested):

cedar-policy-core/src/parser/cst_to_ast.rs
eidelmanjonathan: Totally makes sense! Back in the original tolerant AST [PR](https://github.com/cedar-policy/cedar/pull/1470) there was some back and forth about this and decision was to accept a certain amount of repetition

But I think you're right that here it's a little too much, so I will refactor

cedar-policy-core/src/parser/cst_to_ast/to_ref_or_refs.rs
eidelmanjonathan: Will refactor

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -879,7 +930,11 @@ impl Node<Option<cst::VariableDef>> {
                     }
                 }
             }
-            let eref = rel_expr.to_ref_or_slot(var)?;
+            let eref = match tolerant_ast {
+                TolerantAstSetting::NotTolerant => rel_expr.to_ref_or_slot(var)?,
+                #[cfg(feature = "tolerant-ast")]
+                TolerantAstSetting::Tolerant => rel_expr.to_ref_or_slot_tolerant_ast(var)?,
```
adpaco-aws, line 936: You don't need an `enum` for this, right? You can simply use `#[cfg(not(feature = "tolerant-ast"))]` and `#[cfg(feature = "tolerant-ast")]` to handle each case separately.

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -834,9 +841,34 @@ enum PrincipalOrResource {
     Resource(ResourceConstraint),
 }
 
+enum TolerantAstSetting {
+    NotTolerant,
+    #[cfg(feature = "tolerant-ast")]
+    Tolerant,
+}
```
adpaco-aws, line 848: Why are we adding an `enum` like this?

eidelmanjonathan: I think this is a nice way to support both tolerant and non-tolerant ASTs even if the feature is enabled. Is there a better way to handle this that you know of?

cedar-policy-core/src/parser/cst_to_ast.rs
adpaco-aws: I see, that's what I was missing. Thanks!

adpaco-aws (approved):

cdisselkoen (approved):
Looks good

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -834,9 +841,34 @@ enum PrincipalOrResource {
...
+
+    #[cfg(feature = "tolerant-ast")]
+    fn to_principal_constraint_tolerant_ast(&self) -> Result<PrincipalConstraint> {
+        match self.to_principal_or_resource_constraint(
+            ast::Var::Principal,
+            TolerantAstSetting::Tolerant,
+        )? {
             PrincipalOrResource::Principal(p) => Ok(p),
             PrincipalOrResource::Resource(_) => Err(self
                 .to_ast_err(ToASTErrorKind::IncorrectVariable {
```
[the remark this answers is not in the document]
john-h-kastner-aws: Sounds like a nice design to me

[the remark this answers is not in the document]
eidelmanjonathan: Works for me, I'll refactor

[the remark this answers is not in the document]
eidelmanjonathan: Gonna do this for all the non public methods. Will leave the existing approach for public things so we don't change the existing API

PR #1548, opened 2025-03-21
add public API to get the valid request envs for a schema

cdisselkoen:
## Description of changes

We already had `Policy::get_valid_request_envs()` and `Template::get_valid_request_envs()`, but these return only the request envs in which a particular policy/template is not trivially false. (They also involve a bunch of typechecking work.)  This PR provides a public API to simply get all of the request envs for a schema.

### DIFF
--- a/cedar-policy-validator/src/schema.rs
+++ b/cedar-policy-validator/src/schema.rs
@@ -39,7 +39,8 @@ use crate::{
     cedar_schema::SchemaWarning,
     json_schema,
     partition_nonempty::PartitionNonEmpty,
-    types::{Attributes, EntityRecordKind, OpenTag, Type},
+    types::{Attributes, EntityRecordKind, OpenTag, RequestEnv, Type},
+    ValidationMode,
 };
 
 mod action;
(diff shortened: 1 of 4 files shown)


cedar-policy-validator/src/schema.rs
```diff
@@ -271,6 +272,44 @@ impl ValidatorSchema {
...
     }
 
+    /// Returns an iterator over every valid `RequestEnv` in the schema
+    pub fn unlinked_request_envs(
+        &self,
+        mode: ValidationMode,
+    ) -> impl Iterator<Item = RequestEnv<'_>> + '_ {
+        // Gather all of the actions declared in the schema.
+        let all_actions = self
+            .action_ids()
```
john-h-kastner-aws, line 282: This could probably just be `self.action_ids()` of maybe `self.action_ids.values()`

john-h-kastner-aws (approved):

cedar-policy-validator/src/schema.rs
cdisselkoen: Ha, I inherited this code (this is just moved), but you're right that the existing code is quite inefficient.  Will fix.

shaobo-he-aws (approved):
LGTM

PR #1555, opened 2025-03-26
add `AsRef` impls to a bunch of public-api types

cdisselkoen:
## Description of changes

Adds `AsRef` impls to a bunch of public-api types, that convert into the corresponding Core/Validator types.  This is possibly controversial because it could allow folks to peek behind our abstraction layer and get at the (internal, unstable) Core/Validator types.  However, I claim this PR is at least worth consideration because:

- I guarded all of these with `doc(hidden)` so they won't show in the docs -- you have to know they're there or read the code
- The reverse transformations for most of these are already publicly available via the `RefCast` trait (which even appears in the docs! it's not `doc(hidden)`
- Transforming between the public types and the Core/Validator types is already possible with the public Protobuf API, by converting to the appropriate `proto::model` type and back, since we have public conversions from (eg) `Policy` to/from `model::Policy` to/from `ast::Policy`.  This PR just makes that more official and much more efficient than going via the `proto::model` types.

### DIFF
--- a/cedar-policy/src/api/id.rs
+++ b/cedar-policy/src/api/id.rs
@@ -38,7 +38,7 @@ use std::str::FromStr;
 /// ```
 /// # use cedar_policy::EntityId;
 /// let id : EntityId = "my-id".parse().unwrap_or_else(|never| match never {});
-/// # assert_eq!(id.as_ref(), "my-id");
+/// # assert_eq!(<EntityId as AsRef<str>>::as_ref(&id), "my-id");
 /// ```
 ///
 /// `EntityId` does not implement `Display`, partly because it is unclear
@@ -50,6 +50,13 @@ use std::str::FromStr;
 #[allow(clippy::module_name_repetitions)]
 #[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord, RefCast)]
 pub struct EntityId(ast::Eid);
+
+#[doc(hidden)] // because this converts to a private/internal type
+impl AsRef<ast::Eid> for EntityId {
+    fn as_ref(&self) -> &ast::Eid {
+        &self.0
+    }
+}
 
 impl EntityId {
     /// Construct an [`EntityId`] from a source string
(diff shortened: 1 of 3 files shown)


adpaco-aws (approved):
This is fine with me, the main reason being it's already possible to do without it so it's mostly an ergonomic improvement.

katherine-hough (approved):
I agree with adpaco-aws's reasoning for permitting this

cedar-policy/src/api/id.rs
```diff
@@ -38,7 +38,7 @@ use std::str::FromStr;
 /// ```
 /// # use cedar_policy::EntityId;
 /// let id : EntityId = "my-id".parse().unwrap_or_else(|never| match never {});
-/// # assert_eq!(id.as_ref(), "my-id");
+/// # assert_eq!(<EntityId as AsRef<str>>::as_ref(&id), "my-id");
```
john-h-kastner-aws, line 41: Are changes like this likely to be necessary in users code?

john-h-kastner-aws (commented):
looks reasonable, but is there a specific motivation for this change?

cedar-policy/src/api/id.rs
cdisselkoen: Shouldn't be necessary if Rust can determine the type of `x.as_ref()` needs to be `str` (or conversely, needs to be `ast::Eid`).  Here inside `assert_eq!`, there is no expectation of what the type should be, so we need changes like this.  I do not know what portion of users' usages are likely to fall into one camp or the other

cdisselkoen:
> is there a specific motivation for this change?

Analysis would like to accept the public types, but operate on the internal AST

cedar-policy/src/api/id.rs
cdisselkoen: We could make a `.unescaped()` method on `EntityId` to go along with the current `.escaped()` method.  `unescaped()` would be redundant with `.as_ref()` but avoid this issue of which `AsRef` implementation was meant.  That's probably a good idea regardless; I'll add it to this PR

john-h-kastner-aws (approved):

john-h-kastner-aws (approved):

PR #1557, opened 2025-03-28
Add unknown_entities method to PartialResponse

vladcebo:
Introduces unknown_entities method to the PartialResponse.

## Description of changes

Introduces a method  `uknown_entities` on the `PartialResponse`.
During partial evaluation, knowing unknown entity ids can help clients to fetch the missing entities
and run the evaluation again.

This is achievable using the existing code via:
```rust
    let unknowns = partial_response
        .all_residuals()
        .fold(HashSet::new(), |mut unknowns, p| {
            unknowns.extend(p.unknown_entities().into_iter());
            unknowns
        });

```

Unfortunately this is slow for large policy sets due to `Policy::from_ast` [relying on pretty printing.](https://github.com/cedar-policy/cedar/blob/5dd810be8cf74cf71500f4c2723751a66ff51f34/cedar-policy/src/api.rs#L3599)

so calling that in a loop  doing entity discovery is just very expensive.

## Issue #, if available

N/A

### DIFF
--- a/cedar-policy-core/src/ast/policy.rs
+++ b/cedar-policy-core/src/ast/policy.rs
@@ -23,7 +23,11 @@ use miette::Diagnostic;
 use nonempty::{nonempty, NonEmpty};
 use serde::{Deserialize, Serialize};
 use smol_str::SmolStr;
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::{HashMap, HashSet},
+    str::FromStr,
+    sync::Arc,
+};
 use thiserror::Error;
 
 #[cfg(feature = "wasm")]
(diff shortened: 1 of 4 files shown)


vladcebo:
Looking more at the API, there are many other methods that map the inner result via`Policy::from_ast`.

For example, say I'm interested in the policy ids that are definitely satisfied:
```rust
    let satisfied: HashSet<PolicyId> = cedar_response
        .definitely_satisfied()
        .map(|policy| policy.id().clone())
        .collect();
```

Well, you if you have lots of satisfied policies then there's lots of extra working in pretty printing that stuff.
Maybe can think of a more generic approach to expose that information in a cheaper way?

adpaco-aws:
Hi @vladcebo , thanks for the PR!

> Maybe can think of a more generic approach to expose that information in a cheaper way?

Yes, it sounds like at the very least we should test the assumption that pretty-printing is cheaper than building the `est` with `est::Policy::from(ast.clone())`. I'm also not sure if these interfaces will remain the same with [Type-Aware Partial Evaluation](https://github.com/cedar-policy/rfcs/pull/95). @shaobo-he-aws any thoughts?

shaobo-he-aws:
https://github.com/cedar-policy/rfcs/pull/95 should be an independent implementation in parallel to the existing PE. Besides, there's no explicit unknown in TPE. So, we should be fine here.

vladcebo:
Added a test for coverage 👍

adpaco-aws (approved):

shaobo-he-aws (approved):

cdisselkoen (approved):

adpaco-aws:
Thanks for the contribution, @vladcebo ! I'll merge it now.

PR #1562, opened 2025-04-04
Add `Entities::len` and `Entities::is_empty`

### LINKED ISSUE #1523: Addition of `Entities::num_of_entities` function similar to `PolicySet::num_of_policies`
### Category

A `PolicySet` has a `num_of_policies` function that gets the size of the policy set in constant time. I'd like to do the same to entities, for logging purposes, without iterating through every entity using `.iter`. 

You could use `Entities::iter` and then run `.count`, but that's linear time and consumes the iterator. 

### Additional context

_No response_

adpaco-aws:
## Description of changes

Add the `Entities::len` method requested in #1523. This new API returns the number of `Entity`s in the `Entities`.
Also add the `Entities::is_empty` method as recommended by `clippy`.

The PR uses both `len` and `is_empty` in some existing unit tests and includes new tests specific to them.

Resolves #1523 

### DIFF
--- a/cedar-policy-core/src/entities.rs
+++ b/cedar-policy-core/src/entities.rs
@@ -260,7 +260,7 @@ impl Entities {
 
     /// Returns `true` if the `Entities` object is empty
     pub fn is_empty(&self) -> bool {
-        self.entities.len() == 0
+        self.entities.is_empty()
     }
 
     /// Convert an `Entities` object into a JSON value suitable for parsing in
--- a/cedar-policy-core/src/entities.rs
+++ b/cedar-policy-core/src/entities.rs
@@ -258,6 +258,11 @@ impl Entities {
         self.entities.len()
     }
 
+    /// Returns `true` if the `Entities` object is empty
+    pub fn is_empty(&self) -> bool {
+        self.entities.len() == 0
+    }
+
     /// Convert an `Entities` object into a JSON value suitable for parsing in
     /// via `EntityJsonParser`.
     ///
@@ -2096,6 +2101,20 @@ mod entities_tests {
         )
         .expect("Failed to construct entities");
         assert_eq!(es.len(), 4);
+        assert!(!es.is_empty());
+    }
+
+    #[test]
+    fn test_is_empty() {
+        let es = Entities::from_entities(
+            vec![],
+            None::<&NoEntitiesSchema>,
+            TCComputation::ComputeNow,
+            Extensions::all_available(),
+        )
+        .expect("Failed to construct entities");
+        assert_eq!(es.len(), 0);
+        assert!(es.is_empty());
     }
 
     #[test]
--- a/cedar-policy-core/src/entities.rs
+++ b/cedar-policy-core/src/entities.rs
@@ -253,6 +253,11 @@ impl Entities {
         })
     }
 
+    /// Returns the length of the `Entities` object
+    pub fn len(&self) -> usize {
+        self.entities.len()
+    }
+
     /// Convert an `Entities` object into a JSON value suitable for parsing in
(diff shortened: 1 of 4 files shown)


shaobo-he-aws (approved):

cedar-policy/src/api.rs
```diff
@@ -776,6 +776,11 @@ impl Entities {
         Some(entity.ancestors().map(EntityUid::ref_cast))
     }
 
+    /// Returns the number of `Entity`s in the `Entities`
+    pub fn num_of_entities(&self) -> usize {
```
john-h-kastner-aws, line 780: I don't like the name `num_of_entities`. `len` is more consistent with the, e.g., `HashMap` api. For `PolicySet` the more explicit name is required to differentiate between policies and templates, but we don't need that here (though, now that I think about it, I'd prefer if we'd called those `num_policies` and `num_templates` instead).

cedar-policy-core/src/entities.rs
```diff
@@ -253,6 +253,11 @@ impl Entities {
         })
     }
 
+    /// Returns the length of the `Entities` object
+    pub fn len(&self) -> usize {
```
john-h-kastner-aws, line 257: IIRC clippy will complain if `Entities` has `len` but not `is_empty`.

john-h-kastner-aws (commented):
Complaining about the name

john-h-kastner-aws:
> It's unclear to me if num_of_entities (or the underlying len method) should be specialized for the partial-evaluation case. 

What would `.iter().count()` do? IMO these should give the same number unless we have a good reason.

cdisselkoen (commented):
Looks good to me, modulo John's comments, which I endorse.

cedar-policy/src/api.rs
chaluli: What if instead we called them: `len_policies` and `len_templates`? Then it would fall even more in line when naming this function `len`.

cedar-policy-core/src/entities.rs
adpaco-aws: Thanks, added `is_empty` too.

adpaco-aws:
> What would `.iter().count()` do? IMO these should give the same number unless we have a good reason.

Ok, I'll leave it as is.

cedar-policy-core/src/entities.rs
```diff
@@ -258,6 +258,11 @@ impl Entities {
         self.entities.len()
     }
 
+    /// Returns `true` if the `Entities` object is empty
+    pub fn is_empty(&self) -> bool {
+        self.entities.len() == 0
```
john-h-kastner-aws, line 263: nit: `self.entities.is_empty()`

cedar-policy/src/api.rs
chaluli: Totally fair.

cedar-policy/src/api.rs
```diff
@@ -777,10 +777,15 @@ impl Entities {
     }
 
     /// Returns the number of `Entity`s in the `Entities`
-    pub fn num_of_entities(&self) -> usize {
+    pub fn len(&self) -> usize {
         self.0.len()
     }
 
+    /// Returns the number of `Entity`s in the `Entities`
```
john-h-kastner-aws, line 784:
```suggestion
    /// Returns true if this `Entities` contains no `Entity`s.
```

john-h-kastner-aws (approved):

cedar-policy/src/api.rs
adpaco-aws: Used a similar text for this doc.

PR #1575, opened 2025-04-16
TPE implementation

shaobo-he-aws:

### DIFF
--- a/cedar-partial-evaluation/Cargo.toml
+++ b/cedar-partial-evaluation/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "cedar-partial-evaluation"
+rust-version.workspace = true
+version.workspace = true
+homepage.workspace = true
+keywords.workspace = true
+categories.workspace = true
+license.workspace = true
+edition.workspace = true
+repository.workspace = true
+
+[dependencies]
+cedar-policy-core = { version = "=4.4.0", path = "../cedar-policy-core" }
+cedar-policy-validator = { version = "=4.4.0", path = "../cedar-policy-validator" }
+serde = "1.0.219"
+serde_json = "1.0.140"
+serde_with = "3.12.0"
+smol_str = "0.3.2"
+
+[lints]
+workspace = true
--- a/cedar-partial-evaluation/src/entities.rs
+++ b/cedar-partial-evaluation/src/entities.rs
@@ -1,0 +1,74 @@
+/*
+ * Copyright Cedar Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+use std::collections::HashMap;
+
+use cedar_policy_core::{entities::EntityUidJson, jsonvalue::JsonValueWithNoDuplicateKeys};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+use smol_str::SmolStr;
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+#[serde_as]
+#[serde(transparent)]
+pub struct DeduplicatedMap {
+    #[serde_as(as = "serde_with::MapPreventDuplicates<_,_>")]
+    pub map: HashMap<SmolStr, JsonValueWithNoDuplicateKeys>
+}
+
+/// Serde JSON format for a single entity
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub struct EntityJson {
+    /// UID of the entity, specified in any form accepted by `EntityUidJson`
+    uid: EntityUidJson,
+    /// attributes, whose values can be any JSON value.
+    /// (Probably a `CedarValueJson`, but for schema-based parsing, it could for
+    /// instance be an `EntityUidJson` if we're expecting an entity reference,
+    /// so for now we leave it in its raw json-value form, albeit not allowing
+    /// any duplicate keys in any records that may occur in an attribute value
+    /// (even nested).)
+    #[serde(default)]
+    // the annotation covers duplicates in this `HashMap` itself, while the `JsonValueWithNoDuplicateKeys` covers duplicates in any records contained in attribute values (including recursively)
+    attrs: Option<DeduplicatedMap>,
+    #[serde(default)]
+    /// Parents of the entity, specified in any form accepted by `EntityUidJson`
+    parents: Option<Vec<EntityUidJson>>,
+    #[serde(default)]
+    // the annotation covers duplicates in this `HashMap` itself, while the `JsonValueWithNoDuplicateKeys` covers duplicates in any records contained in tag values (including recursively)
+    // Note that unlike the concrete JSON entity format, when the `tags` field
+    // is missing, it means `tags` are unknown
+    // This is because we need to represent `tags` being unknowns
+    tags: Option<DeduplicatedMap>,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::EntityJson;
+
+    #[test]
+    fn basic () {
+        let json = serde_json::json!(
+            {
+                "uid" : {
+                    "type" : "A",
+                    "id" : "",
+                },
+                "tags" : null,
+            }
+        );
+        let ejson : EntityJson = serde_json::from_value(json).expect("should parse");
+     }
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ members = [
 	"cedar-policy-cli",
 	"cedar-testing",
 	"cedar-wasm"
-]
+, "cedar-partial-evaluation"]
 
 resolver = "2"
 
--- a/cedar-partial-evaluation/src/lib.rs
+++ b/cedar-partial-evaluation/src/lib.rs
@@ -1,0 +1,6 @@
+pub mod entities;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+}


luxas:
Hi @shaobo-he-aws! 👋 How is the implementation shaping up?
Is it ready enough for me to try out? Any way I can help? 😄

cdisselkoen:
Hi @luxas!  @shaobo-he-aws is now out of town for several weeks.  This PR is still in progress and someone else on the team may pick it up in the coming weeks.

cedar-partial-evaluation/src/entities.rs
```diff
@@ -0,0 +1,640 @@
...
+            if let Some(tags) = &self.tags {
+                if !tags.is_empty() {
+                    return Err(anyhow!("actions can't have tags"));
+                }
+            }
+            if let Some(action) = core_schema.action(uid) {
+                if let Some(ancestors) = &self.ancestors {
+                    let schema_ancestors: HashSet<EntityUID> =
+                        action.ancestors().cloned().collect();
+                    if &schema_ancestors != ancestors {
```
luxas: What is the use-case for consuming ancestor information from JSON, if/as that information already exists in the schema?
If I read this correctly, we could consume it here from the schema, but then we'd have to use &mut self here, or initialize self from a schema.

cedar-partial-evaluation/src/entities.rs
```diff
@@ -0,0 +1,640 @@
...
+        value: serde_json::Value,
+        schema: &ValidatorSchema,
+    ) -> anyhow::Result<Self> {
+        let entities: Vec<EntityJson> = serde_json::from_value(value)
+            .map_err(|e| anyhow::anyhow!("failed to parse entities: {}", e))?;
+        let mut partial_entities = PartialEntities::default();
+        for e in entities {
+            let partial_entity = parse_ejson(e, schema)?;
+            partial_entity.validate(schema)?;
+            partial_entities
```
luxas: should we make sure there are not duplicates in the entities JSON here?

cedar-partial-evaluation/src/evaluator.rs
```diff
@@ -0,0 +1,1657 @@
...
+
+#[derive(Debug)]
+pub struct Evaluator<'e> {
+    pub(crate) request: PartialRequest,
+    pub(crate) entities: &'e PartialEntities,
+    pub(crate) extensions: &'e Extensions<'e>,
+}
+
+impl Evaluator<'_> {
+    pub fn interpret(&self, e: &Expr<Option<Type>>) -> Residual {
```
luxas: why is this `Expr<Option<Type>>` and not just `Expr<Type>`? I see it's coming from the validator, but would it be better to change the validator to always just return a fully typed expr, as the data will never be None?

cedar-partial-evaluation/src/request.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// `None` means that variable will result in a residual for partial evaluation.
+    pub(crate) context: Option<Arc<BTreeMap<SmolStr, Value>>>,
+}
+
+impl PartialRequest {
+    /// Create a well-formed `PartialRequest` (i.e., it conforms to the schema)
+    pub fn new(
+        principal: PartialEntityUID,
+        resource: PartialEntityUID,
+        action: EntityUID,
```
luxas: action before resource to yield PARC?

cedar-partial-evaluation/src/evaluator.rs
```diff
@@ -0,0 +1,1657 @@
...
+                                value: ValueKind::Lit(ast::Literal::Bool(true)),
+                                ..
+                            },
+                        ..
+                    } => self.interpret(right),
+                    Residual::Concrete { ty, .. } => Residual::Error(ty.clone()),
+                    Residual::Partial { .. } => Residual::Partial {
+                        kind: ResidualKind::And {
+                            left: Arc::new(left),
+                            right: Arc::new(self.interpret(right)),
```
luxas: This does not fold `residual && false` => `false`, or `residual && true` => `residual`
I implemented that easily locally, and verified it works.

cedar-partial-evaluation/src/evaluator.rs
```diff
@@ -0,0 +1,1657 @@
...
+                                value: ValueKind::Lit(ast::Literal::Bool(false)),
+                                ..
+                            },
+                        ..
+                    } => self.interpret(right),
+                    Residual::Concrete { ty, .. } => Residual::Error(ty.clone()),
+                    Residual::Partial { .. } => Residual::Partial {
+                        kind: ResidualKind::Or {
+                            left: Arc::new(left),
+                            right: Arc::new(self.interpret(right)),
```
luxas: This does not fold `residual || false` => `residual`, or `residual || true` => `true`
I guess we want that as well as folding for `&&`

cedar-partial-evaluation/src/evaluator.rs
```diff
@@ -0,0 +1,1657 @@
...
+                            attr: attr.clone(),
+                        },
+                        ty,
+                    },
+                    Residual::Error(_) => Residual::Error(ty),
+                }
+            }
+            // PANIC SAFETY: TPE does not expect explicit unknowns in policies
+            #[allow(clippy::unreachable)]
+            ExprKind::Unknown { .. } => unreachable!("we should not unexpect unknowns"),
```
luxas: If I read it correctly, it's safe to panic here, as we know that the typechecker before this must have returned `TypecheckAnswer::fail` if it saw an unknown?

cedar-partial-evaluation/src/residual.rs
```diff
@@ -0,0 +1,204 @@
...
+    // i.e., we can't do the dedup of duplicates until all of the `Expr`s are
+    // evaluated into `Value`s
+    Set(Arc<Vec<Residual>>),
+    /// Anonymous record (whose elements may be arbitrary expressions)
+    Record(Arc<BTreeMap<SmolStr, Residual>>),
+}
+
+/// Conversion from `Residual` to `Expr` so that we can use the concrete evaluator for re-authorization
+impl From<Residual> for Expr {
+    fn from(value: Residual) -> Expr {
```
luxas: Should we have a roundtrip test here to make sure that `Expr::from(evaluator.interpret(expr)) == expr`, for arbitrary exprs, when entities are fully unknown, and `resource is Foo`, `action == Bar` and `principal is Baz` expressions are not present in expr (and maybe something else that I forgot that isn't round-trippable.

And/or residual -> expr -> residual.

cedar-partial-evaluation/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+keywords.workspace = true
+categories.workspace = true
+license.workspace = true
+edition.workspace = true
+repository.workspace = true
+
+[dependencies]
+anyhow = "1.0.98"
+cedar-policy-core = { version = "=4.4.0", path = "../cedar-policy-core" }
+cedar-policy-validator = { version = "=4.4.0", path = "../cedar-policy-validator" }
```
luxas: This needs a rebase now, as validator moved into core.

shaobo-he-aws:
> Super awesome work @shaobo-he-aws 🎉!
> 
> I can happily confirm that the demo use-case for Kubernetes that I have seems to work correctly with this setup. 💯🚀
> 
> It will need the transformer that was discussed in the proposal, to substitute partially unknown attributes with an entity reference. Not sure, but if I find some time, maybe I could contribute some part of that, as I anyways need to create some version of it.
> 
> Wrt out-of-town, absolutely, no worries and/or stress with this. 😊 I'll be on PTO as well the two last weeks of June. I could maybe add the constant foldings that I pointed out here, and take a stab at the TODO, but otherwise I'm not sure what's left to do here, other than maybe integration with the DRT / Lean infrastructure?

@luxas Thank you. I'm back from vacation and will resume this PR soon.

cedar-partial-evaluation/Cargo.toml
shaobo-he-aws: Done. I think I will eventually move it into core.

cedar-partial-evaluation/src/evaluator.rs
shaobo-he-aws: It's not safe to reduce `residual && false` into `false` because `residual` can fail. The other transformation should be doable.

cedar-partial-evaluation/src/evaluator.rs
shaobo-he-aws: Same as above.

cedar-partial-evaluation/src/evaluator.rs
shaobo-he-aws: IIUC, the optional type may be necessary for other validator features. Changing the type definition is complicated so I personally don't think it's necessary. That being said, I can confirm (via DRT) that strict validation always produces a typed expression.

cedar-partial-evaluation/src/entities.rs
shaobo-he-aws: Good catch. I've changed the logic such that actions info comes from the schema only.

cedar-partial-evaluation/src/residual.rs
shaobo-he-aws: I will make a DRT target to test this roundtrip property.

shaobo-he-aws:
Hi @luxas, I've addressed your comments. Please take another look at this PR when you have time.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -0,0 +1,1709 @@
...
+        let ty = e
+            .data()
+            .clone()
+            .expect("type checked should provide a type");
+        match e.expr_kind() {
+            ExprKind::Lit(l) => Residual::Concrete {
+                value: l.clone().into(),
+                ty,
+            },
+            ExprKind::Var(Var::Action) => Residual::Concrete {
```
chaluli, line 57: Why is this guaranteed to be concrete? What if the `PartialRequest` does not have a concrete action? Do we require that the partial request has to have _at least_ the action provided?

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -0,0 +1,1709 @@
...
+                    },
+                    Residual::Concrete {
+                        value:
+                            Value {
+                                value: ValueKind::Lit(ast::Literal::Bool(true)),
+                                ..
+                            },
+                        ..
+                    } => self.interpret(right),
+                    Residual::Concrete { ty, .. } => Residual::Error(ty.clone()),
```
chaluli, line 122: Should we improve the error by also providing the original expression and the value?

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -0,0 +1,1709 @@
...
+                        } => left,
+                        right => Residual::Partial {
+                            kind: ResidualKind::And {
+                                left: Arc::new(left),
+                                right: Arc::new(right.clone()),
+                            },
+                            ty,
+                        },
+                    },
+                    Residual::Error(_) => Residual::Error(ty),
```
chaluli, line 140: What change the error? Won't that obfuscate the actual error?

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -0,0 +1,1709 @@
...
+                        Residual::Concrete { value: v1, .. },
+                        Residual::Concrete { value: v2, .. },
+                    ) => match op {
+                        BinaryOp::Eq | BinaryOp::Less | BinaryOp::LessEq => {
+                            if let Ok(v) =
+                                crate::evaluator::binary_relation(*op, v1, v2, self.extensions)
+                            {
+                                Residual::Concrete { value: v, ty }
+                            } else {
+                                Residual::Error(ty)
```
chaluli, line 305: This error isn't informative. It's missing the reason why the expression failed to evaluate.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -0,0 +1,1709 @@
...
+                                }
+                            } else {
+                                Residual::Error(ty)
+                            }
+                        }
+                        BinaryOp::Contains => match &v1.value {
+                            ValueKind::Set(Set { fast: Some(h), .. }) => Residual::Concrete {
+                                value: match v2.try_as_lit() {
+                                    Some(lit) => (h.contains(lit)).into(),
+                                    None => false.into(),
```
chaluli: Why is this false if v2 is not a lit? Is this relying on the fact that sets can only contain literals? A comment would be helpful to describe why this is true.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -0,0 +1,1709 @@
...
+                }
+            }
+            ExprKind::ExtensionFunctionApp { fn_name, args } => {
+                let args = args.iter().map(|a| self.interpret(a)).collect::<Vec<_>>();
+                if let Ok(vals) = args
+                    .iter()
+                    .map(|a| Value::try_from(a.clone()))
+                    .collect::<std::result::Result<Vec<_>, _>>()
+                {
+                    if let Ok(ext_fn) = self.extensions.func(fn_name) {
```
chaluli, line 464: Comments explaining what this block is doing would be helpful.

cedar-policy-core/src/tpe/evaluator.rs
shaobo-he-aws: Yes, TPE requires concrete actions.

cedar-policy-core/src/tpe/evaluator.rs
shaobo-he-aws: The `Residual::Error` simply indicates that TPE produces an error and its type (or root cause) does not matter because it does not for the result of reauthorization. So, I don't think we need to incorporate the original expression.

cedar-policy-core/src/tpe/evaluator.rs
shaobo-he-aws: Please see my previous comment. The type of `Residual::Error` changes because we want to make sure the resulting residual is well-typed.

cedar-policy-core/src/tpe/evaluator.rs
shaobo-he-aws: Same as above.

cedar-policy-core/src/tpe/evaluator.rs
shaobo-he-aws: Correct. If the `fast` field is `Some`, it means all elements are literals. If `v2` is not a literal, then the LHS can't contain it. Will add a comment.

shaobo-he-aws: Actually, I should've used the new API `Set::contains` that hides the implementation details. Updated the code.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -423,73 +423,26 @@ impl Evaluator<'_> {
...
+                            ValueKind::Set(s) => Residual::Concrete {
+                                value: s.contains(v2).into(),
                                 ty,
                             },
-                            ValueKind::Set(Set {
-                                fast: None,
-                                authoritative,
-                            }) => Residual::Concrete {
-                                value: (authoritative.contains(v2)).into(),
+                            _ => Residual::Error(ty),
```
chaluli, line 430: Looks like a great refactor and simplification. :)

chaluli:
You may want to merge main in to fix some of the CI failures.

cedar-policy-core/src/tpe.rs
```diff
@@ -0,0 +1,489 @@
...
+use crate::{ast::PolicySet, extensions::Extensions};
+
+use crate::tpe::{
+    entities::PartialEntities, evaluator::Evaluator, request::PartialRequest, residual::Residual,
+};
+
+/// Type-aware partial-evaluation on a `PolicySet`.
+/// Both `request` and `entities` should be valid and hence be constructed
+/// using their safe constructors.
+/// Polices must be static.
```
cdisselkoen:
```suggestion
/// Policies must be static.
```

cedar-policy-core/src/tpe/entities.rs
```diff
@@ -0,0 +1,780 @@
...
+    #[serde(default)]
+    /// Parents of the entity, specified in any form accepted by `EntityUidJson`
+    parents: Option<Vec<EntityUidJson>>,
+    #[serde(default)]
+    // the annotation covers duplicates in this `HashMap` itself, while the `JsonValueWithNoDuplicateKeys` covers duplicates in any records contained in tag values (including recursively)
+    // Note that unlike the concrete JSON entity format, when the `tags` field
+    // is missing, it means `tags` are unknown
+    // This is because we need to represent `tags` being unknowns
+    tags: Option<DeduplicatedMap>,
+}
```
cdisselkoen, line 89: Is this duplicated from somewhere else in Core?  If not, how does it differ from the entity JSON format we accept in the CLI?

cedar-policy-core/src/tpe/entities.rs
```diff
@@ -0,0 +1,780 @@
...
+        if let Some(attrs) = &self.attrs {
+            let other_attrs = entity
+                .attrs()
+                .map(|(a, pv)| match pv {
+                    PartialValue::Value(v) => Ok((a.clone(), v.clone())),
+                    PartialValue::Residual(_) => Err(UnknownAttributeError {
+                        uid: self.uid.clone(),
+                        attr: a.clone(),
+                    }
+                    .into()),
```
cdisselkoen, line 119: Am I understanding correctly that an entity with residuals in its attributes can never be consistent with any PartialEntity?

cedar-policy-core/src/tpe/entities.rs
```diff
@@ -0,0 +1,780 @@
...
+        }
+    }
+
+    fn reset_edges(&mut self) {}
+}
+
+impl PartialEntity {
+    pub(crate) fn add_ancestor(&mut self, uid: EntityUID) {
+        // PANIC SAFETY: this method should be only called on entities that have known ancestors
+        #[allow(clippy::expect_used)]
```
cdisselkoen, line 317: Since we're relying on callers to behave a certain way for panic safety, we should put a note in the doc comment on this function about what we require from callers

cedar-policy-core/src/tpe/entities.rs
```diff
@@ -0,0 +1,780 @@
...
+        self.ancestors
+            .as_mut()
+            .expect("should not be unknown")
+            .insert(uid);
+    }
+
+    /// Valiate `self` according to `schema`
```
cdisselkoen:
```suggestion
    /// Validate `self` according to `schema`
```

cedar-policy-core/src/tpe/entities.rs
```diff
@@ -0,0 +1,780 @@
...
+            }
+        }
+    }
+    Ok(())
+}
+
+/// The partial entity store
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct PartialEntities {
+    /// Important internal invariant: for any `Entities` object that exists,
```
cdisselkoen, line 445: should this comment say `PartialEntities`?

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -0,0 +1,1667 @@
...
+    tpe::request::PartialRequest,
+    tpe::residual::{Residual, ResidualKind},
+};
+
+/// The partial evaluator
+#[derive(Debug)]
+pub struct Evaluator<'e> {
+    pub(crate) request: PartialRequest,
+    pub(crate) entities: &'e PartialEntities,
+    pub(crate) extensions: &'e Extensions<'e>,
```
cdisselkoen: tiny nit: `request` could also be a `&'e` reference. I notice that at least one caller of `Evaluator` ends up having to (needlessly?) clone their `PartialRequest` because of this.

cedar-policy-core/src/tpe/evaluator.rs
cdisselkoen: When would this error occur anyway?  It looks like this would only happen if the LHS of an `And` evaluates to a concrete value but not `true` or `false`.  Can that ever happen in a validated policy?

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -0,0 +1,1667 @@
...
+                    Residual::Concrete { ty, .. } => Residual::Error(ty.clone()),
+                    Residual::Partial { .. } => match &self.interpret(&right) {
+                        Residual::Concrete {
+                            value:
+                                Value {
+                                    value: ValueKind::Lit(ast::Literal::Bool(true)),
+                                    ..
+                                },
+                            ..
+                        } => left,
```
cdisselkoen, line 131: should we have a special-case for `false`, too?

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -0,0 +1,1667 @@
...
+                    Residual::Concrete { ty, .. } => Residual::Error(ty.clone()),
+                    Residual::Partial { .. } => match &self.interpret(&right) {
+                        Residual::Concrete {
+                            value:
+                                Value {
+                                    value: ValueKind::Lit(ast::Literal::Bool(false)),
+                                    ..
+                                },
+                            ..
+                        } => left,
```
cdisselkoen, line 174: likewise here

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -0,0 +1,1667 @@
...
+                    },
+                    Residual::Error(_) => Residual::Error(ty),
+                }
+            }
+            // PANIC SAFETY: TPE does not expect explicit unknowns in policies
+            #[allow(clippy::unreachable)]
+            ExprKind::Unknown { .. } => unreachable!("we should not unexpect unknowns"),
+            // PANIC SAFETY: TPE currently only works on static policies
+            #[allow(clippy::unreachable)]
+            ExprKind::Slot(_) => unreachable!("we should not unexpect slot for now"),
```
cdisselkoen: I think `panic!` would fit better than `unreachable!` here.  (I have no problem with the `panic` though, this is an experimental feature)

cedar-policy-core/src/tpe/evaluator.rs
shaobo-he-aws: I don't think so. We could replace it with a unreachable statement but I'd like to avoid such refactoring because the annotation burden. And the code is more consistent if we treat all possible errors (e.g., extension function call) as `Residual::Error`, even though type errors are not possible.

cdisselkoen: Great.  (My point is not that we replace it with an unreachable statement, but that if this error should never occur, we shouldn't make a huge deal about improving the error message)

cedar-policy-core/src/tpe/entities.rs
shaobo-he-aws: No, it'd be great if I can just reuse core but there are differences in the two formats. The prominent one is shown in the comment,

```
// Note that unlike the concrete JSON entity format, when the `tags` field
// is missing, it means `tags` are unknown
// This is because we need to represent `tags` being unknowns
```

shaobo-he-aws: Besides, `attrs`, `parents` and `tags` are also optional, whereas they must present in the concrete entity format.

cedar-policy-core/src/tpe/entities.rs
shaobo-he-aws: Correct. A `PartialEntity` can only be consistent with an `Entity` that does not have any unknowns.

cedar-policy-core/src/tpe/evaluator.rs
shaobo-he-aws: Got it.

cedar-policy-core/src/tpe/evaluator.rs
shaobo-he-aws: I don't think we can do anything about the `false` case: `<residual> && false` cannot be soundly reduced to `false` because `<residual>` can evaluate to an error.

cedar-policy-core/src/tpe/evaluator.rs
shaobo-he-aws: Please see the comment above.

cedar-policy-core/src/tpe/evaluator.rs
cdisselkoen: Ah I see.  I noticed we can now soundly reduce `true && <residual>` to `<residual>`, but I suppose that is sound in a way that this would not be.  TPE knows that `<residual>` cannot be a type error, but it can't rule out other errors

shaobo-he-aws: Correct.

cedar-policy-core/src/lib.rs
```diff
@@ -35,6 +35,8 @@ pub mod extensions;
 pub mod fuzzy_match;
 pub mod jsonvalue;
 pub mod parser;
+#[cfg(feature = "tpe")]
+pub mod tpe;
```
luxas, line 39: nice that you could move it to the core crate now instead of making a new one 👍

cedar-policy/src/api.rs
```diff
@@ -5050,6 +5050,209 @@ pub fn eval_expression(
...
+                .0
+                .validate_request(&request.0, Extensions::all_available())
+                .map_err(|err| TPEReauthorizationError::RequestValidation(err.into()))?;
+            let core_schema = CoreSchema::new(&self.schema.0);
+            let entities_checker =
+                EntitySchemaConformanceChecker::new(&core_schema, Extensions::all_available());
+            for entity in entities.0.iter() {
+                entities_checker.validate_entity(&entity)?;
+            }
+            let _ = self.entities.0.check_consistency(&entities.0)?;
```
luxas, line 5216: curious: is there some reason to have the `let _ = ` prefix here, as the return value of `Ok` is `()` afaics? Just style or am I missing something?

cedar-policy/src/api.rs
```diff
@@ -5050,6 +5050,209 @@ pub fn eval_expression(
...
+            let authorizer = Authorizer::new();
+            Ok(authorizer.is_authorized(&request, &self.ps, &entities))
+        }
+    }
+
+    impl PolicySet {
+        /// Perform type-aware partial evaluation on this [`PolicySet`]
+        /// If successful, the result is a [`PolicySet`] containing residual
+        /// policies ready for re-authorization
+        pub fn tpe<'a>(
```
luxas, line 5227: maybe `partially_authorize` would be more understandable for those not deeply familiar with this feature? I guess the old partial evaluation will go away soon-ish anyways, and then this is the only partial feature that exists, and it is typed from the context (it takes a schema).

cedar-policy/src/api.rs
```diff
@@ -5050,6 +5050,209 @@ pub fn eval_expression(
...
+                None
+            }
+        }
+
+        /// Get residual policies
+        pub fn residuals(&self) -> impl Iterator<Item = &Policy> {
+            self.ps.policies()
+        }
+
+        /// Re-authorize residuals with consistent concrete request and entities
```
luxas, line 5199: I think I understand the definition of "consistency", as I've read the RFC and checked this PR, but it would be nice to in a comment briefly explain what it means for different types?

On another note, I might need sometimes to do tpe multiple times over, with an increasing amount of entities on a shrinking policyset; it's probably a niche use-case to have multi-pass tpe, so it doesn't probably need its own function. I guess I will construct a new PolicySet from `Residuals.residuals()`, and then run `tpe` on that again, that works well (even though I have to clone every item in the policyset, as opposed to if the `ps` field was public, or `tpe` was available on `Residuals` as well.

cedar-policy/src/api.rs
```diff
@@ -5050,6 +5050,209 @@ pub fn eval_expression(
...
+            Self(tpe::request::PartialEntityUID {
+                ty: ty.0,
+                eid: id.map(|id| <EntityId as AsRef<ast::Eid>>::as_ref(&id).clone()),
+            })
+        }
+    }
+
+    /// A partial [`Request`]
+    /// Its principal and resource could have unknown [`EntityId`]; its action
+    /// must be know; and its context could be unknown
```
luxas, line 5092: nit, typo known

also suggesting: "its principal, action and resource types must be known" and "its context must either be fully known or unknown"

cedar-policy/src/api.rs
```diff
@@ -5050,6 +5050,209 @@ pub fn eval_expression(
...
+                    ast::Context::Value(m) => Ok(m),
+                })
+                .transpose()?;
+            tpe::request::PartialRequest::new(principal.0, action.0, resource.0, context, &schema.0)
+                .map(Self)
+                .map_err(|e| PartialRequestCreationError::Validation(e.into()))
+        }
+    }
+
+    /// Partial [`Entities`]
```
luxas, line 5120: nit: as with request above, it would be nice with a short recap of what invariants need to hold for partial entities (e.g. fully known or unknown tags, attrs and ancestors, and ancestors, if specified, must have recursively known ancestors)

cedar-policy-core/src/tpe/request.rs
```diff
@@ -0,0 +1,587 @@
...
+        }
+        Ok(())
+    }
+}
+
+/// A request builder based on a [`PartialRequest`]
+/// Users should use it to iteratively construct a [`Request`] using methods
+/// `add_*`
+#[derive(Debug, Clone)]
+pub struct RequestBuilder<'s> {
```
luxas, line 300: is this meant only for the low-level core create, or was it meant to be available for the high-level cedar-policy API surface?

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -0,0 +1,1667 @@
...
+                            attr: attr.clone(),
+                        },
+                        ty,
+                    },
+                    Residual::Error(_) => Residual::Error(ty),
+                }
+            }
+            // PANIC SAFETY: TPE does not expect explicit unknowns in policies
+            #[allow(clippy::panic)]
+            ExprKind::Unknown { .. } => panic!("we should not unexpect unknowns"),
```
luxas, line 594: this reminds me of the translation layer of the mixed-unknown attributes case and TPE; I will need to make a small wrapper of the translation for the k8s use case. Is implementation of that translation layer in core planned as a follow up?

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -0,0 +1,1667 @@
...
+                        ty,
+                    }
+                } else {
+                    Residual::Partial {
+                        kind: ResidualKind::Var(Var::Context),
+                        ty,
+                    }
+                }
+            }
+            ExprKind::And { left, right } => {
```
luxas, line 100: We concluded that it's not safe to fold `<residual> && false` into `false` due to errors, but would it be possible to have some mode where one could annotate whether (assuming no errors), the expr would/could have a definite true/false value? I could build this independently of this code though as well, as a dedicated AST walker doing only that check.

luxas (commented):
Awesome work @shaobo-he-aws, looks really clean 💯🎉
I was OOO, so didn't get to this before now, but already from the first pass there were no major comments.

I'll take this for a spin now 👍

cedar-policy-core/src/tpe/request.rs
shaobo-he-aws: So far it's meant to be for the core crate. We will export it to the public APIs when we have a better understanding how it should behave.

cedar-policy/src/api.rs
shaobo-he-aws: Good news. #1812 implements a solution for the incremental TPE use case: Now TPE computes a residual from a residual. That being said, I will expand the comment to elaborate what consistency means.

cedar-policy-core/src/tpe/evaluator.rs
shaobo-he-aws: Can you remind me of what the wrapper is?

cedar-policy-core/src/tpe/evaluator.rs
shaobo-he-aws: Like we discussed, rewriting is currently the workaround. There's unfortunately no such annotation mechanism.

cedar-policy/src/api.rs
shaobo-he-aws: Done.

cedar-policy/src/api.rs
shaobo-he-aws: Done.

PR #1595, opened 2025-05-06
derive `PartialOrd` and `Ord` for `PolicyId`

cdisselkoen:
## Description of changes

Seems to have been an oversight -- these were already derived for the other types in this file.

I don't think this is significant enough to merit a changelog entry, but happy to make one if others think it would be helpful.

### DIFF
--- a/cedar-policy/src/api/id.rs
+++ b/cedar-policy/src/api/id.rs
@@ -341,7 +341,7 @@ impl From<ast::EntityUID> for EntityUid {
 /// ```
 #[repr(transparent)]
 #[allow(clippy::module_name_repetitions)]
-#[derive(Debug, PartialEq, Eq, Clone, Hash, Serialize, Deserialize, RefCast)]
+#[derive(Debug, PartialEq, Eq, Clone, Hash, PartialOrd, Ord, Serialize, Deserialize, RefCast)]
 #[cfg_attr(feature = "wasm", derive(tsify::Tsify))]
 #[cfg_attr(feature = "wasm", tsify(into_wasm_abi, from_wasm_abi))]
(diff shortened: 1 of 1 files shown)


adpaco-aws (commented):
Is it an oversight or were we trying to avoid defining it for some reason?

If I'm not mistaken, policy IDs are expected to be (at least in the most common case) of the form "policy{num}". Deriving like this implies lexicographic order, which would result in policies being reordered if we do something like sorting by policy ID.

cdisselkoen:
@adpaco-aws what would you prefer?

In my code that uses cedar-policy, I'm currently working around the lack of `Ord` on `PolicyId` by writing my own `Ord` implementation for structs containing `PolicyId`, which is basically the autoderived `Ord`, but uses `.as_ref()` to turn each `PolicyId` into an `&str` before comparing.  This derived `Ord` would be equivalent to that.

shaobo-he-aws:
I think it should be fine to derive these because they are ordered by nature. It should be at users' discretion to leverage the ordering or not.

adpaco-aws:
It depends on how we rely on policy IDs being of the form "policy{num}":
 1. If it's considered an invariant, then I think it'd make more sense to just extract `{num}`, convert it and compare with it.
 2. If it's only the common case, then we can go with the current approach.

It's still not clear to me what's the expectation w.r.t. policy IDs in Cedar. Given this API exists, I suppose we are in case (2). Just wanted us to be aware of this behavior and make sure it's fine too.

adpaco-aws (approved):

cdisselkoen:
It's not considered an invariant.  I think we envision that in production use-cases, our users will use our APIs to control the policy IDs, if they care about policy IDs at all.   The `policy{num}` IDs are handy for getting started quickly, and for folks who do not care at all about the policy IDs.

PR #1601, opened 2025-05-06
Refactor set operations

shaobo-he-aws:
## Description of changes
This PR puts the operations on sets into `Set` from the evaluator, which allows consumers of core to reuse them easily. This PR also makes changes in the implementations of these operations.

* Use `BTreeSet` methods instead of iterators
* Optimize the case where a set is all-literal and the other operand is a non-literal value
* Optimize the case where a set is not all-literal and the other operand is all-literal

### DIFF
--- a/cedar-policy-core/src/ast/value.rs
+++ b/cedar-policy-core/src/ast/value.rs
@@ -437,6 +437,10 @@ impl Set {
             // both sets are in fast form, ie, they only contain literals.
             // Fast hashset-based implementation.
             (Some(ls1), Some(ls2)) => ls1.is_subset(ls2.as_ref()),
+            // `self` contains non-literal(s), `other` is all-literal.
+            // The invariant about `Set::fast` should allow us to conclude that
+            // the result is `false`
+            (None, Some(_)) => false,
             // one or both sets are in slow form, ie, contain a non-literal.
             // Fallback to slow implementation.
             _ => self.authoritative.is_subset(&other.authoritative),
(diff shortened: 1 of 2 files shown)


cedar-policy-core/src/ast/value.rs
```diff
@@ -430,6 +430,46 @@ impl Set {
...
     }
+
+    /// Subset test
+    pub fn is_subset(&self, other: &Set) -> bool {
+        match (&self.fast, &other.fast) {
+            // both sets are in fast form, ie, they only contain literals.
+            // Fast hashset-based implementation.
+            (Some(ls1), Some(ls2)) => ls1.is_subset(ls2.as_ref()),
+            // one or both sets are in slow form, ie, contain a non-literal.
+            // Fallback to slow implementation.
```
john-h-kastner-aws, line 441: Could we optimize `(None, Some(_)) => false`? I.e., a set which contains a non-literal cannot be a subset of a set that doesn't contain any non-literals

john-h-kastner-aws (approved):
good change

cedar-policy-core/src/ast/value.rs
shaobo-he-aws: Good catch. Done.

cdisselkoen (approved):

cedar-policy-core/src/ast/value.rs
john-h-kastner-aws: From the coverage report, look like we don't have a test this specific case. Would be good to add

PR #1602, opened 2025-05-07
Add a typechecking function to check a `Template` by a single env

shaobo-he-aws:


shaobo-he-aws:
I think we're safe to ignore the coverage failure. We can add an test for the API in order to meet the coverage criteria but I'm not sure how much it can buy us.

john-h-kastner-aws:
> I think we're safe to ignore the coverage failure. We can add an test for the API in order to meet the coverage criteria but I'm not sure how much it can buy us given that this PR essentially is a refactoring.

Maybe `typecheck_by_single_request_env` doesn't need to exist and whatever callers you have in mind can just call `single_env_typechecking` directly. The names are already similar enough to be confusing

adpaco-aws (approved):

PR #1604, opened 2025-05-08
Refactor entity conformance checking

shaobo-he-aws:
This commit also fixes a bug where Cedar doesn't performance checks on tags.

### DIFF
--- a/cedar-policy/src/test/test.rs
+++ b/cedar-policy/src/test/test.rs
@@ -1403,6 +1403,7 @@ mod ancestors_tests {
 /// schema-based parsing.
 mod entity_validate_tests {
     use super::*;
+    use cedar_policy_core::entities::conformance::err::EntitySchemaConformanceError;
     use cool_asserts::assert_matches;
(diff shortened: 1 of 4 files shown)


chaluli (approved):

cedar-policy/src/test/test.rs
```diff
@@ -2189,6 +2190,57 @@ action "g" appliesTo {
         "#;
         assert_matches!(Entity::from_json_str(entity_str, Some(&schema)), Ok(_));
     }
+
+    #[test]
+    fn from_entities_tags() {
```
john-h-kastner-aws, line 2195: Should probably include a test that conformance check succeeds if the correct type of tag is provided.

john-h-kastner-aws (approved):

cedar-policy/src/test/test.rs
shaobo-he-aws: Will do it in another PR so that I can use the APIs now.

PR #1607, opened 2025-05-09
Adding ability to convert from Expr<T> to Expr

lianah:
## Description of changes

Added a function to convert from `Expr<T>` to `Expr` by dropping the data to allow creating a new `PolicySet` from a type-checked `Expr` using `Policy::from_when_clause`. I considered updating `from_when_clause` to take `Expr<T>` but that required a lot of invasive changes. 

### DIFF
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -15,7 +15,11 @@
  */
 
 #[cfg(feature = "tolerant-ast")]
-use super::expr_allows_errors::AstExprErrorKind;
+use {
+    super::expr_allows_errors::AstExprErrorKind, crate::parser::err::ToASTError,
+    crate::parser::err::ToASTErrorKind,
+};
+
 use crate::{
     ast::*,
     expr_builder::{self, ExprBuilder as _},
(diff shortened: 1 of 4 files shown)


lianah:
Fixed a merge conflict with https://github.com/cedar-policy/cedar/pull/1602 that also exposed functionality to type-check in just one request environment.

cedar-policy-core/src/ast/expr.rs
```diff
@@ -397,6 +397,122 @@ impl<T> Expr<T> {
             ExprKind::Error { .. } => None,
         }
     }
+
+    /// Convert an Expr<T> to an Expr by dropping the data.
+    pub fn to_expr(&self) -> Expr {
```
cdisselkoen, line 402: should this be `pub fn into_expr(self) -> Expr`, ie, consuming `self` so that we can avoid all/many of the `.clone()`s?

john-h-kastner-aws: probably not since the `T` could be large and isn't cloned

john-h-kastner-aws, line 402: Looking at this code it occurs to me that we could use the `ExprBuilder` trait to merge this definition with https://github.com/cedar-policy/cedar/blob/35182b6f306722ee6f54772ef2854da82695b610/cedar-policy-core/src/est/expr.rs#L1069

lianah: @john-h-kastner-aws can you elaborate on why `T`consuming `self` to drop the `T` that could be large is a problem? Are you thinking of something else that is keeping the data `T` around? For my use case the `T` would not be used post conversion.

john-h-kastner-aws: Right, if a caller needs to keep ownership of `T`, they'll have to `clone` everything even though the current implementation doesn't clone the data.

cdisselkoen: I was imagining that anyone calling this method does not need `T` after calling this

cdisselkoen: Avoiding all the `.clone()s` in this method is potentially a significant performance win

lianah: I updated the pull request to use `ExprBuilder` and avoid code duplication. Note that for the `ExprKind::Error` the `est` implementation is not just reusing the error in the `ast::Expr` but it’s creating a [new  error](https://github.com/cedar-policy/cedar/blob/35182b6f306722ee6f54772ef2854da82695b610/cedar-policy-core/src/est/expr.rs#L1133). I retained the behavior of the `est` conversion because this kind seems to be under the `tolerant-ast` feature flag which I don't think makes sense for`SymEval`.

cedar-policy-core/src/ast/expr.rs
```diff
@@ -397,6 +401,116 @@ impl<T> Expr<T> {
...
+                .slot(slot),
+            ExprKind::Unknown(u) => B::new()
+                .with_maybe_source_loc(source_loc.as_ref())
+                .unknown(u),
+            ExprKind::If {
+                test_expr,
+                then_expr,
+                else_expr,
+            } => B::new().with_maybe_source_loc(source_loc.as_ref()).ite(
+                Expr::<V>::into_expr::<V, B>(Arc::unwrap_or_clone(test_expr)),
```
john-h-kastner-aws, line 428: Would this work as `Arc::unwrap_or_clone(test_expr).into_expr()`? Same comment at other call sites

cedar-policy-core/src/ast/expr.rs
```diff
@@ -397,6 +401,116 @@ impl<T> Expr<T> {
             ExprKind::Error { .. } => None,
         }
     }
+
+    /// Converts an `Expr<V>` to `B::Expr` using the provided builder.
+    ///
+    /// Preserves source location information and recursively transforms each expression node.
+    /// Note: Data may be cloned if the source expression is retained elsewhere.
+    pub fn into_expr<V, B: expr_builder::ExprBuilder>(expr: Expr<V>) -> B::Expr
```
john-h-kastner-aws, line 409: I think we shouldn't need the generic `V` here since where inside `impl<T> Expr<T>`. This can probably be `pub fn into_expr<B: expr_builder::ExprBuilder>(self) -> B::Expr`

cedar-policy-core/src/ast/expr.rs
john-h-kastner-aws: Makes sense to me

cedar-policy-core/src/ast/expr.rs
cdisselkoen: Seconded -- it should take `self`

cedar-policy-core/src/ast/expr.rs
```diff
@@ -397,6 +401,116 @@ impl<T> Expr<T> {
...
+    where
+        V: Clone,
+    {
+        let source_loc = expr.source_loc().cloned();
+        match expr.into_expr_kind() {
+            ExprKind::Lit(lit) => B::new().with_maybe_source_loc(source_loc.as_ref()).val(lit),
```
cdisselkoen, line 415: Since all of these branches start with `B::new().with_maybe_source_loc(...)`, can we pull that out above the `match` as `let builder = ...` or something?

cedar-policy-core/src/ast/expr.rs
```diff
@@ -1288,7 +1414,8 @@ pub mod expression_construction_errors {
     }
 }
 
-/// A new type wrapper around `Expr` that provides `Eq` and `Hash`
+/// A
+/// new type wrapper around `Expr` that provides `Eq` and `Hash`
```
cdisselkoen, line 1418: Probably unintentional?

cedar-policy-core/src/est.rs
```diff
@@ -379,7 +379,7 @@ impl From<ast::Template> for Policy {
 
 impl<T: Clone> From<ast::Expr<T>> for Clause {
     fn from(expr: ast::Expr<T>) -> Clause {
-        Clause::When(expr.into())
+        Clause::When(ast::Expr::<T>::into_expr::<T, Builder>(expr))
     }
```
cdisselkoen, line 383: I think we should be able to make this `expr.into_expr()` if we remove the `V` parameter from `into_expr()`?

cdisselkoen (approved):

cedar-policy-core/src/ast/expr.rs
lianah: That is what I tried initially, but the issue is that the `T` type has to implement the `Clone` trait. I tried to update `Expr<T : Clone = ()>` but that required adding the `Clone` trait on `T` in a lot of places. I just changed it to:
```
    pub fn into_expr<B: expr_builder::ExprBuilder>(self) -> B::Expr
    where
        T: Clone,
```
and it seems to work. I didn't realize I could constrain the class type parameter just on one method. Rust is cool :) Will update the pull request.

cedar-policy-core/src/ast/expr.rs
lianah: Still needs the `B` type parameter to make the compiler happy, but I was able to drop the `V` one per @cdisselkoen and your suggestion.

cedar-policy-core/src/ast/expr.rs
lianah: Great suggestion, will do.

cedar-policy-core/src/ast/expr.rs
lianah: Yes, will fix.

cedar-policy-core/src/ast/expr.rs
cdisselkoen: Yep this is the recommended convention -- no constraints on type parameters except on the particular methods (or `impl` blocks) that need them. See https://stackoverflow.com/questions/49229332/should-trait-bounds-be-duplicated-in-struct-and-impl/66369912#66369912

john-h-kastner-aws (approved):
looks great

PR #1625, opened 2025-05-16
Make typed expression generation consistent with Lean

shaobo-he-aws:
We produce Boolean literals when and/or operations short-circuit.

### DIFF
--- a/cedar-policy-validator/src/typecheck.rs
+++ b/cedar-policy-validator/src/typecheck.rs
@@ -577,12 +577,10 @@ impl<'a> SingleEnvTypechecker<'a> {
                 );
                 ans_left.then_typecheck(|typ_left, capability_left| {
                     match typ_left.data() {
-                        // LHS argument is false, so short circuit the `&&` to `False` _without_
-                        // typechecking the RHS.
+                        // LHS argument is false, so short circuit the `&&` to
+                        // `False` _without_ typechecking the RHS.
                         Some(Type::False) => TypecheckAnswer::success(
-                            ExprBuilder::with_data(Some(Type::False))
-                                .with_same_source_loc(e)
-                                .val(false),
+                            typ_left.with_maybe_source_loc(e.source_loc().cloned()),
                         ),
                         _ => {
                             // Similar to the `then` branch of an `if`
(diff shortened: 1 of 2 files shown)


cedar-policy-validator/src/typecheck.rs
```diff
@@ -673,15 +671,11 @@ impl<'a> SingleEnvTypechecker<'a> {
                 );
                 ans_left.then_typecheck(|ty_expr_left, capability_left| match ty_expr_left.data() {
                     // LHS argument is true, so short circuit the `|| to `True` _without_
-                    // typechecking the RHS. We still need to build an annotated `||` with
-                    // some RHS, so we use a literal `true`.  Contrary to `&&`, we keep a
-                    // capability  when short circuiting `||`. The left operand is `true`,
-                    // so its capability is maintained. The right operand is not evaluated,
-                    // so its capability is not considered.
+                    // typechecking the RHS.
```
cdisselkoen, line 674: I think we should still keep the comment about capabilities that begins "Contrary to `&&`" ?

john-h-kastner-aws (changes requested):
Rewriting to `true` or `false` will reintroduce the bug fixed by #1462. We need to keep `left` in short circuiting case since it is evaluated, but was can drop `right` and don't need to construct and `&&` or `||` expression.

john-h-kastner-aws:
Here are some new test cases for `level_validate.rs` that catch the bug. Existing test cases for entity manifests check that `if` short circuiting was handled correctly,b but not `&&` and `||`.

```rust
    #[test]
    fn short_circuiting_checks_evaluated_expr() {
        assert_requires_level(
            r#"permit(principal, action, resource) when { (principal.bool || true) || true};"#,
            ["principal.bool"],
            1,
        );
        assert_requires_level(
            r#"permit(principal, action, resource) when { (principal.bool && false) && false};"#,
            ["principal.bool"],
            1,
        );
        assert_requires_level(
            r#"permit(principal, action, resource) when { if (principal.bool && false) then true else false };"#,
            ["principal.bool"],
            1,
        );
        assert_requires_level(
            r#"permit(principal, action, resource) when { if (principal.bool && false) then true else false };"#,
            ["principal.bool"],
            1,
        );
        assert_requires_level(
            r#"permit(principal, action, resource) when { if true then principal.bool else false };"#,
            ["principal.bool"],
            1,
        );
        assert_requires_level(
            r#"permit(principal, action, resource) when { if false then true else principal.bool };"#,
            ["principal.bool"],
            1,
        );
    }

    #[test]
    fn short_circuiting_skips_unevaluated_expr() {
        assert_requires_level(
            r#"permit(principal, action, resource) when { true || principal.bool};"#,
            [],
            0,
        );
        assert_requires_level(
            r#"permit(principal, action, resource) when { false && principal.bool};"#,
            [],
            0,
        );
        assert_requires_level(
            r#"permit(principal, action, resource) when { if false then principal.bool else false };"#,
            [],
            0,
        );
        assert_requires_level(
            r#"permit(principal, action, resource) when { if true then true else principal.bool };"#,
            [],
            0,
        );
    }
```

john-h-kastner-aws (approved):

cdisselkoen (approved):

PR #1626, opened 2025-05-16
Make location info optional

adpaco-aws:
## Description of changes

Makes location information optional throughout most of the code base. Note that location information was already optional in some places. This change mainly focuses on the CST's `Loc`, but requires propagating the change to other structures and function/method calls like those where we expected a `miette::SourceSpan`. Whenever we can, we make our best to compute spans like we do in other parts of the code base.

### DIFF
--- a/cedar-policy-core/src/ast/name.rs
+++ b/cedar-policy-core/src/ast/name.rs
@@ -547,7 +547,7 @@ impl Name {
                         src: s.to_string(),
                         normalized_src,
                     },
-                    Loc::new(diff_byte, s.into()),
+                    Some(Loc::new(diff_byte, s.into())),
                 )))
             }
         }
@@ -570,11 +570,11 @@ impl From<ReservedNameError> for ParseError {
     fn from(value: ReservedNameError) -> Self {
         ParseError::ToAST(ToASTError::new(
             value.clone().into(),
-            match &value.0.loc {
-                Some(loc) => loc.clone(),
+            match value.0.loc {
+                loc @ Some(_) => loc,
                 None => {
                     let name_str = value.0.to_string();
-                    Loc::new(0..(name_str.len()), name_str.into())
+                    Some(Loc::new(0..(name_str.len()), name_str.into()))
                 }
             },
         ))
--- a/cedar-policy-core/src/ast/name.rs
+++ b/cedar-policy-core/src/ast/name.rs
@@ -570,13 +570,10 @@ impl From<ReservedNameError> for ParseError {
     fn from(value: ReservedNameError) -> Self {
         ParseError::ToAST(ToASTError::new(
             value.clone().into(),
-            match value.0.loc {
-                loc @ Some(_) => loc,
-                None => {
-                    let name_str = value.0.to_string();
-                    Some(Loc::new(0..(name_str.len()), name_str.into()))
-                }
-            },
+            value.0.loc.clone().or_else(|| {
+                let name_str = value.0.to_string();
+                Some(Loc::new(0..(name_str.len()), name_str.into()))
+            }),
         ))
     }
 }
--- a/cedar-policy-core/src/error_macros.rs
+++ b/cedar-policy-core/src/error_macros.rs
@@ -68,10 +68,10 @@ macro_rules! impl_diagnostic_from_two_source_loc_fields {
 
         fn labels(&self) -> Option<Box<dyn Iterator<Item = miette::LabeledSpan> + '_>> {
             let spans: Vec<_> = [&self.$i, &self.$j]
-            .into_iter()
-            .filter_map(|loc| loc.as_ref())
-            .map(|loc| miette::LabeledSpan::underline(loc.span))
-            .collect();
+                .into_iter()
+                .filter_map(|loc| loc.as_ref())
+                .map(|loc| miette::LabeledSpan::underline(loc.span))
+                .collect();
 
             if spans.is_empty() {
                 None
--- a/cedar-policy-core/src/error_macros.rs
+++ b/cedar-policy-core/src/error_macros.rs
@@ -21,13 +21,14 @@
 macro_rules! impl_diagnostic_from_source_loc_field {
     ( $i:ident ) => {
         fn source_code(&self) -> Option<&dyn miette::SourceCode> {
-            Some(&self.$i.src as &dyn miette::SourceCode)
+            self.$i.as_ref().map(|l| &l.src as &dyn miette::SourceCode)
         }
 
         fn labels(&self) -> Option<Box<dyn Iterator<Item = miette::LabeledSpan> + '_>> {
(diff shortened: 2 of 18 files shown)


cedar-policy-core/src/ast/name.rs
```diff
@@ -570,11 +570,11 @@ impl From<ReservedNameError> for ParseError {
...
             value.clone().into(),
-            match &value.0.loc {
-                Some(loc) => loc.clone(),
+            match value.0.loc {
+                loc @ Some(_) => loc,
                 None => {
                     let name_str = value.0.to_string();
-                    Loc::new(0..(name_str.len()), name_str.into())
+                    Some(Loc::new(0..(name_str.len()), name_str.into()))
                 }
```
cdisselkoen, line 578: if you want, instead of `match` this could be `value.0.loc.or_else(|| ...)`

cedar-policy-core/src/error_macros.rs
```diff
@@ -21,13 +21,14 @@
 macro_rules! impl_diagnostic_from_source_loc_field {
     ( $i:ident ) => {
         fn source_code(&self) -> Option<&dyn miette::SourceCode> {
-            Some(&self.$i.src as &dyn miette::SourceCode)
+            self.$i.as_ref().map(|l| &l.src as &dyn miette::SourceCode)
```
cdisselkoen, line 24: shouldn't we just use the `impl_diagnostic_from_source_loc_opt_field` macro (below) instead of changing this one?

If we do change this one, let's update the comment to indicate `$i` should be type `Option<Loc>`

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -425,7 +427,7 @@ impl Node<Option<cst::Policy>> {
                         slot.clone(),
                         if is_when { "when" } else { "unless" },
                     ),
-                    slot.loc.unwrap_or_else(|| c.loc.clone()),
+                    slot.loc.or_else(|| self.loc.clone()),
```
cdisselkoen, line 430: intentional change from `c.loc` to `self.loc`?

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -523,35 +540,50 @@ impl cst::PolicyImpl {
...
         // this position to indicate where to fill in vars if we're missing one.
-        let mut end_of_last_var = self.effect.loc.end();
+        let mut end_of_last_var = self.effect.loc.as_ref().map(|loc| loc.end());
 
         let mut vars = self.variables.iter();
         let maybe_principal = if let Some(scope1) = vars.next() {
-            end_of_last_var = scope1.loc.end();
-            scope1.to_principal_constraint(TolerantAstSetting::Tolerant)
+            end_of_last_var = scope1.loc.as_ref().map(|loc| loc.end()).or(end_of_last_var);
+            scope1.to_principal_constraint(TolerantAstSetting::NotTolerant)
```
cdisselkoen, line 548: curious why the change from `Tolerant` to `NotTolerant` (also a few places below)

cdisselkoen (approved):

cedar-policy-core/src/parser/cst_to_ast.rs
adpaco-aws: This was a copy-paste mistake 😢

cedar-policy-core/src/parser/cst_to_ast.rs
adpaco-aws: Not really, good catch.

cedar-policy-core/src/ast/name.rs
adpaco-aws: I did this then decided against it because then you need to compute `name_str` twice (since you're in the closure) or pre-compute it. Open to suggestions if you know how to make it clean.

chaluli (approved):

cedar-policy-core/src/ast/name.rs
cdisselkoen:
```
value.0.loc.or_else(|| {
  let name_str = value.0.to_string();
  Some(Loc::new(0..(name_str.len()), name_str.into()))
});
```

cedar-policy-core/src/error_macros.rs
cdisselkoen: Yeah. Is it used in any other crates we control?

cedar-policy-core/src/parser.rs
```diff
@@ -73,8 +73,14 @@ pub fn parse_policyset_and_also_return_policy_text(
...
     let texts = cst
         .with_generated_policyids()
-        .expect("shouldn't be None since parse_policies() and to_policyset() didn't return Err")
-        .map(|(id, policy)| (id, &text[policy.loc.start()..policy.loc.end()]))
+        .expect("shouldn't be `None` since `parse_policies` and `to_policyset` didn't return `Err`")
+        .map(|(id, policy)| {
+            if let Some(loc) = &policy.loc {
+                (id, &text[loc.start()..loc.end()])
+            } else {
+                (id, "<undefined>")
```
john-h-kastner-aws, line 81: Could this mess up the policy `Display` impl in `cedar-policy`? I think this text might be stored and returned as the lossless representation of polices parsed from text.

adpaco-aws: Sorry, I don't see how this and the `Display` impl are related... the impl relies on `LosslessPolicy` which I don't see using this API. But in general, the alternative here is to throw an error if a location is not found. Would you prefer that?

john-h-kastner-aws: This bit. Looks like if `loc` is `None` the `LosslessPolicy` will contain `<undefined>`

https://github.com/cedar-policy/cedar/blob/40be99c42cd6e102515b28db6bfb4c0b1dfc9421/cedar-policy/src/api.rs#L2234-L2243

I think what we'll need is to add a `None` variant to `LosslessPolicy` for the fast parsing path. So, you won't get lossless roundtripping if you used fast parsing.

cedar-policy-core/src/error_macros.rs
adpaco-aws: I can't find them anywhere in this repo nor in cedar-spec, so I'm removing it. Also, I was doing the same thing with the other macro... rewriting the non-optional. So I'm removing both non-optional, and using the optional versions only.

shaobo-he-aws (approved):

cedar-policy-core/src/parser.rs
adpaco-aws: Opened #1627 for this, we'll expect locations here for now.

john-h-kastner-aws (approved):
Looks good then. We'll address the opened issue before releasing fast parsing

PR #1639, opened 2025-05-23
Introduce `MaybeLoc` as a compile-time definition for locations

adpaco-aws:
## Description of changes

Introduce the `MaybeLoc` type as a compile-time definition for locations.  This definition is controlled by a new feature flag called `fast-parsing`. The definition is as follows:
 * When `fast-parsing` is disabled, `MaybeLoc` is simply `Option<Loc>`
 * When `fast-parsing` is enabled, `MaybeLoc` is `Option<Box<Loc>>`

In other words, when `fast-parsing` is enabled, we store the location is the heap (if any). The idea is for `fast-parsing` to also avoid storing locations, thus making any location-carrying structure smaller and boosting parsing performance. **Note that we don't actually enable any fast-parsing API here, but just introduce the type and flags required to enable it in a follow-up PR.**

### Performance

The reason we need a compile-time definition for `MaybeLoc` is that, while using the heap to store locations reduces the size of structures containing them, we are also introducing an overhead when allocating in it. Because of that, using the regular parsing API with `fast-parsing` is significantly slower (+20ms in the largest case):

```
example             fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ reg. parsing (fast-parsing OFF)│               │               │               │         │
   ├─ 1             24.54 µs      │ 1.157 ms      │ 27.04 µs      │ 28.63 µs      │ 1000    │ 1000
   ├─ 100           224 µs        │ 286.4 µs      │ 226.8 µs      │ 228.8 µs      │ 1000    │ 1000
   ├─ 1000          2.028 ms      │ 2.637 ms      │ 2.057 ms      │ 2.069 ms      │ 1000    │ 1000
   ├─ 5000          11.51 ms      │ 13.56 ms      │ 11.76 ms      │ 11.84 ms      │ 1000    │ 1000
   ├─ 10000         24.69 ms      │ 27.71 ms      │ 25.14 ms      │ 25.27 ms      │ 1000    │ 1000
   ╰─ 20000         51.47 ms      │ 77.83 ms      │ 52.57 ms      │ 52.79 ms      │ 1000    │ 1000

example             fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ reg. parsing (fast-parsing ON) │               │               │               │         │
   ├─ 1             24.66 µs      │ 1.199 ms      │ 27.58 µs      │ 28.92 µs      │ 1000    │ 1000
   ├─ 100           273.6 µs      │ 376.5 µs      │ 275.9 µs      │ 278.1 µs      │ 1000    │ 1000
   ├─ 1000          2.63 ms       │ 3 ms          │ 2.668 ms      │ 2.683 ms      │ 1000    │ 1000
   ├─ 5000          14.23 ms      │ 17.66 ms      │ 14.48 ms      │ 14.62 ms      │ 1000    │ 1000
   ├─ 10000         30.8 ms       │ 36.42 ms      │ 31.41 ms      │ 31.52 ms      │ 1000    │ 1000
   ╰─ 20000         63.03 ms      │ 141.1 ms      │ 72.22 ms      │ 72.68 ms      │ 1000    │ 1000
```

But if we define `MaybeLoc` as `Option<Box<Loc>>` AND always make those values be `None`, the size of structures like `Node<Option<T>>` is much smaller so they're easier to cache and pass around:
```
example             fastest       │ slowest       │ median        │ mean          │ samples │ iters
╰─ fast parsing (fast-parsing ON) │               │               │               │         │
   ├─ 1             22.95 µs      │ 310.3 µs      │ 23.83 µs      │ 25.19 µs      │ 1000    │ 1000
   ├─ 100           201.9 µs      │ 367.9 µs      │ 222.1 µs      │ 221.1 µs      │ 1000    │ 1000
   ├─ 1000          1.688 ms      │ 5.055 ms      │ 1.718 ms      │ 1.752 ms      │ 1000    │ 1000
   ├─ 5000          9.071 ms      │ 10.78 ms      │ 9.23 ms       │ 9.28 ms       │ 1000    │ 1000
   ├─ 10000         18.83 ms      │ 42.04 ms      │ 19.36 ms      │ 19.58 ms      │ 1000    │ 1000
   ╰─ 20000         38.68 ms      │ 50.41 ms      │ 39.71 ms      │ 39.89 ms      │ 1000    │ 1000
```

Note that this PR doesn't include an end-to-end workflow to produce the last table since we're hard-coding `false` for the `is_fast` flags that we pass to the grammar. So we're essentially refactoring and don't need additional tests.

### Callouts

I would appreciate the reviews to focus on two things:
 1. **Make sure we're not propagating the `MaybeLoc` definition too far, especially around API-related structs.** We only make limited changes to the `cedar-policy` crate, but since some public-facing structures like error structs have fields which are `[#from]`, I'm worried we might be introducing a breaking change with the `fast-parsing` flag which we could avoid.
 2. **Review carefully the implementations for `IntoMaybeLoc` and similar traits that I've written to manage locations in a transparent way.** Despite the more performant results, I suspect we're introducing an overhead somewhere. That's because in the experiment I did prior to this PR I was able to get ~35ms in the largest case (`MaybeLoc` was fixed to `Option<Box<Loc>>` and `is_fast = true`). My intent was to have the implementations inlined so we didn't get any overhead, but even then we have a ~5ms difference. I'll investigate further on my end too.

Note that there is a change in `parse_policyset_and_also_return_policy_text` that I need for evaluations. I'm attempting to fix that separately in #1634 , so the change shouldn't be included here.

Also, it just occurred to me this should update the changelog to mention the new flag. That change is pending.


shaobo-he-aws (commented):
If we don't actually store `Loc` when `fast-parsing` is enabled, I wonder why we need to box it

Btw, I think `OptionalLoc` sounds better than `MaybeLoc`.

adpaco-aws:
> If we don't actually store `Loc` when `fast-parsing` is enabled, I wonder why we need to box it

You still need `Loc` when `fast-parsing` is enabled and use the standard parsing APIs. While not recommended, users may want to do it anyway (e.g., to back off to standard parsing in the error case for better debugging).

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -33,12 +33,13 @@
...
 
-use super::cst;
 use super::err::{parse_errors, ParseError, ParseErrors, ToASTError, ToASTErrorKind};
-use super::loc::Loc;
+use super::loc::CloneMaybeLoc;
 use super::node::Node;
 use super::unescape::{to_pattern, to_unescaped_string};
 use super::util::{flatten_tuple_2, flatten_tuple_3, flatten_tuple_4};
+use super::{cst, MaybeLoc};
+use super::{AsLocRef, IntoMaybeLoc, Loc};
```
cdisselkoen, line 42: tiny nit: could put the `CloneMaybeLoc` import next to the `IntoMaybeLoc` import

cedar-policy-core/src/parser/cst_to_ast/to_ref_or_refs.rs
```diff
@@ -20,6 +20,8 @@ use super::Result;
 use crate::ast;
 use crate::ast::EntityReference;
 use crate::ast::EntityUID;
+use crate::parser::loc::AsLocRef;
+use crate::parser::IntoMaybeLoc;
 use crate::parser::{
     cst::{self, Literal},
     err::{self, ParseErrors, ToASTError, ToASTErrorKind},
```
cdisselkoen, line 27: super nit: could merge these imports into the big import from `crate::parser`

cedar-policy-core/src/parser/err.rs
```diff
@@ -31,11 +31,12 @@ use thiserror::Error;
 
 use crate::ast::{self, ReservedNameError};
 use crate::parser::fmt::join_with_conjunction;
-use crate::parser::loc::Loc;
+use crate::parser::loc::{Loc, MaybeLoc};
 use crate::parser::node::Node;
 use crate::parser::unescape::UnescapeError;
 
 use super::cst;
+use super::loc::AsLocRef;
```
cdisselkoen, line 39: inconsistent that we import this from `super::loc` but other items from `crate::parser::loc`

cedar-policy-core/src/parser/loc.rs
```diff
@@ -16,6 +16,14 @@
 
 use std::sync::Arc;
 
+/// Represents an optional `Loc`
+#[cfg(not(feature = "fast-parsing"))]
+pub type MaybeLoc = Option<Loc>;
+
+/// Represents an optional `Loc`
+#[cfg(feature = "fast-parsing")]
+pub type MaybeLoc = Option<Box<Loc>>;
```
cdisselkoen, line 25: Perhaps a comment here about why `Box` is actually faster.  I know it's in the PR description but would probably be valuable to immortalize here in code comments as well

cedar-policy-core/src/parser/loc.rs
```diff
@@ -74,6 +82,101 @@ impl From<&Loc> for miette::SourceSpan {
...
+    /// Automatic clone to `MaybeLoc`
+    fn clone_maybe_loc(&self) -> MaybeLoc;
+}
+
+impl CloneMaybeLoc for MaybeLoc {
+    #[inline]
+    fn clone_maybe_loc(&self) -> MaybeLoc {
+        self.clone()
+    }
+}
```
cdisselkoen, line 178: why not just use the `Clone` trait for this?

cedar-policy-core/src/parser/macros.rs
```diff
@@ -0,0 +1,18 @@
...
+            {
+                Some(Box::new($loc))
+            }
+            #[cfg(not(feature = "fast-parsing"))]
+            {
+                Some($loc)
+            }
+        }
+    };
+}
```
cdisselkoen, line 18: I wonder if this is an opportunity for further performance gain.  I think uses of this `maybe_loc!` macro still always compute the second argument (`loc`) even if just to discard it.  In particular that might involve incrementing an `Arc`.  Perhaps we need the macro to take the components with which to construct a `Loc`, rather than an already-constructed `Loc`, and only construct the `Loc` if actually needed (i.e., if `flag` is false)

cdisselkoen (approved):
Approving as-is; my comments are all minor

cedar-policy/Cargo.toml
```diff
@@ -52,8 +52,9 @@ corpus-timing = []
 
 # Experimental features.
 # Enable all experimental features with `cargo build --features "experimental"`
-experimental = ["partial-eval", "permissive-validate", "partial-validate", "entity-manifest", "protobufs", "tolerant-ast", "extended-schema", "deprecated-schema-compat"]
+experimental = ["partial-eval", "permissive-validate", "partial-validate", "entity-manifest", "protobufs", "tolerant-ast", "extended-schema", "deprecated-schema-compat", "fast-parsing"]
```
john-h-kastner-aws, line 55: Long-term this should probably not be "experimental" if we're going to commit to the API

cedar-policy-core/src/parser/macros.rs
john-h-kastner-aws: I think this looks fine as written. Macro expansion won't eval `loc`

cedar-policy-core/src/parser/loc.rs
```diff
@@ -74,6 +82,101 @@ impl From<&Loc> for miette::SourceSpan {
...
+            self
+        }
+        #[cfg(feature = "fast-parsing")]
+        {
+            self.map(|loc| Box::new(loc))
+        }
+    }
+}
+
+impl IntoMaybeLoc for Option<&Loc> {
```
john-h-kastner-aws, line 139: This looks like a potential performance footgun since it's the only trait impl that needs to clone.

john-h-kastner-aws (approved):
looks good

cedar-policy-core/src/parser/err.rs
adpaco-aws: Fixed - we don't use `loc` imports anymore.

cedar-policy-core/src/parser/loc.rs
adpaco-aws: Right, I think it's because I started with a trait with multiple `impl`s (for `Option<Loc>`, `Option<Box<Loc>>`...), then changed it to `MaybeLoc` and didn't realize it wasn't needed anymore. Anyway, it's been removed.

cdisselkoen (approved):

cedar-policy-core/src/parser/macros.rs
```diff
@@ -0,0 +1,18 @@
+/// Return a `MaybeLoc` value that depends on the flag
```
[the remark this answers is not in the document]
adpaco-aws: Added, thanks!

cedar-policy/Cargo.toml
john-h-kastner-aws: I should also say that this breaks our CI which runs with `--features experimental` and not `--all-features`

adpaco-aws: Right, I've removed it from `experimental` in [5d01fd46](https://github.com/cedar-policy/cedar/pull/1639/commits/5d01fd46a903c8273c40fde7c38852140e3c5c6a) but now I'm not sure you actually wanted me to do it. 

It's unclear to me what the contract is for features which are not `experimental`. I think we want to give `fast-parsing` some time to adjust (assuming `experimental` is the only way allowed for the API to change), and because of that it makes sense to leave it there for now. But you're right that it should be "stabilized" at some point.

adpaco-aws: > I should also say that this breaks our CI which runs with `--features experimental` and not `--all-features`

Isn't `--all-features` tested by `cargo hack` in the `msrv` jobs? For example, [this one](https://github.com/cedar-policy/cedar/actions/runs/15305209793/job/43056023453?pr=1639).

cedar-policy-core/src/parser/loc.rs
adpaco-aws: Good catch. IIRC there might be places where this could be avoided. But I'd like to avoid making those changes now, let me open an issue so we don't lose track.

adpaco-aws: https://github.com/cedar-policy/cedar/issues/1642

cedar-policy/Cargo.toml
adpaco-aws: Alright, I'm keeping [5d01fd46](https://github.com/cedar-policy/cedar/pull/1639/commits/5d01fd46a903c8273c40fde7c38852140e3c5c6a) so `fast-parsing` is not in `experimental`.

adpaco-aws:
Thanks for the reviews @cdisselkoen @john-h-kastner-aws ! I've made a couple more changes to remove the tweak I needed for evaluations (we should complete that in #1634) and adding a line to the changelog.

cdisselkoen (approved):

PR #1641, opened 2025-05-28
New `run-tests` command in Cedar CLI to run policy tests in JSON format

zhengyao-lin:
## Description of changes

This PR adds a utility in Cedar CLI to run test cases in the following JSON format
```
[
  {
    "name": String, // Optional
    "request": {
      "principal": String, // Parsed as <EntityType>::"<ID>"
      "action": String,
      "resource": String,
      "context": { .. }
    },
    "entities": { .. } // Same as the main entity store format
    "decision": "allow" / "deny",
    "reason": [String],
    "num_errors": Number,
  },
]
```

The additional command `run-tests` can be called with
```
$ cedar run-tests --policies <policy set> --schema <schema> --tests <tests in JSON>
```
Which runs every test (request + entity store) against the given policy set.

And it gives output like the following
```
running 3 test(s)
  test (unamed) ... error(s): missing reason(s): `policy2`
  test (unamed) ... fail: expected Deny, got Allow
  test (unamed) ... ok
results: 1 passed, 2 failed
```

Currently, we report test failure if:
- The decision (allow/deny) does not match the expected one in the test; or
- If there's unmatched number of runtime errors; or
- If some `reason` specified in the test case is not in the `reason` set in the response.


cedar-policy-cli/Cargo.toml
```diff
@@ -19,6 +19,7 @@ serde_json = "1.0"
 miette = { version = "7.6.0", features = ["fancy"] }
 thiserror = "2.0"
 semver = "1.0.26"
+colored = "3.0.0"
```
john-h-kastner-aws: From `cargo-deny` in CI, `colored` doesn't have a compatible license: 

```

error[rejected]: failed to satisfy license requirements
  ┌─ colored 3.0.0 (registry+https://github.com/rust-lang/crates.io-index):4:12
  │
4 │ license = "MPL-2.0"
  │            ^^^^^^^
  │            │
  │            license expression retrieved via Cargo.toml `license`
  │            rejected: license is considered copyleft
  │
  = MPL-2.0 - Mozilla Public License 2.0:
  =   - OSI approved
  =   - FSF Free/Libre
  =   - Copyleft
  = colored v3.0.0
    └── cedar-policy-cli v4.4.0
```

So, we can't use `colored`. Looking at miette, it uses https://github.com/owo-colors/owo-colors which is MIT licensed, so let's use that instead.

zhengyao-lin: fixed!

cedar-policy-cli/sample-data/tiny_sandboxes/sample7/tests-combined.json
john-h-kastner-aws, line 1: I'd prefer to keep these sample test case files smaller and nicely formatted.

cedar-policy-cli/src/lib.rs
```diff
@@ -1453,6 +1472,270 @@ pub fn partial_authorize(args: &PartiallyAuthorizeArgs) -> CedarExitCode {
...
+                    .map(|r| format!("`{}`", r))
+                    .collect::<Vec<_>>()
+                    .join(", ")
+            ));
+        }
+
+        // Check that evaluation errors are expected
+        let has_error = ans.diagnostics().errors().next().is_some();
+        if has_error && !test.has_error {
+            warnings.push(format!(
```
john-h-kastner-aws: Looks like this bit isn't covered. Seems interesting enough to be worth adding a test

cedar-policy-cli/src/lib.rs
```diff
@@ -1453,6 +1472,270 @@ pub fn partial_authorize(args: &PartiallyAuthorizeArgs) -> CedarExitCode {
...
+    } else {
+        CedarExitCode::Success
+    })
+}
+
+pub fn run_tests(args: &RunTestsArgs) -> CedarExitCode {
+    match run_tests_inner(args) {
+        Ok(status) => status,
+        Err(e) => {
+            println!("{e:?}");
```
john-h-kastner-aws, line 1618: IIRC, you need to `println!("{:?}", Report::new(e))` to get miette to pretty print the error

cedar-policy-cli/src/lib.rs
```diff
@@ -1453,6 +1472,270 @@ pub fn partial_authorize(args: &PartiallyAuthorizeArgs) -> CedarExitCode {
...
+    RequestValidationError(#[from] RequestValidationError),
+    #[error("error when parsing entities")]
+    EntitiesError(#[from] EntitiesError),
+    #[error("missing field `{0}`")]
+    MissingField(String),
+}
+
+impl TestCase {
+    /// Parse a `TestCase` from a JSON value.
+    fn from_json_value(
```
john-h-kastner-aws: Could we simplify this a bit by having `load_partial_tests` parse into a struct like 

```rust
struct TestCaseJson {
  request: RequestJSON,
  entities: Value,
  ...
}
```

Should reduce some repetition here  

should avoid

zhengyao-lin: For `load_partial_tests`, I was trying to isolate format errors within each test, so that if, e.g., a test is missing the `request` field, that won't just cause the entire testing run to fail.

are you thinking that I should move some of the parsing to be macro generated?

cedar-policy-cli/src/lib.rs
zhengyao-lin: `e` is already a `Report` though. Maybe I'm not familiar with miette; should I use other error types and then only convert to Report at the end?

cedar-policy-cli/src/lib.rs
```diff
@@ -1453,6 +1472,270 @@ pub fn partial_authorize(args: &PartiallyAuthorizeArgs) -> CedarExitCode {
     }
 }
 
+enum TestResult {
+    Pass,
+    Warning,
```
adpaco: Why are `Warning`s considered a result?

cedar-policy-cli/src/lib.rs
```diff
@@ -1453,6 +1472,270 @@ pub fn partial_authorize(args: &PartiallyAuthorizeArgs) -> CedarExitCode {
...
+                    .into_iter()
+                    .map(|r| format!("`{}`", r))
+                    .collect::<Vec<_>>()
+                    .join(", ")
+            ));
+        }
+
+        // Check that evaluation errors are expected
+        let has_error = ans.diagnostics().errors().next().is_some();
+        if has_error && !test.has_error {
```
adpaco: Why don't you check evaluation errors like you do warnings? You could have a list of declared errors and check that they appear as actual errors. Or are there any challenges in doing so?

cedar-policy-cli/src/lib.rs
```diff
@@ -1453,6 +1472,270 @@ pub fn partial_authorize(args: &PartiallyAuthorizeArgs) -> CedarExitCode {
...
+            warnings.push(format!(
+                "unexpected runtime error(s): {}",
+                ans.diagnostics()
+                    .errors()
+                    .map(|e| e.to_string())
+                    .collect::<Vec<_>>()
+                    .join(", ")
+            ));
+        } else if !has_error && test.has_error {
+            warnings.push("expected error(s) but none were found".to_string());
```
adpaco: In cases where errors don't match the expected ones, shoudn't we rather fail?

cedar-policy-cli/src/lib.rs
```diff
@@ -1453,6 +1472,270 @@ pub fn partial_authorize(args: &PartiallyAuthorizeArgs) -> CedarExitCode {
...
+    } else {
+        print!("  test (unamed) ... ");
+    }
+
+    let test = match TestCase::from_json_value(test.clone(), schema).into_diagnostic() {
+        Ok(test) => test,
+        Err(e) => {
+            println!("{}:", "error".red());
+            println!("{e:?}");
+            return TestResult::Fail;
```
adpaco: No. You won't be able to distinguish tests failing because they didn't parse from tests failing because of other reasons. This function should probably return a `Result<TestResult>`.

cedar-policy-cli/src/lib.rs
```diff
@@ -1453,6 +1472,270 @@ pub fn partial_authorize(args: &PartiallyAuthorizeArgs) -> CedarExitCode {
...
+    }
+
+    println!(
+        "results: {} {}, {} {}, {} {}",
+        tests.len() - total_fails - total_warnings,
+        if total_fails == 0 && total_warnings == 0 {
+            "passed".green().to_string()
+        } else {
+            "passed".to_string()
+        },
```
adpaco: What are we doing here? If everything passes we get colored output otherwise we don't?

cedar-policy-cli/src/lib.rs
```diff
@@ -1453,6 +1472,270 @@ pub fn partial_authorize(args: &PartiallyAuthorizeArgs) -> CedarExitCode {
...
+
+    let schema = args.schema.get_schema()?;
+    let tests = load_partial_tests(&args.tests)?;
+
+    let mut total_fails: usize = 0;
+    let mut total_warnings: usize = 0;
+
+    println!("running {} test(s)", tests.len());
+    for test in tests.iter() {
+        match run_one_test(&policies, Some(&schema), test) {
```
adpaco: Can we run these in parallel?

cedar-policy-cli/src/lib.rs
adpaco: It doesn't look like there's a good reason to not derive `Deserialize`. You could even convert parsing errors to your custom ones if you wanted.

zhengyao-lin:
Hi @adpaco-aws, thank you so much for the comments!
I'll make changes to address your suggestions, but to answer some high-level design decisions:

- Having entities in the tests: Currently, each test has its own entity store. This is primarily because the overall goal is to test and improve the Cedar policy itself, rather than Cedar policy + entity store. Maybe this is up to debate if you think that people want to test on a specific entity store in most use cases; but by having the entity stores attached, we can enlarge the space of tests, and can make sure that the policy set behave robustly on, e.g., edge cases of entity hierarchies.

    If you think this is a good idea, we can potentially also add a flag to override the entity stores in the test cases.

- Aligning `reason` and `has_error` with `Diagnostics`: One external reason for this is that it is currently hard to obtain the exact error type from the SMT encoding, from which we actually generate the tests (since it encodes potentially erroring expressions with an `Option` type).


Otherwise, I think converting some of the warnings to errors makes sense; although @john-h-kastner-aws I remember you had some thoughts on this?

cedar-policy-cli/src/lib.rs
zhengyao-lin: This is more of an external reason, but on the test generation side, there is currently no straightforward way to obtian the exact evaluation error from the SMT encoding (which encodes all evaluation results as an option type)

cedar-policy-cli/src/lib.rs
zhengyao-lin: I'm just trying to improve user experience here :)
Basically if there is no error/warning, the passed is printed in green to visually indicate that everything is good. Similarly, if there are warnings/errors, the corresponding word like "failed" and "warning(s)" will be highlighted.

cedar-policy-cli/src/lib.rs
zhengyao-lin: We can, but I think these tests typically run pretty fast and I don't want to complicate this first version with optimizations.

cedar-policy-cli/src/lib.rs
adpaco: Sounds like I'm missing the bigger picture here, but could we do better? Number of evaluation errors maybe?

cedar-policy-cli/src/lib.rs
adpaco: Sounds good, I liked colored output. However, you should check that the output goes to a terminal if you're doing colored, otherwise you'll end up printing escape characters if the output is piped to a file.

cedar-policy-cli/src/lib.rs
zhengyao-lin: Yeah number of errors should be doable. I'll try to add it on both sides.

zhengyao-lin:
@adpaco-aws I addressed some of your comments:

- All test warnings are now errors
- TestCase is parsed with the Deserialize trait (although since we can't pass schema into the deserializer, the `--schema` flag is now removed); in the future, we can potentially add an optional --schema flag for stricter validation.
- Added missing checks for whether the terminal supports color
- Distinguish parsing error from test failure

For the purpose of this PR, maybe we can use the `has_error` field for now? To enforce the number of errors, we can add an optional `num_errors` field in the future, which won't conflict with the current setup. Using `has_error` is also more friendly for people handwriting test cases.

There's also a test failure that I'm not sure how to fix: https://github.com/cedar-policy/cedar/actions/runs/15736219258/job/44349687703?pr=1641
It doesn't seem to be related to this PR

adpaco:
Thanks for addressing my comments!

> we can add an optional `num_errors` field in the future, which won't conflict with the current setup.

But then you'll have a field that you don't need, right? Users would be able to write things like `has_error = true` and `num_errors = 0`, which requires you to do additional checks.

> There's also a test failure that I'm not sure how to fix: https://github.com/cedar-policy/cedar/actions/runs/15736219258/job/44349687703?pr=1641
It doesn't seem to be related to this PR

I think it is, actually. There is a `test_format_samples` test in `cedar-policy-cli/tests/sample.rs` that's checking the new samples are properly formatted. It's looking like sample10 is not well-formatted.

zhengyao-lin:
@adpaco-aws @john-h-kastner-aws I have fixed all of the suggestions above (including replacing `has_error` with `num_errors`)

Is it ok to merge this?

cedar-policy-formatter/tests/cli-snapshots/cedar_policy_formatter__pprint__fmt__tests__format_files@tiny_sandboxes__sample10__policy.cedar.snap
adpaco, line 1: Is this file supposed to be here?

adpaco-aws (commented):
Almost there 😄

Can you please update the PR description to match what's in your contribution? In particular, the format needs to be updated.

cedar-policy-formatter/tests/cli-snapshots/cedar_policy_formatter__pprint__fmt__tests__format_files@tiny_sandboxes__sample10__policy.cedar.snap
zhengyao-lin: It seems to be required for fixing this CI error: https://github.com/cedar-policy/cedar/actions/runs/15736219258/job/44349687703?pr=1641

@john-h-kastner-aws could you please confirm?

zhengyao-lin:
@adpaco-aws good catch! it should be fixed now

cedar-policy-formatter/tests/cli-snapshots/cedar_policy_formatter__pprint__fmt__tests__format_files@tiny_sandboxes__sample10__policy.cedar.snap
john-h-kastner-aws: correct. We use automatically use CLI example files as test cases for the policy formatter

PR #1644, opened 2025-05-30
Expose raw parsing APIs for policies

adpaco-aws:
## Description of changes

Expose `parse_raw` APIs for `PolicySet`, `Policy` and `Template`.

Internally, these call new `parse*_raw` counterparts in `cedar-policy-core` that end up calling `parse_collect_errors`. The main difference there is that we're now making use of the additional `is_fast` flag which we pass to the grammar parsing function. We also rename the feature to `raw-parsing` which is more accurate.

The PR includes some tests for the new APIs, both in `cedar-policy` and `cedar-policy-core`. Feel free to ask for additional testing if you they're insufficient.


cedar-policy-core/src/parser/loc.rs
```diff
@@ -29,9 +29,9 @@ pub type MaybeLoc = Option<Loc>;
 /// - `Some` case: Slightly slower due to heap allocation and indirect access
 /// - `None` case: More memory efficient as no space is reserved for the location data
 ///
-/// When the `fast-parsing` feature is enabled, we use this type and avoid
+/// When the `lossy-parsing` feature is enabled, we use this type and avoid
 /// storing locations during parsing, maximizing the parsing performance.
-#[cfg(feature = "fast-parsing")]
+#[cfg(feature = "lossy-parsing")]
```
chaluli: I guess it's probably because I'm relatively new to Rust. Why does using `Option<Box<Loc>>` achieve not storing locations during parsing while `Option<Loc>` does not?

I would have suspected that "No location" would be represented by `None` in both situations.

john-h-kastner-aws: They're both `None`, but they need to reserve the same amount of memory as the `Some` variant. So it's `sizeOf Box<Loc>` instead of `sizeOf Loc` and boxes are just a pointer, so smaller than our `Loc` struct.

adpaco: The `Option<Box<Loc>>` type does not avoid storing locations by itself. The `Option<Box<Loc>>` type is smaller than `Option<Loc>` because `Box` is just a pointer to some `Loc` in the heap. This reduction in size makes objects containing locations easier to pass around and cache, but we're also introducing an overhead when allocating/retrieving data from the heap. The trick here is that `parse_lossy` always has `None`s for locations, so you get maximum benefits because you never need to allocate or pull any location data from the heap. It does however make regular parsing slower, see a table in https://github.com/cedar-policy/cedar/pull/1639

cedar-policy-core/src/parser/text_to_cst.rs
```diff
@@ -46,10 +46,11 @@ fn parse_collect_errors<'a, P, T>(
         bool,
         &'a str,
     ) -> Result<T, err::RawParseError<'a>>,
+    is_fast: bool,
```
cdisselkoen, line 49: probably rename this to `is_raw` too?

cedar-policy/src/api.rs
```diff
@@ -3681,6 +3747,29 @@ impl Policy {
...
 
+    /// Attempt to parse a [`Policy`] from source, without retaining source information.
+    /// If `id` is Some, the policy will be given that Policy Id.
+    /// If `id` is None, then "policy0" will be used.
+    /// The behavior around None may change in the future.
+    ///
+    /// This can fail if the policy fails to parse.
+    /// It can also fail if a template was passed in, as this function only accepts static
+    /// policies.
+    /// Similar to [`Policy::parse`], but does not retain the original source
```
cdisselkoen: Probably a blank `///` line between these two lines, like in the other functions

cedar-policy-core/src/parser/loc.rs
chaluli: Oh, I didn't know that a none requires the same space allocation as the some type. Then that makes a lot of since. Thanks!

cedar-policy-core/src/parser/text_to_cst.rs
adpaco: Actually, I intend on renaming this flag to something like `keep_locations`. This means we'll also need to flip the boolean values everywhere, so I'd rather do it in a follow up PR.

cedar-policy/src/api.rs
adpaco: Done.

cedar-policy-core/src/parser.rs
```diff
@@ -54,6 +54,13 @@ pub fn parse_policyset(text: &str) -> Result<ast::PolicySet, err::ParseErrors> {
     cst.to_policyset()
 }
 
+/// Like [`parse_policyset`], but without retaining source information.
+#[cfg(feature = "raw-parsing")]
```
john-h-kastner-aws, line 58: Do this an other similar internal functions need to be behind the `raw-parsing` flag?

john-h-kastner-aws (approved):

cedar-policy-core/src/parser.rs
adpaco: No, they don't need to. But I think it's good to add them since they should only be used from `parse_raw`.

john-h-kastner-aws: I was thinking it might just be easier to test, but this way is fine

john-h-kastner-aws (approved):

PR #1648, opened 2025-06-03
Add expression visitor used by LSP implementation

john-h-kastner-aws:
## Description of changes

Adds an `ExprVisitor` trait to `cedar-policy-core` that is used in implementing some LSP functionality

### DIFF
--- a/cedar-policy-core/src/ast.rs
+++ b/cedar-policy-core/src/ast.rs
@@ -54,3 +54,5 @@ mod expr_iterator;
 pub use expr_iterator::*;
 mod annotation;
 pub use annotation::*;
+mod expr_visitor;
+pub use expr_visitor::*;
--- a/cedar-policy-core/src/ast/expr_visitor.rs
+++ b/cedar-policy-core/src/ast/expr_visitor.rs
@@ -1,0 +1,406 @@
+use std::{collections::BTreeMap, sync::Arc};
+
+use crate::{
+    ast::{
+        expr_allows_errors::AstExprErrorKind, BinaryOp, EntityType, Expr, ExprKind, Literal, Name,
+        Pattern, SlotId, UnaryOp, Unknown, Var,
+    },
+    parser::Loc,
+};
+use smol_str::SmolStr;
+
+/// A visitor trait for traversing Cedar Policy Abstract Syntax Trees (ASTs).
+///
+/// This trait enables type-safe traversal of Cedar policy expressions in the language server.
+/// Implementers can selectively override methods to process specific expression types
+/// while inheriting default behavior for others.
+///
+/// # Usage
+///
+/// Implement this trait and override the methods for the expression types you want to process.
+///
+/// # Traversal Behavior
+///
+/// By default, the visitor will traverse the expression tree depth-first, stopping and
+/// returning the first non-None result. Implementers can override this behavior by
+/// providing custom implementations for specific visit methods.
+pub trait ExprVisitor {
+    /// The output type for this visitor.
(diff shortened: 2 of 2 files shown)


cedar-policy-core/src/ast.rs
```diff
@@ -54,3 +54,5 @@ mod expr_iterator;
 pub use expr_iterator::*;
 mod annotation;
 pub use annotation::*;
+mod expr_visitor;
+pub use expr_visitor::*;
```
cdisselkoen, line 58: is there duplicate functionality between `expr_iterator` and `expr_visitor`? could one be implemented in terms of the other?

cedar-policy-core/src/ast/expr_visitor.rs
```diff
@@ -0,0 +1,408 @@
...
+///
+/// # Traversal Behavior
+///
+/// By default, the visitor will traverse the expression tree depth-first, stopping and
+/// returning the first non-None result. Implementers can override this behavior by
+/// providing custom implementations for specific visit methods.
+pub trait ExprVisitor {
+    /// The output type for this visitor.
+    ///
+    /// By default, one a `visit_*` function returns `Some`, this visitor will
```
cdisselkoen, line 45: typo "once"

cedar-policy-core/src/ast.rs
john-h-kastner-aws: It's not immediately obvious.

The expression iterator hardcodes a preorder traversal while the visitor allows customization. On the other hand the iterator is implemented without recursion and can incrementally retrieve the next expression without doing the full traversal.

cedar-policy-core/src/ast/expr_visitor.rs
```diff
@@ -0,0 +1,408 @@
+use std::{collections::BTreeMap, sync::Arc};
```
adpaco, line 17: Needs copyright.

cedar-policy-core/src/ast/expr_visitor.rs
```diff
@@ -0,0 +1,408 @@
...
+
+    /// Entry point for visiting an expression.
+    ///
+    /// This is typically called to begin traversal and dispatches to `visit_expr_kind`.
+    fn visit_expr(&mut self, expr: &Expr) -> Option<Self::Output> {
+        self.visit_expr_kind(expr.expr_kind(), expr.source_loc())
+    }
+
+    /// Dispatches to the appropriate visit method based on the expression kind.
+    fn visit_expr_kind(&mut self, expr_kind: &ExprKind, loc: Option<&Loc>) -> Option<Self::Output> {
```
adpaco: Why do we have `loc ` as an argument? Most expressions have their own `loc` so why do they need it? And I guess you could always keep the `loc` in the custom implementation if you really care about it.

For example, we could have a `visit_expr_inner` method instead with a single `expr: &Expr` argument that matches on `expr.expr_kind()` at first, right? I don't see why the inner method is based on kind.

john-h-kastner-aws: I think I addressed this by getting rid of the `visit_expr_kind` function. Now the `match` happens directly in `visit_expr`. The other visitor method still have a `loc` argument which is used by the calling code in the LSP

cedar-policy-core/src/ast/expr_visitor.rs
john-h-kastner-aws: added

adpaco-aws (approved):

PR #1651, opened 2025-06-05
Address type mismatch false positive by implementing context validation

mishjude:
## Description of changes

Address type mismatch false positive for context validation by:
-Creating a new method validate_context to determine whether a context is valid against the provided schema and action 
-Calling validate_context in check_parse_context and returned CheckParseAnswer::failure if an error was returned 
-Writing a test case to address a faulty context situation and ensuring it passes locally

(I am like 70% sure the documentation was written under the assumption that the issue I was trying to fix was not an issue (and that context validation worked as intended), so I don't think the documentation needs to be updated, but I'm also not 100% sure)


cedar-policy-core/src/validator/coreschema.rs
```diff
@@ -260,6 +260,41 @@ impl ast::RequestSchema for ValidatorSchema {
         }
         Ok(())
     }
+
+    /// Validate a context against a schema for a specific action
+    fn validate_context<'a>(
```
john-h-kastner-aws, line 251: Should be able to refactor `validate_request` to call this function

cedar-policy/src/api.rs
```diff
@@ -4688,6 +4688,32 @@ impl Context {
     ) -> Result<Self, ContextCreationError> {
         Self::from_pairs(self.into_iter().chain(other_context))
     }
+
+    /// Validates this context against the provided schema
+    /// Returns Ok(()) if the context is valid according to the schema, or an error otherwise
+    pub fn validate_context(
```
john-h-kastner-aws: since this is a method on `Context`, let's just call it `validate`

cedar-policy/src/ffi/check_parse.rs
```diff
@@ -534,4 +547,152 @@ mod test {
...
+                                ],
+                                "resourceTypes": [
+                                    "Photo"
+                                ],
+                                "context": {
+                                    "type": "ContextType"
+                                }
+                            }
+                        },
+                        "createPhoto": {
```
john-h-kastner-aws: For simplicity, we should be able to remove the action other than `viewPhoto` from the test schema. They won't impact context validation. Likewise we can delete the attributes on the entity types and probably some of the entity types completely to a get.

cedar-policy/src/api.rs
```diff
@@ -4688,6 +4688,32 @@ impl Context {
...
+                // Call the validate_context function from coreschema.rs
+                RequestSchema::validate_context(
+                    &schema.0,
+                    &self.0,
+                    action.as_ref(),
+                    &Extensions::all_available(),
+                )
+                .map_err(|e| format!("Context validation failed: {}", e))
+            }
+            _ => Ok(()), // If no schema or action provided, we can't validate, so return Ok
```
john-h-kastner-aws: Let's just make them non-optional and handle the case when their absent in the calling code

cedar-policy/src/api.rs
```diff
@@ -4688,6 +4688,32 @@ impl Context {
     ) -> Result<Self, ContextCreationError> {
         Self::from_pairs(self.into_iter().chain(other_context))
     }
+
+    /// Validates this context against the provided schema
+    /// Returns Ok(()) if the context is valid according to the schema, or an error otherwise
```
john-h-kastner-aws: nits for nicer rendering in rustdocs:

```suggestion
    /// Validates this context against the provided schema.
    ///
    /// Returns `Ok(())` if the context is valid according to the schema, or an error otherwise
```

also we should have a sentence mentioning that this validation is also performed by `Request::new`, so you don't need to call this function if you're already validating the whole request

cedar-policy/src/api.rs
```diff
@@ -4688,6 +4688,32 @@ impl Context {
...
+    pub fn validate_context(
+        &self,
+        schema: Option<&crate::Schema>,
+        action: Option<&EntityUid>,
+    ) -> std::result::Result<(), String> {
+        match (schema, action) {
+            (Some(schema), Some(action)) => {
+                // Get the context schema for this action
+                let _context_schema = Self::get_context_schema(schema, action)
+                    .map_err(|e| format!("Context schema error: {}", e))?;
```
john-h-kastner-aws: I don't think this line is necessary. It will return an error if `action` isn't in the schema, but so should `RequestSchema::validate_context`.

cedar-policy/src/api.rs
```diff
@@ -4688,6 +4688,32 @@ impl Context {
...
         Self::from_pairs(self.into_iter().chain(other_context))
     }
+
+    /// Validates this context against the provided schema
+    /// Returns Ok(()) if the context is valid according to the schema, or an error otherwise
+    pub fn validate_context(
+        &self,
+        schema: Option<&crate::Schema>,
+        action: Option<&EntityUid>,
+    ) -> std::result::Result<(), String> {
```
john-h-kastner-aws: IMO makes sense to make this return type `Result<(), RequestValidationError>` (using the `RequestValidationError` type defined in `api/err.rs`)

cedar-policy-core/src/validator/coreschema.rs
```diff
@@ -260,6 +246,41 @@ impl ast::RequestSchema for ValidatorSchema {
...
+        &self,
+        context: &ast::Context,
+        action: &ast::EntityUID,
+        extensions: &Extensions<'a>,
+    ) -> std::result::Result<(), RequestValidationError> {
+        // Following the same logic in validate_request
```
cdisselkoen: I think this comment is outdated now that validate_request calls this function

cedar-policy-core/src/validator/coreschema.rs
```diff
@@ -260,6 +246,41 @@ impl ast::RequestSchema for ValidatorSchema {
...
+        // Get the action ID
+        let action_arc = Arc::new(action.clone());
```
cdisselkoen: can we avoid cloning the action and constructing an `Arc` on the happy path?  I think the clone and Arc are only needed on error paths

cedar-policy/src/api.rs
```diff
@@ -32,6 +32,7 @@ pub use cedar_policy_core::validator::entity_manifest::{
 };
 use cedar_policy_core::validator::json_schema;
 use cedar_policy_core::validator::typecheck::{PolicyCheck, Typechecker};
+pub use err::RequestValidationError;
```
cdisselkoen: Why is this necessary?  I think there is already a `pub use err::*` below?

cdisselkoen (approved):
Looks good

cedar-policy-core/src/validator/coreschema.rs
```diff
@@ -254,14 +254,12 @@ impl ast::RequestSchema for ValidatorSchema {
...
-        // Following the same logic in validate_request
         // Get the action ID
-        let action_arc = Arc::new(action.clone());
-        let validator_action_id = self.get_action_id(&action_arc).ok_or_else(|| {
-            request_validation_errors::UndeclaredActionError {
-                action: action_arc.clone(),
-            }
-        })?;
+        let validator_action_id =
+            self.get_action_id(&Arc::new(action.clone()))
```
cdisselkoen: Can this just be `let validator_action_id = self.get_action_id(action)`, or does that not work?

john-h-kastner-aws: Yeah, I think this should work

mishjude: noted! made that change!

cdisselkoen (approved):

john-h-kastner-aws (approved):

PR #1652, opened 2025-06-06
allow entity attributes and tags to contain Action-typed entities

### LINKED ISSUE #304: Schema Validation for Storing Actions in Entities

cdisselkoen:
## Description of changes

Allow entity attributes and tags to contain Action-typed entities.  This was always supported in the evaluator, but is now also supported in policy validation, entity validation, and request validation.  Any action UID literals that appear in attribute values or tag values must refer to actions that actually exist in the schema.

## Issue #, if available

Fixes #304 

### DIFF
--- a/cedar-policy-core/src/validator/schema.rs
+++ b/cedar-policy-core/src/validator/schema.rs
@@ -827,6 +827,7 @@ impl ValidatorSchema {
         // entity types or actions. Any keys left in the map are therefore
         // undeclared.
         Self::check_for_undeclared(
+            &all_defs,
             &entity_types,
             entity_children.into_keys(),
             &action_ids,
@@ -848,19 +849,13 @@ impl ValidatorSchema {
     /// This function assumes that all entity types are fully qualified, which
     /// is indicated by the use of the [`EntityType`] and [`EntityUID`] types.
     fn check_for_undeclared(
+        all_defs: &AllDefs,
         entity_types: &HashMap<EntityType, ValidatorEntityType>,
         undeclared_parent_entities: impl IntoIterator<Item = EntityType>,
         action_ids: &HashMap<EntityUID, ValidatorActionId>,
         undeclared_parent_actions: impl IntoIterator<Item = EntityUID>,
         common_types: impl IntoIterator<Item = ValidatorType>,
     ) -> Result<()> {
-        // For the purposes of this function, the set of declared entity types
-        // includes the types of all declared actions.
-        let declared_entity_types: HashSet<&EntityType> = entity_types
-            .keys()
-            .chain(action_ids.keys().map(|action_uid| action_uid.entity_type()))
-            .collect();
-
         // When we constructed `entity_types`, we removed entity types from  the
         // `entity_children` map as we encountered a declaration for that type.
         // Any entity types left in the map are therefore undeclared. These are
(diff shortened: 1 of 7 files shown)


cedar-policy-core/src/validator/schema.rs
```diff
@@ -607,6 +607,18 @@ impl ValidatorSchema {
...
 
+        // Add entity-type declarations for the `Action` type in each namespace
+        // that contains actions.  This allows schemas to, for instance, have
+        // attributes of type `Action` or `Set<Action>` etc.
+        let action_types: HashSet<InternalName> = all_defs
+            .action_defs
+            .iter()
+            .map(|action_def| action_def.entity_type().as_ref().as_ref().clone())
+            .collect();
+        for action_type in action_types {
```
katherine-hough, line 618: Could you use for_each instead of the collect + loop?

katherine-hough (approved):

cedar-policy-core/src/validator/schema.rs
cdisselkoen: The original reason for the `.collect()` is to avoid mutating-while-iterating.  Stated in terms of the borrow checker, the first pass over `action_types` (lines 613-617) needs an immutable borrow of `all_defs`, while the second pass (lines 618-620) needs a mutable borrow of `all_defs`.  So, we have to completely finish iterating `all_defs` in the first pass before the second pass can start; we can't make the `mark_as_defined_entity_type()` calls in the same iterator chain.

I'm certainly open to a better solution if there's a better pattern to use here.

john-h-kastner-aws: We could probably obtain the set of defined action types without iterating over `all_defs.action_defs` since there's only ever one action type defined per name space. Maybe that'd be a little more brittle than basing it off the actual actions, but it could save some time if there are a lot of actions in one namespace

katherine-hough: That makes sense. Thank you!

cedar-policy-core/src/validator/schema.rs
```diff
@@ -844,6 +854,13 @@ impl ValidatorSchema {
         undeclared_parent_actions: impl IntoIterator<Item = EntityUID>,
         common_types: impl IntoIterator<Item = ValidatorType>,
     ) -> Result<()> {
+        // For the purposes of this function, the set of declared entity types
+        // includes the types of all declared actions.
+        let declared_entity_types: HashSet<&EntityType> = entity_types
```
john-h-kastner-aws: Could this possibly be pulled from `all_defs` since you updated that to include that action entity types already?

cedar-policy-core/src/validator/schema.rs
cdisselkoen: > there's only ever one action type defined per name space

How I've implemented it in this PR, there are either 0 or 1 action types defined per namespace.  I have a test which demonstrates that if there are no actions defined in a namespace, the corresponding action type in that namespace is not defined either.  This is important for name resolution to work if you want to refer to unqualified `Action` from within a namespace block

cedar-policy-core/src/validator/schema.rs
cdisselkoen: Yep; done this in the latest rev of this PR

katherine-hough (approved):

PR #1658, opened 2025-06-12
Address error reporting principal, action, and resource schema validation

mishjude:
## Description of changes

Address principal, action, and resource schema validation by:
-Creating a new method validate_scope_variables to determine whether the principal, action, and resource are valid against the provided schema
-Creating a new struct ScopeVariablesParsingCall to take non-optional principal, action, resource, and schema variables for validate_scope_variables
-Calling validate_scope_variables in check_parse_scope_variables and returned CheckParseAnswer::failure if an error was returned
-Writing 3 test cases to address faulty principal, action, and resource inputs and ensuring it passes locally

(I believe it should be invisible as it should just improve error reporting.)


cedar-policy-core/src/validator/coreschema.rs
```diff
@@ -176,75 +176,18 @@ impl ast::RequestSchema for ValidatorSchema {
...
-                // expensive than we want for this validation step.
-                // Instead, we just let the above checks (that principal and
-                // resource are of types that at least _exist_ in the schema)
-                // suffice.
-            }
+        // Only validate entities if principal, action, and resource UIDs are all available
+        if let (Some(principal_uid), Some(action_uid), Some(resource_uid)) =
+            (request.principal().uid(), a_uid, request.resource().uid())
+        {
+            self.validate_scope_variables(principal_uid, action_uid, resource_uid)
```
adpaco: This doesn't match the behavior we had previously, right? We would validate each component separately if we could, which is less restrictive than having all 3. Why are we changing it?

cedar-policy-core/src/validator/coreschema.rs
```diff
@@ -292,6 +234,62 @@ impl ast::RequestSchema for ValidatorSchema {
...
+
+        let validator_action_id = self.get_action_id(action).ok_or_else(|| {
+            request_validation_errors::UndeclaredActionError {
+                action: Arc::new(action.clone()),
+            }
+        })?;
+
+        let principal_type = principal.entity_type();
+        validator_action_id.check_principal_type(principal_type, &Arc::new(action.clone()))?;
+        let principal_type = resource.entity_type();
```
adpaco: Did you mean
```suggestion
        let resource_type = resource.entity_type();
```
?

cedar-policy/src/api.rs
```diff
@@ -844,6 +844,30 @@ impl Entities {
...
+pub fn validate_scope_variables(
+    principal: &EntityUid,
+    action: &EntityUid,
+    resource: &EntityUid,
+    schema: &Schema,
+) -> std::result::Result<(), RequestValidationError> {
+    let principal_uid = &principal.0; //EntityUID::from(principal.clone());
+    let action_uid = &action.0; //EntityUID::from(action.clone());
+    let resource_uid = &resource.0; //EntityUID::from(resource.clone());
+                                    // Call the validate_context function from coreschema.rs
```
adpaco: Were these comments supposed to be removed? Also, you can directly pass `&var.0` to `validate_scope_variables`.

cedar-policy/src/api.rs
```diff
@@ -844,6 +844,30 @@ impl Entities {
...
+    Ok(RequestSchema::validate_scope_variables(
+        &schema.0,
+        principal_uid,
+        action_uid,
+        resource_uid,
+    )?)
```
adpaco: If you have an `Ok(call?)` you can do just `call`, right?

cedar-policy/src/ffi/check_parse.rs
```diff
@@ -92,6 +92,71 @@ pub fn check_parse_schema_json_str(json: &str) -> Result<String, serde_json::Err
     serde_json::to_string(&ans)
 }
 
+/// Check whether a set of scope variables successfully parses.
+pub fn check_parse_scope_variables(call: ScopeVariablesParsingCall) -> CheckParseAnswer {
+    //move this to other top level function instead, use new parsingcall struct that you make
+    let schema_ref = match call.schema.parse() {
```
adpaco: Does this still need to be done?

cedar-policy/src/ffi/check_parse.rs
```diff
@@ -92,6 +92,71 @@ pub fn check_parse_schema_json_str(json: &str) -> Result<String, serde_json::Err
...
+    //all three should be provided, so just call validate
+    if let Err(err) =
+        crate::api::validate_scope_variables(&principal, &action, &resource, &schema_ref)
+    {
+        return CheckParseAnswer::Failure {
+            errors: vec![miette::Report::msg(err).into()],
+        };
+    }
+
+    CheckParseAnswer::Success
```
adpaco: Don't do a `return` at this point, you can just have an `if-then-else` for each case.

cedar-policy/src/ffi/check_parse.rs
```diff
@@ -249,6 +314,19 @@ pub struct EntitiesParsingCall {
     schema: Option<Schema>,
 }
 
+/// Struct containing the input data for [`check_parse_scope_variables()`]
+#[derive(Serialize, Deserialize, Debug)]
+pub struct ScopeVariablesParsingCall {
+    /// Optional Principal entity for schema-based validation
+    principal: EntityUid,
+    /// Optional Action entity for schema-based validation
+    action: EntityUid,
```
adpaco: These are not optional?

cedar-policy-core/src/validator/coreschema.rs
```diff
@@ -176,75 +176,18 @@ impl ast::RequestSchema for ValidatorSchema {
...
-                    }
-                }
-            } else {
-                return Err(request_validation_errors::UndeclaredResourceTypeError {
-                    resource_ty: resource_type.clone(),
-                }
-                .into());
-            }
-        }
+        let a_uid = request.action().uid(); //used twice so only call once
```
adpaco: Nit: I'd recommend calling it `action_uid` and getting rid of the comment.

cedar-policy-core/src/validator/coreschema.rs
mishjude: I might be interpreting the code/your comment wrong, so let me know if I am, but from what I see from the original logic, it evaluates the principal, action, and resource from the request (which I believe those three aren't optional when making a request) separately. So, when validate_scope_variables is called, I believe it evaluates each component separately as it did in the original validate_request since I formatted my logic directly from validate_request. My initial reasoning to why principal, action, and resource should be required parameters for validate_scope_variables was due to the fact that, from looking at the Request struct, principal, action, and resource were not optional parameters, so there wouldn't be an instance where there would be no value when pulling them from request. So to detail this a little easier:

Request (has non-optional principal, action, and resource parameters, so it will always have some value there) **->** pass in principal, action, resource parameters into validate_scope_variables (keeping these non-optional as well since I didn't think there would be a scenario which they aren't provided given the format of the Request struct) **->** validate_scope_variables checks each component separately, as validate_request did.

Hopefully I was able to explain my train of thought, but let me know if I have any misconceptions about the logic!

cedar-policy/src/api.rs
mishjude: Yes, they were supposed to be removed! I'm realising that I'm a little bit sloppy with editing/removing comments once I no longer need them, so I'm going to be extra vigilant about that going forward!

cedar-policy/src/ffi/check_parse.rs
```diff
@@ -92,6 +92,70 @@ pub fn check_parse_schema_json_str(json: &str) -> Result<String, serde_json::Err
...
+                errors: vec![err.into()],
+            };
+        }
+    };
+
+    //all three should be provided, so just call validate
+    if let Err(err) =
+        crate::api::validate_scope_variables(&principal, &action, &resource, &schema_ref)
+    {
+        return CheckParseAnswer::Failure {
```
john-h-kastner-aws: Don't need explicit `return` here and line 140

cedar-policy-core/src/validator/coreschema.rs
john-h-kastner-aws: I think the part you're missing is that there's also an experimental partial evaluation API where a request might not contain a principal, action, or resource. IIUC, the code was written so that it would still validate the principal if the resource wasn't present, and still apply some validation to the principal/resource even without an action.


For the public API/ffi functions I think you made the correct choice in making the variables non-optional. This internal `validate_scope_variables` function should however take `Option`s so that we don't change the behavior for partial evaluation.

cedar-policy/src/ffi/check_parse.rs
```diff
@@ -92,6 +92,70 @@ pub fn check_parse_schema_json_str(json: &str) -> Result<String, serde_json::Err
...
+}
+
+/// Check whether a set of scope variables successfully parses. Input is a JSON
+/// encoding of [`ScopeVariablesParsingCall`] and output is a JSON encoding of
+/// [`CheckParseAnswer`].
+///
+/// # Errors
+///
+/// Will return `Err` if the input JSON cannot be deserialized as a
+/// [`Entities`].
```
adpaco: We are not attempting to deserialize as `Entities`, right?

cedar-policy/src/api.rs
```diff
@@ -844,6 +844,27 @@ impl Entities {
...
+///
+/// This validation is already handled by `Request::new`, so there is no need to separately call
+/// if you are validating the whole request
+pub fn validate_scope_variables(
+    principal: &EntityUid,
+    action: &EntityUid,
+    resource: &EntityUid,
+    schema: &Schema,
+) -> std::result::Result<(), RequestValidationError> {
+    // Call the validate_context function from coreschema.rs
```
adpaco: Need to update/remove this comment.

cedar-policy/src/ffi/check_parse.rs
mishjude: Good point! There were a couple sections initially where I mistakenly though entities were equivalent/interchangeable to principal, action, and resource, so I don't think it should be written as such here. Do you think it would be better to update that description to be:

"Will return `Err` if the input JSON cannot be deserialized as principal, action, or resource."
or 
"Will return `Err` if the input JSON cannot be deserialized as a scope variable."
(Alternatively, if you have a different suggestion, let me know!)

adpaco: "scope variables" sounds good since that's what you've been using until now.

cedar-policy/src/ffi/check_parse.rs
```diff
@@ -92,6 +92,70 @@ pub fn check_parse_schema_json_str(json: &str) -> Result<String, serde_json::Err
...
+
+    //all three should be provided, so just call validate
+    if let Err(err) =
+        crate::api::validate_scope_variables(&principal, &action, &resource, &schema_ref)
+    {
+        CheckParseAnswer::Failure {
+            errors: vec![miette::Report::msg(err).into()],
+        }
+    } else {
+        CheckParseAnswer::Success
```
john-h-kastner-aws, line 140: Just noticed we don't have a test covering the success case. Should add one before merging

mishjude: on it o7

john-h-kastner-aws (approved):

PR #1669, opened 2025-06-23
Cedar Generalized Templates

alanwang67:
## Description of changes

Changes for [Cedar Generalized Templates ](https://github.com/cedar-policy/rfcs/pull/98#event-18532517610). Currently this this works for when a Schema is provided. The API end point that is exposed to end users are a function that allows for the creation of generalized slots and generalized_link which allows users to also provide values besides EntityUIDs. 

- [] A bug fix or other functionality change requiring a patch to `cedar-policy`.
- [] A change "invisible" to users (e.g., documentation, changes to "internal" crates like `cedar-policy-core`, `cedar-validator`, etc.)

- [] Does not update the CHANGELOG because my change does not significantly impact released code.

- [] Does not require updates because my change does not impact the Cedar formal model or DRT infrastructure.
- [] Requires updates, but I do not plan to make them in the near future. (Make sure that your changes are hidden behind a feature flag to mark them as experimental.)
- [] I'm not sure how my change impacts `cedar-spec`. (Post your PR anyways, and we'll discuss in the comments.)

- [] Does not require updates because my change does not impact the Cedar language specification.
- [X] Requires updates, and I have made / will make these updates myself. (Please include in your description a timeline or link to the relevant PR in [`cedar-docs`](https://github.com/cedar-policy/cedar-docs). PRs should be targeted at a `staging-X.Y` branch, not `main`.)


cedar-policy-core/src/ast/name.rs
```diff
@@ -307,6 +312,19 @@ impl SlotId {
...
+        matches!(self, Self(ValidSlotId::GeneralizedSlot(_)))
+    }
+
+    /// Returns the id if the slot is a `generalized slot`
+    pub fn extract_id_out_of_generalized_slot(&self) -> Option<Id> {
+        match self {
+            Self(ValidSlotId::GeneralizedSlot(id)) => Some(id.clone()),
+            _ => None,
+        }
+    }
```
cdisselkoen: I suggest returning `Option<&Id>` from this function, to avoid the clone.  (Caller can always clone if they need an owned `Id`.)

cedar-policy-core/src/ast/name.rs
```diff
@@ -324,20 +342,22 @@ impl std::fmt::Display for SlotId {
...
-/// Two possible variants for Slots
-#[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+/// Three possible variants for Slots
+#[derive(Debug, Clone, Eq, PartialEq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 pub(crate) enum ValidSlotId {
     #[serde(rename = "?principal")]
     Principal,
     #[serde(rename = "?resource")]
     Resource,
+    GeneralizedSlot(Id),
```
cdisselkoen: Would be good to have a doc comment here explaining what a GeneralizedSlot is, maybe link to the RFC

cdisselkoen (approved):
Looks good, just a couple minor comments.

I also wonder if we can write more tests for this functionality yet, or does that have to wait for a future PR that completes the CST-to-AST functionality?

cedar-policy-core/src/ast/name.rs
alanwang67: Gotcha yeah, I wasn't sure if using the pointer here would make that much of a difference compared to ```Id``` since the documentation mentioned that the SmolStr stored inside ```Id``` was O(1) clone.

alanwang67:
> Looks good, just a couple minor comments.
> 
> I also wonder if we can write more tests for this functionality yet, or does that have to wait for a future PR that completes the CST-to-AST functionality?

I think tests for the EST portion of the changes will need to wait until the CST-to-AST function is completed.

cedar-policy-core/src/ast/name.rs
```diff
@@ -324,20 +342,22 @@ impl std::fmt::Display for SlotId {
...
     Resource,
+    GeneralizedSlot(Id), // Slots for generalized templates, for more info see [RFC 98](https://github.com/cedar-policy/rfcs/pull/98).
 }
 
 impl std::fmt::Display for ValidSlotId {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let s = match self {
             ValidSlotId::Principal => "principal",
             ValidSlotId::Resource => "resource",
+            ValidSlotId::GeneralizedSlot(id) => &id.to_smolstr(),
```
john-h-kastner-aws: Might be able to call `as_ref()` instead

cedar-policy-core/src/ast/policy.rs
```diff
@@ -1511,6 +1511,7 @@ pub enum EntityReference {
     EUID(Arc<EntityUID>),
     /// Template Slot
     Slot(
+        Option<Id>, // If a slot stores None then it is a principal/resource, otherwise it is a generalized slot
```
john-h-kastner-aws: Could we change this to be a `SlotId` instead? That might make this comment and the logic in `into_expr` easier to follow

alanwang67: I was thinking of making it a ```SlotId``` but it seems like the convention followed so far is that the type of slot ```principal``` or ```resource``` is determined by whether or not the entity reference is nested within a ```PrincipalConstraint``` or a ```ResourceConstraint```. So I think a potential downside by having the user supply the exact slot type is that you can have error states where you have a ```resource slot``` nested within a ```PrincipalConstraint```. Would changing the Option type to have more descriptive variants such as 
`pub enum SlotKind {
        PrincipalOrResource,
        Generalized(Id)
    }` make it easier to follow or would it still be best to just change it to using a ```SlotId``` instead?

john-h-kastner-aws: > that you can have error states where you have a resource slot nested within a PrincipalConstraint

This doesn't feel like a huge issue if it substantially simplifies other code. 

> Would changing the Option type to have more descriptive variants 

IMO this could be a bit nicer, but it wouldn't help all that much.

I think my main concern isn't actually with this enum variant. It's more with the `into_expr` function taking a `SlotId` which it just ignores when the stored `Id` is `Some`. 

Not a huge issue, so this is fine as is if you've thought about it and don't think there's a great alternative

PR #1673, opened 2025-06-26
add missing Protobuf conversions for `api::ValidationMode`

cdisselkoen:
Also reserves `2` to represent the `Partial` variant in Protobuf


cdisselkoen:
cargo-semver-checks flags a breaking change, but it's for a `proto` type, which is marked experimental and thus we allow breaking changes.

john-h-kastner-aws (approved):

Issue #1684, opened 2025-06-30
Removing default-on cdylib target from cedar-policy crate


Mark-Simulacrum:
### Category

Other

### Describe the feature you'd like to request

The cedar-policy crate currently declares a cdylib crate type (for WASM, I guess):

https://github.com/cedar-policy/cedar/blob/7731f7f41ffb7410023d6f9535dfb65364a6f31f/cedar-policy/Cargo.toml#L73

Today this means that consumers *not* on wasm are unable to benefit from Cargo's pipelining ([rough docs](https://internals.rust-lang.org/t/evaluating-pipelined-rustc-compilation/10199)) and given that the cedar crates can be quite slow to compile -- in one build I'm seeing 36s for cedar-policy-core, 28s for cedar-policy-validator, and 16s for cedar-policy -- it'd be nice to avoid blocking the build on LLVM's codegen completing for cedar-policy.

It's not clear to me where the wasm consumers of the cedar-policy crate are or whether it's possible to have a separate crate for that use case. I haven't tried locally patching Cedar to drop this to evaluate the impact -- and it would depend on how much codegen is being done in the final crate -- but it's almost guaranteed to be a win.

### Describe alternatives you've considered

None at this time.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [x] ⚠️ This feature might incur a breaking change

chaluli:
As far as I am aware there is no way to specify in the `Cargo.toml` or `build.rs` to optionally compile as a `cdylib` when the `wasm` feature is provided.

Though, it is possible to compile as an rlib by default and additionally run `cargo rustc --crate-type cdylib` to compile as a `cdylib`.

I did some testing (all times on my personal VM) to determine how long it takes to compile `cedar-policy` crate.

Compiling with both `"rlib"` and `"cdylib"` by default took ~47 seconds.
Compiling with just `"rlib"` takes ~41 seconds.
Compiling with `cargo rustc --crate-type cdylib` takes ~43 seconds.
Compiling with `cargo rustc --crate-type cdylib` after building just `"rlib"` takes ~4 seconds (~45 seconds total).

All times were to compile `cedar-policy` crate after running `cargo clean` with default features.

PR #1689, opened 2025-07-02
Only compile cedar-policy as rlib by default.

chaluli:
## Description of changes

Remove flag in `cedar-policy/Cargo.toml` to compile as a `cdylib` crate (i.e., compile as a dynamic library .so file).

It seems that `cdylib` is only required for `cedar-wasm` which has it's own `Cargo.toml` which specifies to compile as `cdylib`.

Addresses request in issue #1684 which points out that we do not need to compile `cdylib`. Removing this compilation target by default improves compiler performance by ~14% in local testing on my development machine.


cdisselkoen (approved):
LGTM given that CI passes

adpaco-aws (approved):

PR #1690, opened 2025-07-02
[Draft] Entity manifest as a DAG to support Cedar tags

oflatt:
## Update: Don't review

Emina and I have come up with a new plan for slicing, so this PR might not be needed. The downside of this PR is that the entity DAG can be O(n^2) in the size of the program when you have repeated `getTag` calls.

## Description of changes

This PR rewrites the EntityManifest to be a DAG instead of a tree, flipping all the pointers to point to data dependents instead of dependendees. This allows us to extend them to work with (dynamically computed) entity tags while still deduplicating data paths.

The main trick is to use a hash cons to deduplicate paths, and storing data dependencies lets us do that. The `EntityLoader` api is fairly unchanged, and we still compute access tries for convenience there.

Overall I find this a much better implementation than the old tree-based one, eliminating ugly tags along the trie in favor of new kinds of nodes in the dag.

The implementation of equivalence between these new dags is particularly tricky, as it has to map from one dag to another.

For testing and convenience, there's a new human-readable entity manifest stored as a list of cedar expressions.

One awkward thing about this code is that it uses an array encoding of a graph, so there are
lots of accesses to `unwrap`. Watch out for these, I added PANIC_SAFETY comments to
all of them.


cedar-policy-core/src/validator/entity_manifest.rs
```diff
@@ -30,49 +31,216 @@ use serde_with::serde_as;
...
+    pub fn access_paths_mut(&mut self) -> &mut AccessPaths {
+        &mut self.access_paths
+    }
+
+    /// Add all paths from `other` to `self`
+    fn union_with(&mut self, other: &Self) -> PathMapping {
+        let mut path_mapping = PathMapping::new();
+        // First, add all paths from the other manifest to this manifest
+        // and build a mapping from paths in the other manifest to paths in this manifest
+        for i in 0..other.dag.manifest_store.len() {
```
john-h-kastner-aws: Can this be `for (i, variant) in other.dag.manifest_store.iter().enumerate()`? Then maybe this could be rewritten to `collect` into the `PathMapping` instead of building it up with `insert`s

cedar-policy-core/src/validator/entity_manifest.rs
```diff
@@ -30,49 +31,216 @@ use serde_with::serde_as;
...
+            #[allow(clippy::panic)]
+            let variant = &other.dag.manifest_store.get(i).unwrap();
+            let mapped_variant = self.map_variant(variant, &mut path_mapping);
+            let new_path = self.dag.add_path(mapped_variant);
+            path_mapping.path_map.insert(i, new_path.id);
+        }
+
+        // Map each path from the other manifest to this manifest
+        for path in &other.access_paths.paths {
+            if let Some(mapped_path) = path_mapping.map_path(path) {
```
john-h-kastner-aws: Could use `filter_map` here

cedar-policy-core/src/validator/entity_manifest.rs
```diff
@@ -30,49 +31,216 @@ use serde_with::serde_as;
...
+    /// Variables are included in the request.
+    /// We prune these from the set of access paths required.
+    fn prune_leafs(&mut self) {
+        let mut to_remove = Vec::new();
+        for path in &self.access_paths.paths {
+            if path.is_leaf(&self.dag) {
+                to_remove.push(path.clone());
+            }
+        }
+        for path in to_remove {
```
john-h-kastner-aws: Could this loop over `self.access_paths.paths.iter().filter(...)` to avoid building the `Vec`?

cedar-policy-core/src/validator/entity_manifest.rs
```diff
@@ -30,49 +31,216 @@ use serde_with::serde_as;
...
+    paths: HashSet<AccessPath>,
+}
+
+/// Represents a path of data involving a sequence of
+/// attribute or tag accesses, ending in a cedar variable or literal.
+/// Internally represented as a single integer into a backing store
+/// (a directed acyclic graph).
+/// Hashing an [`AccessPath`] is extremely cheap, so resulting data can be cached.
+#[doc = include_str!("../../experimental_warning.md")]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
```
john-h-kastner-aws: Can make this `Copy` since it's just a `usize`. This should let you omit some of the explicit `clones` later on

cedar-policy-core/src/validator/entity_manifest.rs
```diff
@@ -30,49 +31,216 @@ use serde_with::serde_as;
...
+/// or a record in another entity.
+///
+/// After construction, the dag is annotated with the types of all of the access paths.
+#[serde_as]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct AccessDag {
+    /// A map from [`AccessPathInternal`] to the [`AccessPath`], which
+    /// indexes the `manifest_store`.
+    /// This allows us to de-duplicate equivalent access paths using the "hash cons"
```
john-h-kastner-aws: Is there a way to do this that doesn't mean storing two copies of every access path variant?

cedar-policy-core/src/validator/entity_manifest.rs
```diff
@@ -268,10 +370,220 @@ pub enum EntityManifestFromJsonError {
...
+impl Eq for EntityManifest {}
+
+impl PathsForRequestType {
+    /// Map a variant from the source manifest to the target manifest
+    ///
+    /// This method recursively maps a variant and its children from the source manifest
+    /// to the target manifest, creating new paths in the target manifest as needed.
+    fn map_variant(
+        &mut self,
+        variant: &AccessPathVariant,
```
john-h-kastner-aws: Could probably be owned, avoid clones later on

cedar-policy-core/src/validator/entity_manifest.rs
```diff
@@ -268,10 +370,220 @@ pub enum EntityManifestFromJsonError {
...
+    ) -> AccessPath {
+        // Check if the path is already mapped
+        if let Some(mapped_path) = path_mapping.map_path(path) {
+            return mapped_path;
+        }
+
+        // If not, get the variant for this path from the source manifest
+        // and recursively map it to create a new path in the target manifest
+        if let Ok(variant) = path.get_variant(&self.dag).clone() {
+            let variant = variant.clone();
```
john-h-kastner-aws: Cloning `variant` twice is suspicious

cedar-policy-core/src/validator/entity_manifest.rs
```diff
@@ -97,181 +273,212 @@ impl Display for EntityRoot {
...
+            let mapped_variant = self.map_variant(&variant, path_mapping);
+            let new_path = self.dag.add_path(mapped_variant);
+            path_mapping.path_map.insert(path.id, new_path.id);
+            new_path
+        } else {
+            // If we can't find the variant, create a safe default
+            // This should rarely happen in practice
+            let default_variant = AccessPathVariant::Var(Var::Principal);
```
john-h-kastner-aws: This needs some better justification. Why is returning `principal` safe?  If it's rare, when does it happen?

cedar-policy-core/src/validator/entity_manifest.rs
```diff
@@ -97,181 +273,212 @@ impl Display for EntityRoot {
...
+    /// A manifest is a subset of another if all access paths from one are reachable from the other.
+    fn is_subset_of(&self, other: &Self) -> bool {
+        // For each request type, check that the other is a subset
+        for (request_type, my_paths) in &self.per_action {
+            let Some(other_paths) = other.per_action.get(request_type) else {
+                return false;
+            };
+            let mut other_clone = other_paths.clone();
+
+            let mapping = other_clone.union_with(my_paths);
```
john-h-kastner-aws: What's going on here? You create a mutable copy of `other_paths`m presumably update it in `union_with`, and then never use it again?

cedar-policy-core/src/validator/entity_manifest.rs
```diff
@@ -97,181 +273,212 @@ impl Display for EntityRoot {
...
+
+            // now find all the reachable paths in other
+            let mut reachable = HashSet::new();
+            for path in other_paths.access_paths.paths() {
```
john-h-kastner-aws: Could probably `.flat_map(...).collect()` directly into the set

cedar-policy-core/src/validator/entity_manifest.rs
```diff
@@ -97,181 +273,212 @@ impl Display for EntityRoot {
...
+
+        return true;
+    }
+
+    /// Union this entity manifest with another entity manifest
+    ///
+    /// This adds all paths from the other manifest to this manifest,
+    /// and updates the per_action map to include the paths from the other manifest.
+    ///
+    /// Returns a mapping from paths in the other manifest to paths in this manifest.
```
john-h-kastner-aws: actually returns unit

cedar-policy-core/src/validator/entity_manifest.rs
```diff
@@ -301,143 +508,97 @@ impl EntityManifest {
...
-        let mut primary_map = HashMap::new();
-
-        // special case: if the path is completely empty,
-        // no need to insert anything
-        if current != AccessTrie::new() {
-            primary_map.insert(self.root.clone(), current);
+    /// Get the immediate children of this path
+    pub fn children(&self, store: &AccessDag) -> Vec<AccessPath> {
+        // Get the variant for this path
+        if let Some(variant) = store.manifest_hash_cons.iter().find_map(|(variant, path)| {
```
john-h-kastner-aws: I'm confused by this lookup. Is it not the same as `store.manifest_store.get(self.id)`?

Also, the current lookup would probably be clearer as just a `find` binding like `let Some((variant, _)) =`

cedar-policy-core/src/validator/entity_manifest.rs
```diff
@@ -454,118 +615,157 @@ pub fn compute_entity_manifest(
...
     for policy in policies.policies() {
         // typecheck the policy and get all the request environments
         let request_envs = typechecker.typecheck_by_request_env(policy.template());
         for (request_env, policy_check) in request_envs {
-            let new_primary_slice = match policy_check {
+            let request_type = request_env
+                .to_request_type()
+                .ok_or(PartialRequestError {})?;
+
+            let mut per_request = match manifest.per_action.entry(request_type.clone()) {
```
john-h-kastner-aws: is this just doing `.remove().unwrap_or()`?

cedar-policy-core/src/validator/entity_manifest/type_annotations.rs
```diff
@@ -16,162 +16,184 @@
...
-                        .iter()
-                        .map(|(key, val)| Ok((key.clone(), val.to_typed(key, schema)?)))
-                        .collect::<Result<
-                            HashMap<RequestType, RootAccessTrie>,
-                            MismatchedEntityManifestError,
-                        >>()?,
-            },
-        )
+        // Create a new entity manifest with the same structure
+        let mut typed_manifest = self.clone();
```
john-h-kastner-aws: could probably just take owned self

cedar-policy-core/src/validator/entity_manifest/type_annotations.rs
```diff
@@ -16,162 +16,184 @@
...
-                            }
-                        },
-                    ))
-                })
-                .collect::<Result<HashMap<EntityRoot, AccessTrie>, MismatchedEntityManifestError>>(
-                )?,
-        })
+    ) -> Result<PathsForRequestType, MismatchedEntityManifestError> {
+        // Create a new PathsForRequestType with the same structure
+        let mut typed_paths = self.clone();
```
john-h-kastner-aws: could possibly take owned self

cedar-policy-core/src/validator/entity_manifest/type_annotations.rs
```diff
@@ -16,162 +16,184 @@
...
-                    if let Some(ty) = attributes.get_attr(field) {
-                        new_children.insert(
-                            field.clone(),
-                            Box::new(child.to_typed(request_type, &ty.attr_type, schema)?),
-                        );
+                // Store the type in the types vector
+                types[path.id] = Some(ty);
+            }
+            AccessPathVariant::Literal(lit) => {
+                let ty = &Type::euid_literal(lit, schema).ok_or_else(|| {
```
john-h-kastner-aws: Takes reference and then immediately clones it

cedar-policy-core/src/validator/entity_manifest/type_annotations.rs
```diff
@@ -16,162 +16,184 @@
...
+                    }
+                    _ => {
+                        // Non-entity and non-record types don't have attributes
+                        // This should be caught by the typechecker, but we'll handle it gracefully
                     }
                 }
-                new_children
+            }
+            AccessPathVariant::Tag { of: _, tag: _ } => {
+                // Tags are not fully supported yet, but we'll handle them as strings
```
john-h-kastner-aws: This is surprising since tags can have arbitrary types

cedar-policy-core/src/validator/entity_manifest/entity_manifest_tests.rs
john-h-kastner-aws, line 1: If this is a simple copy-paste of existing tests into a new file, I'd like to have that change first as a separate PR.

Other small refactoring split out into new PRs is also appreciated if it's easy

cedar-policy-core/src/validator/entity_manifest/analysis.rs
```diff
@@ -191,103 +316,279 @@ impl WrappedAccessPaths {
...
+/// The [`WrappedAccessPaths`] contains the result's access paths
+/// and any access paths encountered during the analysis.
+pub(crate) fn analyze_expr_access_paths(
+    expr: &Expr<Option<Type>>,
+    store: &mut AccessDag,
+) -> Result<Rc<WrappedAccessPaths>, EntityManifestError> {
+    Ok(match expr.expr_kind() {
+        ExprKind::Slot(slot_id) => {
+            if slot_id.is_principal() {
+                WrappedAccessPaths::from_root(EntityRoot::Var(Var::Principal), store)
```
john-h-kastner-aws: don't think this is correct. `?principal` slot isn't always equal to principal.  though, this shouldn't matter with the current limited templates

oflatt:
Thanks for feedback @john-h-kastner-aws! I'll address all of it. I haven't found an easy way to break this PR up yet.

I didn't intend this PR to be ready for review, I just put it up so people can see what I'm working on.

cedar-policy-core/src/validator/entity_manifest.rs
oflatt: If we want de-duplication we need a map from path variants to their id (AccessPath) and vice versa. I think this means storing them all twice unfortunately.

cedar-policy-core/src/validator/entity_manifest.rs
oflatt: I think we clone less stuff in some branches as a reference

cedar-policy-core/src/validator/entity_manifest.rs
oflatt: Actually, this case can't happen. Added a panic

cedar-policy-core/src/validator/entity_manifest.rs
oflatt: Yeah this is tricky, I'll add comments.

We just needed the mapping from the self dag to the other dag so we can detect if all the access paths are reachable *in other*.

cedar-policy-core/src/validator/entity_manifest.rs
oflatt: Oops good point, also I've changed it to not default to empty children and instead panic
Old code from before we could just look them all up

cedar-policy-core/src/validator/entity_manifest/type_annotations.rs
oflatt: Oops, old comment (we won't handle them as strings)
I'm working on handling tags

cedar-policy-core/src/validator/entity_manifest.rs
oflatt: yep

cedar-policy-core/src/validator/entity_manifest.rs
oflatt: actually I added a panic- the map_path call should return Some

cedar-policy-core/src/validator/entity_manifest.rs
oflatt: yep!

cedar-policy-core/src/validator/entity_manifest.rs
oflatt: yep!

cedar-policy-core/src/validator/entity_manifest/analysis.rs
oflatt: Hmm I can't remember this super well- it's how the current code works I believe

cedar-policy-core/src/validator/entity_manifest/entity_manifest_tests.rs
oflatt: I did end up editing the tests to the new human format, and changing the slicing tests.

Cargo.lock
```diff
@@ -1374,7 +1374,7 @@ checksum = "e04d7f318608d35d4b61ddd75cbdaee86b023ebe2bd5a66ee0915f0bf93095a9"
 dependencies = [
  "hermit-abi",
  "libc",
- "windows-sys",
+ "windows-sys 0.52.0",
```
oflatt, line 1377: not sure why my local build insists on adding this, happy to remove

oflatt:
Working on the fix for DRT!

oflatt:
PR to fix drt testing, but not sure how to get it to run in this CI
https://github.com/cedar-policy/cedar-spec/pull/670

oflatt:
Don't review this! We have a new plan and might close this PR.

oflatt:
Closing in favor of batched evaluation
https://github.com/cedar-policy/cedar-spec/pull/681

PR #1696, opened 2025-07-09
Adds more conversion functions needed for updating DRT to public types

chaluli:
## Description of changes

Adds more conversion functions needed for updating DRT to public types

ast::json_schema::SchemaFragment -> api::SchemaFragment
api::EvalResult -> api::Expression

These changes don't impact DRT. But I did make these changes due to (not yet pushed) changes I made to DRT.

### DIFF
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -379,13 +379,6 @@ impl From<cedar_policy_core::entities::Entities> for Entities {
     }
 }
 
-#[doc(hidden)]
-impl From<Entities> for cedar_policy_core::entities::Entities {
-    fn from(entities: Entities) -> Self {
-        entities.0
-    }
-}
-
 use entities_errors::EntitiesError;
 
 impl Entities {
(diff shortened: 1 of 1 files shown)


cedar-policy/src/api.rs
```diff
@@ -1504,6 +1511,25 @@ impl
     }
 }
 
+#[doc(hidden)] // because this converts to a private/internal type
```
cdisselkoen: I think this converts _from_ a private/internal type?

cedar-policy/src/api.rs
```diff
@@ -379,6 +379,13 @@ impl From<cedar_policy_core::entities::Entities> for Entities {
     }
 }
 
+#[doc(hidden)]
+impl From<Entities> for cedar_policy_core::entities::Entities {
+    fn from(entities: Entities) -> Self {
+        entities.0
+    }
+}
```
cdisselkoen: I think this might be the first instance of a From<API type> for <core type>; we do already have an `AsRef` for this conversion.  But if you do need owned `cedar_policy_core::entities::Entities`, this is more efficient than `.as_ref().clone()`, so seems reasonable.  Should we have similar instances for all other cases where we currently have `AsRef` instances?

chaluli: I wrote this early on and refactored out the part that used this. I've removed it a following commit.

PR #1703, opened 2025-07-14
Initial push of cedar symbolic compiler.

chaluli:
Contains the Rust version of the Cedar Symbolic compiler.

More PRs to come:
Add CI to build and run `cedar-symcc`.
Update `cedar-policy-cli` to have an analyze command matching `cedar-lean-cli`.
Update `cedar-symmcc`'s README to with setup instructions and usage examples.
Improve comments on public types/functions with what it does in isolation with requiring knowledge of Lean `symcc`.
Add a `cedar-symcc` specific changelog.

### DIFF
--- a/cedar-policy-symcc/Cargo.toml
+++ b/cedar-policy-symcc/Cargo.toml
@@ -3,7 +3,7 @@ name = "cedar-policy-symcc"
 edition.workspace = true
 rust-version.workspace = true
 
-version.workspace = true
+version = "0.1.0"
 license.workspace = true
 categories.workspace = true
 description = "Cedar Symbolic Compiler: translates queries about Cedar Policies to SMT"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -499,7 +499,7 @@ dependencies = [
 
 [[package]]
 name = "cedar-policy-symcc"
-version = "4.5.0"
+version = "0.1.0"
 dependencies = [
  "anyhow",
  "async-recursion",


.github/workflows/build_and_test.yml
```diff
@@ -55,6 +55,24 @@ jobs:
       - run: ./panic_safety.sh
       - run: cargo doc --all-features --no-deps
 
+      # Required to run cedar-symcc tests
+      - name: Install cvc5
+        shell: bash
+        run: |
+              ARCH=$(uname -m)
+              if [ "$ARCH" = "x86_64" ]; then
```
shaobo-he-aws, line 63: Is it possible to select the arch of the container running CI? I feel like we only need to support x86_64 for CI.

cedar-policy-symcc/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "cedar-policy-symcc"
+edition.workspace = true
+rust-version.workspace = true
+
+version.workspace = true
```
shaobo-he-aws: I recall that we don't want symcc to have the same version with other packages because its APIs are subject to more frequent major changes.

chaluli: Good catch. I was a bit overzealous with copying from other Cargos.

.github/workflows/build_and_test.yml
chaluli: It is possible. But I don't think it's necessary. Why not support both x86_64 and arm?

shaobo-he-aws: It's just more maintenance burden. I believe that a successful run on x86_64 should be suffice to indicate the correctness. arm is generally not stable so I expect we will run into dependency issues a lot. That being said, I don't have strong opinions.

PR #1712, opened 2025-07-21
Chaluli/cedar symcc datetime

chaluli:
## Description of changes

Adds (most of) the code needed for encoding Datetime and Duration Extension Types and Operators. This code is missing code for 3 `BitVec` operators which are anticipated to be provided by @lianah 's PR for the encoding of the IP extension.

### DIFF
--- a/cedar-policy-symcc/src/symcc/compiler.rs
+++ b/cedar-policy-symcc/src/symcc/compiler.rs
@@ -203,11 +203,19 @@ pub fn compile_app2(op2: &BinaryOp, t1: Term, t2: Term, es: &SymEntities) -> Res
             }
         }
         (Less, Bitvec { n: 64 }, Bitvec { n: 64 }) => Ok(some_of(factory::bvslt(t1, t2))),
+        (Less, Ext { xty: DateTime }, Ext { xty: DateTime }) => Ok(some_of(factory::bvslt(
+            factory::ext_datetime_val(t1),
+            factory::ext_datetime_val(t2),
+        ))),
         (Less, Ext { xty: Duration }, Ext { xty: Duration }) => Ok(some_of(factory::bvslt(
             factory::ext_duration_val(t1),
             factory::ext_duration_val(t2),
         ))),
         (LessEq, Bitvec { n: 64 }, Bitvec { n: 64 }) => Ok(some_of(factory::bvsle(t1, t2))),
+        (LessEq, Ext { xty: DateTime }, Ext { xty: DateTime }) => Ok(some_of(factory::bvsle(
+            factory::ext_datetime_val(t1),
+            factory::ext_datetime_val(t2),
+        ))),
         (LessEq, Ext { xty: Duration }, Ext { xty: Duration }) => Ok(some_of(factory::bvsle(
             factory::ext_duration_val(t1),
             factory::ext_duration_val(t2),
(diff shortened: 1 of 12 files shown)


shaobo-he-aws (approved):
LGTM

cedar-policy-symcc/src/symcc/factory.rs
```diff
@@ -609,9 +626,43 @@ pub fn ext_ipaddr_prefix_v6(t: Term) -> Term {
...
+            ret_ty: TermType::Ext {
+                xty: ExtType::DateTime,
+            },
+        },
+    }
+}
+
 pub fn ext_duration_val(t: Term) -> Term {
     match t {
-        Term::Prim(TermPrim::Ext(Ext::Duration { d })) => Ext::Duration { d }.into(),
```
chaluli, line 614: Previous implementation was incorrect. It acted as an identity function when T was a constant Duration. When instead it should have been transforming the constant duration to the "equivalent" constant `BitVec 64`.

cedar-policy-symcc/src/symcc/compiler.rs
```diff
@@ -734,4 +807,469 @@ mod tests {
...
+        //            r#"datetime("1969-12-31T10:30:00.001Z").toDate()"#,
+        //            -86400000,
+        //        );
+        //        test_valid_datetime_simpl_expr(
+        //            r#"
+        //            datetime("1969-12-31T10:30:00.001Z").toDate()
+        //            .offset(duration("1d1h29m59s999ms"))
+        //            "#,
+        //            86400000,
+        //        );
```
adpaco, line 1196: Do we need to enable these tests?

cedar-policy-symcc/src/symcc/compiler.rs
```diff
@@ -734,4 +807,469 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_duration_simpl_expr() {
+        // Tests with toTime will not work until BitVec::srem is implemented
+        // test_valid_duration_simpl_expr(r#"datetime("1970-01-01").toTime()"#, 0);
+        // test_valid_duration_simpl_expr(r#"datetime("1969-12-31T00:01:00Z").toTime()"#, 60000);
```
adpaco, line 1203: Same goes for these, `srem` is implemented right?

cedar-policy-symcc/src/symcc/factory.rs
```diff
@@ -609,9 +626,43 @@ pub fn ext_ipaddr_prefix_v6(t: Term) -> Term {
...
+        Term::Prim(TermPrim::Ext(Ext::Duration { d })) => Term::Prim(TermPrim::Bitvec(BitVec {
+            width: 64,
+            v: i128::from(d.to_milliseconds()),
```
lianah, line 664: I was looking at the corresponding Lean code and I didn't see a corresponding `d.to_milliseconds()` call: https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/Cedar/SymCC/Factory.lean#L321C1-L323C77 Am I missing some implicit conversion in the Lean code?

chaluli: `Duration::to_milliseconds` Just returns the underlying (private) i64 val of the Duration. It's not necessary in the lean code because lean doesn't have private members.

cedar-policy-symcc/src/symcc/compiler.rs
chaluli: These tests will be used once the implementation of `srem`, `umod`, `smod` are implemented. They are not currently implemented and will be implemented in a future PR that provides a refactor of the BitVec library. I didn't think it was prudent to redo work that should be done shortly.

cedar-policy-symcc/src/symcc/compiler.rs
chaluli: No, `srem` is not implemented. I didn't not implement them myself because a PR by @lianah should be coming soon that does a fairly large refactor of the BitVec library and I did not want to perform redundant work.

cedar-policy-symcc/src/symcc/factory.rs
cdisselkoen: Would it be useful to create a function in the Lean with the appropriate name, and call it, even if all it does is `id`?  just for clarity of correspondence between Rust and Lean

chaluli: Note. Term::from(val: impl<into Literal>) when applied to `Ext::Duration { d }` elements returns `Term::Prim(TermPrim::Ext(Ext::Duration { d }))`. But in the Lean which uses implicit coercion on `d.val` will cast into `Term.Prim (TermPrim.BitVec 64 d.val)`

PR #1720, opened 2025-07-28
TPE: Do not simplify false && <residual>

luxas:
## Description of changes

Do not simplify `false && <residual>` in TPE, neither way. Before `false && <residual> => false` but `<residual> && false` was left unchanged (desired behavior).

Leftover from https://github.com/cedar-policy/cedar/pull/1575

@shaobo-he-aws 

### DIFF
--- a/cedar-policy-core/src/tpe/evaluator.rs
+++ b/cedar-policy-core/src/tpe/evaluator.rs
@@ -107,10 +107,27 @@ impl Evaluator<'_> {
                                 ..
                             },
                         ..
-                    } => Residual::Concrete {
-                        value: false.into(),
-                        ty,
+                    } => match self.interpret(&right) {
+                        // false && <bool> => false
+                        Residual::Concrete {
+                            value:
+                                Value {
+                                    value: ValueKind::Lit(ast::Literal::Bool(_)),
+                                    ..
+                                },
+                            ..
+                        } => left,
+                        // cannot simplify false && <residual>
+                        right => Residual::Partial {
+                            kind: ResidualKind::And {
+                                left: Arc::new(left),
+                                right: Arc::new(right),
+                            },
+                            ty,
+                        },
                     },
+                    // true && <bool> => <bool>
+                    // true && <residual> => <residual>
                     Residual::Concrete {
                         value:
                             Value {
@@ -120,7 +137,8 @@ impl Evaluator<'_> {
                         ..
                     } => self.interpret(right),
                     Residual::Concrete { ty, .. } => Residual::Error(ty.clone()),
-                    Residual::Partial { .. } => match &self.interpret(&right) {
+                    Residual::Partial { .. } => match self.interpret(&right) {
+                        // <residual> && true => <residual>
                         Residual::Concrete {
                             value:
                                 Value {
@@ -129,10 +147,12 @@ impl Evaluator<'_> {
                                 },
                             ..
                         } => left,
+                        // cannot simplify <residual> && false
+                        // cannot simplify <residual> && <residual>
                         right => Residual::Partial {
                             kind: ResidualKind::And {
                                 left: Arc::new(left),
-                                right: Arc::new(right.clone()),
+                                right: Arc::new(right),
                             },
                             ty,
                         },


shaobo-he-aws:
Hey @luxas, can I ask what the motivation is? `false && <residual>` should be able to soundly reduced to `false` due to Cedar's short-circuiting behaviors.

luxas:
If that is by design, then I'm ok with it (and I can close), but I had just expected this behavior to be commutative, as TPE traverses the whole AST. I e.g. made some unit tests with a deny when unless clause, just because "deny when unless" subjectively sounded more logical than "deny unless when". If the non-commutative, short-circuiting behavior is desired, all good. Sending the PR to discuss was as fast as sending you a Slack message, so I figured I'd capture the discussion in public if other people stumble onto the same thing 😁

However, because the Kubernetes integration I'm building does "TPE under the hood" without the user knowing what information is unavailable at what request stages, the behavior must be commutative. It is also required to fold `<residual> && false => false`, losing a little bit of precision of error behavior wrt integer overflow (but I guess that's about it, as the policyset is already validated by this point), in order to keep compability with existing k8s rules.

I will anyways most likely work around this on my side by walking the AST post-TPE to add that folding behavior (if such opt-in post-processing is common enough I can donate such a function upstream here). That will yield the commutative behavior, so I do not need this PR per se, just wanted to discuss.

shaobo-he-aws:
Yes. I also observed exactly the same behavior. I think that it's unavoidable given Cedar's short-circuiting behavior. That is, `false && <expr>` always evaluates to false but `<expr> && false` evaluates to an error when `<expr>` does. I also agrees with you that this behavior is not ideal, particularly when `<expr>` does not lead to errors that validation is not able to rule out. For my particular use case, it causes partial authorization decisions to depend on the order of clauses. Given the following policy and a partial request `principal = { a: true }`, the policy partially evaluates to false and hence the partial authorizer can make an `Allow` decision. If we flip the order of the two clauses, the decision becomes `None` because we can't reduce `resource.b && false`.

```cedar
forbid(principal, action, resource)
unless { principal.a }
when { resource.b }
```

I think pattern matching on the AST is work-around. That being said, I wonder if a general solution is to leverage [the symbolic compiler](https://www.cedarpolicy.com/blog/introducing-cedar-analysis) to check if there is a satisfying input. This approach works on the semantics level and should be fast enough for most policies.

@emina probably has better ideas.

emina:
> Yes. I also observed exactly the same behavior. I think that it's unavoidable given Cedar's short-circuiting behavior. That is, `false && <expr>` always evaluates to false but `<expr> && false` evaluates to an error when `<expr>` does. I also agrees with you that this behavior is not ideal, particularly when `<expr>` does not lead to errors that validation is not able to rule out. For my particular use case, it causes partial authorization decisions to depend on the order of clauses. Given the following policy and a partial request `principal = { a: true }`, the policy partially evaluates to false and hence the partial authorizer can make an `Allow` decision. If we flip the order of the two clauses, the decision becomes `None` because we can't reduce `resource.b && false`.
> 
> ```
> forbid(principal, action, resource)
> unless { principal.a }
> when { resource.b }
> ```
> 
> I think pattern matching on the AST is work-around. That being said, I wonder if a general solution is to leverage [the symbolic compiler](https://www.cedarpolicy.com/blog/introducing-cedar-analysis) to check if there is a satisfying input. This approach works on the semantics level and should be fast enough for most policies.
> 
> @emina probably has better ideas.

Calling the symbolic compiler wouldn't quite help here, because the symbolic compiler assumes that the input entities and context are _closed_, meaning there are no dangling / undefined references.  As a result, the symbolic compiler will not detect errors that are due to missing entities (they are ruled out by assumption).

If we allow inputs with dangling references, then it is not safe to rewrite `<expr> && false` to `false` because the evaluation of `expr` can always error due to missing entities.  For example:

```
permit(principal, action, resource)
unless { principal.a }
when { Datum::"Foo".b }
```

Here, we can't safely swap the evaluation order of these two clauses without assuming (like the symbolic compiler does) that the provided entities will always include a (well-typed) binding for `Datum::"Foo"`.

It is reasonable to assume, like the symbolic compiler does, that the _entire_ entity store should be closed (that is, free of dangling references) and that all policies defined on that store are also closed with respect to the store (that is, don't reference undefined entities).  In fact, applications of Cedar should ensure this for the same reason that we don't want to have dangling references in programs :)  So, the analysis guarantees are with respect to the contents of an entire, well-formed entity store.

However, it is not necessarily desirable to assume that the input to TPE is a full entity store!  In fact, you usually want it to be a slice of the entity store, which means that we do want to allow inputs with dangling references.  This is precisely what entity slicing does.  It produces a slice of a well-formed entity store, where the slice usually has dangling references, because the slicing algorithm has proven that the attributes / ancestors of the dangling references won't be accessed.

So, safely rewriting `<expr> && false` to just `false` is quite subtle.  We would need additional assumptions on the final slice against which the residual is evaluated that says something along these lines:  evaluating the residual against the slice produces the same result as evaluating the residual against the full well-formed entity store.  This assumption is necessary no matter what.

Now, on top of this assumption, we could use the symbolic compiler to show that `<expr>` is also free of other errors that the validation doesn't guard against (arithmetic and extension errors).  _Or_ we could avoid calling the symbolic compiler and write a static analysis that says expressions with no arithmetic or error-throwing extension operators are safe.  This analysis would be conservative (e.g., it would reject some `<expr>`s that contain safe arithmetic).  The burden then would be to continue to maintain this extra analysis and proofs as the language evolves (and more errors are added that aren't caught by validation).

emina:
This is all to say that the semantics of the short-circuiting `&&` and `||` operators are not commutative (in Cedar or elsewhere).  That's why their partial semantics isn't commutative either.  So, in that sense, this behavior is by design:  the partial semantics mirrors the concrete semantics.

shaobo-he-aws:
@emina thank you for the detailed clarification. @luxas I suggest maintaining the current behavior, allowing users to decide for themselves when to simplify residuals like `<expr> && false` in their code, e.g., via AST rewriting.

luxas:
Thanks a lot for the detailed discussion and rationale 👏 (again, discussion was the aim of sending this, not the PR itself).
Indeed, thinking deeper about it, it definitely is reasonable (and the only right thing to do) to keep both short-circuiting behaviors the same, between full and partial evaluation. And dealing with errors (especially in forbid policies) is certainly hard make both intuitive and correct.

The use-case I'm experimenting with for this is in detail covered in [my thesis](https://github.com/luxas/research/blob/main/msc_thesis.pdf), but on a high level, Kubernetes has two request stages (of importance within this context): authorization (metadata-only) and admission (metadata+request body). At the authorization stage, in the absence of the request body Cedar entity, it is desired to reject any such request, which cannot become authorized later (when the request body is deserialized). This in order to prevent DoS, where an attacker could overload the server with large bogus write (or read) requests that never pass the second stage, but use resources in between the first and the second one.

Currently, it is not possible to write policies conditional to the request body in the authorization stage. However, it's a useful feature that is being explored currently. There are a couple of ways to achieve this, but what I have currently is to allow policy authors to write one unified policy across both stages (without possibly even realizing it). In other words, one could do 
```
permit(principal, action == create, resource is core::Secret)
unless { resource.request.data.isSecure }
when { resource.namespace == "foobar" }
```

where `resource.request` is unknown at the authorization stage, but all other metadata is known, in other words, a `create secret` request in the `prod` namespace shall be denied, even though this order of `unless when` gives `<residual> && false` in that case.

I will do the AST rewrite that I anyways planned, in order to ask the question "can this expression become true?". For `true` and `<residual>`, the answer is (trivially) yes, for TPE output of form `false`, `<residual> && false`, no. This is used to find at least one permit policy that is or _can become_ true in the already authorization stage. If we know already at the authorization stage that the two possible outcomes for a permit policy are either `false` or error, we do not consider that policy for determining whether the request should be authorized to proceed to the next stage.

Likewise, I'd ask "would this expression _necessarily_ evaluate to true?", to which the answer is yes for `true` and `<residual> || true`, and no for `false` and `<residual>`. This is used to check whether there is a forbid policy that always fires. By Cedar design, I know that both permit and forbid policies short-circuit on errors and ignore the policy. I'm considering to (at least initially) restrict the expressiveness of the forbid policies such that they cannot error, to keep Cedar error semantics. Do I remember correctly that the only two (current) expression operations that may error are arithmetic operations and extensions?

If this "fail fast" method is useful upstream in core as some extra library function, I'm happy to open a PR for it, but if it's too narrow a use-case (with potential footguns if used incorrectly, e.g. by not restricting deny policy expressitivity) I'm also ok to keep this downstream. The idea is to contribute the Rust PoC I'm implementing now (if all goes) to https://github.com/cedar-policy/cedar-access-control-for-k8s, so it can also be staged there as a library function (with related safety invariants) until we find other use-cases potentially.

emina:
> The use-case I'm experimenting with for this is in detail covered in [my thesis](https://github.com/luxas/research/blob/main/msc_thesis.pdf), but on a high level, Kubernetes has two request stages (of importance within this context): authorization (metadata-only) and admission (metadata+request body). At the authorization stage, in the absence of the request body Cedar entity, it is desired to reject any such request, which cannot become authorized later (when the request body is deserialized). This in order to prevent DoS, where an attacker could overload the server with large bogus write (or read) requests that never pass the second stage, but use resources in between the first and the second one.
> 
> Currently, it is not possible to write policies conditional to the request body in the authorization stage. However, it's a useful feature that is being explored currently. There are a couple of ways to achieve this, but what I have currently is to allow policy authors to write one unified policy across both stages (without possibly even realizing it). In other words, one could do
> 
> ```
> permit(principal, action == create, resource is core::Secret)
> unless { resource.request.data.isSecure }
> when { resource.namespace == "foobar" }
> ```
> 
> where `resource.request` is unknown at the authorization stage, but all other metadata is known, in other words, a `create secret` request in the `prod` namespace shall be denied, even though this order of `unless when` gives `<residual> && false` in that case.
> 
> I will do the AST rewrite that I anyways planned, in order to ask the question "can this expression become true?". For `true` and `<residual>`, the answer is (trivially) yes, for TPE output of form `false`, `<residual> && false`, no. This is used to find at least one permit policy that is or _can become_ true in the already authorization stage. If we know already at the authorization stage that the two possible outcomes for a permit policy are either `false` or error, we do not consider that policy for determining whether the request should be authorized to proceed to the next stage.
> 
> Likewise, I'd ask "would this expression _necessarily_ evaluate to true?", to which the answer is yes for `true` and `<residual> || true`, and no for `false` and `<residual>`. This is used to check whether there is a forbid policy that always fires. By Cedar design, I know that both permit and forbid policies short-circuit on errors and ignore the policy. I'm considering to (at least initially) restrict the expressiveness of the forbid policies such that they cannot error, to keep Cedar error semantics. Do I remember correctly that the only two (current) expression operations that may error are arithmetic operations and extensions?
> 
> If this "fail fast" method is useful upstream in core as some extra library function, I'm happy to open a PR for it, but if it's too narrow a use-case (with potential footguns if used incorrectly, e.g. by not restricting deny policy expressitivity) I'm also ok to keep this downstream. The idea is to contribute the Rust PoC I'm implementing now (if all goes) to https://github.com/cedar-policy/cedar-access-control-for-k8s, so it can also be staged there as a library function (with related safety invariants) until we find other use-cases potentially.

Ah, thank you for sharing the details of your use case.  The symbolic compiler can answer these kinds of questions (more detail below).  You can, of course, also answer them with TPE together with sound rewrites too.  Which approach is better depends on how much time you can spend answering these questions and how much precision you need.  For example, if precision is more important than efficiency, the symbolic compiler would be a good solution.  Otherwise, rewrites would be a better choice.

Depending on the exact details of what can be (un)known in your policies, you can answer these questions without any partial evaluation at all, using the symbolic compiler alone.  That's because the symbolic compiler itself is a partial evaluator, and you can fix (make constant) some of its inputs while leaving other inputs symbolic. Or you can specify what's constant with additional constraints (assumptions) about the inputs.  

To see how this would work, let's consider your example policy:

```
permit(principal, action == create, resource is core::Secret)
unless { resource.request.data.isSecure }
when { resource.namespace == "foobar" }
```

We can use the symbolic compiler can answer the following questions about this policies:
1. Assuming that principal == User::"Alice", resource.namespace == "bar", can this policy ever evaluate to "true"?  The answer is "no". 
2. Assuming that  principal == User::"Alice", resource.namespace == "foobar", can this policy ever evaluate to "true"?  The answer is "yes". 
3. Assuming that  principal == User::"Alice", resource.namespace == "foobar", can this policy ever evaluate to "false" or "error"?  The answer is "yes". 
4. We can write even richer assumptions (not just equalities).  For instance,  assuming that  principal == User::"Alice", resource.namespace != "foobar", can this policy ever evaluate to "true"?  The answer is "no".

Granted, the above queries aren't supported by the CLI, so we'd have to write some new glue code to emit them for you :)  But if it's something that sounds interesting (and your use case can afford the added latency compared to TPE), it can be done.

luxas:
Thanks for giving a concrete example on how the symbolic compiler can be used for this too, that's very interesting!

In the couple of days here in between, I implemented the most of the TPE flow, and it seems to work pretty well 👍 Latency is of concern, so intuitively (based on the quick testing of the symbolic compiler from before, but no real side-by-side benchmarking) I'd say that latency is more important than precision.

For now, what I did was to fold `<residual> && false` into `false` for allow policies (as they would anyways be skipped if `<residual>` errored), and (most likely) both `<residual> && false => false` and `<residual> || true => true` for deny policies, which has more limited syntax at policy validation time, deny policies cannot use arithmetic nor extension functions, to make sure that `<residual>` cannot error, to make sure that we can fold safely.

Another interesting use-case I encountered, was that some values are always present (e.g. a resource's name), and a policy author can use that when writing policies, with `__cedar::String` type. However, some requests might return a list of objects, where the returned name is unrestricted. In this case, I keep `resource.name` unknown when performing TPE. A policy author can authorize such a request by not putting any predicate (neither in permit or forbid policies) on `resource.name`, and if the policies are such that no residual is returned, and the decision otherwise is allowed, the request is allowed. However, if we for this method did not expect a residual, one can conclude that the policy author put a constraint (even sth like an allow policy's `resource.name != "foo"`, and thus must the decision be deny, as the current request (at least in theory) could return a resource with name foo, as name is unrestricted.

I'll close the PR, but it'd be good to continue the discussion on what parts of these TPE use-cases/features we want to expose in upstream Cedar Rust API.

PR #1722, opened 2025-07-29
Allow principal resource slots in condition

alanwang67:
## Description of changes
Removes the check that prevents `?principal` and `?resource` slots from appearing in the scope of the policy. Instead now these slots can be used in the condition of the template as long as they also appear in the scope. Since we are not longer going with having user defined slots in the scope of the template, the only way to allow using the slots in the scope in the condition of the template will be to use `?principal` and `?resource`. 

## Issue #, if available
https://github.com/cedar-policy/rfcs/pull/98

- [X] Requires updates, and I have made / will make these updates myself. (Please include in your description a timeline or link to the relevant PR in [`cedar-docs`](https://github.com/cedar-policy/cedar-docs). PRs should be targeted at a `staging-X.Y` branch, not `main`.)

### DIFF
--- a/cedar-policy-core/src/est.rs
+++ b/cedar-policy-core/src/est.rs
@@ -452,10 +452,12 @@ impl std::fmt::Display for Clause {
 #[cfg(test)]
 mod test {
     use super::*;
+    use crate::ast::SlotId;
     use crate::parser::{self, parse_policy_or_template_to_est};
     use crate::test_utils::*;
     use cool_asserts::assert_matches;
     use serde_json::json;
+    use std::collections::HashMap;
 
     /// helper function to just do EST data structure --> JSON --> EST data structure.
     /// This roundtrip should be lossless for all policies.
@@ -530,6 +532,23 @@ mod test {
             .node
             .expect("Node should not be empty");
         cst.try_into().expect("Failed to convert to EST")
+    }
+
+    #[track_caller]
+    fn cst_to_est_link_roundtrip(
+        policy_str: &str,
+        expected_policy_str: &str,
+        vals: &HashMap<SlotId, EntityUidJson>,
+    ) {
+        let cst = parser::text_to_cst::parse_policy(policy_str)
+            .unwrap()
+            .node
+            .unwrap();
+        let est: Policy = cst.try_into().unwrap();
+
+        let new_policy = est.link(vals).unwrap();
+
+        assert_eq!(new_policy.to_string(), expected_policy_str);
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


cedar-policy-core/src/est.rs
```diff
@@ -312,25 +315,44 @@ impl Policy {
...
-            .into())
-        } else {
-            Ok(e)
         }
+        Ok(e)
     }
+
     /// `id` is the ID of the policy the clause belongs to, used only for reporting errors
-    fn try_into_ast(self, id: &ast::PolicyID) -> Result<ast::Expr, FromJsonError> {
+    fn try_into_ast(
```
cdisselkoen: bonus points for adding description of the `has_principal` and `has_resource` parameters to the doc comment here. (what they mean, why they're here)

cdisselkoen (approved):

cedar-policy-core/src/est.rs
```diff
@@ -312,25 +315,47 @@ impl Policy {
...
-    fn filter_slots(e: ast::Expr, is_when: bool) -> Result<ast::Expr, FromJsonError> {
-        let first_slot = e.slots().next();
-        if let Some(slot) = first_slot {
-            Err(parse_errors::SlotsInConditionClause {
-                slot,
-                clause_type: if is_when { "when" } else { "unless" },
+    fn filter_slots(
+        e: ast::Expr,
+        has_principal: bool,
+        has_resource: bool,
```
john-h-kastner-aws: Here and elsewhere: these should be more explicit names IMO, `has_principal_slot` and `has_resource_slot`

cedar-policy-core/src/parser/cst_to_ast.rs
```diff
@@ -320,24 +322,37 @@ impl Node<Option<cst::Policy>> {
         let maybe_scope = policy.extract_scope();
 
         // convert conditions
-        let maybe_conds = ParseErrors::transpose(policy.conds.iter().map(|c| {
-            let (e, is_when) = c.to_expr::<ast::ExprBuilder<()>>()?;
+        let maybe_conds = match policy.extract_scope() {
+            Ok((p, _, r)) => {
+                let slots_in_scope: HashSet<ast::Slot> =
```
john-h-kastner-aws, line 327: Constructing the `Expr`s and iterating over their ast feels unnecessarily heavy when we could instead do pattern matching on the scope variants.

john-h-kastner-aws (commented):
noticed this which probably needs to change 
https://github.com/cedar-policy/cedar/blob/b0e6683aa878a0c41359c6ba0f68026eea15a208/cedar-policy-core/src/est.rs#L159-L162

alanwang67:
> noticed this which probably needs to change
> 
> https://github.com/cedar-policy/cedar/blob/b0e6683aa878a0c41359c6ba0f68026eea15a208/cedar-policy-core/src/est.rs#L159-L162

Thanks for pointing that out, just updated it.

cedar-policy-core/src/ast/policy.rs
```diff
@@ -1384,6 +1384,18 @@ impl PrincipalConstraint {
             _ => self,
         }
     }
+
+    /// If the principal constraint has a slot, return it
+    pub fn get_slot(self) -> Option<Slot> {
+        match self.constraint {
+            PrincipalOrResourceConstraint::Eq(EntityReference::Slot(l))
+            | PrincipalOrResourceConstraint::In(EntityReference::Slot(l)) => Some(Slot {
```
john-h-kastner-aws: Needs to handle `PrincipalOrResourceConstraint::IsIn`

john-h-kastner-aws: which raises the question why `with_filled_slot` above this doesn't handle that case. bug?

alanwang67: Yeah think it's a bug, seems to only touch the conversion from of a linked template to est and validating slots with templates that use an `IsIn` operator. I'll just make a separate PR to fix `with_filled_slot`.

PR #1732, opened 2025-07-31
Internal data structures for Generalized templates

alanwang67:
## Description of changes
Introduce internal data structures for generalized templates. The main changes are having a policy now store a map from slots to types. The differences in the changes between [Allow principal resource slots in condition](https://github.com/cedar-policy/cedar/pull/1722) and this PR are [here](https://github.com/alanwang67/cedar/compare/allow_principal_resource_slots_in_condition...alanwang67:cedar:generalized_templates_branch). 


cedar-policy-core/src/ast/generalized_slots_annotation.rs
```diff
@@ -0,0 +1,122 @@
...
+use crate::extensions::Extensions;
+use crate::validator::{
+    json_schema::Type as JSONSchemaType, types::Type as ValidatorType, RawName, SchemaError,
+    ValidatorSchema,
+};
+use serde::{Deserialize, Serialize};
+
+/// Struct storing the pairs of SlotId's and their corresponding type
+#[derive(Clone, Eq, PartialEq, PartialOrd, Ord, Debug, Hash, Serialize, Deserialize)]
+pub struct GeneralizedSlotsAnnotation(BTreeMap<SlotId, JSONSchemaType<RawName>>);
```
cdisselkoen: Instead of `Annotation` (which means something else in Cedar), maybe this could be `Declaration`?

cedar-policy-core/src/ast/generalized_slots_annotation.rs
```diff
@@ -0,0 +1,122 @@
...
+
+impl From<BTreeMap<SlotId, JSONSchemaType<RawName>>> for GeneralizedSlotsAnnotation {
+    fn from(value: BTreeMap<SlotId, JSONSchemaType<RawName>>) -> Self {
+        Self(value)
+    }
+}
+
+/// Struct storing the pairs of SlotId's and their corresponding validator types
+#[derive(Clone, Eq, PartialEq, PartialOrd, Ord, Debug, Hash)]
+pub struct ValidatorGeneralizedSlotsAnnotation(BTreeMap<SlotId, ValidatorType>);
```
cdisselkoen: Same here about "Annotation"

cedar-policy-core/src/ast/policy.rs
```diff
@@ -266,15 +299,18 @@ impl Template {
     pub fn check_binding(
         template: &Template,
         values: &HashMap<SlotId, EntityUID>,
+        generalized_values: &HashMap<SlotId, RestrictedExpr>,
```
cdisselkoen, line 306: I could imagine callers might not always know when to use `values` vs `generalized_values`.  Do they use `generalized_values` for all slots other than `principal` and `resource`?  or only for slots whose type isn't `EntityUID`?  It would be good to at least have a comment about this.  But I wonder if we couldn't just have all values, including EntityUIDs, be passed as a single map from `SlotId` to `RestrictedExpr`

cedar-policy-core/src/ast/policy.rs
```diff
@@ -370,6 +523,45 @@ pub enum LinkingError {
...
+        /// invalid slots
+        invalid: Vec<SlotId>,
+    },
+
+    /// Invalid slots in generalized_env (Principal and Resource slots)
+    #[error(fmt = describe_invalid_slot_in_env_error)]
+    InvalidSlotIdInGeneralizedEnv {
+        /// invalid slots
+        invalid: Vec<SlotId>,
+    },
```
cdisselkoen: did the error messages get switched for these two?

cedar-policy-core/src/ast/policy.rs
```diff
@@ -370,6 +523,45 @@ pub enum LinkingError {
...
+    #[error("value provided `{value}` for `{slot}` is not an EntityUID")]
+    ValueProvidedForSlotInScopeNotEntityUID {
+        /// Slot
+        slot: SlotId,
+        /// Value provided
+        value: RestrictedExpr,
+    },
+
+    /// Value provided for a slot is not of the type annotated
+    #[error("value provided `{value}` for `{slot}` is not of type annotated `{ty}`")]
```
cdisselkoen:
```suggestion
    #[error("value provided `{value}` for `{slot}` is not of declared type `{ty}`")]
```

cedar-policy-core/src/ast/policy.rs
```diff
@@ -370,6 +523,45 @@ pub enum LinkingError {
...
+        /// Value provided
+        value: RestrictedExpr,
+        /// Expected Type
+        ty: ValidatorType,
+    },
+
+    /// Types provided are not found in the Schema
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    AnnotatedTypeNotFoundInSchema(#[from] SchemaError),
```
cdisselkoen: Same here with `Annotated`.  `Declared`?

cedar-policy-core/src/ast/policy.rs
```diff
@@ -399,6 +603,28 @@ fn describe_arity_error(
...
+    )
+}
+
+fn describe_invalid_slot_in_generalized_env_error(
+    invalid: &[SlotId],
+    fmt: &mut std::fmt::Formatter<'_>,
+) -> std::fmt::Result {
+    write!(
+        fmt,
+        "The following slots should not be in the generalized_values hashmap: {}",
```
cdisselkoen: This error message would probably be pretty inscrutable to an end-user.  Could this error ever occur for someone using our public APIs, or only if someone was misusing Core APIs?

cedar-policy-core/src/ast/policy.rs
```diff
@@ -640,6 +897,8 @@ pub struct LiteralPolicy {
     link_id: Option<PolicyID>,
     /// Values of the slots
     values: SlotEnv,
+    /// Generalized values of the slots
+    generalized_values: GeneralizedSlotEnv,
```
cdisselkoen: Would it be simpler to store the values of all slots in a `GeneralizedSlotEnv`?  `EntityUID`s could be stored as a `RestrictedExpr`

cedar-policy-core/src/ast/policy.rs
alanwang67: Will add a comment, currently the convention is that all values that are for `?principal` and `?resource` go into `values` and everything else goes into `generalized_values`. I think for our internal API we can flatten the two, but for backwards compatibility we probably would still need this distinction?

cedar-policy-core/src/ast/generalized_slots_annotation.rs
alanwang67: Gotcha, will change all occurrences.

cedar-policy-core/src/ast/policy.rs
alanwang67: Yeah, think it did will fix this.

cedar-policy-core/src/ast/policy.rs
alanwang67: I think it could occur for an end user, since currently principal, resource, and generalized slots are all of the same type so the type system isn't enforcing anything about the only values belonging in `generalized_vals` being a generalized slot.

cedar-policy-core/src/ast/policy.rs
alanwang67: Just saw the last comment, I think yeah I can just switch it so that `generalized_values` stores everything and have the link API just expose `generalized_values`.

john-h-kastner-aws (commented):
Can you say a little more about the relation between this and #1722. And I guess lay out more the implementation plan so I know what to expect coming in future PRs.

alanwang67:
> Can you say a little more about the relation between this and #1722. And I guess lay out more the implementation plan so I know what to expect coming in future PRs.

The changes in #1722 only allow the current ```?principal``` and ```?resource``` slots to appear in the condition of the template. The changes there more or less don't touch any of the other stuff with user defined slots. The other parts of the PR are [here](https://github.com/cedar-policy/cedar/pull/1738). The changes there are to the typechecker (the typechecker struct stores a map from slot id's to validator types and uses that when typechecking slots), changes to the parser, EST, and evaluator (whenever we evaluate we carry around another the ```generalized_values``` map).

cedar-policy-core/src/ast/policy.rs
alanwang67: I spent some more time thinking about this and I think one of the consequences would be that now every time we have code involving templates we would have to handle the regular templates and the generalized case differently since the values of their slots appear in different fields and we would also have to enforce the invariant that only one of these ```HashMaps``` can have values in it. I was considering making the default value field of type ```HashMap<SlotId, RestrictedExpr>``` however, that would cause issues with the EST format. 

One alternative solution in terms of making it easier for user's of the API not needing to supply values in two ```HashMaps``` and the potentially confusing error messages is to have the API handle the translation to our internal data structures.

cdisselkoen: If I understand correctly, you think it would be better to have a single map for internal data structures, two maps for the EST, and handle the conversion during translation.  That sounds fine to me.

cedar-policy-core/src/ast/policy.rs
cdisselkoen: Based on out-of-band discussion, my understanding is we'll keep the distinction between `values` and `generalized_values` in our internal structures, so that we can have the more precise value type of `EntityUID` for `values`.  Thanks for adding the comment to this function.

cedar-policy-core/src/ast/policy.rs
```diff
@@ -289,27 +344,119 @@ impl Template {
...
+
     /// Attempt to create a template-linked policy from this template.
     /// This will fail if values for all open slots are not given.
     /// `new_instance_id` is the `PolicyId` for the created template-linked policy.
     pub fn link(
         template: Arc<Template>,
         new_id: PolicyID,
         values: HashMap<SlotId, EntityUID>,
+        generalized_values: HashMap<SlotId, RestrictedExpr>,
+        schema: Option<&ValidatorSchema>,
```
cdisselkoen, line 443: It would be great to also have a comment on this function explaining that `?principal` and `?resource` must be in `values` (if present at all) and all other slots must be in `generalized_values`

cedar-policy-core/src/ast/policy.rs
```diff
@@ -421,23 +593,35 @@ pub struct Policy {
     /// The constructor `new` is only visible in this module,
     /// so it is the responsibility of callers to maintain
     values: HashMap<SlotId, EntityUID>,
+    /// All of the generalized slots in `template` MUST
+    /// be bound by `generalized values`.
+    /// The SlotId for generalized_values should be disjoint from
+    /// values as well
+    generalized_values: HashMap<SlotId, RestrictedExpr>,
```
cdisselkoen: I think we could say, more strongly, not just that the SlotIds need to be disjoint, but that all SlotIds other than `?principal` and `?resource` need to be in `generalized_values`.

cedar-policy-core/src/ast/policy.rs
```diff
@@ -421,23 +593,35 @@ pub struct Policy {
...
     /// Link a policy to its template
     /// INVARIANT (values total map):
     /// `values` must bind every open slot in `template`
-    fn new(template: Arc<Template>, link_id: Option<PolicyID>, values: SlotEnv) -> Self {
+    fn new(
+        template: Arc<Template>,
+        link_id: Option<PolicyID>,
+        values: SlotEnv,
+        generalized_values: GeneralizedSlotEnv,
+    ) -> Self {
```
cdisselkoen: this `INVARIANT` comment should be updated

cedar-policy-core/src/ast/policy.rs
cdisselkoen: Can you update this comment thread based on our out-of-band discussion?

cdisselkoen (approved):

cedar-policy-core/src/ast/policy.rs
alanwang67: After implementing a version of this PR that used a single HashMap to store all slot values, there were issues involved with need to introduce many ```unwrap``` calls to translate between `RestrictedExpr` and `EntityUID`. Therefore, the choice ultimately made is to internally represent values of templates using two HashMap's and for user convenience at the API level to expose a single HashMap from `SlotId` to `RestrictedExpr`. Internally, we will do this conversion to use two HashMap's and so therefore we have to maintain the invariant that in value there are only bindings for SlotId's ```?principal``` and ```?resource``` and in generalized_values there are only bindings for generalized slots.

cedar-policy-core/src/ast/generalized_slots_declaration.rs
```diff
@@ -0,0 +1,122 @@
...
+use crate::extensions::Extensions;
+use crate::validator::{
+    json_schema::Type as JSONSchemaType, types::Type as ValidatorType, RawName, SchemaError,
+    ValidatorSchema,
+};
+use serde::{Deserialize, Serialize};
+
+/// Struct storing the pairs of SlotId's and their corresponding type
+#[derive(Clone, Eq, PartialEq, PartialOrd, Ord, Debug, Hash, Serialize, Deserialize)]
+pub struct GeneralizedSlotsDeclaration(BTreeMap<SlotId, JSONSchemaType<RawName>>);
```
john-h-kastner-aws, line 29: On a first pass it's not clear why we need this and `ValidatorGeneralizedSlotsDeclaration`. Probably worth a comment explaining this design

cedar-policy-core/src/validator/schema.rs
```diff
@@ -537,6 +547,39 @@ impl ValidatorSchema {
         )
     }
 
+    /// Construct a [`Type`] from [`json_schema::Type<RawName>`]
+    pub fn json_schema_type_to_validator_type(
```
john-h-kastner-aws, line 551: Does this function not already exist somewhere else? I feel like it must

john-h-kastner-aws (approved):
a couple questions, but looks good

cedar-policy-core/src/ast/generalized_slots_declaration.rs
alanwang67: Gotcha, yeah will include a comment. The reason why we need both is that ```JSONSchemaType<RawName>``` can't be used by the typechecker and needs to be converted to a validator type by combing information from the Schema.

cedar-policy-core/src/validator/schema.rs
alanwang67: I think the other place where a similar conversion occurs is the function ```from_schema_fragments``` in ```cedar-policy-core/src/validator/schema.rs``` but it's basically inlined within it.

PR #1733, opened 2025-07-31
Export some interface for compiling to and manipulating `Term` directly.

zhengyao-lin:
## Description of changes

This PR exports some interface (`Term`, `TermVar`, `TermType`, `symcc::factory`, `verify_*`, and `check_(un)sat`) for manipulating terms directly.
Currently these exports are gated behind the `term` feature flag.

Two motivations for this change:
- We might need it for SymCC DRT (@chaluli)
- Some other unpublished fancy tools might depend on it

However, at the same time, I'm not entirely sure about this change as it's quite easy to construct a ill-typed query by combining these functions.

In any case, thanks for reviewing!


chaluli:
Looks good, but I don't think this is actually sufficient for DRT with the Lean model.

Specifically, we cannot actually destruct a term because some of the underlying types (e.g., `Op`, `ExtOp`, `Ext`, etc.) are private. If we are worried about users foot gunning themselves, why don't we make an additional "trusted" `Term` type that is a thin wrapper around the Term type, but notable cannot be constructed using our public API from the "untrusted" Term. But could be converted into an "untrusted" term.

The `compile_*` can return a trusted `Term`. The `check_unsat`/`check_sat` then requires a trusted `Term`.

A end user can convert a trusted term to an untrusted term and then modify and manipulate the untrusted term as much as they like. But most notibly cannot convert an untrusted `Term` to an trusted `Term`. Then we shouldn't have any problem making the Term type completely public.

This also allows us to change the internal Term Type

We can expose all of the internal types without letting a user running SymCC on those modified terms.

Alternatively, we have a function that converts the "untrusted" terms to "trusted" terms by checking that the term is well formed.

cdisselkoen:
This sounds great to me 👍🏻

zhengyao-lin:
Yes this sounds better! Although we might need to do a `WellFormedAsserts` instead of `WellFormedTerm`, and then we can add a `WellFormedAsserts::asserts(&self) -> Asserts`, and export all relevant types for terms like `Op`.

PR #1747, opened 2025-08-04
SymCC: Make `Term`-related interface more difficult to be misused

zhengyao-lin:
## Description of changes

As suggested by @chaluli, this PR exports more `Term`-related types and functions, and also introduces a `WellFormedAsserts` type to wrap the results of `CedarSymCompiler::compile_*`, so that it's less likely to be misused (and as a result, I also removed the `term` feature flag):

```
pub struct WellFormedAsserts<'a> {
    asserts: Asserts,
    /// Policies that have been used to generate the asserts.
    footprint: Vec<&'a cedar_policy_core::ast::Policy>,
}
```

### DIFF
--- a/cedar-policy-symcc/src/lib.rs
+++ b/cedar-policy-symcc/src/lib.rs
@@ -20,7 +20,14 @@ mod symcc;
 use solver::Solver;
 use symcc::SymCompiler;
 pub use symcc::{solver, Env, Environment, Interpretation, SmtLibScript, SymEnv};
-use symcc::{well_typed_policies, well_typed_policy};
+use symcc::{
+    verify_always_allows, verify_always_denies, verify_disjoint, verify_equivalent, verify_implies,
+    verify_never_errors, well_typed_policies, well_typed_policy,
+};
+
+pub use symcc::factory as term;
+pub use symcc::type_abbrevs::*;
+pub use symcc::{Asserts, BitVec, Ext, ExtOp, Op, Term, TermType, TermVar, Uuf};
 
 use cedar_policy::{Policy, PolicySet, RequestEnv, Schema};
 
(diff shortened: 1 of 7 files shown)


cedar-policy-symcc/src/lib.rs
```diff
@@ -327,153 +334,191 @@ impl<S: Solver> CedarSymCompiler<S> {
...
+    ///
+    /// NOTE: Constructing [`WellFormedAsserts`] this way may lead to errors.
+    pub fn from_asserts_unchecked<'a>(
+        asserts: Asserts,
+        policies: impl Iterator<Item = &'a cedar_policy_core::ast::Policy>,
+    ) -> Self {
+        WellFormedAsserts {
+            asserts,
+            footprint: policies.map(|p| p.condition()).collect(),
         }
```
cdisselkoen: does the footprint need to be all _subexpressions_ of these policies?

cedar-policy-symcc/src/symcc.rs
```diff
@@ -269,10 +266,15 @@ impl<S: Solver> SymCompiler<S> {
...
+        let footprint = policies1
+            .policies()
+            .chain(policies2.policies())
+            .map(|p| p.condition())
+            .collect::<Vec<_>>();
         self.check_sat(
             |symenv| verify_implies(policies1, policies2, symenv),
             symenv,
-            policies1.policies().chain(policies2.policies()),
+            footprint.iter(),
```
cdisselkoen, line 270: Should be able to avoid the `.collect()` here I think (and thus the allocation of a `Vec`)?

cedar-policy-symcc/src/symcc.rs
```diff
@@ -293,10 +295,14 @@ impl<S: Solver> SymCompiler<S> {
...
     ) -> Result<Option<Env>> {
+        let footprint = policies
+            .policies()
+            .map(|p| p.condition())
+            .collect::<Vec<_>>();
         self.check_sat(
             |symenv| verify_always_allows(policies, symenv),
             symenv,
-            policies.policies(),
+            footprint.iter(),
```
cdisselkoen, line 298: Same here about avoiding the `.collect()`?

cdisselkoen: And in other places below

cedar-policy-symcc/src/symcc/decoder.rs
```diff
@@ -143,12 +143,12 @@ pub enum SExpr {
...
-///     \u{d₁d₀}
-///     \u{d₂d₁d₀}
-///     \u{d₃d₂d₁d₀}
-///     \u{d₄d₃d₂d₁d₀}
+///   \ud₃d₂d₁d₀  
+///   \u{d₀}
+///   \u{d₁d₀}
+///   \u{d₂d₁d₀}
+///   \u{d₃d₂d₁d₀}
+///   \u{d₄d₃d₂d₁d₀}
```
cdisselkoen: could make this a bulleted list instead (in Markdown)

cdisselkoen (approved):

cedar-policy-symcc/src/lib.rs
zhengyao-lin: Yes, that process is done in some lower-level functions like `repair_as_counterexample` and [`concretize`](https://github.com/cedar-policy/cedar/blob/c152d99d7b3193c3f972bfd398eb84ba5c7ab043/cedar-policy-symcc/src/symcc/concretize.rs#L480), which `SymCompiler::check_sat` calls.

zhengyao-lin: So this is kind of split up into two steps. maybe that's not entirely desirable

cedar-policy-symcc/src/symcc.rs
zhengyao-lin: `Policy::condition` produces an owned value, so it's kind of difficult to get its reference without storing it somewhere first? Previously this `collect` is done withing `check_sat`, so we are technically not doing more allocation in this PR.

cedar-policy-symcc/src/lib.rs
cdisselkoen: No, that sounds great to me, storing all subexpressions would be a potentially exponentially large amount of memory I think

cedar-policy-symcc/src/symcc.rs
cdisselkoen: Ah I see. `check_sat` needs an iterator of `&Expr` not of `Expr`

cedar-policy-symcc/src/lib.rs
```diff
@@ -327,153 +346,197 @@ impl<S: Solver> CedarSymCompiler<S> {
...
-            Ok(verify_implies(
-                pset1.policy_set(),
-                pset2.policy_set(),
-                symenv,
-            )?)
+    /// Creates a new [`WellFormedAsserts`] from the given [`Asserts`]
+    /// without checking if it is well-formed.
+    ///
+    /// NOTE: Constructing [`WellFormedAsserts`] this way may lead to errors.
+    pub fn from_asserts_unchecked<'b>(
```
chaluli, line 385: Why may this lead to errors?

Can we analyze the inputs to ensure that they won't lead to an error when encoded / executed by the SMT solver?

If so, can we do that analysis and change the return type to `Result<Self, ...>`.

Though, I don't think we "need" to do this analysis in this PR.  But it would be nice if that is possible. Would also be good to know what errors we expect may be possible.

zhengyao-lin: Yeah I left this as future work since it might be a bit complicated for the following reasons.

The user may construct arbitrary non-well-formed terms as inputs for this function, and we currently don't have an executable check to see if a Term is well-formed (there is only a [Prop](https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/Cedar/Thm/SymCC/Data/Basic.lean#L463) in the Lean model). But this should be relatively straightforward.

What complicates this more is the footprint computation. The user may have constraints that talk about entity expressions that are missing from the footprint, and then try to call `check_sat` with them, which might produce incorrect results.

I think one long term solution is to refactor Rust/Lean versions to compute footprint from `Term` directly (instead of `Expr`), so that instead of depending on a user-provided footprint, we can directly compute it.
With this, we can probably also decouple the acyclicity/transitivity constraints from `Asserts`, and then only generate them in `check_unsat` or `check_sat` directly from the input `Asserts`/`Term`; this can allow users to more freely compose results of various `verify_*` functions.
But this might require a medium refactor in the meta-theory results, and I'm not sure if @emina would agree.

zhengyao-lin: We could have some limited way to compose `WellFormedAsserts` without manipulating `Asserts` directly, I think.

For example, it should be sound to take the conjunction/disjunction of two `WellFormedAsserts`s (and maybe negations too), provided that their `symenv`s are equal and the resulting footprint is the union of their respective footprints.

Maybe this can help reducing the number of queries for, e.g., Cedar Analyzer

I can try to add it if you think that's a good idea!

chaluli: Sounds good. I agree that it is too much work to check that the term is well formed in this PR. We can add this as an issue for future work.

Perhaps, we should add a comment in the code saying how to use the function correctly (e.g., provide the policies used to construct the terms) and state that if the user doesn't do that it may result in errors.

I think the idea of separating the footprint construction is interesting, but certainly requires a medium to big change to the meta-theory proofs in Lean.

For now, I don't think we need any other ways to construct the WellFormedAsserts at this point.

emina: > I think one long term solution is to refactor Rust/Lean versions to compute footprint from `Term` directly (instead of `Expr`), so that instead of depending on a user-provided footprint, we can directly compute it. With this, we can probably also decouple the acyclicity/transitivity constraints from `Asserts`, and then only generate them in `check_unsat` or `check_sat` directly from the input `Asserts`/`Term`; this can allow users to more freely compose results of various `verify_*` functions. But this might require a medium refactor in the meta-theory results, and I'm not sure if @emina would agree.

I've tried this before, and it's a very difficult refactoring.  Proof of correctness become very hard when there is no direct relation between expressions and their corresponding footprint terms, as established by the current algorithm (which walks the expression and gathers the footprint terms---so you can always figure out which expression a footprint term came from).

To get an intuition for where this get hard, suppose the footprint is computed from `t` instead, like this:

```
def Term.footprint (t : Term) : Set Term :=
  match t with
  | .prim _
  | .var _               => ofLeaf
  | .none _              => Set.empty
  | .some tᵢ             => tᵢ.footprint
  | .set (Set.mk ts) _   => ts.attach.mapUnion λ ⟨tᵢ, _⟩ => tᵢ.footprint
  | .record (Map.mk ats) => ats.attach₃.mapUnion λ ⟨(_, tᵢ), _⟩ => tᵢ.footprint
  | .app o ts ty         => ofApp o ty ∪ ts.attach.mapUnion λ ⟨tᵢ, _⟩ => tᵢ.footprint
where
  ofLeaf := if t.typeOf.isEntityType then Set.singleton (⊙t) else Set.empty
  ofApp  : Op → TermType → Set Term
    | .record.get _, .option (.entity _) => Set.singleton t
    | .record.get _, .entity _           => Set.singleton (⊙t)
    | _, _                               => Set.empty
```

(The code above predates tags so ignore the fact that they aren't being handled :)

Now, consider this key lemma footprint proven [here](https://github.com/cedar-policy/cedar-spec/blob/7650a698e2a796b8c5b4118ac93d9e2874bc0807/cedar-lean/Cedar/Thm/SymCC/Enforcer/Footprint.lean#L146):

```
private theorem mem_footprint_exists_wf_prop {p : Expr → Prop} {x : Expr} {tₑ : Term} {εnv : SymEnv}
  (hwε : εnv.WellFormedFor x)
  (hp  : p x)
  (hin : tₑ ∈ footprint x εnv)
  (hite  : ∀ {x₁ x₂ x₃}, p (Expr.ite x₁ x₂ x₃) → p x₁ ∧ p x₂ ∧ p x₃)
  (hand  : ∀ {x₁ x₂}, p (Expr.and x₁ x₂) → p x₁ ∧ p x₂)
  (hor   : ∀ {x₁ x₂}, p (Expr.or x₁ x₂) → p x₁ ∧ p x₂)
  (happ₁ : ∀ {o x₁}, p (Expr.unaryApp o x₁) → p x₁)
  (happ₂ : ∀ {o x₁ x₂}, p (Expr.binaryApp o x₁ x₂) → p x₁ ∧ p x₂)
  (hget  : ∀ {a x₁}, p (Expr.getAttr x₁ a) → p x₁)
  (hhas  : ∀ {a x₁}, p (Expr.hasAttr x₁ a) → p x₁)
  (hset  : ∀ {xs}, p (Expr.set xs) → ∀ x ∈ xs, p x)
  (hrec  : ∀ {axs}, p (Expr.record axs) → ∀ ax ∈ axs, p ax.snd)
  (hcall : ∀ {f xs}, p (Expr.call f xs) → ∀ x ∈ xs, p x) :
  ∃ xₑ, εnv.WellFormedFor xₑ ∧ p xₑ ∧ compile xₑ εnv = .ok tₑ
:= by
``` 

It requires us to establish that every term `tₑ` in the footprint of `x` has a corresponding expression `xₑ` from which it was compiled.  This is straightforward to establish when the footprint is computed by walking the AST of `x`: an easy induction shows that `xₑ` is some subexpression of `x` in this case.

If the footprint is computed from `t` instead, then we have to prove this instead:

```
private theorem mem_footprint_exists_wf_prop {p : Expr → Prop} {x : Expr} {tₑ t : Term} {εnv : SymEnv}
  (hwε : εnv.WellFormedFor x)
  (hp  : p x)
  (ht  : reduce x εnv = .ok t)
  (hin : tₑ ∈ t.footprint)
  (hite  : ∀ {x₁ x₂ x₃}, p (Expr.ite x₁ x₂ x₃) → p x₁ ∧ p x₂ ∧ p x₃)
  (hand  : ∀ {x₁ x₂}, p (Expr.and x₁ x₂) → p x₁ ∧ p x₂)
  (hor   : ∀ {x₁ x₂}, p (Expr.or x₁ x₂) → p x₁ ∧ p x₂)
  (happ₁ : ∀ {o x₁}, p (Expr.unaryApp o x₁) → p x₁)
  (happ₂ : ∀ {o x₁ x₂}, p (Expr.binaryApp o x₁ x₂) → p x₁ ∧ p x₂)
  (hget  : ∀ {a x₁}, p (Expr.getAttr x₁ a) → p x₁)
  (hhas  : ∀ {a x₁}, p (Expr.hasAttr x₁ a) → p x₁)
  (hset  : ∀ {xs}, p (Expr.set xs) → ∀ x ∈ xs, p x)
  (hrec  : ∀ {axs}, p (Expr.record axs) → ∀ ax ∈ axs, p ax.snd)
  (hcall : ∀ {f xs}, p (Expr.call f xs) → ∀ x ∈ xs, p x) :
  ∃ xₑ, εnv.WellFormedFor xₑ ∧ p xₑ ∧ reduce xₑ εnv = .ok tₑ
:= by
```

Observe that we have now lost the correspondence between the subexpressions of `x` and the terms ` tₑ` that are in the footprint of `t`.  How do we materialize _some_ ` xₑ` that corresponds to ` tₑ`?  This becomes hard because there can be  a ` tₑ` in the footprint of `t` that doesn't actually correspond to any subexpression of `x`.  

For example, consider the expression `context.friends` where `.friends` is of type `Set<User>` and `User` is an entity.  Now, suppose that the symbolic `context` binds `.friends` to the literal `Term.set (Set.make [User::"Alice", User::"Bob"])`.  The footprint of this term includes `User::"Alice"` and `User::"Bob"`.  But neither of these are actually subexpressions of `x`, so we have to materialize expressions that correspond to these literals (and, of course, establish that these literals are well-formed, from assumptions).  

Note also, that the term-based footprint is _bigger_ than the expression-based footprint for `context.friends`.  The expression-based one would just be the empty set.

To avoid this inflation in the size of the footprint, we'd have to develop an even more precise / tricker algorithm than what I showed above---and, of course, all the proofs would get even harder as the result.

This is all to say that such a refactoring is possible, but it's a major one that we should think carefully about undertaking (in terms of complexity / benefits tradeoffs).

cedar-policy-symcc/src/lib.rs
```diff
@@ -327,153 +346,206 @@ impl<S: Solver> CedarSymCompiler<S> {
...
-        pub async fn check_unsat(&mut self, asserts: Asserts, symenv: &SymEnv) -> Result<bool> {
-            Ok(self.symcc.check_unsat(|_| Ok(asserts), symenv).await?)
-        }
+/// Well-formed assertions generated by the symbolic compiler.
+#[derive(Debug, Clone)]
+pub struct WellFormedAsserts<'a> {
+    symenv: &'a SymEnv,
+    asserts: Asserts,
+    /// All [`cedar_policy_core::ast::Expr`]s that have been compiled when generating the asserts.
+    footprint: Vec<cedar_policy_core::ast::Expr>,
```
lianah, line 355: Should this be a set?

cedar-policy-symcc/src/lib.rs
zhengyao-lin: Thank you so much for the detailed explanation!

I didn't know that this was already tried before; in that case it's probably not worth it.

cedar-policy-symcc/src/lib.rs
zhengyao-lin: Yes I did consider that, but decided to use Vec for now since this is usually a list `[p.condition() for p in policy set]`, so there's probably not much redundancy.

I can change it in another PR if you think HashSet or BTreeSet makes more sense here. We can also maybe store the results of footprint as `BTreeSet<Term>`, which could be much smaller than the current version as it only includes terms of entity types if I understood correctly.

PR #1748, opened 2025-08-04
Allow serializing `Unknown` values as `CedarValueJSON`.

chaluli:
Update construction of `CedarValueJson` from `BorrowedRestrictedExpr` to serialize unknown as an extension function with name "unknown" and arg the name of the unknown. This matches the already existing pattern for deserializing an extension function with name "unknown" as an Unknown `RestrictedExpr`.

### DIFF
--- a/cedar-policy-core/src/entities.rs
+++ b/cedar-policy-core/src/entities.rs
@@ -2196,6 +2196,28 @@ mod json_parsing_tests {
             });
         });
     }
+
+    #[test]
+    fn serialize_unknown_no_error() {
+        let test = serde_json::json!([{
+            "uid" : { "type" : "A", "id" : "b" },
+            "attrs": {
+                "age":  {
+                    "__extn": {
+                        "fn": "unknown",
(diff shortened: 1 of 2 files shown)


john-h-kastner-aws (approved):
Change looks good, but worth looking closer at other code dealing with restricted expressions.

The restricted expression invariant is specified as 

https://github.com/cedar-policy/cedar/blob/2132364464a74e4bce088369d7f5039835010057/cedar-policy-core/src/ast/restricted_expr.rs#L36-L43

but then we have the function

https://github.com/cedar-policy/cedar/blob/2132364464a74e4bce088369d7f5039835010057/cedar-policy-core/src/ast/restricted_expr.rs#L102-L106

which seems to violate the invariant (but claims `unknown` is a valid restricted expression). Is there anywhere else that we're not handling the unkown case properly when matching on a restricted expression.

john-h-kastner-aws (commented):
also should have a test

PR #1755, opened 2025-08-05
Template analysis

alanwang67:
## Description of changes
Introduced some APIs to do analysis on templates. Also merged in Cedar main and changed the Cargo.toml file to use the local version because I needed to use a new API to create a request environment that also included type information for principal and resource slots.

- [] A breaking change requiring a major version bump to `cedar-policy` (e.g., changes to the signature of an existing API).
- X ] A backwards-compatible change requiring a minor version bump to `cedar-policy` (e.g., addition of a new API).

- [] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).


zhengyao-lin:
Very nice PR! Not a full review but I'm curious about how you handled the soundness of the SMT encoding. Right now it seems like principal and resource slots are symbolically represented as two `TermVar`s of the relevant types, which would correspond to the same instantiation for each template (I think?). But I wonder if there might be cases where either (a) a property holds for one instantiation, but doesn't hold for multiple instantiations; (b) a property holds if all templates use the same instantiation, but doesn't hold if we allow different instantiations.

~I'm not familiar enough with templates to come up with counterexamples unfortunately; so there might be some meta-theory results that rule out these cases.~

edit: okay maybe one example would be
```
entity E enum [ "A", "B" ];
action view appliesTo { principal: [E], resource: [E] };

permit(principal == ?principal, action, resource);
```
If only one instance is allowed, this policy set would never always allow;
but if we allow two instances, this statically linked version always allows any request
```
permit(principal == E::"A", action, resource);
permit(principal == E::"B", action, resource);
```
but I'm not sure if there's an example without enums.

alanwang67:
> Very nice PR! Not a full review but I'm curious about how you handled the completeness of the SMT encoding. Right now it seems like principal and resource slots are symbolically represented as two `TermVar`s of the relevant types, which would correspond to the same instantiation for each template (I think?). But I wonder if there might be cases where either (a) a property holds for one instantiation, but doesn't hold for multiple instantiations; (b) a property holds if all templates use the same instantiation, but doesn't hold if we allow different instantiations.
> 
> ~I'm not familiar enough with templates to come up with counterexamples unfortunately; so there might be some meta-theory results that rule out these cases.~
> 
> edit: okay maybe one example would be
> 
> ```
> entity E enum [ "A", "B" ];
> action view appliesTo { principal: [E], resource: [E] };
> 
> permit(principal == ?principal, action, resource);
> ```
> 
> If only one instance is allowed, this policy set would never always allow; but if we allow two instances, this statically linked version always allows any request
> 
> ```
> permit(principal == E::"A", action, resource);
> permit(principal == E::"B", action, resource);
> ```
> 
> but I'm not sure if there's an example without enums.

If I understand correctly, currently, the only properties that I included in the API are properties that make sense for all instantiation of the template. So I didn't write an API for always allows since, we would probably want something closer along the lines of if there exists an instantiation of a slot that results in the template always allowing the request which I don't think we can express without quantifiers.

PR #1761, opened 2025-08-06
SymCC: Error type refactoring

zhengyao-lin:
## Description of changes

This PR refactors error types in SymCC to be more idiomatic.

Major changes are:
- Merged `Error` in `src/symcc.rs` into `src/err.rs`
- Separated `BitVecError` and `IPError` from `CompileError` (previously `symcc::result::Error`)
- Made error messages more standard according to [common conventions](https://rust-lang.github.io/api-guidelines/interoperability.html#examples-of-error-messages).
- All error types should be exported now (see `src/err.rs` for a full list)
- Added `EncodeError` for errors from the encoder

### DIFF
--- a/cedar-policy-symcc/Cargo.toml
+++ b/cedar-policy-symcc/Cargo.toml
@@ -26,6 +26,7 @@ tokio = { version = "1.0", features = ["io-util", "process"] }
 num-bigint = "0.4"
 num-traits = "0.2"
 chrono = "0.4.41"
+miette = { version = "7.6.0" }
 
 [dev-dependencies]
 miette = { version = "7.5.0", features = ["fancy"] }
--- a/cedar-policy-symcc/Cargo.toml
+++ b/cedar-policy-symcc/Cargo.toml
@@ -26,7 +26,7 @@ tokio = { version = "1.0", features = ["io-util", "process"] }
 num-bigint = "0.4"
 num-traits = "0.2"
 chrono = "0.4.41"
-miette = { version = "7.6.0" }
+miette = "7.6.0"
 
 [dev-dependencies]
 miette = { version = "7.5.0", features = ["fancy"] }
--- a/cedar-policy-symcc/src/err.rs
+++ b/cedar-policy-symcc/src/err.rs
@@ -17,9 +17,8 @@ use miette::Diagnostic;
  */
 use thiserror::Error;
 
-pub use crate::{
-    solver::SolverError,
-    symcc::{BitVecError, CompileError, ConcretizeError, DecodeError, IPError},
+pub use crate::symcc::{
+    BitVecError, CompileError, ConcretizeError, DecodeError, EncodeError, IPError, SolverError,
 };
 
 /// Top-level errors from the whole `cedar-policy-symcc` crate.
@@ -34,9 +33,9 @@ pub enum Error {
     /// Errors during symbolic compilation.
     #[error("symbolic compilation failed")]
     CompileError(#[from] CompileError),
-    /// Errors in the SMT encoder.
+    /// Errors from the SMT encoder.
     #[error("failed to encode SMT terms")]
-    EncodeError(#[source] anyhow::Error),
+    EncodeError(#[from] EncodeError),
     /// Solver-related errors.
     #[error(transparent)]
     SolverError(#[from] SolverError),
--- a/cedar-policy-symcc/src/err.rs
+++ b/cedar-policy-symcc/src/err.rs
@@ -1,3 +1,5 @@
+use cedar_policy_core::validator::ValidationError;
+use miette::Diagnostic;
 /*
  * Copyright Cedar Contributors
  *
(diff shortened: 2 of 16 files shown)


cedar-policy-symcc/Cargo.toml
```diff
@@ -26,6 +26,7 @@ tokio = { version = "1.0", features = ["io-util", "process"] }
 num-bigint = "0.4"
 num-traits = "0.2"
 chrono = "0.4.41"
+miette = { version = "7.6.0" }
```
cdisselkoen: nit: This can just be written `miette = "7.6.0"`

cedar-policy-symcc/src/err.rs
```diff
@@ -15,35 +17,41 @@
...
+pub use crate::{
+    solver::SolverError,
+    symcc::{BitVecError, CompileError, ConcretizeError, DecodeError, IPError},
+};
+
+/// Top-level errors from the whole `cedar-policy-symcc` crate.
+#[derive(Debug, Diagnostic, Error)]
+#[diagnostic(
+    help("if you are not using *_unchecked functions, this is likely due to a bug in the symbolic compiler or cvc5")
+)]
```
cdisselkoen: Is this true for all of the variants here -- I'm particularly looking at `SolverError` which can be thrown in different ways depending on the particular `Solver` implementation?

cedar-policy-symcc/src/symcc/compiler.rs
```diff
@@ -365,16 +374,18 @@ pub fn compile_or(t1: Term, r2: Result<Term>) -> Result<Term> {
...
+        (_, _) => Err(CompileError::TypeError),
     }
 }
 
 pub fn compile_set(ts: Vec<Term>) -> Result<Term> {
     if ts.is_empty() {
-        Err(Error::UnsupportedError) // Reject empty set literals
+        Err(CompileError::UnsupportedFeature(
+            "empty set literals are not supported".to_string(),
+        )) // Reject empty set literals
```
cdisselkoen: probably don't need the comment anymore

cedar-policy-symcc/src/symcc/extension_types/ipaddr.rs
```diff
@@ -22,15 +22,27 @@
...
+    bitvec::{BitVec, BitVecError},
     type_abbrevs::{nat, Fin, Nat, Width},
 };
 
-type Result<T> = std::result::Result<T, Error>;
+/// Errors in [`IPNet`] operations.
+#[derive(Debug, Diagnostic, Error)]
+pub enum IPError {
+    /// Errors in [`BitVec`] operations.
+    #[error("bit-vector error when manipulating ip addresses")]
```
cdisselkoen, line 38: We should include the text from the underlying `BitVecError` here.

```suggestion
    #[error("bit-vector error when manipulating ip addresses: {0}")]
```

cedar-policy-symcc/src/symcc/result.rs
```diff
@@ -17,22 +17,36 @@
...
-    Unreachable(String),
+    /// Unsupported features.
+    #[error("unsupported feature in SymCC")]
+    UnsupportedFeature(String),
+    /// Bit-vector error.
+    #[error("bit-vector error ")]
+    BitVecError(#[from] BitVecError),
+    /// IP address error.
+    #[error("IP address error")]
+    IPError(#[from] IPError),
```
cdisselkoen, line 48: similarly here we should include the text of the underlying error

cdisselkoen (approved):

cedar-policy-symcc/src/symcc/extension_types/ipaddr.rs
zhengyao-lin: Ah that's where I wasn't particularly sure about.

Since we derive the `thiserror::Error` trait, it should implement [`Error::source`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source), which points to the underlying error (BitVecError in this case). And then at the top-level, when we do `format!("{:?}", miette::Report::from(err))`, we should get the entire chain of errors, making the `: {0}` part redundant?

But if the user just do `format!("{:?}", err)`, it just prints the top-level error. So I'm not sure what's the best practice here

cedar-policy-symcc/src/err.rs
zhengyao-lin: Ah I see. Maybe I'll remove this help message for now

cedar-policy-symcc/src/symcc/extension_types/ipaddr.rs
cdisselkoen: ooh I see

cedar-policy-symcc/src/symcc/bitvec.rs
```diff
@@ -34,7 +33,24 @@ pub struct BitVec {
...
+    ExtractOutOfBounds,
+    /// Attempting to create a bit-vector with zero width
+    #[error("cannot create a bit-vector with zero width")]
+    ZeroWidthBitVec,
+    /// Mismatched bit-vector widths in various operations
+    #[error("mismatched bit-vector widths in {0}")]
+    MismatchedWidths(String),
+    /// Shifting by a value greater than the bit-width
+    #[error("shifting by a value greater than the bit-width")]
+    ShiftAmountTooLarge,
```
lianah, line 50: Minor, the error message is not quite right: this error happens when the shift value cannot fit in u32. It will not happen if the shift amount does fit in u32 but the shift value is greater than the current bit-width.

zhengyao-lin: Thanks for catching that! I changed it to `shift amount too large to fit in u32`

PR #1764, opened 2025-08-07
Fixes for Rust 1.89.0

chaluli:
## Description of changes

Fixes compilation errors caused by upgrade to rust 1.89.0

### DIFF
--- a/cedar-language-server/src/utils.rs
+++ b/cedar-language-server/src/utils.rs
@@ -63,16 +63,17 @@ fn to_lsp_severity(severity: miette::Severity) -> lsp_types::DiagnosticSeverity 
 /// (a source range with a message). Errors wanting to provide more information
 /// can use this function while building a richer diagnostic with something like
 /// `to_lsp_diagnostic(error, src).map(|d| Diagnostic { data: my_data, ..d })`
-#[allow(clippy::unwrap_used, reason = "LSP code is experimental")]
 pub(crate) fn to_lsp_diagnostics<'a>(
     diagnostic: &'a dyn miette::Diagnostic,
     src: &'a str,
 ) -> Vec<lsp_types::Diagnostic> {
     let mut message = diagnostic.to_string();
     if let Some(source) = diagnostic.source() {
+        #[allow(clippy::unwrap_used, reason="writing string cannot fail")]
         write!(&mut message, ". {source}").unwrap();
     }
     if let Some(help) = diagnostic.help() {
+        #[allow(clippy::unwrap_used, reason="writing string cannot fail")]
         write!(&mut message, ". {help}").unwrap();
     }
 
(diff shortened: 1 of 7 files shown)


cedar-language-server/src/utils.rs
```diff
@@ -63,6 +63,7 @@ fn to_lsp_severity(severity: miette::Severity) -> lsp_types::DiagnosticSeverity
 /// (a source range with a message). Errors wanting to provide more information
 /// can use this function while building a richer diagnostic with something like
 /// `to_lsp_diagnostic(error, src).map(|d| Diagnostic { data: my_data, ..d })`
+#[allow(clippy::unwrap_used, reason = "LSP code is experimental")]
```
john-h-kastner-aws: The two unwraps here are safe. Move allow onto the expressions with comment says writing to a String can't fail

cedar-policy/src/ffi/convert.rs
```diff
@@ -29,7 +29,7 @@ use wasm_bindgen::prelude::wasm_bindgen;
 #[cfg(feature = "wasm")]
 extern crate tsify;
 
-/// Takes a PolicySet represented as string and return the policies
+/// Takes a `PolicySet`` represented as string and return the policies
```
john-h-kastner-aws: extra back-tick

cedar-policy-core/src/lib.rs
```diff
@@ -17,6 +17,10 @@
 //! Implementation of the Cedar parser and evaluation engine in Rust.
 #![warn(missing_docs)]
 #![cfg_attr(feature = "wasm", allow(non_snake_case))]
+#![allow(
```
john-h-kastner-aws, line 20: not very happy allowing this at the crate level

chaluli: Me neither. But it's the only choice allowed by rust 1.89.0.

chaluli: I even tried to only allow when compiling in test mode but that isn't enough to sate rust 1.89.0.

cdisselkoen (approved):

cedar-policy-core/src/validator/schema.rs
```diff
@@ -1098,6 +1098,10 @@ impl ValidatorSchema {
 #[derive(Debug, Clone, Deserialize)]
 #[serde(bound(deserialize = "N: Deserialize<'de> + From<RawName>"))]
 #[serde(transparent)]
+#[allow(
+    dead_code,
+    reason = "Not actually dead, but linter mistakenly things this code is dead"
```
adpaco:
```suggestion
    reason = "Not actually dead, but linter mistakenly thinks this code is dead"
```

PR #1766, opened 2025-08-07
Change policy to expression

shaobo-he-aws:
Construct conjunctions that associate to the right

### DIFF
--- a/cedar-policy-core/src/ast/policy.rs
+++ b/cedar-policy-core/src/ast/policy.rs
@@ -1194,16 +1194,16 @@ impl TemplateBody {
     pub fn condition(&self) -> Expr {
         match self {
             TemplateBody::TemplateBody(TemplateBodyImpl { .. }) => Expr::and(
+                self.principal_constraint_expr(),
                 Expr::and(
+                    self.action_constraint_expr(),
                     Expr::and(
-                        self.principal_constraint_expr(),
-                        self.action_constraint_expr(),
+                        self.resource_constraint_expr(),
+                        self.non_scope_constraints().clone(),
                     )
                     .with_maybe_source_loc(self.loc().into_maybe_loc()),
-                    self.resource_constraint_expr(),
                 )
                 .with_maybe_source_loc(self.loc().into_maybe_loc()),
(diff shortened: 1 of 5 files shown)


cedar-policy-core/src/ast/policy.rs
```diff
@@ -1194,16 +1194,16 @@ impl TemplateBody {
...
-                        self.principal_constraint_expr(),
-                        self.action_constraint_expr(),
+                        self.resource_constraint_expr(),
+                        self.non_scope_constraints().clone(),
                     )
                     .with_maybe_source_loc(self.loc().into_maybe_loc()),
-                    self.resource_constraint_expr(),
                 )
                 .with_maybe_source_loc(self.loc().into_maybe_loc()),
-                self.non_scope_constraints().clone(),
```
cdisselkoen, line 1206: should we use the `and_nary()` Expr constructor here to make this more concise?

cdisselkoen (approved):

cedar-policy-core/src/ast/policy.rs
shaobo-he-aws: `and_nary` doesn't work here because the accumulator of its foldl is the LHS of an and op. So we get something like `(c3 && c2) && c1` for c1, c2, c3

cdisselkoen: Should we change `and_nary` as well?

chaluli: Would it be better to have it fold into a vec/list/etc. and then pass the result as input to `and_nary`?

shaobo-he-aws: > Should we change `and_nary` as well?

I renamed it to `and_naryl` to better reflect its naming.

shaobo-he-aws: > Would it be better to have it fold into a vec/list/etc. and then pass the result as input to `and_nary`?

The intended foldr cannot be easily expressed using `and_nary` here.

cdisselkoen (approved):

Issue #105, opened 2023-05-31
Improve precision when typechecking `in`


john-h-kastner-aws:
### Category

Cedar validation features

### Describe the feature you'd like to request

The validator sometimes gives imprecise types to expressions using `in`.

* `principal in []` will type as `Bool` when principal is Unspecified, although we could give the more precise type of `False`.
* Expressions like `action in action` or `principal in principal` will type as `Bool`, even though we could give them type `True`.
* If the action hierarchy says that `Action::"read"` is a descendant of `Action::"read_or_write"`, then `Action::"read"` in `Action::"read_or_write"` will type as `Bool`, even though we could type it as `True`. 

### Describe the solution you'd like

n/a

### Describe alternatives you've considered

n/a

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

john-h-kastner-aws:
From #310, a substantial improvement can be had by implementing singleton action types. This would make giving precise types to `in` expression involving actions trivial, even simplifying some of the existing code. This will also support precise types for `in` expression in policy bodies that aren't directly applied to action literals.

john-h-kastner-aws:
Implementing this in a backwards compatible manner requires #539

john-h-kastner-aws:
Related to #638

cdisselkoen:
Partially resolved by #1769.  Some of the cases identified in this issue still remain, like `principal in []`

john-h-kastner-aws:
#1772 was revised before merging so this issue isn't actually resolved

john-h-kastner-aws:
Calling this resolved with #1931. There are probably more ways to improve precision, but the current Rust implementation now matches what was modeled and proven correct in Lean, making it slightly more precise than before. Any other extension will need to be carefully considered (i.e., and RFC) and shown to be sound and analyzable before implementation in Rust.

PR #1769, opened 2025-08-08
More precise policy validation for action hierarchy

zhengyao-lin:
## Description of changes

This PR makes the policy validation more precise, mimicking the behavior of a [recent Lean validator change](https://github.com/cedar-policy/cedar-spec/pull/648#discussion_r2172668108).
Basically, when type checking `x in y`, when x has an action entity type, and any action of x's type does not have an ancestor of y's entity type, then we type `x in y` as the singleton false type.

This this should only make the validator more precise, no existing validated policy should break.

**Not sure**

### DIFF
--- a/cedar-policy-core/src/validator/typecheck.rs
+++ b/cedar-policy-core/src/validator/typecheck.rs
@@ -1817,6 +1817,18 @@ impl<'a> SingleEnvTypechecker<'a> {
         }
     }
 
+    /// Checks if `lhs_ety` may be a descendant of `rhs_ety` in the action hierarchy.
+    /// Lean counterpart: https://github.com/cedar-policy/cedar-spec/blob/7e231a68b0e0eb1b8ce1362e81de4568671a668a/cedar-lean/Cedar/Validation/Types.lean#L202
+    fn check_action_descendant(&self, lhs_ety: &EntityType, rhs_ety: &EntityType) -> bool {
+        lhs_ety == rhs_ety
+            || self.schema.action_ids().any(|action| {
+                action.name().entity_type() == rhs_ety
+                    && action
+                        .descendants()
+                        .any(|desc| desc.entity_type() == lhs_ety)
+            })
+    }
+
     /// Handles typechecking of `in` expressions. This is complicated because it
     /// requires searching the schema to determine if an `in` expression
     /// consisting of variables and literals can ever be true. When we find that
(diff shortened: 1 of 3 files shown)


cedar-policy-core/src/validator/typecheck.rs
```diff
@@ -1929,17 +1941,86 @@ impl<'a> SingleEnvTypechecker<'a> {
...
+                        // could be a collection of entity types.
+                        (
+                            Some(Type::EntityOrRecord(EntityRecordKind::ActionEntity {
+                                name: lhs_ety,
+                                ..
+                            })),
+                            Some(Type::EntityOrRecord(EntityRecordKind::Entity(rhs_etys))),
+                        ) if rhs_etys
+                            .iter()
+                            .all(|rhs_ety| !self.check_action_descendant(lhs_ety, rhs_ety)) =>
```
cdisselkoen: is this missing a check for `EntityRecordKind::ActionEntity`?  In the single case, we check that it's `ActionEntity`, but in the multiple case, we do not, which seems inconsistent

cdisselkoen:
noting for context/posterity that this addresses some (but not all) of #105

cedar-policy-core/src/validator/typecheck.rs
zhengyao-lin: Yes thanks for catching that! I also realized that I missed the case when the LHS could be `EntityRecordKind::Entity`. But I think I have fixed all of these cases in the new commit below

cedar-policy-core/src/validator/typecheck.rs
```diff
@@ -1817,6 +1817,64 @@ impl<'a> SingleEnvTypechecker<'a> {
         }
     }
 
+    /// Checks if `lhs_ety` may be a descendant of `rhs_ety` in the action hierarchy.
+    /// Lean counterpart: <https://github.com/cedar-policy/cedar-spec/blob/7e231a68b0e0eb1b8ce1362e81de4568671a668a/cedar-lean/Cedar/Validation/Types.lean#L202>
+    fn check_action_in_entity_type(&self, lhs_ety: &EntityType, rhs_ety: &EntityType) -> bool {
```
cdisselkoen: can we be super explicit about the preconditions for this function, in the comment? It takes two `EntityType`s; does it assume that LHS, RHS, or both are Action types?

cdisselkoen (approved):

cedar-policy-core/src/validator/typecheck.rs
zhengyao-lin: Sounds good. I added more comments below. `lhs_ety` is expected to be an action entity type; `rhs_ety` can be any entity type.

shaobo-he-aws (approved):
Lgtm

PR #1771, opened 2025-08-08
SymCC: Fix bvso implementation

zhengyao-lin:
## Description of changes

Integer overflow checks for literals would previously always succeed in SymCC (found by @shaobo-he-aws in DRT!)


shaobo-he-aws (approved):
LGTM

PR #1776, opened 2025-08-12
Fix well typed policy construction

shaobo-he-aws:

### DIFF
--- a/cedar-policy-core/src/validator/rbac.rs
+++ b/cedar-policy-core/src/validator/rbac.rs
@@ -39,8 +39,8 @@ use super::{schema::*, Validator};
 impl Validator {
     /// Validate if a [`Template`] contains entities of enumerated entity types
     /// but with invalid UIDs
-    pub(crate) fn validate_enum_entity<'a>(
-        &'a self,
+    pub fn validate_enum_entity<'a>(
+        schema: &'a ValidatorSchema,
         template: &'a Template,
     ) -> impl Iterator<Item = ValidationError> + 'a {
         policy_entity_uids(template)
@@ -49,7 +49,7 @@ impl Validator {
                 if let Some(ValidatorEntityType {
                     kind: ValidatorEntityTypeKind::Enum(choices),
                     ..
-                }) = self.schema.get_entity_type(e.entity_type())
+                }) = schema.get_entity_type(e.entity_type())
                 {
                     match is_valid_enumerated_entity(&Vec::from(choices.clone().map(Eid::new)), e) {
                         Ok(_) => {}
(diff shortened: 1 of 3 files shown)


cedar-policy-symcc/src/symcc.rs
```diff
@@ -409,6 +409,28 @@ fn well_typed_policy_inner(
     schema: &Schema,
 ) -> Result<Policy> {
     let validator_schema = schema.as_ref();
+    // We need to perform these three checks like what the validator does here: https://github.com/cedar-policy/cedar/blob/82784864c01b5096cb73885dd2df5643074355ed/cedar-policy-core/src/validator.rs#L178-L185
+    // We don't need `validate_template_action_application` because existence of `env` already serves as evidence
+    let errs: Vec<_> = cedar_policy_core::validator::Validator::validate_entity_types(
```
john-h-kastner-aws, line 414: We shouldn't need to perform these checks for every request env. Can you lift them outside whatever loop enumerates `RequestEnv`s?

cedar-policy-core/src/validator/rbac.rs
john-h-kastner-aws, line 1: Not really a fan of changing the signature to all these functions if we can avoid it. The way the validator module is set up you should construct a `Validator` wrapping the schema, and then, if you need a schema ref, you can call `Validator::schema`

cedar-policy-symcc/src/symcc.rs
shaobo-he-aws: A well-typed policy is constructed according to a request environment.

cedar-policy-core/src/validator/rbac.rs
shaobo-he-aws: I'm afraid we can't pass a validator without cloning the schema.

john-h-kastner-aws: We should be able to if we push it up the stack enough and pass around `&Validator` instead of `&Schema` in some of the `symcc` code. E.g., `WellTypedPolicies::from_policies` could take a `&Validator`

cedar-policy-symcc/src/symcc.rs
john-h-kastner-aws: hmm, so this is a larger refactoring? should definitely be possible at some level

cedar-policy-core/src/validator/rbac.rs
cdisselkoen: Backing up, why the distinction between the `Validator` and `Schema` types anyway?  One is just a wrapper around the other, is it causing headaches for no reason?

shaobo-he-aws: `Schema` can't perform validation.

shaobo-he-aws: > We should be able to if we push it up the stack enough and pass around `&Validator` instead of `&Schema` in some of the `symcc` code. E.g., `WellTypedPolicies::from_policies` could take a `&Validator`

I'll leave this design decision to the SymCC developers. @cdisselkoen @chaluli. IMO, the status quo looks better.

cdisselkoen: I understand that today, the validation methods are on `Validator` or require `Validator`.  Would it make sense / simplify things to move those methods to `ValidatorSchema` etc?  Is the separation causing more headache than it's worth?

chaluli: I personally don't care which direction we go. I'm not sure why there is a separation between `Validator` and `ValidatorSchema` for these actions currently.

In terms of what to do for SymCC, I think it doesn't really matter. We could wrap the `ValidatorSchema` in a `Validator` up front and pass that around rather than the `ValidatorSchema`.

PR #1781, opened 2025-08-13
Update release workflow with step to push tags to other repos

adpaco-aws:
## Description of changes

Updated the release workflow with a new step to push tags to other repos, namely `cedar-spec` and `cedar-integration-tests`. This step uses the new GitHub App approach which required to be setup. It also uses the `github-script@v7` action to create the release branch in `cedar-integration-tests` and the tags in release branches.

I have not tested this in my fork because it would require setting up a GitHub App there too.

### DIFF
--- a/.github/workflows/build_release.yml
+++ b/.github/workflows/build_release.yml
@@ -8,11 +8,20 @@ on:
 jobs:
   create_related_repo_tags:
     runs-on: ubuntu-latest
+    if: ${{ vars.APP_ID && secrets.APP_PRIVATE_KEY }}
     steps:
+      - name: Generate GitHub App token
+        id: app-token
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: ${{ vars.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+          repositories: cedar-spec,cedar-integration-tests
+
       - name: Create tag in cedar-spec
         uses: actions/github-script@v7
         with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+          github-token: ${{ steps.app-token.outputs.token }}
           script: |
             const tag = context.ref.replace('refs/tags/', '');
             await github.rest.git.createRef({
@@ -29,7 +38,7 @@ jobs:
       - name: Create tag in cedar-integration-tests
         uses: actions/github-script@v7
         with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+          github-token: ${{ steps.app-token.outputs.token }}
           script: |
             const tag = context.ref.replace('refs/tags/', '');
             await github.rest.git.createRef({
--- a/.github/workflows/build_release.yml
+++ b/.github/workflows/build_release.yml
@@ -38,47 +38,6 @@ jobs:
               })).data.commit.sha
             });
 
-      - name: Create release branch in cedar-integration-tests
-        uses: actions/github-script@v7
-        with:
-          github-token: ${{ steps.app-token.outputs.token }}
-          script: |
-            const tag = context.ref.replace('refs/tags/', '');
-            const version = tag.replace('v', '');
-            const [major, minor, patch] = version.split('.');
-            const releaseBranch = `release/${major}.${minor}.x`;
-
-            // Only create branch for patch releases
-            if (patch === '0') {
-              // Check if branch already exists
-              try {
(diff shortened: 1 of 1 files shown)


katherine-hough:
Should this be tagging main opposed to the corresponding release branch on cedar-drt?

adpaco:
> Should this be tagging main opposed to the corresponding release branch on cedar-drt?

This code is already tagging the `main` branches. Or am I missing something?

katherine-hough:
Sorry, what I meant to ask is whether we should be tagging the release branch instead of main

.github/workflows/build_release.yml
```diff
@@ -6,6 +6,52 @@ on:
...
+          script: |
+            const tag = context.ref.replace('refs/tags/', '');
+            await github.rest.git.createRef({
+              owner: context.repo.owner,
+              repo: 'cedar-spec',
+              ref: `refs/tags/${tag}`,
+              sha: (await github.rest.repos.getBranch({
+                owner: context.repo.owner,
+                repo: 'cedar-spec',
+                branch: 'main'
```
chaluli: I don't think we should create the tag from main. It should create the tag from `release/<TAG>` because cedar-spec has release branches.

In general this should be safe, but in some cases (e.g., patch releases and minors that are cherry-picked), cedar-spec main may be incompatible with release <TAG> for cedar.

chaluli (changes requested):
I don't think we should create the tag from main. It should create the tag from release/<TAG> because cedar-spec has release branches.

In general this should be safe, but in some cases (e.g., patch releases and minors that are cherry-picked), cedar-spec main may be incompatible with release for cedar.

adpaco:
Makes sense. There should be a release branch in `cedar-spec` at this point in the process, right? But not for `cedar-integration-tests`.

chaluli:
> Makes sense. There should be a release branch in `cedar-spec` at this point in the process, right? But not for `cedar-integration-tests`.

Yes, to my knowledge the release process we follow creates the release branch of `cedar-spec` by the time the tags are created.

adpaco:
Alright, we're targeting the release branches now. I've also added another step to create the release branch in `cedar-integration-tests` (only for non-patch releases). I'm way less confident that's going to work, but we can check if at least the `cedar-spec` tagging does.

.github/workflows/build_release.yml
```diff
@@ -24,31 +24,78 @@ jobs:
...
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ steps.app-token.outputs.token }}
+          script: |
+            const tag = context.ref.replace('refs/tags/', '');
+            const version = tag.replace('v', '');
+            const [major, minor, patch] = version.split('.');
+            const releaseBranch = `release/${major}.${minor}.x`;
+
+            // Only create branch for patch releases
```
katherine-hough: Should this say non-patch releases?

chaluli (approved):
Looks good modulo Katie's comment.

.github/workflows/build_release.yml
```diff
@@ -6,6 +6,99 @@ on:
...
+          repositories: cedar-spec,cedar-integration-tests
+
+      - name: Create tag in cedar-spec
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ steps.app-token.outputs.token }}
+          script: |
+            const tag = context.ref.replace('refs/tags/', '');
+            const version = tag.replace('v', '');
+            const [major, minor] = version.split('.');
```
john-h-kastner-aws: Will this just drop the third element when splitting something like `4.5.1`? Seems reasonable, but IDK

john-h-kastner-aws (approved):

adpaco:
Just talked to @katherine-hough , we're going to update the process so the two branches are already created at this point.  Thus I've removed the step to create the release branch for `cedar-integration-tests`.

.github/workflows/build_release.yml
adpaco: Yes, I'd expect this code to do that. But I've updated it to use `[major, minor, _]` so it's even clearer that there will be three elements and we just don't care about the last one.

chaluli:
> Just talked to @katherine-hough , we're going to update the process so the two branches are already created at this point. Thus I've removed the step to create the release branch for `cedar-integration-tests`.

Good suggestion.

chaluli (approved):

katherine-hough (approved):

PR #1788, opened 2025-08-15
Avoid cache term types and terms that has built-in SMT-LIB encoding

shaobo-he-aws:

### DIFF
--- a/cedar-policy-symcc/src/symcc/encoder.rs
+++ b/cedar-policy-symcc/src/symcc/encoder.rs
@@ -267,13 +267,21 @@ impl<S: tokio::io::AsyncWrite + Unpin + Send> Encoder<'_, S> {
             Some(enc) => Ok(enc.clone()),
             None => {
                 let enc = match ty {
-                    TermType::Bool => "Bool".to_string(),
-                    TermType::String => "String".to_string(),
-                    TermType::Bitvec { n } => format!("(_ BitVec {n})"),
-                    TermType::Option { ref ty } => {
-                        format!("(Option {})", self.encode_type(ty).await?)
+                    TermType::Bool => {
+                        return Ok("Bool".to_string());
                     }
-                    TermType::Set { ty } => format!("(Set {})", self.encode_type(ty).await?),
+                    TermType::String => {
+                        return Ok("String".to_string());
+                    }
+                    TermType::Bitvec { n } => {
+                        return Ok(format!("(_ BitVec {n})"));
+                    }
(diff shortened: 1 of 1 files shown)


shaobo-he-aws:
On a second thought, maybe it's Lean that should be changed to cache these.

chaluli (approved):

cedar-policy-symcc/src/symcc/encoder.rs
john-h-kastner-aws, line 1: I'm staring at this and cannot figure out what's actually changed lol. Where is the caching happening?

cedar-policy-symcc/src/symcc/encoder.rs
```diff
@@ -555,8 +570,12 @@ impl<S: tokio::io::AsyncWrite + Unpin + Send> Encoder<'_, S> {
                 ["(none)".to_string(), "(some (val X))".to_string()],
             )
             .await?;
+        let mut ids: Vec<_> = Vec::new();
         for t in ts {
             let id = self.encode_term(t).await?;
+            ids.push(id);
```
chaluli, line 576: @john-h-kastner-aws I believe by caching he means this line. Where he stores the ids of the top-level asserts and later adds the asserts after all of the definitions required by all top-level asserts have been encoded.

shaobo-he-aws: It's actually these `return`s.

cdisselkoen: So to check my understanding -- for many types, we now early-return from `encode_type()`, which means we just return the encoding without modifying `self.types`. Whereas, for types without early-return (or prior to this PR, for all types), we call `self.types.insert()` before returning the encoding, thus caching the encoding if we see this type in the future.  Presumably the gamble is that these type encodings are cheap enough to generate (most of them are one `format!` call) that it's not worth inserting them into a hashtable?  

I'm not totally sure I buy that this gamble pays off -- hashtable insertions are cheap; we don't actually save any hashtable lookups (we always check the cache regardless); `format!()` calls are not free, and some of the early-returns even require recursion.

shaobo-he-aws: Porting Emina's reply here: `self.types` is also used by the decoder, which may assume that only entity or record types are there. I don't think we don't get much performance improvement if we cache the types. That being said, I'm also fine either way as long as we don't break the decoder.

PR #1796, opened 2025-08-21
SymCC docs

zhengyao-lin:
## Description of changes

Adds some missing doc comments for exported types and functions (and minor tweaks for clippy warnings).
Sorry if this conflicts with any work in progress, in which case feel free to ignore this PR.

### DIFF
--- a/cedar-policy-symcc/src/symcc/extension_types/datetime.rs
+++ b/cedar-policy-symcc/src/symcc/extension_types/datetime.rs
@@ -18,7 +18,8 @@
 //! It is based on
 //! <https://github.com/cedar-policy/cedar-spec/blob/main/cedar-lean/Cedar/Spec/Ext/Datetime.lean>
 
-use std::{num::ParseIntError, str::FromStr};
+use std::num::ParseIntError;
+use std::str::FromStr;
 
 use chrono::{DateTime, NaiveDate, NaiveDateTime, Utc};
 use miette::Diagnostic;


cedar-policy-symcc/src/symcc/bitvec.rs
```diff
@@ -56,10 +56,12 @@ pub enum BitVecError {
 type Result<T> = std::result::Result<T, BitVecError>;
 
 impl BitVec {
+    /// Converts an (unsigned) [`Nat`] into a [`BitVec`] of the given width.
     pub fn of_nat(width: Width, v: Nat) -> Result<Self> {
         BitVec::new(width, v)
     }
 
+    /// Converts a (signed) [`Int`] into a [`BitVec`] of the given width.
     pub fn of_int(width: u32, v: Int) -> Result<Self> {
```
cdisselkoen: noticed that one of these takes `width` as `u32` and the other as `Width`. Not this PR's fault obviously

cedar-policy-symcc/src/symcc/extension_types/decimal.rs
```diff
@@ -28,59 +28,64 @@ static DECIMAL_DIGITS: u32 = 4;
...
-                            l_prime + r_prime
-                        } else {
-                            l_prime - r_prime
-                        };
-                        match i.try_into() {
-                            Ok(i) => Some(Decimal(i)),
-                            Err(_) => None,
+impl Decimal {
+    /// Parses a [`Decimal`] from a string.
+    pub fn parse(s: &str) -> Option<Decimal> {
```
cdisselkoen: while we're renaming this to `Decimal::parse()`, we really should consider just making it an imlpementation of `std::str::FromStr` instead

cedar-policy-symcc/src/symcc/extension_types/ipaddr.rs
```diff
@@ -374,19 +395,26 @@ impl IPNet {
...
         self.in_range(match self {
             IPNet::V4(_) => &MULTICAST_CIDR_V4,
             IPNet::V6(_) => &MULTICAST_CIDR_V6,
         })
     }
+
+    /// Parses an [`IPNet`] from a string.
+    pub fn parse(s: &str) -> Option<IPNet> {
+        parse_ipv4_net(s).or_else(|| parse_ipv6_net(s))
+    }
```
cdisselkoen: same with this.

cdisselkoen (approved):
Looks great, this is awesome. Thanks for this PR.

cedar-policy-symcc/src/lib.rs
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#![warn(missing_docs)]
```
chaluli, line 17: Nice, it's good to be warned if we are missing a doc string.

cedar-policy-symcc/src/symcc/bitvec.rs
chaluli: It looks like we consistently use Width in many of the other functions. We should probably make the change here to `Width`.

If you would, the change would be much appreciated! :)

chaluli (approved):
Looks like a great change! Thanks for this!

zhengyao-lin:
Thanks for the suggestions! I replaced the `parse` functions with `FromStr` impls and changed the `of_int` argument to `Width`.

cdisselkoen (approved):

PR #1798, opened 2025-08-21
Fix SymCC parsing of `duration` function application

shaobo-he-aws:

### DIFF
--- a/cedar-policy-symcc/src/symcc/compiler.rs
+++ b/cedar-policy-symcc/src/symcc/compiler.rs
@@ -973,7 +973,7 @@ mod datetime_tests {
         assert!(
             matches!(
                 compile(&datetime_lit(str), &sym_env),
-                Err(CompileError::TypeError),
+                Err(CompileError::ExtError(_)),
             ),
             "{msg}"
         );
--- a/cedar-policy-symcc/src/symcc/compiler.rs
+++ b/cedar-policy-symcc/src/symcc/compiler.rs
@@ -422,12 +422,12 @@ pub fn compile_record(ats: Vec<(Attr, Term)>) -> Result<Term> {
     ))
 }
 
-pub fn compile_call0(mk: impl Fn(String) -> Option<Ext>, arg: Term) -> Result<Term> {
+pub fn compile_call0(mk: impl Fn(String) -> Option<Ext>, arg: Term, fn_name: &str) -> Result<Term> {
     match arg {
         Term::Some(t) => match Arc::unwrap_or_clone(t) {
             Term::Prim(TermPrim::String(s)) => match mk(s) {
                 Some(v) => Ok(some_of(v.into())),
-                None => Err(CompileError::TypeError),
+                None => Err(CompileError::ExtensionFunctionApp(fn_name.to_owned())),
             },
             _ => Err(CompileError::TypeError),
         },
@@ -515,7 +515,7 @@ pub fn compile_call(xfn: &cedar_policy_core::ast::Name, ts: Vec<Term>) -> Result
     match (xfn.to_string().as_str(), ts.len()) {
         ("decimal", 1) => {
             let t1 = extract_first(ts);
-            compile_call0(Ext::parse_decimal, t1)
+            compile_call0(Ext::parse_decimal, t1, "decimal")
         }
         ("lessThan", 2) => {
             let (t1, t2) = extract_first2(ts);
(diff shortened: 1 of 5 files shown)


shaobo-he-aws:
@zhengyao-lin are you also interesting in fixing this issue?

zhengyao-lin:
Yes, happy to take a look tonight

zhengyao-lin:
Maybe we can just use `i64::try_from(digits.parse::<i128>()?)?` here instead of `parse::<i64>`? (this is where the error came from, I think) https://github.com/cedar-policy/cedar/blob/1b0f697841eaea87bd7e24bc7887068f3f1ee6da/cedar-policy-symcc/src/symcc/extension_types/datetime.rs#L325

`cedar-policy-core` seems to be using i128 when parsing the `ms` part of the duration: https://github.com/cedar-policy/cedar/blob/1b0f697841eaea87bd7e24bc7887068f3f1ee6da/cedar-policy-core/src/extensions/datetime.rs#L492

Lean model [uses `Nat`](https://github.com/cedar-policy/cedar-spec/blob/c68f13028ac563508df4ac6dc71ba89f57b26454/cedar-lean/Cedar/Spec/Ext/Datetime.lean#L230) so there's no overflow.

Not sure which option is the best

cedar-policy-symcc/src/symcc/compiler.rs
```diff
@@ -793,12 +794,11 @@ mod decimal_tests {
...
-            matches!(
-                compile(&dec_lit(str), &sym_env),
-                Err(CompileError::TypeError),
-            ),
-            "{msg}"
+        assert_matches!(
+            compile(&dec_lit(str), &sym_env),
+            Err(CompileError::ExtError(_)),
+            "{}",
+            msg
```
cdisselkoen, line 801:
```suggestion
            "{msg}",
```

cedar-policy-symcc/src/symcc/compiler.rs
```diff
@@ -970,12 +970,11 @@ mod datetime_tests {
...
-            matches!(
-                compile(&datetime_lit(str), &sym_env),
-                Err(CompileError::TypeError),
-            ),
-            "{msg}"
+        assert_matches!(
+            compile(&datetime_lit(str), &sym_env),
+            Err(CompileError::ExtError(_)),
+            "{}",
+            msg
```
cdisselkoen, line 977:
```suggestion
            "{msg}",
```

cedar-policy-symcc/src/symcc/compiler.rs
```diff
@@ -993,12 +992,11 @@ mod datetime_tests {
...
-            matches!(
-                compile(&duration_lit(str), &sym_env),
-                Err(CompileError::TypeError),
-            ),
-            "{msg}"
+        assert_matches!(
+            compile(&duration_lit(str), &sym_env),
+            Err(CompileError::ExtError(_)),
+            "{}",
+            msg
```
cdisselkoen, line 999:
```suggestion
            "{msg}",
```

cedar-policy-symcc/src/symcc/extension_types/datetime.rs
```diff
@@ -322,13 +322,11 @@ impl Duration {
                     i64::MIN
                 } else {
                     // Parse absolute value. Any remaining overflow / parse errors
-                    let abs_val = digits.parse::<i64>()?;
+                    let abs_val = digits.parse::<i128>()?;
```
cdisselkoen: should we parse as `u128`, since we've already processed the sign?

cdisselkoen (approved):

cedar-policy-symcc/src/symcc/extension_types/datetime.rs
shaobo-he-aws: Yes. Changed to a version that looks more like Lean.

cedar-policy-symcc/src/symcc/compiler.rs
shaobo-he-aws: We unfortunately can't do this because of how this macro is designed.

cedar-policy-symcc/src/symcc/compiler.rs
shaobo-he-aws: Please see above.

cdisselkoen (approved):

chaluli (approved):

PR #1800, opened 2025-08-22
SymCC: Fix a cex generation bug

### LINKED ISSUE #1799: Spurious counterexamples generated by SymCC

### Bug Category

Other

### Describe the bug

SymCC produces a spurious counterexample when checking always denies on the following schema and policy,

```
Schema: {
      "entityTypes": {
        "a": {
          "shape": {
            "type": "Record",
            "attributes": {
              "r": {
                "type": "Boolean"
              }
            }
          },
          "tags": {
            "type": "Extension",
            "name": "datetime"
          }
        }
      },
      "actions": {
        "action": {
          "appliesTo": {
            "resourceTypes": [
              "a"
            ],
            "principalTypes": [
              "a"
            ]
          },
          "memberOf": []
        }
      }
    }

Policies: permit(
      principal,
      action in [Action::"action"],
      resource
    ) when {
      ((true && (a::"*"["r"])) && true) && true
    };
```

The entities of the counterexample is as follows,
```
entities:
a::"": {"r": true}
Action::"action"
```

It doesn't include the entity literal `a::"*"`. Lean was able to give a correct counterexample.

### Expected behavior

N/A

### Reproduction steps

N/A

### Code Snippet

```rust
// Put your code below this line.

```

### Log output

```
// Put your output below this line

```

### Additional configuration

_No response_

### Operating  System

_No response_

_No response_

zhengyao-lin:
## Description of changes

Fixes #1799, which was due to incorrect EUID collection during concretization: https://github.com/cedar-policy/cedar/blob/1b0f697841eaea87bd7e24bc7887068f3f1ee6da/cedar-policy-symcc/src/symcc/concretizer.rs#L427

This produced an empty footprint in the case of #1799, because `a::"*"["r"]` compiles to a literal `true` term. However, we do need the `a::"*"` entity for a well-formed counterexample.

I changed it to match the Lean model (which is almost always correct, unsurprisingly), so that the EUIDs are directly collected from the [`Expr` without going through `footprint`](https://github.com/cedar-policy/cedar-spec/blob/c68f13028ac563508df4ac6dc71ba89f57b26454/cedar-lean/Cedar/SymCC/Concretizer.lean#L234).

I also did some more refactoring to add an `extractor` module to match the latest version of the Lean model.

### DIFF
--- a/cedar-policy-symcc/src/symcc/concretizer.rs
+++ b/cedar-policy-symcc/src/symcc/concretizer.rs
@@ -420,9 +420,9 @@ impl SymEntities {
 /// An [`ExprVisitor`] to collect all entity UIDs occurring in an expression.
 ///
 /// Corresponds to `Expr.entityUIDs` in `Concretize.lean`.
-struct GetAllUIDsVisitor<'a>(&'a mut BTreeSet<EntityUid>);
+struct EntityUIDCollector<'a>(&'a mut BTreeSet<EntityUid>);
 
-impl ExprVisitor for GetAllUIDsVisitor<'_> {
+impl ExprVisitor for EntityUIDCollector<'_> {
     type Output = ();
 
     fn visit_literal(&mut self, lit: &Literal, _: Option<&Loc>) -> Option<Self::Output> {
@@ -449,7 +449,7 @@ impl SymEnv {
         // Instead of using `footprint` and `Term::get_all_entity_uids`,
         // we collect EUIDs in expressions directly to avoid incorrect
         // short-circuiting in an incomplete entity store.
-        let mut visitor = GetAllUIDsVisitor(&mut uids);
+        let mut visitor = EntityUIDCollector(&mut uids);
         for expr in exprs {
             visitor.visit_expr(expr);
         }


shaobo-he-aws:
I ran DRT on your branch and discovered an error. Cex generation failed for the following schema and policy,

```
Schema: {
"entityTypes": {
"a": { "shape": {  "type": "Record", "attributes": { "ZB": { "type": "Record", "attributes": {}, "required": false }
            }
          }
        }
      },
      "actions": {
        "action": {
          "appliesTo": {
            "resourceTypes": [
              "a"
            ],
            "principalTypes": [
              "a"
            ]
          },
          "memberOf": []
        }
      }
    }
 Policies: @r("\u{1a}\u{19}\u{19}")
    @r00000000("")
    permit(
      principal,
      action in [Action::"action"],
      resource
    ) when {
      true && (a::"" has "ZB")
    };
```

The error message is: `failed to decode model: Unknown SMT literal: R0`

shaobo-he-aws:
Lean also fails on the example above. Error message:

```
LeanBackendError(
        "IO error: Model decoding failed: expected enum id, given \"R0\"\n (\n(define-fun f0 ((_arg_1 E2)) R1 (R1 ((as some (Option R0)) R0)))\n)\n",
    )
```

Tagging @emina to confirm if it's expected.

zhengyao-lin:
Could be wrong but I believe it's because the decoder incorrectly handles empty record values (separate issue from this PR).
In the example, the empty record type in the schema is encoded as
```
; {}
(declare-datatype R0 (
  (R0)))
```
and the model output by cvc5 is
```
(
(define-fun f0 ((_arg_1 E2)) R1 (R1 ((as some (Option R0)) R0)))
)
```
where `R0` is directly used as a symbol as opposed to an application `(R0 ...)` like non-empty records.

In the case of a single symbol, both Lean and Rust decoders only consider it as [true/false/enum literal](https://github.com/cedar-policy/cedar-spec/blob/c68f13028ac563508df4ac6dc71ba89f57b26454/cedar-lean/Cedar/SymCC/Decoder.lean#L224-L226), hence not being able to recognize it as a record value.

Attempts to fix this issue: Rust #1801, Lean https://github.com/cedar-policy/cedar-spec/pull/721

cedar-policy-symcc/src/symcc/concretizer.rs
```diff
@@ -415,17 +417,86 @@ impl SymEntities {
...
-    pub fn concretize<'a>(
+    /// Corresponds to `Prim.entityUIDs` in `Concretize.lean`
+    fn get_all_entity_uids_in_literal(lit: &Literal, uids: &mut BTreeSet<EntityUid>) {
+        if let Literal::EntityUID(euid) = lit {
+            uids.insert(euid.as_ref().clone().into());
+        }
+    }
+
+    /// Corresponds to `Expr.entityUIDs` in `Concretize.lean`
+    fn get_all_entity_uids_in_expr(
```
cdisselkoen: Would it simplify things to define this in terms of the already-existing `ExprVisitor` (see [here](https://github.com/cedar-policy/cedar/blob/1b0f697841eaea87bd7e24bc7887068f3f1ee6da/cedar-policy-core/src/ast/expr_visitor.rs#L4))?

cedar-policy-symcc/src/symcc/extractor.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
+}
+
+impl Interpretation<'_> {
+    /// Our acyclicity constraints only apply to the footprint,
+    /// but SMT solver may choose to add additional elements
+    /// that introduce cycles in the ancestor functions.
+    ///
+    /// This function repairs that by mappinng all inputs in the
+    /// ancestor functions taht are not in the footprint to empty sets
```
cdisselkoen:
```suggestion
    /// ancestor functions that are not in the footprint to empty sets
```

cdisselkoen (approved):

cedar-policy-symcc/src/symcc/concretizer.rs
shaobo-he-aws: I vaguely recall that Andrew wrote something similar.

cdisselkoen: I think the one Andrew wrote is [this](https://docs.rs/cedar-policy/latest/cedar_policy/struct.Policy.html#method.entity_literals)?  For [reasons](https://github.com/cedar-policy/cedar/pull/1648#discussion_r2124731064), that's not implemented in terms of ExprVisitor.  We could use that here too, though, not sure which is better

shaobo-he-aws (approved):

PR #1812, opened 2025-09-03
Batched Evaluation

oflatt:
## Description of changes
This PR adds batched evaluation, a new way to run Cedar authorization without a concrete entity store. Instead, an `EntityLoader` is provided that loads entities in batches.
Besides batched evaluation, this PR adds support for batched evaluation by

- Changing the internals of TPE to match the Lean code. Use `Residual` instead of `Expr`
- Adding a new error variant to TPE for unsupported expressions
- Adding policy level computation to support testing. Batched evaluation should converge to a value with a maximum steps equal to the level of the policy.
- Added policy level computation using a simple algorithm

  Proof: https://github.com/cedar-policy/cedar-spec/pulls?q=is%3Apr+batched+is%3Aclosed
  Algorithm: https://github.com/cedar-policy/cedar-spec/pull/681

### DIFF
--- a/cedar-policy-core/src/tpe/response.rs
+++ b/cedar-policy-core/src/tpe/response.rs
@@ -83,7 +83,7 @@ impl From<ResidualPolicy> for Policy {
 // This struct is akin is to PE's `PartialResponse`
 #[derive(Debug, Clone)]
 pub struct Response<'a> {
-    decison: Option<Decision>,
+    decision: Option<Decision>,
     residuals: HashMap<PolicyID, ResidualPolicy>,
     // All of the [`Effect::Permit`] policies that were satisfied
     satisfied_permits: HashSet<PolicyID>,
@@ -99,16 +99,15 @@ pub struct Response<'a> {
     non_trivial_forbids: HashSet<PolicyID>,
     // request used for this partial evaluation
     request: &'a PartialRequest,
-    // TODO we could make this owned, but would require a clone
-    // entities used for this partial evaluation, if any
-    // this is [`None`] for batched evaluation
+    // entities used for this partial evaluation
     entities: &'a PartialEntities,
     // schema
     schema: &'a ValidatorSchema,
 }
 
 impl<'a> Response<'a> {
-    /// Construct a [`Response`] from an iterator of [`ResidualPolicy`]s
+    /// Construct a [`Response`] from an iterator of [`ResidualPolicy`]s.
+    /// Guaranteed to arrive at a [`Decision`] if all the residuals are not [`Residual::Partial`]
     pub fn new(
         residuals: impl Iterator<Item = ResidualPolicy>,
         request: &'a PartialRequest,
@@ -118,10 +117,10 @@ impl<'a> Response<'a> {
         let mut residual_map = HashMap::new();
         let mut satisfied_permits = HashSet::new();
         let mut false_permits = HashSet::new();
-        let mut non_trivial_permits = HashSet::new();
+        let mut residual_permits = HashSet::new();
         let mut satisfied_forbids = HashSet::new();
         let mut false_forbids = HashSet::new();
-        let mut non_trivial_forbids = HashSet::new();
+        let mut residual_forbids = HashSet::new();
         for rp in residuals {
             let r = rp.get_residual();
             let id = rp.get_policy_id();
@@ -130,55 +129,51 @@ impl<'a> Response<'a> {
                 Effect::Forbid => {
                     if r.is_true() {
                         satisfied_forbids.insert(id);
-                    } else if r.is_false() {
+                    } else if r.is_false() || r.is_error() {
                         false_forbids.insert(id);
                     } else {
-                        non_trivial_forbids.insert(id);
+                        residual_forbids.insert(id);
                     }
                 }
                 Effect::Permit => {
                     if r.is_true() {
                         satisfied_permits.insert(id);
-                    } else if r.is_false() {
+                    } else if r.is_false() || r.is_error() {
                         false_permits.insert(id);
                     } else {
-                        non_trivial_permits.insert(id);
+                        residual_permits.insert(id);
                     }
                 }
             }
         }
-        let decison = if !satisfied_forbids.is_empty() {
-            // there are satsified forbid policies, the decision must be a deny
-            Some(Decision::Deny)
-        } else if !non_trivial_forbids.is_empty() {
-            // satisfied_forbids.is_empty && !non_trivial_forbids.is_empty()
-            // there are residual forbid policies, we can't make any conclusive
-            // authorization answer
-            None
-        } else if !satisfied_permits.is_empty() {
-            // satisfied_forbids.is_empty && non_trivial_forbids.is_empty() && !satisfied_permits.is_empty()
-            // all forbid policies are unsatisified, as long as there's one
-            // permit policy, we can give an allow decision
-            Some(Decision::Allow)
-        } else if non_trivial_permits.is_empty() {
-            // satisfied_forbids.is_empty && non_trivial_forbids.is_empty() && satisfied_permits.is_empty() && non_trivial_permits.is_empty()
-            // there's no satisifed permit and no permit residual, in other
-            // words, all permit residual policies are false, we can give a deny
-            // decision
-            Some(Decision::Deny)
-        } else {
-            // fallback option
-            None
+
+        let decision = match (
+            !satisfied_forbids.is_empty(),
+            !satisfied_permits.is_empty(),
+            !residual_permits.is_empty(),
+            !residual_forbids.is_empty(),
+        ) {
+            // Any true forbids means we will deny
+            (true, _, _, _) => Some(Decision::Deny),
+            // No potentially or trivially true permits, means we default deny
+            (_, false, false, _) => Some(Decision::Deny),
+            // Potentially true forbids, means we can't know (as that forbid may evaluate to true, overriding any permits)
+            (false, _, _, true) => None,
+            // No true permits, but some potentially true permits + no true/potentially true forbids means we don't know
+            (false, false, true, false) => None,
+            // At least one trivially true permit, and no trivially or possible true forbids, means we allow
+            (false, true, _, false) => Some(Decision::Allow),
         };
+
         Self {
-            decison,
+            decision,
             residuals: residual_map,
             satisfied_permits,
             false_permits,
-            non_trivial_permits,
+            non_trivial_permits: residual_permits,
             satisfied_forbids,
             false_forbids,
-            non_trivial_forbids,
+            non_trivial_forbids: residual_forbids,
             request,
             entities,
             schema,
@@ -222,7 +217,7 @@ impl<'a> Response<'a> {
 
     /// Attempt to get the authorization decision
     pub fn decision(&self) -> Option<Decision> {
-        self.decison
+        self.decision
     }
 
     /// Perform reauthorization
--- a/cedar-policy-core/src/authorizer/partial_response.rs
+++ b/cedar-policy-core/src/authorizer/partial_response.rs
@@ -216,7 +216,7 @@ impl PartialResponse {
 
     /// Returns the set of non-trivial (meaning more than just `true` or `false`) residuals expressions
     pub fn nontrivial_residuals(&'_ self) -> impl Iterator<Item = Policy> + '_ {
-        self.nontrival_permits().chain(self.nontrival_forbids())
+        self.nontrivial_permits().chain(self.nontrivial_forbids())
     }
 
     /// Returns the set of ids of non-trivial (meaning more than just `true` or `false`) residuals expressions
@@ -227,7 +227,7 @@ impl PartialResponse {
     }
 
     /// Returns the set of non-trivial (meaning more than just `true` or `false`) residuals expressions from [`Effect::Permit`]
-    fn nontrival_permits(&self) -> impl Iterator<Item = Policy> + '_ {
+    fn nontrivial_permits(&self) -> impl Iterator<Item = Policy> + '_ {
         self.residual_permits
             .iter()
             .map(|(id, (expr, annotations))| {
@@ -236,7 +236,7 @@ impl PartialResponse {
     }
 
     /// Returns the set of non-trivial (meaning more than just `true` or `false`) residuals expressions from [`Effect::Forbid`]
-    pub fn nontrival_forbids(&self) -> impl Iterator<Item = Policy> + '_ {
+    pub fn nontrivial_forbids(&self) -> impl Iterator<Item = Policy> + '_ {
         self.residual_forbids
             .iter()
             .map(|(id, (expr, annotations))| {


oflatt:
Does adding new error variants constitute a breaking change?

cdisselkoen:
> Does adding new error variants constitute a breaking change?

- If the error enum is not publicly exported in `cedar-policy`, no.  Most enums that are publicly exported in `cedar-policy` have a comment saying so at the top of the enum.  You can confirm via docs.rs/cedar-policy
- If the error enum is marked `#[non_exhaustive]`, no.
- Otherwise yes

cedar-policy-core/src/validator.rs
```diff
@@ -155,6 +155,34 @@ impl Validator {
         )
     }
 
+    /// Calculate the minimum level at which the policy set will
+    /// validate. Does this very simply as a loop starting from 0.
+    ///
+    /// Returns a [`ValidationResult`] which contains errors if there are any type errors.
+    ///
+    // PANIC SAFETY: Cedar programs are less than u32::MAX in size
```
oflatt: Is this an okay assumption?

cedar-policy-core/src/tpe/response.rs
```diff
@@ -92,10 +97,14 @@ pub struct Response<'a> {
...
-    // entities
-    entities: &'a PartialEntities,
+    // TODO we could make this owned, but would require a clone
+    // request used for this partial evaluation
+    // this is [`None`] for batched evaluation
+    request: Option<&'a PartialRequest>,
+    // TODO we could make this owned, but would require a clone
+    // entities used for this partial evaluation, if any
+    // this is [`None`] for batched evaluation
+    entities: Option<&'a PartialEntities>,
```
oflatt: This is awkward, but I didn't want to introduce a clone here. It's used for sanity checking.

cedar-policy/src/api.rs
```diff
@@ -5299,9 +5313,9 @@ mod tpe {
     /// A partial [`Response`]
     #[repr(transparent)]
     #[derive(Debug, Clone, RefCast)]
-    pub struct Response<'a>(pub(crate) tpe::response::Response<'a>);
+    pub struct TPEResponse<'a>(pub(crate) tpe::response::Response<'a>);
```
oflatt, line 5310: I did this rename because it seems like `Response` was already taken in our api. I think this is okay because tpe is still experimental.

I'm not sure how people have been importing this `Reponse` currently

cdisselkoen: Yes, breaking changes are acceptable for items gated by an experimental feature flag

oflatt:
Looks like integration tests are not counted in the coverage, so I'll add more testing.

shaobo-he-aws:
QQ: Lean's `PartialEntityData` can be present or absent. Do we need the same in Rust?

oflatt:
Actually I realized it isn't necessary and I'll be removing it from the Lean side. PR coming soon

oflatt:
PR here: https://github.com/cedar-policy/cedar-spec/pull/736

cedar-policy-core/src/tpe.rs
```diff
@@ -70,26 +70,183 @@ pub fn is_authorized<'a>(
...
-        entities,
+        residuals?.into_iter(),
+        Some(request),
+        Some(entities),
         schema,
     ))
 }
 
+/// Internal version of [`EntityLoader`]
+pub trait EntityLoaderInternal {
```
shaobo-he-aws: Nit: Let's call it `EntityLoader` without internal.

cedar-policy-core/src/tpe.rs
```diff
@@ -70,26 +70,183 @@ pub fn is_authorized<'a>(
...
+    };
+
+    Ok(PartialRequest::new(
+        principal, action, resource, context, schema,
+    )?)
+}
+
+/// Perform authorization using loader function instead
+/// of an [`Entities`] store.
+pub fn is_authorized_batched<'a>(
```
shaobo-he-aws: Let's move it along with relevant definitions to a dedicated module.

shaobo-he-aws:
I wonder if it makes sense to make batched evaluation as an experimental feature that depends on `tpe`.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -39,27 +40,31 @@ pub struct Evaluator<'e> {
...
-            .data()
-            .clone()
-            .expect("type checked should provide a type");
-        match e.expr_kind() {
-            ExprKind::Lit(l) => Residual::Concrete {
-                value: l.clone().into(),
-                ty,
-            },
-            ExprKind::Var(Var::Action) => Residual::Concrete {
+    pub fn interpret(&self, r: &Residual) -> Residual {
```
shaobo-he-aws, line 52: Does it make sense to take an owned `Residual` here? It will save us a lot of clones.

cedar-policy-core/src/tpe/residual.rs
```diff
@@ -60,6 +67,115 @@ impl Residual {
...
+        match self {
+            Residual::Partial { ty, .. } => ty,
+            Residual::Concrete { ty, .. } => ty,
+            Residual::Error(ty) => ty,
+        }
+    }
+
+    /// Convert an expression to a residual, returning an error if it can't be converted
+    /// due to slots, unknowns, or missing type annotations.
+    pub fn from_expr(expr: &Expr<Option<Type>>) -> std::result::Result<Self, ExprToResidualError> {
```
shaobo-he-aws: Implement it as `TryFrom`?

cedar-policy-core/src/validator.rs
```diff
@@ -155,6 +155,33 @@ impl Validator {
...
+    #[allow(clippy::panic)]
+    pub fn calculate_minimum_level<'a>(
```
shaobo-he-aws: We should optimize this function. A low-hanging fruit is to find the maximum reference level by traversing the AST. Tagging @john-h-kastner-aws

oflatt: I think if we optimize this function we need to prove it's correct in lean as well. I'll leave that as a future task.

oflatt: Also does this handle literals in the program?
The lean version doesn't handle literals I think

oflatt:
> I wonder if it makes sense to make batched evaluation as an experimental feature that depends on `tpe`.

I'd prefer not to, since it's already proven correct and new feature flags are a pain. Happy to do it if it's necessary

cedar-policy-core/src/tpe/evaluator.rs
oflatt: I think the `Arc<Residual>` children of residual make it so that a reference makes more sense, otherwise we end up doing a lot of copying as we traverse.

cedar-policy-core/src/tpe/residual.rs
oflatt: good idea

cedar-policy-core/src/validator.rs
oflatt: Ah I see that it errors on literals. I think that's okay.

cedar-policy-core/src/validator.rs
john-h-kastner-aws: Would `usize::Max` be a better assumption?

cedar-policy-core/src/validator.rs
john-h-kastner-aws: regarding algorithm: `check_entity_deref_target_level` actually implements this already in order to give nicer error message (`this policy requires level 3, which exceeds the maximum allowed level (1)`)

cedar-policy-core/src/validator.rs
john-h-kastner-aws: but, also, I see you've removed the need for this comment

cedar-policy-core/src/ast/value.rs
```diff
@@ -196,6 +201,33 @@ impl ValueKind {
             _ => None,
         }
     }
+
+    /// All literal uids referenced by this value kind
+    pub fn all_literal_uids(&self) -> HashSet<EntityUID> {
+        match self {
+            ValueKind::Lit(Literal::EntityUID(uid)) => {
+                let mut uids = HashSet::new();
+                uids.insert((**uid).clone());
```
john-h-kastner-aws: If this is an `Arc` i think would work and is nicer `uid.as_ref().clone()` is nicer

cedar-policy-core/src/ast/value.rs
```diff
@@ -196,6 +201,33 @@ impl ValueKind {
             _ => None,
         }
     }
+
+    /// All literal uids referenced by this value kind
+    pub fn all_literal_uids(&self) -> HashSet<EntityUID> {
+        match self {
+            ValueKind::Lit(Literal::EntityUID(uid)) => {
+                let mut uids = HashSet::new();
```
john-h-kastner-aws: I would prefer `HashSet::from([$elem])`

cedar-policy-core/src/ast/value.rs
```diff
@@ -196,6 +201,33 @@ impl ValueKind {
...
+                uids.insert((**uid).clone());
+                uids
+            }
+            ValueKind::Lit(_) => HashSet::new(),
+            ValueKind::Set(set) => {
+                let mut uids = HashSet::new();
```
john-h-kastner-aws: `set.iter().flat_map(|v| v.all_literal_uid()).collect()`

cedar-policy-core/src/tpe/batched_evaluator.rs
```diff
@@ -0,0 +1,212 @@
...
+use crate::validator::ValidatorSchema;
+use crate::{ast::PolicySet, extensions::Extensions};
+
+use crate::tpe::{entities::PartialEntities, evaluator::Evaluator};
+
+/// Loads entities based on their EntityUID.
+/// See the public version in `api.rs`
+pub trait EntityLoader {
+    /// Load all entities for the given set of entity UIDs.
+    /// Returns a map from [`EntityUID`] to Option<Entity>, where `None` indicates
```
john-h-kastner-aws:
```suggestion
    /// Returns a map from [`EntityUID`] to `Option<Entity>`, where `None` indicates
```

cedar-policy-core/src/tpe/batched_evaluator.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// the entity does not exist.
+    fn load_entities(
+        &mut self,
+        uids: &std::collections::HashSet<EntityUID>,
+    ) -> std::collections::HashMap<EntityUID, Option<Entity>>;
```
john-h-kastner-aws: any reason not to `use` these?

cedar-policy-core/src/tpe/batched_evaluator.rs
```diff
@@ -0,0 +1,212 @@
...
+    let mut residuals = residuals_res?;
+
+    // PANIC SAFETY: residuals and policy set contain the same policy ids
+    #[allow(clippy::unwrap_used)]
+    for _i in 0..max_iters {
+        let ids = residuals.iter().flat_map(|r| r.all_literal_uids());
+        let mut to_load = HashSet::new();
+        // filter to_load for already loaded entities
+        for uid in ids {
+            if !entities.entities.contains_key(&uid) {
```
john-h-kastner-aws, line 158: Could be a `filter` and then `collect` to build the map

cedar-policy-core/src/tpe/batched_evaluator.rs
```diff
@@ -0,0 +1,212 @@
...
+                to_load.insert(uid);
+            }
+        }
+        // Subtle: missing entities are equivalent empty entities in both normal and partial evaluation.
+        let loaded_entities = loader.load_entities(&to_load);
+
+        // check that all requested entities were loaded and return error otherwise
+
+        let mut loaded_partial = vec![];
+        let mut loaded_empty_partial = vec![];
```
john-h-kastner-aws: Why do we need to construct this intermediate vecs and then add to `entities` rather than inserting into `entities` directly as we iterate?

cedar-policy-core/src/tpe/entities.rs
```diff
@@ -525,6 +531,68 @@ impl PartialEntities {
...
+
+    /// Add a set of partial entities to this store,
+    /// erroring on duplicates.
+    pub fn add_entities(
+        &mut self,
+        entity_mappings: impl Iterator<Item = (EntityUID, PartialEntity)>,
+        schema: &ValidatorSchema,
+        tc_computation: TCComputation,
+    ) -> std::result::Result<(), EntitiesError> {
+        for (uid, entity) in entity_mappings {
```
john-h-kastner-aws: Is this loop just `add_entities_trusted`?

cedar-policy-core/src/tpe/err.rs
```diff
@@ -295,6 +366,20 @@ pub struct UnknownEntityError {
     pub(super) uid: EntityUID,
 }
 
+/// Error thrown when some requested entities were not loaded
+#[derive(Debug, Error)]
+#[error("Failed to load entities: {}", .missing_entities.iter().map(|uid| uid.to_string()).collect::<Vec<_>>().join(", "))]
```
john-h-kastner-aws: I'd probably just move this into an `impl std::fmt::Display`

cedar-policy-core/src/tpe/evaluator.rs
john-h-kastner-aws: Could this be an owned `Arc`? That would let us do `unwrap_or_clone`, saving the clone if the `Arc` is actually unique

cedar-policy-core/src/tpe/residual.rs
```diff
@@ -72,6 +97,93 @@ impl TryFrom<Residual> for Value {
...
+
+        // Otherwise, convert to a partial residual
+        let kind = match expr.expr_kind() {
+            ast::ExprKind::Var(var) => ResidualKind::Var(*var),
+            ast::ExprKind::If {
+                test_expr,
+                then_expr,
+                else_expr,
+            } => ResidualKind::If {
+                test_expr: Arc::new(Self::try_from(&(**test_expr))?),
```
john-h-kastner-aws: Does `test_expr.as_ref()` work here? Also could potentially have this function take `Arc<..>` itself, depending on how it's called elsewhere.

cedar-policy-core/src/tpe/residual.rs
```diff
@@ -170,6 +282,61 @@ pub enum ResidualKind {
     Record(Arc<BTreeMap<SmolStr, Residual>>),
 }
 
+impl ResidualKind {
+    /// All literal uids referenced by this residual kind
+    pub fn all_literal_uids(&self) -> HashSet<EntityUID> {
```
john-h-kastner-aws, line 287: Am I missing something or will this never return any uids?

cedar-policy-core/src/tpe/batched_evaluator.rs
oflatt: I tried a filter and collect but had borrowing issues. The problem is that the closure needs a reference to entities.
Do you know how to make this work?

cedar-policy-core/src/tpe/batched_evaluator.rs
oflatt: Good point!

oflatt:
> Could this be an owned Arc then. That would let us do unwrap_or_clone, saving the clone if the Arc is actually unique

It's not really clear to me that this is faster- it introduces a lot of `Arc` clones as we traverse the residual so we can get an owned Arc. And in these cases, there is guaranteed to be multiple references to the Arc.

Perhaps a `&Arc<Residual>` is better? I'm not sure, but I think we may be in premature optimization territory. After all, residuals are small since their children are wrapped in `Arc`s.

cedar-policy-core/src/tpe/residual.rs
oflatt: `.as_ref()` works, thanks

Does having it take an `Arc<..>` buy us anything here? Sorry if I'm missing something

cedar-policy-core/src/tpe/residual.rs
oflatt: Values contain literal ids, so this is all to visit all the values and get the literal ids from them.

cedar-policy-core/src/tpe/entities.rs
oflatt: Good point

cedar-policy-core/src/tpe/batched_evaluator.rs
oflatt: nope, fixed

cedar-policy-core/src/tpe/batched_evaluator.rs
john-h-kastner-aws: I'd expect it to be able to implicitly capture the reference as long as you don't try to return the iterator, but this is fine if it tricky to get working

john-h-kastner-aws:
> > Could this be an owned Arc then. That would let us do unwrap_or_clone, saving the clone if the Arc is actually unique
> 
> It's not really clear to me that this is faster- it introduces a lot of `Arc` clones as we traverse the residual so we can get an owned Arc. And in these cases, there is guaranteed to be multiple references to the Arc.
> 
> Perhaps a `&Arc<Residual>` is better? I'm not sure, but I think we may be in premature optimization territory. After all, residuals are small since their children are wrapped in `Arc`s.

If you'd need to clone the arcs then, yeah, I expect `unwrap_or_clone` wouldn't help us. I don't think there's much reason to pass `&Arc` since it's cheap enough to unconditionally clone it if we need an owned Arc. So, probably not anything worth changing here unless we can push owned values up all up the stack

cedar-policy-core/src/tpe/response.rs
shaobo-he-aws: Why they are `None` for batched evaluation?

oflatt: We construct a new `PartialEntities` store, so I can't give a reference with a `'a` lifetime

shaobo-he-aws: I think maybe we should have a variant of `tpe` that batched evaluation can use? Batched evaluation doesn't need heavy lifting in `Response`'s constructor.

cedar-policy-core/src/tpe/batched_evaluator.rs
```diff
@@ -0,0 +1,206 @@
...
+
+/// Perform authorization using loader function instead
+/// of an [`Entities`] store.
+pub fn is_authorized_batched<'a>(
+    request: &Request,
+    ps: &PolicySet,
+    schema: &'a ValidatorSchema,
+    loader: &mut dyn EntityLoader,
+    max_iters: u32,
+) -> Result<Response<'a>, BatchedEvalError> {
```
shaobo-he-aws: I wonder if we should error on the case where we can't get a concrete response, as opposed to returning a partial response.

oflatt: edit: saw slack message, will return a concrete response

oflatt: Settled for returning a `Decision` for now instead of a `Response` since TPE doesn't give a good enough error message

shaobo-he-aws: Yes, it makes sense.

cedar-policy-core/src/tpe/response.rs
shaobo-he-aws: Can we restore it given that `is_authorized_batched` returns a decision now?

cedar-policy-core/src/tpe/response.rs
```diff
@@ -132,55 +129,51 @@ impl<'a> Response<'a> {
...
-            // satisfied_forbids.is_empty && non_trivial_forbids.is_empty() && satisfied_permits.is_empty() && non_trivial_permits.is_empty()
-            // there's no satisifed permit and no permit residual, in other
-            // words, all permit residual policies are false, we can give a deny
-            // decision
-            Some(Decision::Deny)
-        } else {
-            // fallback option
-            None
+
+        let decision = match (
```
john-h-kastner-aws, line 150: I don't really find this easier to read the the previous control flow. No concrete suggestions and not blocking the PR on this, but I think it could be improved

cedar-policy-core/src/tpe/response.rs
```diff
@@ -132,55 +129,51 @@ impl<'a> Response<'a> {
...
         };
+
         Self {
-            decison,
+            decision,
             residuals: residual_map,
             satisfied_permits,
             false_permits,
-            non_trivial_permits,
+            non_trivial_permits: residual_permits,
```
john-h-kastner-aws, line 173: does it make sense to apply the rename to the struct as well?

oflatt: yeah I think `residual` makes more sense

cedar-policy-core/src/tpe/response.rs
oflatt: I think it is more precise than the old version and matches the old PE code
Though it is ugly I agree

cedar-policy-core/src/tpe/response.rs
oflatt: (follow up pr)

Issue #1829, opened 2025-09-17
Stateful policy set and schema for WASM


MarcBridner:
### Category

User level API features/changes

### Describe the feature you'd like to request

Presently, when using the WASM version of Cedar the entire schema and policy set needs to sent along with each authorization request. In our case that is several megabytes of JSON that is created for each authorization request, leading to performance issues.

The suggested solution is to implement `stateful_is_authorized` that uses a cached copy of the schema and policy set.

A PR is on the way, and this solution has been discussed with @Swolebrain out of band.

### Describe alternatives you've considered

We thought about running the Rust authorizer natively in a Lambda layer to get a little bit more performance, but it seems more reasonable to cache the policy set and schema as a first step.

### Is this something that you'd be interested in working on?

- [x] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #1831, opened 2025-09-17
Stateful isAuthorization method

MarcBridner:
## Description of changes
Adds `stateful_is_authorized`, `preparse_policy_set` and `preparse_schema` that can be used to pre-load an immutable schema/policy set.

This way, we don't have to build a massive JSON struct on each authorization request.

In our tests where we perform approximately 50000 evaluations the execution time went from:

```
real    13m33.222s
user    132m49.959s
sys     1m43.130s
```
to
```
real    0m52.297s
user    7m55.203s
sys     1m8.939s
```

Or, in other words, a 15x performance increase.

## Issue #, if available
https://github.com/cedar-policy/cedar/issues/1829

### DIFF
--- a/cedar-policy/CHANGELOG.md
+++ b/cedar-policy/CHANGELOG.md
@@ -24,6 +24,7 @@ Starting with version 3.2.4, changes marked with a star (*) are _language breaki
 - Implemented variadic `ipaddr.isInRange` that returns true if the child `ipaddr` is in range for any of the arguments [RFC 99](github.com/cedar-policy/rfcs/pull/99)
 - Implemented batched evaluation under the experimental flag `tpe`. Batched evaluation allows for permission queries against large databases of entities. (#1812) 
 - Added `calculate_minimum_level` which computes the minimum level at which policies validate. (#1812)
+- Added `stateful_is_authorized`, `preparse_policy_set` and `preparse_schema` to support stateful evaluation using a cached policy set and schema, in the `ffi` module. (#1829)
 
 ### Changed
 
--- a/cedar-policy/src/ffi/is_authorized.rs
+++ b/cedar-policy/src/ffi/is_authorized.rs
@@ -18,6 +18,7 @@
 //! is generated from the [`is_authorized()`] function in this file.
 
 #![allow(clippy::module_name_repetitions)]
+use super::check_parse::CheckParseAnswer;
 #[cfg(feature = "partial-eval")]
 use super::utils::JsonValueWithNoDuplicateKeys;
 use super::utils::{Context, DetailedError, Entities, EntityUid, PolicySet, Schema, WithWarnings};
@@ -25,7 +26,6 @@ use crate::{Authorizer, Decision, PolicyId, Request};
 use cedar_policy_core::validator::cedar_schema::SchemaWarning;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
-#[cfg(feature = "partial-eval")]
 use std::collections::HashMap;
 use std::collections::HashSet;
 #[cfg(feature = "wasm")]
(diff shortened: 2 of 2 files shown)


cedar-policy/src/ffi/is_authorized.rs
```diff
@@ -21,12 +21,13 @@
...
 use super::utils::JsonValueWithNoDuplicateKeys;
 use super::utils::{Context, DetailedError, Entities, EntityUid, PolicySet, Schema, WithWarnings};
+use super::check_parse::CheckParseAnswer;
 use crate::{Authorizer, Decision, PolicyId, Request};
 use cedar_policy_core::validator::cedar_schema::SchemaWarning;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
-#[cfg(feature = "partial-eval")]
 use std::collections::HashMap;
+#[cfg(feature = "partial-eval")]
```
Swolebrain: Why'd we move this?

MarcBridner: I think that was an oopsie - can no longer use the the feature on hash-map as far as I can tell, since it will always be required. I thought that the HashSet was conditionally required, but it appears it was not - will update

cedar-policy/CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ Starting with version 3.2.4, changes marked with a star (*) are _language breaki
 - Implemented variadic `ipaddr.isInRange` that returns true if the child `ipaddr` is in range for any of the arguments [RFC 99](github.com/cedar-policy/rfcs/pull/99)
 - Implemented batched evaluation under the experimental flag `tpe`. Batched evaluation allows for permission queries against large databases of entities. (#1812) 
 - Added `calculate_minimum_level` which computes the minimum level at which policies validate. (#1812)
+- Added `stateful_is_authorized`, `preparse_policy_set` and `preparse_schema` to support stateful evaluation using a cached policy set and schema (#1829)
```
chaluli: We should note that this is specifically for the FFI.

chaluli (changes requested):
Please update the CHANGELOG to specify that this is a change in the FFI/wasm. Also, fix the formatting issues.

cedar-policy/CHANGELOG.md
MarcBridner: Updated

cedar-policy/src/ffi/is_authorized.rs
```diff
@@ -88,6 +98,44 @@ pub fn is_authorized_json_str(json: &str) -> Result<String, serde_json::Error> {
     serde_json::to_string(&ans)
 }
 
+/// Preparse and cache a policy set in thread-local storage
```
john-h-kastner-aws, line 101: Probably worth explicitly documenting that these don't have side effects on error.

cedar-policy/src/ffi/is_authorized.rs
```diff
@@ -605,6 +714,101 @@ impl AuthorizationCall {
     }
 }
 
+impl StatefulAuthorizationCall {
```
john-h-kastner-aws, line 726: This, `AuthorizationCall` and `PartialAuthorizationCall` are substantial similar. Wondering if there's any parsing logic we can share

cedar-policy/src/ffi/is_authorized.rs
```diff
@@ -524,6 +606,33 @@ pub struct PartialAuthorizationCall {
     entities: Entities,
 }
 
+/// Stateful authorization using preparsed schemas and policy sets
+/// This function works like is_authorized but retrieves schemas and policy sets
```
john-h-kastner-aws: Here and elsewhere, clippy is complaining about `is_authorized` without back-ticks (most clippy lints aren't errors, but we have some configured to be errors in public facing docs)

cedar-policy/src/ffi/is_authorized.rs
MarcBridner: Will do

cedar-policy/src/ffi/is_authorized.rs
MarcBridner: I saw in the clippy run - having trouble running it locally so I will update and hope for the best

cedar-policy/src/ffi/is_authorized.rs
MarcBridner: I went for the "rather have some duplication than break something existing" approach for this PR.
I'm not very familiar with the idiosyncrasies of Rust so I figured this'd be safer. I can give it a try if required.

john-h-kastner-aws:
Looks good. Merging!

PR #1834, opened 2025-09-18
Updated PolicySet to use LinkedHashMap

lianah:
## Description of changes

Updated ast::PolicySet, set::PolicySet and api::PolicySet to use LinkedHashMap and LinkedHashSet for deterministic iteration order.

### DIFF
--- a/cedar-policy/src/proto/api.rs
+++ b/cedar-policy/src/proto/api.rs
@@ -67,14 +67,6 @@ impl From<&api::Policy> for models::Policy {
     }
 }
 
-// impl TryFrom<&models::Policy> for api::Policy {
-//     type Error = cedar_policy_core::ast::ReificationError;
-//     fn try_from(v: &models::Policy) -> Result<Self, Self::Error> {
-//         let p = cedar_policy_core::ast::Policy::try_from(v)?;
-//         Ok(Self::from_ast(p))
-//     }
-// }
-
 impl From<&api::PolicySet> for models::PolicySet {
     fn from(v: &api::PolicySet) -> Self {
         Self::from(&v.ast)
--- a/cedar-policy-core/Cargo.toml
+++ b/cedar-policy-core/Cargo.toml
@@ -39,8 +39,6 @@ serde-wasm-bindgen = { version = "0.6", optional = true }
 tsify = { version = "0.4.5", optional = true }
 wasm-bindgen = { version = "0.2.97", optional = true }
 
-
-
 # datetime extension requires chrono
 chrono = { version = "0.4.42", optional = true, default-features = false}
 
--- a/cedar-policy/src/proto/policy.rs
+++ b/cedar-policy/src/proto/policy.rs
@@ -60,14 +60,6 @@ impl From<&models::Policy> for ast::LiteralPolicy {
     }
 }
 
-// impl TryFrom<&models::Policy> for ast::Policy {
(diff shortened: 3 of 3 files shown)


shaobo-he-aws:
I'm personally not very comfortable with the new dependencies two reasons.

1. They are not very popular, meaning much less tested.
2. Their major version number is still 0, meaning breaking changes can occur frequently.

Would `BTreeSet` or `BTreeMap` just work?

lianah:
>     1. They are not very popular, meaning much less tested.
I wouldn't say they are not popular, `linked_hash_map` has 150 million downloads, and over 300k in the last month. 

>     2. Their major version number is still 0, meaning breaking changes can occur frequently.
I doubt that there will be breaking changes, `linked_hash_map` is in maintenance mode meaning they only do bug fixes and don't implement new features. 

> Would `BTreeSet` or `BTreeMap` just work?

We discussed the option of `BTreeSet` or `BTreeMap` and there were concerns about performance. Happy to revisit that conversation.

adpaco:
> We discussed the option of `BTreeSet` or `BTreeMap` and there were concerns about performance. Happy to revisit that conversation.

And those concerns don't apply here? We already use `BTreeSet` and `BTreeMap` is cases where deterministic order is needed or preferred (see #887 for example).

Is there a reason for `LinkedHash*` structures to outperform `BTree*` ones in this case? Or have you compared their performance?

lianah:
> Is there a reason for `LinkedHash*` structures to outperform `BTree*` ones in this case? Or have you compared their performance?

Yes, there is an asymptotic complexity difference. `BTree*` has O(log n) complexity for search, insertion and deletion while `LinkedHash*` has the same O(1) complexity as `Hash*`. The only difference is that it maintains a linked list to preserve insertion order, so there might be a slight memory overhead.

cedar-policy/src/proto/api.rs
```diff
@@ -67,14 +67,6 @@ impl From<&api::Policy> for models::Policy {
     }
 }
 
-impl TryFrom<&models::Policy> for api::Policy {
```
adpaco, line 70: Why are these `protobuf` impls being deleted?

lianah: They are not being used and would panic if called (even before my changes). The reason is that they are trying to convert a `models::Policy` into a `api::Policy` by calling `reify` with an empty `templates` hash map [here](https://github.com/cedar-policy/cedar/blob/b3d9b87616a42716fdde7d93c0647e1b75d2bc0b/cedar-policy/src/proto/policy.rs#L67). This would just panic [here](https://github.com/cedar-policy/cedar/blob/main/cedar-policy-core/src/ast/policy.rs#L749) with a `NoSuchTemplate` error. The problem is the `models::Policy` only has the template id [here](https://github.com/cedar-policy/cedar/blob/0003f5faf898747ed840183b376a480cd0215a73/cedar-policy/protobuf_schema/core.proto#L54) so there is no way to get the policy body that is required for a `api::Policy` without the actual templates. The code that converts from `models::PolicySet` does [this conversion](https://github.com/cedar-policy/cedar/blob/0003f5faf898747ed840183b376a480cd0215a73/cedar-policy/src/proto/api.rs#L76) right. I noticed this because I was updating `reify` to pass in a `LinkedHashMap`.

PR #1838, opened 2025-09-23
Fix validation for action entitites in TPE

luxas:
## Description of changes

In the Cedar+Kubernetes integration, I noticed that I could not use TPE correctly. At first, I did not add the action entities to the entities list when running TPE, but then the residual contained things like `action in Action::"foo"`, which I did not want, as this data is in fact known at this point.

However, adding the actions to the entity list manually [here](https://github.com/upbound/kubernetes-cedar-authorizer/blob/main/src/cedar_authorizer/authorizer.rs#L414-L435) yielded an error, as `PartialEntity.validate` both [errors](https://github.com/cedar-policy/cedar/blob/22ec17872ca100485b235e6fe925b715619ad898/cedar-policy-core/src/tpe/entities.rs#L358-L381) if the action has `None` and `Some` attributes and tags, which I presume is a bug.

With this patch the k8s integration works as expected.

As a side point: it'd be nice to not have to do the extra work of adding the action entities at all, but let that be done automatically by TPE, as it anyways has the schema at hand.

FYI @shaobo-he-aws 

I'll make a unit test once you've confirmed removing this if statement was desired.

- [x] A bug fix that only impacts unreleased or experimental code.

Not sure if this is covered (or should be) by DRT.

### DIFF
--- a/cedar-policy-core/src/tpe/entities.rs
+++ b/cedar-policy-core/src/tpe/entities.rs
@@ -355,12 +355,7 @@ impl PartialEntity {
         let etype = uid.entity_type();
 
         if self.uid.is_action() {
-            if self.attrs.is_none() || self.tags.is_none() {
-                return Err(UnknownActionComponentError {
-                    action: uid.clone(),
-                }
-                .into());
-            }
+            // An Action entity cannot have attributes
             if let Some(attrs) = &self.attrs {
                 if let Some((attr, _)) = attrs.first_key_value() {
                     return Err(EntitySchemaConformanceError::unexpected_entity_attr(
@@ -370,6 +365,7 @@ impl PartialEntity {
                     .into());
                 }
             }
+            // An Action entity cannot have tags
             if let Some(tags) = &self.tags {
                 if let Some((tag, _)) = tags.first_key_value() {
                     return Err(EntitySchemaConformanceError::unexpected_entity_tag(
@@ -379,6 +375,7 @@ impl PartialEntity {
                     .into());
                 }
             }
+            // The Action must exist in the schema, and have the same ancestors as in the schema.
             if let Some(action) = core_schema.action(uid) {
                 if let Some(ancestors) = &self.ancestors {
                     let schema_ancestors: HashSet<EntityUID> =


cedar-policy-core/src/tpe/entities.rs
```diff
@@ -355,12 +355,7 @@ impl PartialEntity {
         let etype = uid.entity_type();
 
         if self.uid.is_action() {
-            if self.attrs.is_none() || self.tags.is_none() {
```
chaluli, line 358: I believe here `none` is used to denote unknown attributes / tags. Which is different from no attributes which would be represented as Some(Empty BTreeMap).

@shaobo-he-aws can confirm.

chaluli: In the code you linked [here](https://github.com/upbound/kubernetes-cedar-authorizer/blob/main/src/cedar_authorizer/authorizer.rs#L414-L435), if you changed lines 425 and 434 to be `Some(BTreeMap::new())` does it work as expected without this patch?

chaluli:
To your aside, I agree we should have a way to automatically add the action entities on a users behalf within TPE because the schema is known (perhaps as a separate API for constructing a partial entities).

cedar-policy-core/src/tpe/entities.rs
shaobo-he-aws: @chaluli is correct.

shaobo-he-aws:
@luxas sorry about the confusion. Like @chaluli said exactly, you will need to make empty attributes and empty tags in action entities. We should be able to add action entities automatically. I'll make a PR.

cedar-policy-core/src/tpe/entities.rs
luxas: I'll check. Don't remember of the top of my head, it's a month or so since I looked at this, I just put a note to myself to check this later

shaobo-he-aws:
> To your aside, I agree we should have a way to automatically add the action entities on a users behalf within TPE because the schema is known (perhaps as a separate API for constructing a partial entities).

As a matter of fact, actions are inserted in `PartialEntities::from_json_value` not in `PartialEntities::from_entities`. I'll update the latter.

chaluli:
Awesome. Thanks @shaobo-he-aws

luxas:
Awesome, https://github.com/cedar-policy/cedar/pull/1844 addresses this then 💯 
Thanks for the response, not sure why I got confused of the `None` vs `Some` when implementing that code a month ago, technically I did know `None` meant "unknown" and not "empty", as I had reviewed the TPE PR 😄
Now, let's move on to the next item I had on that list for small improvements 👍

Issue #1839, opened 2025-09-23
Make `construct_exprs_extended_has` in `cedar-policy-core/src/parser/cst_to_ast.rs` public


luxas:
### Describe the improvement you'd like to request

I have a use-case for vendoring the `construct_exprs_extended_has` function; it'd be nice if we could make it public, or re-export the functionality somewhere else, so I do not need to carry the functionality in my fork file (which indeed preserves the attribution and original license info).

What do you think? I realize that probably not too many people will need this functionality, but maybe someone else also finds it useful.

### Is this something that you'd be interested in working on?

- [x] 👋 I may be able to implement this internal improvement
- [ ] ⚠️ This feature might incur a breaking change

shaobo-he-aws:
Since we don't follow semver in `cedar-policy-core`, we're free to make such changes. I'll put up a PR.

shaobo-he-aws:
@luxas I added a method to `ExprBuilder` in #1842 that should accomplish something similar. Could you take a look and see if it works for you?

PR #1842, opened 2025-09-23
added a new method to expression builder to construct extended has expressions

shaobo-he-aws:
## Description of changes
Even though extended has expression doesn't have AST representation, we should offer an API to construct desugared expressions.
## Issue #, if available
#1839 


cedar-policy-core/src/expr_builder.rs
```diff
@@ -179,6 +180,12 @@ pub trait ExprBuilder: Clone {
     /// attribute on a given `Entity` or record.
     fn has_attr(self, expr: Self::Expr, attr: SmolStr) -> Self::Expr;
 
+    /// Create an `Expr` which tests for the existence of a given
+    /// non-empty list of attributes on a given `Entity` or record.
+    fn extended_has_attr(self, expr: Self::Expr, attrs: &NonEmpty<SmolStr>) -> Self::Expr {
+        construct_exprs_extended_has::<Self>(expr, attrs, None)
```
john-h-kastner-aws: Does it make more sense to just implement the function here and get rid of `construct_exprs_extended_has` in `cst_to_ast.rs`?

shaobo-he-aws: Not really. We'll possibly lose source location information in the constructed expressions.

john-h-kastner-aws: now that you mention source location, this should pass `self.loc()` instead of `None`.

Or, alternatively, if we implement it here, we can get the source location just by calling `self.loc()` instead of using the parameter

shaobo-he-aws: Done.

PR #1846, opened 2025-09-25
Better Error Handling for LocalSolver

katherine-hough:
## Description of changes
Improved error messages in `LocalSolver` when the solver process terminates unexpectedly or closes standard out.

### DIFF
--- a/cedar-policy-symcc/src/symcc/solver.rs
+++ b/cedar-policy-symcc/src/symcc/solver.rs
@@ -38,7 +38,7 @@ use std::future::Future;
 use std::process::Stdio;
 use thiserror::Error;
 use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader, BufWriter};
-use tokio::process::{ChildStderr, ChildStdin, ChildStdout, Command};
+use tokio::process::{Child, ChildStderr, ChildStdin, ChildStdout, Command};
 
 /// Satisfiability decision from the SMT solver.
 #[derive(Clone, Debug, PartialEq, Eq, Ord, PartialOrd)]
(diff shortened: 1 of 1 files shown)


cedar-policy-symcc/src/symcc/solver.rs
```diff
@@ -212,14 +220,26 @@ impl Solver for LocalSolver {
...
-                _ => Err(SolverError::UnrecognizedSolverOutput(output)),
-            },
+impl LocalSolver {
+    async fn process_error_output(&mut self, s: &str) -> SolverError {
+        match s
+            .strip_prefix("(error \"")
+            .and_then(|s| s.strip_suffix("\")\n"))
+        {
+            Some(e) => {
+                if e.starts_with("Parse Error: ") {
```
john-h-kastner-aws: What would be the behavior if cvc5 terminates on something without `Parse Error:` or prints `Parse Error:` and then doesn't terminate? I'm assuming we don't fully understand the exact behavior of cvc5 in all possible error conditions

john-h-kastner-aws: Also, do we support `z3` as a `LocalSolver`? That could also mess up these assumptions

katherine-hough: If cvc5 terminates on something without a Parse Error, we would probably see an I/O error (Broken pipe) when writing/flushing to ChildStdin. I could kill the process on a parse error instead of just waiting for termination, to force the issue. But yeah you make a good point; the more I look at the cvc5 documentation the less confident I am that we can rely on these behaviors.

I think I am going to rewrite this so that if we get an EOF on a read we return an error message indicating that. This will be generic but also more clear than the current error.

john-h-kastner-aws (approved):

shaobo-he-aws (approved):

PR #1857, opened 2025-09-29
Made LocalSolver's behavior after a parse error deterministic by killing the solver process

katherine-hough:
## Description of changes
Made `LocalSolver'`s behavior after a parse error deterministic by killing the solver process. This should fix observed flakiness in the test `symcc::solver::test::parse_error_test`.

## Issue #, if available
N/A

### DIFF
--- a/cedar-policy-symcc/src/symcc/solver.rs
+++ b/cedar-policy-symcc/src/symcc/solver.rs
@@ -176,12 +176,7 @@ impl Solver for LocalSolver {
     }
 
     async fn check_sat(&mut self) -> Result<Decision> {
-        if let Some(status) = self.child.try_wait()? {
-            Err(SolverError::Solver(format!(
-                "Solver process terminated unexpectedly with status: {:?}",
-                status.code()
-            )))?
-        }
+        self.check_child_process_status().await?;
         self.smtlib_input().check_sat().await?;
         self.solver_stdin.flush().await?;
         let mut output = String::new();
@@ -195,12 +190,7 @@ impl Solver for LocalSolver {
     }
 
     async fn get_model(&mut self) -> Result<Option<String>> {
-        if let Some(status) = self.child.try_wait()? {
-            Err(SolverError::Solver(format!(
-                "Solver process terminated unexpectedly with status: {:?}",
-                status.code()
-            )))?
-        }
+        self.check_child_process_status().await?;
         self.smtlib_input().get_model().await?;
         self.solver_stdin.flush().await?;
         let mut output = String::new();
@@ -228,15 +218,25 @@ impl Solver for LocalSolver {
 }
 
 impl LocalSolver {
+    async fn check_child_process_status(&mut self) -> Result<()> {
+        if let Some(status) = self.child.try_wait()? {
+            Err(SolverError::Solver(format!(
+                "Solver process terminated unexpectedly with status: {:?}",
+                status.code()
+            )))?
+        }
+        Ok(())
+    }
+
     async fn read_line(&mut self, buffer: &mut String) -> Result<usize> {
         let len = self.solver_stdout.read_line(buffer).await?;
         if len == 0 {
-            Err(SolverError::Solver(
-                "Encountered EOF while reading from solver output".to_string(),
-            ))
-        } else {
-            Ok(len)
+            // An unexpected EOF was encountered while reading from solver output
+            // Kill the child process and clean up its resources
+            self.clean_up().await?;
+            self.check_child_process_status().await?;
         }
+        Ok(len)
     }
 
     async fn process_error_output(s: &str) -> SolverError {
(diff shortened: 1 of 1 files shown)


cedar-policy-symcc/src/symcc/solver.rs
```diff
@@ -190,17 +185,12 @@ impl Solver for LocalSolver {
             "sat\n" => Ok(Decision::Sat),
             "unsat\n" => Ok(Decision::Unsat),
             "unknown\n" => Ok(Decision::Unknown),
-            s => Err(Self::process_error_output(s).await),
+            s => Err(Self::process_error_output(self, s).await),
```
chaluli: Is there a reason why this isn't `self.process_error_output(s).await`?

cedar-policy-symcc/src/symcc/solver.rs
```diff
@@ -190,17 +185,12 @@ impl Solver for LocalSolver {
...
     }
 
     async fn get_model(&mut self) -> Result<Option<String>> {
-        if let Some(status) = self.child.try_wait()? {
-            Err(SolverError::Solver(format!(
-                "Solver process terminated unexpectedly with status: {:?}",
-                status.code()
-            )))?
-        }
+        self.check_child_process_status().await?;
```
chaluli, line 193: good code abstraction. :)

cedar-policy-symcc/src/symcc/solver.rs
```diff
@@ -222,29 +212,46 @@ impl Solver for LocalSolver {
                 }
                 Ok(Some(output))
             }
-            s => Err(Self::process_error_output(s).await),
+            s => Err(Self::process_error_output(self, s).await),
```
chaluli: Similar question to above. Why not `self.proces_error_output(s)`?

cedar-policy-symcc/src/symcc/solver.rs
```diff
@@ -222,29 +212,46 @@ impl Solver for LocalSolver {
...
         if len == 0 {
-            Err(SolverError::Solver(
-                "Encountered EOF while reading from solver output".to_string(),
-            ))
-        } else {
-            Ok(len)
+            // An unexpected EOF was encountered while reading from solver output
+            // Kill the child process and clean up its resources
+            self.clean_up().await?;
+            self.check_child_process_status().await?;
```
chaluli, line 240: Will this sometimes no longer error on EOF?

Is either 236 or 237 guaranteed to error in the current state?

cedar-policy-symcc/src/symcc/solver.rs
```diff
@@ -222,29 +212,46 @@ impl Solver for LocalSolver {
...
         match s
             .strip_prefix("(error \"")
             .and_then(|s| s.strip_suffix("\")\n"))
         {
-            Some(e) => SolverError::Solver(e.to_string()),
+            Some(e) => {
+                if e.starts_with("Parse Error: ") {
+                    // Parse errors cause CVC5 to quit and need to be handled specially.
+                    // Kill the child process and clean up its resources
+                    let _ = self.clean_up().await;
```
chaluli, line 254: Good change! :)

chaluli (approved):
Looks great. Left a few comments.

When CVC5 gives EOF in readline, are the newly introduced functions guaranteed to error (keeping similar OK/error behaviors), if so It'd be great to get a comment stating that.

cedar-policy-symcc/src/symcc/solver.rs
katherine-hough: We need the reference to self to kill the child process if a parse error is encountered.

cedar-policy-symcc/src/symcc/solver.rs
katherine-hough: We need the reference to self to kill the child process if a parse error is encountered.

chaluli: Yes, but why not use the normal syntax? `self.process_error_output(s)` as opposed to `Self::process_error_output(self, s)`.

cedar-policy-symcc/src/symcc/solver.rs
katherine-hough: It will error with "Solver process terminated unexpectedly with status:". The call to `clean_up` will call `wait` on the child process which will wait for the child to exit completely, returning the status that it exited with. Tokio guarantees that after `wait` is called `wait` and `try_wait` will continue to have the same status/return value. So, this value will be picked up by the call to `try_wait` in `check_child_process_status`

cedar-policy-symcc/src/symcc/solver.rs
katherine-hough: Oh yeah. Whoops.

cedar-policy-symcc/src/symcc/solver.rs
chaluli: Thanks for the explanation.

katherine-hough: I added a comment to the code to clarify this.

Issue #1828, opened 2025-09-16
const EntityTypeName


TroyKomodo:
### Category

User level API features/changes

### Describe the feature you'd like to request

Allow the EntityTypeName to be constructed at compile time.

### Describe alternatives you've considered

Not sure, ideally some way to cache the requests.

### Additional context

```rs
const fn validate_part(part: &str) {
    let invalid_parts = &["true", "false", "if", "then", "else", "in", "is", "like", "has", "__cedar"];
    konst::for_range! {idx in 0..invalid_parts.len() =>
        if konst::string::eq_str(invalid_parts[idx], part) {
            const_panic::concat_panic!("part cannot be a reserved keyword: ", part);
        }
    }
}

const fn validate(input: &str) {
    konst::iter::for_each! {part in konst::string::split(input, "::") =>
        validate_part(part);

        let mut chars = konst::string::chars(part);

        let Some(c) = chars.next() else {
            panic!("part must be non empty");
        };

        if c != '_' && !c.is_ascii_alphabetic() {
            const_panic::concat_panic!("part must start with either a '_' or any alphabetic character but started with: ", c);
        }

        konst::iter::for_each!{c in chars =>
            if c != '_' && !c.is_ascii_alphanumeric() {
                const_panic::concat_panic!("part can only contain '_' or alphanumeric characters but contained: ", c);
            }
        }
    }
}

#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
pub struct EntityTypeName(&'static str);

impl std::fmt::Display for EntityTypeName {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        self.0.fmt(f)
    }
}

impl EntityTypeName {
    pub const fn new(input: &'static str) -> Self {
        validate(input);
        EntityTypeName(input)
    }

    pub const fn as_str(&self) -> &'static str {
        self.0
    }
}

#[macro_export]
macro_rules! entity_type_name {
    ($text:expr) => {
        const { $crate::EntityTypeName::new($text) }
    };
}
```

### Is this something that you'd be interested in working on?

- [x] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

john-h-kastner-aws:
We'd be very interested in seeing something like this added to the library. There's some prior discussion on a similar feature
* here #64,
* and here #744 

[...]

PR #1860, opened 2025-09-29
add static id / paths construction

TroyKomodo:
## Description of changes

Allow for construction of IDs, and entity type names at compile time.

## Issue #, if available

#1828 

### DIFF
--- a/cedar-policy-core/src/ast/id.rs
+++ b/cedar-policy-core/src/ast/id.rs
@@ -50,6 +50,12 @@ impl Id {
         Id(s.into())
     }
 
+    /// Similar to `new_unchecked`, but for static strings which can be `const`
+    /// constructed
+    pub(crate) const fn new_unchecked_from_static(s: &'static str) -> Id {
+        Id(SmolStr::new_static(s))
+    }
+
     /// Get the underlying string
     pub fn into_smolstr(self) -> SmolStr {
         self.0
--- a/cedar-policy-core/src/ast/name.rs
+++ b/cedar-policy-core/src/ast/name.rs
@@ -25,9 +25,9 @@ use smol_str::ToSmolStr;
 use std::collections::HashSet;
 use std::fmt::Display;
 use std::str::FromStr;
-use std::sync::Arc;
 use thiserror::Error;
 
+use crate::ast::Path;
 use crate::parser::err::{ParseError, ParseErrors, ToASTError, ToASTErrorKind};
 use crate::parser::Loc;
 use crate::FromNormalizedStr;
@@ -44,7 +44,7 @@ pub struct InternalName {
     /// Basename
     pub(crate) id: Id,
     /// Namespaces
-    pub(crate) path: Arc<Vec<Id>>,
+    pub(crate) path: Path,
     /// Location of the name in source
     #[educe(PartialEq(ignore))]
     #[educe(Hash(ignore))]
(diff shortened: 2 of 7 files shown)


cedar-policy-core/src/ast/name.rs
```diff
@@ -416,6 +421,173 @@ impl FromStr for Name {
...
+        if let CompileTimeName::Name(name) = &self {
+            assert!(name.0.is_static(), "name should be static when forgetting");
+        }
+        std::mem::forget(self);
+    }
+}
+
+/// This is how we can create a `Name` at compile time
+#[macro_export]
+macro_rules! make_name {
```
TroyKomodo, line 568: @john-h-kastner-aws  this is sort of how I think we should allow for compile time name creation.

cedar-policy-core/src/ast/name.rs
```diff
@@ -638,6 +810,17 @@ impl<'a> arbitrary::Arbitrary<'a> for Name {
 mod test {
     use super::*;
 
+    #[test]
+    fn compile_time_name() {
+        const _: Name = make_name!("foo").expect("should be OK");
```
john-h-kastner-aws, line 815: IIUC, this `expect` is also evaluated at compile time, correct?

cedar-policy-core/src/ast/id.rs
```diff
@@ -50,6 +50,12 @@ impl Id {
         Id(s.into())
     }
 
+    /// Similar to `new_unchecked`, but for static strings which can be `const`
+    /// constructed
+    pub(crate) const fn new_unchecked_from_static(s: &'static str) -> Id {
```
john-h-kastner-aws, line 55: This is kind of nice change on its own. Wouldn't mind splitting it out into a separate small PR

cedar-policy-core/src/ast/id.rs
```diff
@@ -61,6 +67,11 @@ impl Id {
     pub fn is_reserved(&self) -> bool {
         self.as_ref() == RESERVED_ID
     }
+
+    /// Check if the `Id` is static
+    pub const fn is_static(&self) -> bool {
```
john-h-kastner-aws, line 72: Would it be better to use the name `is_heap_allocated` instead since this function will return `true` for an inline smolstr?

cedar-policy-core/src/ast/name.rs
```diff
@@ -416,6 +421,173 @@ impl FromStr for Name {
...
+        match self {
+            NameValidationError::ReservedKeyword(s) => const_panic::concat_panic!(msg, ": part is a reserved keyword: ", s),
+            NameValidationError::Empty => const_panic::concat_panic!(msg, ": part is empty"),
+            NameValidationError::PartStart(c) => const_panic::concat_panic!(msg, ": part starts with invalid char: ", c),
+            NameValidationError::PartContains(c) => const_panic::concat_panic!(msg, ": part contains invalid char: ", c),
+        }
+    }
+}
+
+const fn validate_part(part: &str) -> Option<NameValidationError> {
```
john-h-kastner-aws, line 462: My main concern for merging a final implementation will be the correctness of this function. Ideally some of the logic should be shared (at least the list of reserved identifiers), and then unit tests ofc, but we can also do something relatively lightweight with the `proptest` crate

cedar-policy-core/src/ast/name.rs
john-h-kastner-aws: looks like a reasonable approach to me. It should be trivial to have a wrapper around all this in the `cedar-policy` crate to construct the public `Name` type, right?

cedar-policy-core/src/ast/name.rs
TroyKomodo: yes thats correct, this is not a std::result::Result, so this is calling a different method i defined called expect which is const.

cedar-policy-core/src/ast/name.rs
TroyKomodo: Yes thats correct

cedar-policy-core/src/ast/name.rs
TroyKomodo: I agree, it would be ideal if we could share the logic. Not sure how to go about that tho.

PR #1868, opened 2025-10-01
Fix `Decimal`'s implementation of `Display`

shaobo-he-aws:
We should pad 0s after the decimal point

### DIFF
--- a/cedar-policy-core/src/extensions/decimal.rs
+++ b/cedar-policy-core/src/extensions/decimal.rs
@@ -166,7 +166,7 @@ impl std::fmt::Display for Decimal {
         if self.value.is_negative() {
             write!(f, "-")?;
         }
-        let pow = i128::pow(10, 4);
+        let pow = i128::pow(10, NUM_DIGITS);
         write!(f, "{}.{:04}", abs / pow, abs % pow)
     }
 }


chaluli:
> Can I take a moment and question the fact that we have two identical implementations of std::fmt::Display for Decimal

Except Decimal and Decimal are different types, 😅.

Perhaps that requires it's own question.

john-h-kastner-aws:
> Perhaps that requires it's own question.

More concerning, the both implement decimal parsing, and substantially differently.

adpaco:
> Except Decimal and Decimal are different types

But they share the same layout, right? Is there a reason we can't use the `cedar-policy-core` type?

chaluli:
> But they share the same layout, right? Is there a reason we can't use the `cedar-policy-core` type?

I don't see why not. We should probably just inline the type (possibly re-exporting it in the given scope).

shaobo-he-aws:
> > But they share the same layout, right? Is there a reason we can't use the `cedar-policy-core` type?
> 
> I don't see why not. We should probably just inline the type (possibly re-exporting it in the given scope).

Historically speaking, we should not because decimals are extension types that are not part of core Cedar and were supposedly added by users. So a good engineering practice is to expose common interfaces used by the evaluator/validator and not expose internal implementation details. That's why they are hidden. But they are de facto builtin constructs of Cedar now.

Issue #1902, opened 2025-10-21
[TPE] Add APIs to get residuals as `PolicySet`


shaobo-he-aws:
### Category

User level API features/changes

### Describe the feature you'd like to request

We should provide such APIs so users can work on the residuals directly.

### Describe alternatives you've considered

N/A

### Is this something that you'd be interested in working on?

- [x] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #1906, opened 2025-10-23
Add APIs to get residual policies

shaobo-he-aws:
## Issue #, if available
#1902 

### DIFF
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -5050,17 +5050,13 @@ mod tpe {
     use std::collections::{BTreeMap, HashMap, HashSet};
     use std::sync::Arc;
 
-    use cedar_policy_core::ast::{self, PartialValueToValueError, Value};
+    use cedar_policy_core::ast::{self, Value};
     use cedar_policy_core::authorizer::Decision;
     use cedar_policy_core::batched_evaluator::is_authorized_batched;
     use cedar_policy_core::batched_evaluator::{
         err::BatchedEvalError, EntityLoader as EntityLoaderInternal,
     };
     use cedar_policy_core::tpe;
-    use cedar_policy_core::{
-        entities::conformance::EntitySchemaConformanceChecker, extensions::Extensions,
-        validator::CoreSchema,
-    };
     use itertools::Itertools;
     use ref_cast::RefCast;
     use smol_str::SmolStr;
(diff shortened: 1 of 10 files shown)


cedar-policy/src/api.rs
```diff
@@ -5358,6 +5358,22 @@ mod tpe {
                 .map(Into::into)
                 .map_err(Into::into)
         }
+
+        /// Return an iterator of residual policies
+        pub fn residual_policies(&self) -> impl Iterator<Item = Policy> + '_ {
+            self.0
+                .residual_policies()
+                .into_iter()
```
john-h-kastner-aws: redundant `into_iter`?

cedar-policy/src/api.rs
```diff
@@ -5358,6 +5358,22 @@ mod tpe {
                 .map(Into::into)
                 .map_err(Into::into)
         }
+
+        /// Return an iterator of residual policies
+        pub fn residual_policies(&self) -> impl Iterator<Item = Policy> + '_ {
```
john-h-kastner-aws, line 5363: Is this the behavior users want? It's surprising to me that someone might want to include trivially `true` and `false` policies in the list of residuals. IMO documentation for this should at least call out that you might want `non_trivial_residual_policies`

cedar-policy/src/api.rs
```diff
@@ -5487,7 +5503,7 @@ mod tpe {
                     .0
                     .residual_policies()
                     .into_iter()
-                    .map(Policy::from_ast),
+                    .map(|p| p.clone().into()),
```
john-h-kastner-aws, line 5496: what's this change? not obviously equivalent withing looking at the `into` impl

cedar-policy/src/api.rs
shaobo-he-aws: Sounds good.

cedar-policy/src/api.rs
shaobo-he-aws: Yes.

cedar-policy/src/api.rs
shaobo-he-aws: They are actually equivalent. The `From` implementation from core policy to public API policy uses `from_ast`.

shaobo-he-aws:
The semver failure is a false alarm because the renaming in `TPEResponse` is guarded by the feature flag `tpe`.

PR #1909, opened 2025-10-24
Fix warnings for doc build

shaobo-he-aws:

### DIFF
--- a/cedar-policy-symcc/src/err.rs
+++ b/cedar-policy-symcc/src/err.rs
@@ -58,5 +58,5 @@ pub enum Error {
     ConcretizeError(#[from] ConcretizeError),
 }
 
-/// A result type that potentially returns a SymCC [`Error`].
+/// A result type that potentially returns a SymCC [`enum@Error`].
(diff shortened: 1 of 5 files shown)


cedar-policy-symcc/src/err.rs
```diff
@@ -58,5 +58,5 @@ pub enum Error {
     ConcretizeError(#[from] ConcretizeError),
 }
 
-/// A result type that potentially returns a SymCC [`Error`].
+/// A result type that potentially returns a SymCC [`enum@Error`].
```
john-h-kastner-aws, line 61: what's this?

shaobo-he-aws: There's ambiguity in what `Error` refers to. It can be an enum defined in this file or a trait that's imported.

john-h-kastner-aws: interesting. never seen this syntax before

adpaco-aws (approved):

PR #1912, opened 2025-10-27
Handle residual records in has_attr partial evaluation

phsym:
## Description of changes

Handle residual records in has_attr partial evaluation.

Current partial evaluation of `residual_record({"key": unknown("value")}) has key` always returns residual `residual_record({"key": unknown("value")} has key` instead of returnning `true`.

Same for partial eval of `residual_record({}) has key` that returns residual `residual_record({}) has key` instead of returning `false`.

This PR propose an enhancement so that `has` operator is properly partially evaluated on residual records.

### DIFF
--- a/cedar-policy-core/src/evaluator.rs
+++ b/cedar-policy-core/src/evaluator.rs
@@ -703,7 +703,15 @@ impl<'e> Evaluator<'e> {
                     ],
                     &val,
                 )),
-                PartialValue::Residual(r) => Ok(Expr::has_attr(r, attr.clone()).into()),
+                PartialValue::Residual(r) => match r.expr_kind() {
+                    // If the residual is a known record and is projectable (it’s guaranteed to never error on evaluation),
+                    // then check for the attribute existency.
+                    ExprKind::Record(rec) if r.is_projectable() => {
+                        Ok(rec.contains_key(attr).into())
+                    }
+                    // Otherwise, leave the expression as is.
+                    _ => Ok(Expr::has_attr(r, attr.clone()).into()),
+                },
             },
             ExprKind::Like { expr, pattern } => {
                 let v = self.partial_interpret(expr, slots)?;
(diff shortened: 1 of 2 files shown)


shaobo-he-aws:
@phsym thank you for your interest in Cedar. We unfortunately cannot apply the changes proposed in this PR because it breaks the soundness criteria of partial evaluation. That is, partial evaluation when re-evaluated should produce the same outcome as normal evaluation, given the same inputs. For example, consider the condition expression `{"a": 1 + principal.i} has "a"`. The proposed changes evaluate it to true whereas normal evaluation produces an error when `principal.i` is `i64::MAX`. Note that the evaluation of an `has` expression starts with the LHS and errors if evaluating LHS errors.

phsym:
@shaobo-he-aws understood, thanks for the quick feedback.

In that case, do you think it could still be done if lhs is projectable ?

shaobo-he-aws:
> @shaobo-he-aws understood, thanks for the quick feedback.
> 
> In that case, do you think it could still be done if lhs is projectable ?

Could you elaborate on what it means for a LHS to be projectable? If you mean record values are *all* `Value`s, then it should be Ok. It entails a large refactoring of the PE codebase, but we're happy to discuss it.

phsym:
What I have in mind is calling `Expr::is_projectable()` like it's done for the `get_attr` case in https://github.com/cedar-policy/cedar/blob/main/cedar-policy-core%2Fsrc%2Fevaluator.rs#L880

As far as I understood, `is_projectable` returns true if the expr is guaranteed to not fail on evaluation, meaning it only contains values and unknowns.

phsym:
@shaobo-he-aws I just pushed an update adding the `is_projectable` check, limitting the partial evaluation to only the records that are guaranteed to never error on evaluation.

Also added a unit test for that case

cedar-policy-core/src/evaluator.rs
```diff
@@ -703,7 +703,15 @@ impl<'e> Evaluator<'e> {
                     ],
                     &val,
                 )),
-                PartialValue::Residual(r) => Ok(Expr::has_attr(r, attr.clone()).into()),
+                PartialValue::Residual(r) => match r.expr_kind() {
+                    // If the residual is a known record and is projectable (it’s guaranteed to never error on evaluation),
```
shaobo-he-aws, line 707: "unknown record" is a little bit confusing. Can we just say "If the residual is a record".

shaobo-he-aws (approved):
LGTM. Just some minor comments.

john-h-kastner-aws (commented):
These changes look reasonable to me, but I'm not entirely sure we want to be accepting changes to the old (non-type aware) implementation of partial evaluation. At some point in the future type aware partial evaluation will become the stable, supported version of partial evaluation (currently both are experimental).

cdisselkoen (approved):
These look like good changes and I think we should take them.

PR #1929, opened 2025-10-31
Remove unused action attributes from validator implementation

john-h-kastner-aws:
## Description of changes

This is an entirely backwards compatible change to remove the partial implementation of a feature in the validator which was never publicly exposed. It has always been an error to have a schema where an `action` defines `attributes`, so the publicly visible behavior of the schema does not change with this PR.

The only change to the public API is the deprecation if some error variants. These error variants were already never returned when using the public API. They are now never constructed anywhere, but remain in the enum for to avoid a breaking change. 

### DIFF
--- a/cedar-policy/CHANGELOG.md
+++ b/cedar-policy/CHANGELOG.md
@@ -25,7 +25,7 @@ Cedar Language Version: TBD
   ensuring that a `PartialEntities` object always meets the preconditions required for type aware partial evaluation.
 - Evaluate `has` operation when the LHS record is projectable during partial evaluation. (#1912)
 - Deprecated schema parsing errors `ActionAttributesContainEmptySet`, `UnsupportedActionAttribute`, `ActionAttrEval`, and `ExprEscapeUsed`.
-  These errors are never returned, so it is safe do delete any associated error handling code.
+  These errors are never returned, so it is safe to delete any associated error handling code.
 
 ### Fixed
 


john-h-kastner-aws:
debugging corpus test failure. I did not expect any change, so this is interesting

cedar-policy-core/src/validator/typecheck.rs
```diff
@@ -1849,12 +1822,6 @@ impl<'a> SingleEnvTypechecker<'a> {
             // check if any of the actions of type `rhs_ety` have descendant
             // action entities with type `lhs_ety`. If there is none, we
             // can soundly type it as false.
```
cdisselkoen, line 1824: I think this comment needs updating after this change. The RHS is not (necessarily) action types now

cedar-policy-core/src/validator/typecheck.rs
```diff
@@ -1992,22 +1955,6 @@ impl<'a> SingleEnvTypechecker<'a> {
...
-                                Some(rhs_ty),
-                            ) if !self.check_action_in_type(lhs_ety, rhs_ty) => {
-                                TypecheckAnswer::success(
-                                    ExprBuilder::with_data(Some(Type::False))
-                                        .with_same_source_loc(in_expr)
-                                        .is_in(lhs_expr, rhs_expr),
-                                )
-                            }
-
                             // Similar to the case above, but for `EntityRecordKind::Entity`
```
cdisselkoen, line 1958: This comment needs updating

cedar-policy-core/src/validator/types.rs
```diff
@@ -1121,14 +1098,6 @@ pub enum EntityRecordKind {
...
     /// the schema, based on the elements of the [`EntityLUB`].
     Entity(EntityLUB),
-
-    /// We special case action entities, which store their attributes directly, like `Record`s do
-    ActionEntity {
-        /// Type name of the action entity
-        name: EntityType,
-        /// Attributes of the action entity
-        attrs: Attributes,
-    },
```
cdisselkoen, line 1131: Checking my understanding, since we're removing `attrs`, there's no longer a reason to special-case action entities and previous users of `ActionEntity` will just use `Entity` instead?

cedar-policy/CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ Cedar Language Version: TBD
 - For the `tpe` experimental feature, `PartialEntities::from_concrete` now requires a `Schema` and will validate the entities,
   ensuring that a `PartialEntities` object always meets the preconditions required for type aware partial evaluation.
 - Evaluate `has` operation when the LHS record is projectable during partial evaluation. (#1912)
+- Deprecated schema parsing errors `ActionAttributesContainEmptySet`, `UnsupportedActionAttribute`, `ActionAttrEval`, and `ExprEscapeUsed`.
+  These errors are never returned, so it is safe do delete any associated error handling code.
```
cdisselkoen:
```suggestion
  These errors are never returned, so it is safe to delete any associated error handling code.
```

cedar-policy-core/src/validator/types.rs
john-h-kastner-aws: Right. The other way the `ActionEntity` was used was to decide if the entity type _was_ an action, but we can also check that with `EntityType::is_action`. We do need to be careful though, as my first revision of this PR did not added that check where it was needed, leading to a corpus test failure (so, a win for corpus tests).

Issue #104, opened 2023-05-31
Improve precision when typechecking `==`


john-h-kastner-aws:
### Category

Cedar validation features

### Describe the feature you'd like to request

The validator is sometimes able to decide that an `==` expression will always evaluate to `true` of `false`, but we are fairly conservative in this to avoid possible soundness bugs. We should expand the range of expressions this applies to. For example, `1 == "2"` is an equality between different primitive types, so it will always be false.  

### Describe the solution you'd like

n/a

### Describe alternatives you've considered

n/a

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

john-h-kastner-aws:
Implementing this in a backwards compatible manner requires #539

mwhicks1:
Updating this issue with some more rationale. The status quo is:

1. Expression `User::"alice" == Action::"foo"` has type `False` (you can compare two entities of different types)
2. Expression `[1] == 1` is a type error (you cannot compare two non-entity values of different types), not type `False`

The reason for (1) is to handle the cross-product correctly (e.g., when you could have two different principal types you have to allow expressions like `User::"bob" == Admin::"boss"`), and the reason for (2) is to avoid the unsoundness of assuming two values of differing set types must be unequal (since the empty set `[]` can be given many possible `Set<…>` types).

To users, this may appear a little strange. It might appear less strange if we adjusted the behavior to be that `e1 == e2` has
1. Type `False` when `e1` and `e2` have two different types and they are not both `Set<…>` types
2. Type `Bool` when they are both the same `Set<T>` type
3. Type-incorrect otherwise

This is more lenient and potentially a little easier to explain. A third option might be even better, but unfortunately will not work with analysis:
1. Type `False` when `e1` and `e2` have two different types and they are not both `Set<…>` types
2. Type `Bool` otherwise

The reason it won't work is that logical comparisons are strongly typed, i.e., both arguments must have the same type (including the same type for set elements).

emina:
I think there is a way to support the third option in the logical encoding too, at the cost of introducing extra disjunctions into the encoding.  So, let's keep the third option on the table until we have data showing that this extra cost isn't acceptable.

emina:
> I think there is a way to support the third option in the logical encoding too. So, let's keep the third option on the table for now. We have to think some about the encoding idea to make sure it actually works.

[...]

PR #1931, opened 2025-10-31
Make `in`, `==`, and `hasTag` more permissive to match Lean model

john-h-kastner-aws:
## Description of changes

* `==` has type `true` or `false` for comparisons of literals
* `in` has type `true` for `action in` when schema declares the `action` is in the target.
* `in` has type `false` for comparisons between non-literal entity type expression when they the schema states that they cannot be related 
* `hasTag` has type `true` when prior capability already contains tag access (i.e., after a prior `hasTag` check for the same tag)

This doesn't fully resolve the difference with `in`, but these changes match the Lean and are reasonable on their own. Differential testing will confirm that the Rust implementation is still at least as strict as the Lean. 

### DIFF
--- a/cedar-policy-core/src/validator/schema.rs
+++ b/cedar-policy-core/src/validator/schema.rs
@@ -1022,9 +1022,9 @@ impl ValidatorSchema {
     /// Get all action entities in the schema where `action in euids` evaluates
     /// to `true`. This includes all actions which are descendants of some
     /// element of `euids`, and all elements of `euids`.
-    pub(crate) fn get_actions_in_set<'a, 'b, 'c>(
+    pub(crate) fn get_actions_in_set<'a, 'b>(
         &'a self,
-        euids: impl IntoIterator<Item = &'b EntityUID> + 'c,
+        euids: impl IntoIterator<Item = &'b EntityUID>,
     ) -> Option<Vec<&'a EntityUID>> {
         euids
             .into_iter()
--- a/cedar-policy-core/src/validator/typecheck.rs
+++ b/cedar-policy-core/src/validator/typecheck.rs
@@ -2009,10 +2009,10 @@ impl<'a> SingleEnvTypechecker<'a> {
     // the exact set of ancestors for the action in the schema, allowing us to
     // evalute the `in` to `true` or `false`, so the type will be `True` or
     // `False`  respectively.
-    fn type_of_action_in_actions<'b, 'c>(
+    fn type_of_action_in_actions<'b>(
         &self,
         lhs: &EntityUID,
-        rhs: impl IntoIterator<Item = &'b EntityUID> + 'c,
+        rhs: impl IntoIterator<Item = &'b EntityUID>,
         in_expr: &Expr,
         lhs_expr: Expr<Option<Type>>,
         rhs_expr: Expr<Option<Type>>,


john-h-kastner-aws:
Integration test failure is `Expected that validation would fail in /home/runner/work/cedar/cedar/cedar/cedar-testing/../cedar-integration-tests/corpus-tests/14466f88b2efc8be22cfbd0a1219b0a4f952b845.json, but it did not.`

I'm making the validator more permissive, so this is expected, but I still need to look at the exact change here to make sure it's ok

john-h-kastner-aws:
One sample corpus test failure is on

```
permit(
  principal,
  action in [Action::"action"],
  resource == a::""
) when {
  (true && (true == 0)) && false
};
```

The corpus test expects that validation will fail, but it doesn't. This is expected because I changed `true == 0` for being a type-error to having type `False`.


Another failure is

```
@A("")
forbid(
  principal,
  action in [Action::"action"],
  resource
) when {
  ((true && ((if (action in action) then action else resource) in principal)) && false) && false
};
```

Here we previously type `action in action` as `Bool`, but we now type it more precisely as `True`, so there is not type error due to the different types in the `if` branches.

---

IRRC, the procedure for fixing the corpus test failure is to merge this change with the failure and then pull update corpus tests after they're generated from our nightly fuzz testing.

cedar-policy-core/src/validator/typecheck.rs
```diff
@@ -2226,6 +2234,27 @@ impl<'a> SingleEnvTypechecker<'a> {
         }
     }
 
+    /// Check if the action is in the list of descendant action. Return the singleton
+    /// type false if it is not, and true otherwise.
+    fn action_in_descendants<'b, 'c>(
```
cdisselkoen: I think the use of `descendants` is unnecessary here. I was wondering if we got it backwards and meant `ancestors`, but in reality, the implementation here just checks if `lhs_entity` is in the `rhs` iterator. So let's call the function `action_in_set` or something, and the RHS parameter just `rhs`.

cdisselkoen: Actually, there's no reason the function needs to have `action` in the name either, nothing about it is specific to actions

cedar-policy-core/src/validator/typecheck/test/expr.rs
```diff
@@ -838,16 +858,47 @@ fn action_literal_in_action_typechecks_true() {
     "#
     .parse()
     .expect("Expected that schema would parse");
+    let src = r#"Action::"action" in action"#;
     assert_typechecks(
```
cdisselkoen: Can we get another test, where `action in Action::"action"`, and where `action in Action::"action_group"`

cedar-policy-core/src/validator/typecheck/test/strict.rs
```diff
@@ -234,10 +234,13 @@ fn eq_strict_types_mismatch() {
         assert_types_must_match(
             s,
             &q,
-            &Expr::from_str(r#"1 == "foo""#).unwrap(),
-            r#"1 == "foo""#,
+            &Expr::from_str(r#"principal == "foo""#).unwrap(),
+            r#"principal == "foo""#,
             Type::primitive_boolean(),
```
cdisselkoen, line 239: Why is `1 == "foo"` now false, but `principal == "foo"` still boolean? what's the difference? can't `principal == "foo"` also be false while we're here? separate PR?

john-h-kastner-aws: Difference in behavior in literals vs vars. The intent of this change is to align with the Lean model which also doesn't check the var case.

cedar-policy-core/src/validator/typecheck.rs
john-h-kastner-aws: the bit specific to actions is that it returns the type `True` when it finds `lhs_entity` which implements the specific typing rule for `in` on actions

cdisselkoen: Understood; that's probably worth a code comment

john-h-kastner-aws: latest revision inlines `entity_in_descendants` and `action_in_descendants` since these didn't make too much sense as standalone methods now that they're not sharing code anymore

john-h-kastner-aws:
This will resolve #104 as I believe typechecking for `==` should be exactly the same in Rust and Lean. We might be able to make it even more permissive, but that would require Lean model and proofs, so we should consider those changes on a case-by-case basis.

I don't think this is enough to close our #105. I'd need to look closer, but eyeballing the model I think there are still differences.

john-h-kastner-aws:
Posted https://github.com/cedar-policy/cedar-spec/pull/784 to update diff testing to test for exact agreement. Running tests locally, if it doesn't find anything we can call https://github.com/cedar-policy/cedar-spec/issues/126, #104, and #105 resolved

shaobo-he-aws (approved):

cedar-policy-core/src/validator/schema.rs
```diff
@@ -1015,25 +1010,15 @@ impl ValidatorSchema {
     /// includes all entity types that are descendants of the type of `entity`
     /// according  to the schema, and the type of `entity` itself because
     /// `entity in entity` evaluates to `true`.
-    pub(crate) fn get_entity_types_in<'a>(&'a self, entity: &'a EntityUID) -> Vec<&'a EntityType> {
+    pub(crate) fn get_entity_types_in<'a>(&'a self, entity: &'a EntityType) -> Vec<&'a EntityType> {
```
cdisselkoen: Comment needs updating for this change. Also we should consider renaming the parameter

cedar-policy-core/src/validator/typecheck.rs
```diff
@@ -1990,166 +1942,79 @@ impl<'a> SingleEnvTypechecker<'a> {
...
-                    ExprBuilder::with_data(Some(Type::False))
-                        .with_same_source_loc(in_expr)
-                        .is_in(lhs_expr, rhs_expr),
-                )
-            }
+        debug_assert!(
+            lhs_euid.is_action(),
+            "We expect this function is called only when an action entity is on the LHS"
+        );
+        // If there's a at least on action on the right, check if that
```
cdisselkoen:
```suggestion
        // If there's a at least one action on the right, check if that
```

cedar-policy-core/src/validator/typecheck.rs
```diff
@@ -1990,166 +1942,79 @@ impl<'a> SingleEnvTypechecker<'a> {
...
+        // action is an ancestor of the LHS action. We can ignore any
+        // non-actions because we assume action cannot be `in` a non-action.
+        let rhs_actions: Vec<_> = rhs_elems
+            .into_iter()
+            .filter(|e| e.entity_type().is_action())
+            .collect();
+        if !rhs_actions.is_empty() {
```
cdisselkoen: Might be more readable to make the test `if rhs_actions.is_empty()` and switch the branches

john-h-kastner-aws: switched this to `peekable` instead of collecting the vec, so now it's `rhs_actions.peek().is_some()`.

john-h-kastner-aws:
Latest push adds one more fix for level-checking for an issue found by diff testing. The other validation test targets successfully ran over the weekend in the exact-match mode without finding any differences.

Because the Rust impl returns invalid-action-application errors which are never returned by the Lean model, I ignored these errors in the comparison for diff testing. For level checking, this meant the Rust impl would hit this error and stop before doing the actual level checking while the Lean model might continue on to find a level checking error. I ignored this kind of error, so diff testing reported this as Lean finding an error that was missed by the Rust impl (though the validator was still sound because it in fact did produce an error).

I fixed this by having the Rust impl always do level checking even after validation fails. It still skips any specific environments that failed to typecheck. This allows the Rust impl to continue on to find the level checking error

cedar-policy-core/src/validator/schema.rs
```diff
@@ -1005,26 +1005,26 @@ impl ValidatorSchema {
...
     }
 
     /// Get all action entities in the schema where `action in euids` evaluates
     /// to `true`. This includes all actions which are descendants of some
     /// element of `euids`, and all elements of `euids`.
-    pub(crate) fn get_actions_in_set<'a>(
+    pub(crate) fn get_actions_in_set<'a, 'b, 'c>(
         &'a self,
-        euids: impl IntoIterator<Item = &'a EntityUID> + 'a,
+        euids: impl IntoIterator<Item = &'b EntityUID> + 'c,
```
cdisselkoen: I'm surprised the `+ 'c` is necessary?

cedar-policy-core/src/validator/typecheck.rs
```diff
@@ -2015,14 +2009,14 @@ impl<'a> SingleEnvTypechecker<'a> {
     // the exact set of ancestors for the action in the schema, allowing us to
     // evalute the `in` to `true` or `false`, so the type will be `True` or
     // `False`  respectively.
-    fn type_of_action_in_actions<'b>(
+    fn type_of_action_in_actions<'b, 'c>(
         &self,
         lhs: &EntityUID,
-        rhs: impl IntoIterator<Item = &'a EntityUID> + 'a,
+        rhs: impl IntoIterator<Item = &'b EntityUID> + 'c,
```
cdisselkoen: here too

cdisselkoen (approved):

cedar-policy-core/src/validator/schema.rs
john-h-kastner-aws: ah, I was trying to do replace `+ 'a` with `+ '_`, but that didn't work. Looks like I can just omit it entirely.

cdisselkoen (approved):

Issue #1411, opened 2025-01-06
`FromNormalizedStr`'s default implementation is inefficient


jv-garcia:
### Describe the improvement you'd like to request

As I was doing profiling in my team's code, this function happened to be taking a surprisingly big chunk of CPU time. 

Taking a look at the [implementation](https://docs.rs/cedar-policy-core/latest/src/cedar_policy_core/from_normalized_str.rs.html#38), it might seem like it does do more work than necessary, parsing and then comparing the results to `to_string` to see if there are issues.

Notice that my team is using Cedar 2.5, but looking at the code, it seems this issue might still be there in newer versions.

### Describe alternatives you've considered

Haven't tried anything yet, but happy to give a try to a solution that doesn't do the round trip with some guidance.

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this internal improvement
- [ ] ⚠️ This feature might incur a breaking change

jv-garcia:
FWIW, I spent some time today benchmarking this (with the help of an LLM ;)) and found some interesting results.

I created the following toy benchmark:

```rust
use std::str::FromStr;
use std::sync::LazyLock;

use cedar_policy::EntityTypeName;
use criterion::{black_box, criterion_group, criterion_main, Criterion};

fn parse_cedar_entity_name(name: &str) {
    EntityTypeName::from_str(name).ok();
}

use regex::Regex;
use std::collections::HashSet;

static ID_REGEX: LazyLock<Regex> =
    LazyLock::new(|| Regex::new(r"^[_a-zA-Z][_a-zA-Z0-9]*(?:::[_a-zA-Z][_a-zA-Z0-9]*)*$").unwrap());
static RESERVED: LazyLock<HashSet<&str>> = LazyLock::new(|| {
    vec![
        "if",
        "then",
        "else",
        "in",
        "like",
        "has",
        "is",
        "__cedar",
        "true",
        "false",
        "permit",
        "forbid",
        "principal",
        "action",
        "resource",
        "context",
    ]
    .into_iter()
    .collect()
});

fn is_valid_identifier(input: &str) -> bool {
    // First check if the pattern matches
    if !ID_REGEX.is_match(input) {
        return false;
    }

    // Split by "::" and check each segment against reserved keywords
    for segment in input.split("::") {
        if RESERVED.contains(segment) {
            return false;
        }
    }

    true
}

pub fn criterion_benchmark(c: &mut Criterion) {
[...]

PR #1936, opened 2025-11-04
Updated Expr printing to default to printing attributes using . 

lianah:
## Description of changes

Updated Expr printing to default to printing attributes using . notation for valid identifiers. Similarly for `has`. 

### DIFF
--- a/cedar-policy-core/src/est/expr.rs
+++ b/cedar-policy-core/src/est/expr.rs
@@ -1683,6 +1683,11 @@ mod test {
             .into_expr::<Builder>();
         assert_eq!(format!("{expr}"), r#"context["foo-baz"]"#);
 
+        let expr = parse_expr(r#"context["true"]"#)
+            .unwrap()
+            .into_expr::<Builder>();
+        assert_eq!(format!("{expr}"), r#"context["true"]"#);
+
         // Similarly for Expr::HasAttr
         let expr = parse_expr(r#"context has foo"#)
             .unwrap()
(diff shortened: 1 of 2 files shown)


cedar-policy-core/src/est/expr.rs
```diff
@@ -1371,11 +1370,17 @@ impl BoundedDisplay for ExprNoExt {
             }
             ExprNoExt::GetAttr { left, attr } => {
                 maybe_with_parens(f, left, n)?;
-                write!(f, "[\"{}\"]", attr.escape_debug())
+                match parse_ident(attr) {
```
john-h-kastner-aws, line 1373: Could be a regex match if parsing turns out to be slow. This isn't a performance critical part of the code, but I won't be too surprised if one of our fuzz targets finds a slow unit here

lianah:
> Do we need a similar change in cedar-policy-formatter?
> Do we need a similar change in the schema printer?

Great point, I will look into updating those as well for consistency.

chaluli (approved):

cedar-policy-core/src/est/expr.rs
lianah: I can update it to a regex if we get slow units. I guess my thinking was using `parse_ident` would ensure that if we ever change what makes a valid identifier this code won't diverge.

cdisselkoen: I would actually say the other way -- let's accept this change as-is, and if it turns out to be slow, let's update `parse_ident` itself to improve its performance

john-h-kastner-aws: Right, I'm thinking something like #1411 if/when we see slow units (that references `FromNormalizedStr` which doesn't apply to idents, but the same optimization is reasonable)

cedar-policy-core/src/validator/cedar_schema/fmt.rs
```diff
@@ -201,10 +202,13 @@ impl<N: Display> IndentedDisplay for json_schema::RecordType<N> {
                 writeln!(f)?;
             }
             ty.annotations.fmt_indented(f, member_indentation.len())?;
+            let n = match parse_ident(n) {
+                Ok(id) => id.into_smolstr(),
+                Err(_) => format!("\"{}\"", n.escape_debug()).into(),
```
john-h-kastner-aws, line 207: nit: allocating a string here where it wouldn't before

cedar-policy-core/src/validator/cedar_schema/fmt.rs
```diff
@@ -201,10 +202,13 @@ impl<N: Display> IndentedDisplay for json_schema::RecordType<N> {
                 writeln!(f)?;
             }
             ty.annotations.fmt_indented(f, member_indentation.len())?;
+            let n = match parse_ident(n) {
```
john-h-kastner-aws, line 205: Does this correctly handle schema keywords? I don't recall off the top of my head, but I imagine the set of keywords that would need to be quoted isn't the same for policies and schema

lianah: I updated `example.cedarschema` with some Schema specific keywords and it seems they don't need escaping.

john-h-kastner-aws (approved):

lianah:
I have another PR that updates the wasm examples: https://github.com/cedar-policy/cedar-examples/pull/248 (it's currently failing the build because the changes in this PR didn't make it yet)

PR #1938, opened 2025-11-06
Replace `assert_cmd` with `escargot` in cedar-testing

shaobo-he-aws:


shaobo-he-aws:
> looks good. Is it worth standardizing on this crate? We also use `assert_cmd` in the cli crate

I think `assert_cmd` provides better and more ergonomic APIs. It's fine as long as it refers to a binary in the same crate. So, I think we should keep it in the cli crate.

PR #1943, opened 2025-11-10
Use O(1) set and map in SymCC

shaobo-he-aws:


shaobo-he-aws:
> I think the changes could be good (better performances is always good :)), but would the change from BTreeSet to HashSet cause issues in which encoding/decoding terms is not determinstic?

I'm working on benchmarking the changes. The deterministic issue could be easily addressed by using `indexmap`.

shaobo-he-aws:
I think I made symcc much slower with this PR. The reason is that hashing a `BTreeMap` is much more expensive than sorting a `BTreeMap` of `BTreeMap`.

PR #1947, opened 2025-11-14
Performance optimization on the validator

shaobo-he-aws:
## Description of changes

This PR contains three lines of optimizations.

* Use `BTreeSet` to represent capabilities. Existing representation is `HashSet`, which can be slower than `BTreeSet` specifically for composite expressions like set and records, because it needs to iterate every component whereas `BTreeSet` can shortcircuit.
* Wrap the `ty` field of `AttributeType` with `Arc`
* Wrap the `BTreeMap` in `Attributes` with `Arc`

### DIFF
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -1797,7 +1797,8 @@ impl<T> Expr<T> {
                     error_kind: error_kind1,
                 },
             ) => error_kind.cmp(error_kind1),
-            // This should never be reached since we compare variants first
+            // PANIC SAFETY: This should never be reached since we compare variants first
+            #[allow(clippy::unreachable)]
             _ => unreachable!(
                 "Different variants should have been handled by variant_order comparison"
             ),


cedar-policy-core/src/ast/expr.rs
```diff
@@ -1614,10 +1654,159 @@ impl<T> Expr<T> {
...
+
+    /// Implementation of ordering corresponding to equality as implemented by
+    /// `eq_shape`. Must satisfy the usual relationship between equality and
+    /// ordering.
+    pub fn cmp_shape(&self, other: &Expr<T>) -> std::cmp::Ordering {
+        // First compare variants for early short-circuiting using discriminant
+        let self_kind = self.expr_kind();
+        let other_kind = other.expr_kind();
+        if std::mem::discriminant(self_kind) != std::mem::discriminant(other_kind) {
+            return self_kind.variant_order().cmp(&other_kind.variant_order());
```
cdisselkoen, line 1666: Could we just do `self_kind.cmp(&other_kind)`? Since we know they are different variants, the `cmp` will short-circuit?  Then we wouldn't need to maintain our own `variant_order()` method

cedar-policy-core/src/validator/types.rs
```diff
@@ -1173,9 +1174,13 @@ impl EntityRecordKind {
...
             EntityRecordKind::AnyEntity => vec![],
-            EntityRecordKind::Entity(lub) => {
-                lub.get_attribute_types(schema).attrs.into_keys().collect()
-            }
+            EntityRecordKind::Entity(lub) => lub
+                .get_attribute_types(schema)
+                .attrs
+                .as_ref()
+                .clone()
+                .into_keys()
```
cdisselkoen: Might be more efficient to do `.keys().cloned()`, avoiding cloning all the values

cedar-policy-core/src/validator/typecheck.rs
```diff
@@ -776,7 +776,9 @@ impl<'a> SingleEnvTypechecker<'a> {
                         let all_attrs = typ_actual.all_attributes(self.schema);
                         let attr_ty = Type::lookup_attribute_type(self.schema, typ_actual, attr);
                         let annot_expr = ExprBuilder::with_data(
-                            attr_ty.clone().map(|attr_ty| attr_ty.attr_type),
+                            attr_ty
+                                .clone()
+                                .map(|attr_ty| attr_ty.attr_type.as_ref().clone()),
```
john-h-kastner-aws, line 781: more cloning here than I'd expect. Could you drop the first? `attr_ty` is used later, so maybe it'd have to be `attr_ty.as_ref()...`

cedar-policy-core/src/validator/types.rs
john-h-kastner-aws, line 1: The `Type::Set` variant still store a `Box<Type>`. Should that also be an `Arc`?

cedar-policy-core/src/ast/expr.rs
shaobo-he-aws: We can't seem to derive `PartialOrd` and `Ord` for `ExprKind` because it would require `Expr` to derive them, which transitively require the type argument `T` to do the same.

cedar-policy-core/src/validator/types.rs
shaobo-he-aws: Done.

cedar-policy-core/src/validator/typecheck.rs
shaobo-he-aws: Done.

cedar-policy-core/src/validator/types.rs
shaobo-he-aws: Done.

john-h-kastner-aws (approved):

john-h-kastner-aws (approved):

PR #1964, opened 2025-11-25
Fix parsing of small negative decimal literals

chaluli:
## Description of changes

Fixes an issue in decimal parsing.

### DIFF
--- a/cedar-policy-symcc/CHANGELOG.md
+++ b/cedar-policy-symcc/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Fixed parsing of small negative decimal literals. (#1964)
+
 Cedar Language Version: TBD
 
 ## [0.1.1] - 2025-11-20
--- a/cedar-policy/CHANGELOG.md
+++ b/cedar-policy/CHANGELOG.md
@@ -21,6 +21,7 @@ Cedar Language Version: TBD
 ### Fixed
 
 - Fixed authorization and other error messages to correctly display all diagnostic information. (#1944)
+- Fixed parsing of small negative decimal literals. (#1964)
 
 ## [4.8.0] - 2025-11-17
 


cedar-policy-core/src/extensions/decimal.rs
```diff
@@ -122,35 +122,35 @@ impl Decimal {
...
             .as_str();
-        let r = caps
+        let r_str = caps
             .get(2)
             .ok_or_else(|| Error::FailedParse(str.as_ref().to_owned()))?
             .as_str();
 
-        // convert the left component to i64 and multiply by `10 ^ NUM_DIGITS`
-        let l = i64::from_str(l).map_err(|_| Error::Overflow)?;
+        // convert the integer component to i64 and multiply by `10 ^ NUM_DIGITS`
```
john-h-kastner-aws, line 140: could have a comment that the sign will be handled later

cedar-policy-symcc/src/symcc/extension_types/decimal.rs
```diff
@@ -84,7 +84,7 @@ impl FromStr for Decimal {
                             let rlen_u32: u32 = rlen.try_into().unwrap();
                             let r: i128 = r.into();
                             let r_prime = r * 10_i128.pow(DECIMAL_DIGITS - rlen_u32);
-                            let i = if l >= 0 {
+                            let i = if !left.starts_with("-") {
```
john-h-kastner-aws, line 87: Not for now, and I commented this previously on another PR, but I'm still not happy with the duplicated parsing logic.

Good job remembering to apply the fix here as well, but I don't like that we need to remember.

john-h-kastner-aws (approved):

cedar-policy-symcc/src/symcc/extension_types/decimal.rs
chaluli: I tried to remove the parsing here. But I don't see a good way to do that unless we make the decimal type in cedar-policy-core public.

john-h-kastner-aws: > unless we make the decimal type in cedar-policy-core public.

Given the type is just a wrapper around an `i64` I feel like this shouldn't be a problem, but I don't know why we ended up with this design in the first place

PR #2013, opened 2025-12-01
add optimized interface for symcc

cdisselkoen:
## Description of changes

This corresponds to the Lean changes made in https://github.com/cedar-policy/cedar-spec/pull/787.

### DIFF
--- a/cedar-policy-symcc/src/symcc/enforcer.rs
+++ b/cedar-policy-symcc/src/symcc/enforcer.rs
@@ -25,7 +25,6 @@
 //! For more technical details, see comments in SymCC/Enforcer.lean.
 
 use std::{
-    borrow::Borrow,
     collections::{BTreeMap, BTreeSet},
     sync::Arc,
 };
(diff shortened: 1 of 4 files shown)


cedar-policy-symcc/src/lib.rs
```diff
@@ -185,6 +226,9 @@ impl<S: Solver> CedarSymCompiler<S> {
 
     /// Returns true iff [`WellTypedPolicy`] does not error on any well-formed
     /// input in the given symbolic environment.
+    ///
+    /// Consider using the optimized version `check_never_errors_opt()` instead,
+    /// which will allow you to reuse a `CompiledPolicy` across many queries.
```
john-h-kastner-aws, line 231: Worth deleting/deprecating this? Any reason not to call the optimized form?

cdisselkoen: Might be a future where we want to have tests that verify you get the same results on both paths? (We've proven this in Lean, but you know, guarding against mistranslations from Lean to Rust in the symccopt module)

Also I believe that the optimized path is pretty performance-equivalent even when you don't reuse a `CompiledPolicy` or `CompiledPolicies`, but if I'm wrong about that, callers that only want to make a single query might get an advantage by using the unoptimized interface directly

chaluli (approved):

PR #2018, opened 2025-12-02
ensure that symcc unit tests cover both optimized and unoptimized interfaces

cdisselkoen:
This ensures that we have coverage of the new optimized interface introduced in #2013 

Note that there is one small change to `cedar-policy` here -- implementations of `PartialEq` for the `Request` and `Context` types in the public API


cdisselkoen:
These changes invalidate the two failing tests, as those tests are currently written, because the optimized path assumes that you always use the `SymEnv` derived from `envs.req_env` in the ordinary way, while these tests use a `SymEnv` that is modified from that one. Is this a use-case we intend to support? Should we simply delete these tests? Should we modify them in some way so that they only use the unoptimized path?

cdisselkoen:
Per conversation offline, I'll modify the tests to only use the unoptimized path

PR #2037, opened 2025-12-12
fix some warnings with Rust 1.92

cdisselkoen:
## Description of changes

Fixes some warnings that are present with Rust 1.92. In particular, sets one warning class (`unused_assignments`) to `allow` as it seems to erroneously identify some fields in error structs as unused.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ unsafe_code = "forbid"
 unexpected_cfgs = { level = 'deny', check-cfg = ['cfg(kani)', 'cfg(fuzzing)'] }
 missing_debug_implementations = "deny"
 rust-2018-idioms = "deny"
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -39,6 +39,7 @@ unsafe_code = "forbid"
 unexpected_cfgs = { level = 'deny', check-cfg = ['cfg(kani)', 'cfg(fuzzing)'] }
 missing_debug_implementations = "deny"
 rust-2018-idioms = "deny"
+unused_assignments = "allow" # In Rust 1.92 this one erroneously identifies some fields in error structs as unused, when they are actually used in the `#[error]` attribute
```
john-h-kastner-aws, line 42: happy to merge this to unblock CI, but we should open an issue to track removing this when rust fixes the false positives

cdisselkoen: #2038

PR #2045, opened 2025-12-15
Fixing some unicode issues 

aaronjeline:
## Description of changes
Found some issues int he language server's unicode handling:
1. The `get_policy_scope_ranges` function incorrectly uses `char_indices()` which returns byte indices, but stores them as `Position.character` which should be character indices. This causes scope detection to fail for policies containing Unicode characters.
2. The `to_range` function panics when given a `SourceSpan` with a byte offset that falls in the middle of a multi-byte UTF-8 character. The function uses byte slicing (`&src[..offset]`) without validating that the offset is at a character boundary.
3. The `get_word_at_position` function panics when the text contains multi-byte UTF-8 characters and `position.character` points to a byte offset that falls inside a multi-byte character. The function incorrectly treats `position.character` as a byte offset when slicing the string.
4. The `get_operator_at_position` function panics when the text contains multi-byte UTF-8 characters and `position.character` points to a byte offset that falls inside a multi-byte character. This is the same bug as in `get_word_at_position` - the function incorrectly treats `position.character` as a byte offset when slicing the string.

Also adds some new tests that pass, happy to remove them if you don't want them. Let me know if there's any noise at all in the request that you don't want.

## Issue #, if available
N/A

### DIFF
--- a/cedar-language-server/src/policy/format.rs
+++ b/cedar-language-server/src/policy/format.rs
@@ -49,3 +49,33 @@ pub(crate) fn format_policy(policy: &str) -> Option<Vec<TextEdit>> {
 
     Some(vec![edit])
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use proptest::prelude::*;
+
+    proptest! {
+        #[test]
+        fn format_policy_idempotent(policy in "permit\\(principal, action, resource\\);") {
+            if let Some(edits) = format_policy(&policy) {
+                let formatted = &edits[0].new_text;
+                if let Some(edits2) = format_policy(formatted) {
+                    prop_assert_eq!(&edits2[0].new_text, formatted, "formatting should be idempotent");
+                }
+            }
+        }
+
+        #[test]
+        fn format_policy_range_covers_input(policy in "permit\\(principal, action, resource\\);\n?") {
+            if let Some(edits) = format_policy(&policy) {
+                let range = &edits[0].range;
+                prop_assert_eq!(range.start.line, 0);
+                prop_assert_eq!(range.start.character, 0);
+                let input_lines = policy.lines().count() as u32;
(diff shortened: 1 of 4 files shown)


cedar-language-server/src/utils.rs
```diff
@@ -99,19 +99,22 @@ pub(crate) fn to_lsp_diagnostics<'a>(
 /// The function will panic if `source_span.offset()` or
 /// `source_space().offset() + source_span.len()` is not a valid index into str.
```
cdisselkoen: should we remove this comment? did we remove all the possible panics?

cedar-language-server/src/policy/format.rs
```diff
@@ -49,3 +49,33 @@ pub(crate) fn format_policy(policy: &str) -> Option<Vec<TextEdit>> {
 
     Some(vec![edit])
 }
+
+#[cfg(test)]
+mod tests {
```
john-h-kastner-aws, line 54: any actual value in these prop tests? These utils tests look worth keeping though

cedar-language-server/src/utils.rs
```diff
@@ -155,34 +158,28 @@ pub(crate) fn get_word_at_position(position: Position, text: &str) -> Option<&st
...
+        .char_indices()
+        .nth(char_pos)
+        .map(|(i, _)| i)
+        .unwrap_or(line.len());
 
     // Helper function to check if a character is part of a word
     let is_word_char = |c: char| c.is_alphanumeric() || c == '_' || c == ':' || c == '=';
 
     // Find the start of the word
-    #[expect(
```
john-h-kastner-aws, line 167: need to keep `expects`

cedar-language-server/src/policy/format.rs
aaronjeline: They did not find any bugs, so just a question of if you want to keep them for regression testing. Up to you.

PR #2046, opened 2025-12-16
fix more lints in test code

cdisselkoen:
## Description of changes

Fixes more lints that appear in test code. Continuing #2044. As noted there, these lints don't seem to show up in our CI (with our current configuration), but they do show up in IDEs depending on your configuration.

### DIFF
--- a/cedar-language-server/src/policy/completion/provider.rs
+++ b/cedar-language-server/src/policy/completion/provider.rs
@@ -158,11 +158,11 @@ impl PolicyCompletionProvider {
     }
 }
 
+#[cfg(test)]
 #[expect(
     clippy::literal_string_with_formatting_args,
     reason = "strings used by LSP"
 )]
-#[cfg(test)]
 pub(crate) mod tests {
     use std::vec;
 
--- a/cedar-policy-symcc/tests/utils/mod.rs
+++ b/cedar-policy-symcc/tests/utils/mod.rs
@@ -14,10 +14,11 @@
  * limitations under the License.
  */
 
-#![allow(
+#![expect(
     dead_code,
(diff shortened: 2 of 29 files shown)


cedar-language-server/src/policy/completion/provider.rs
```diff
@@ -158,11 +158,11 @@ impl PolicyCompletionProvider {
     }
 }
 
+#[cfg(test)]
```
john-h-kastner-aws, line 161: is this just a style thing or does the order matter?

cedar-policy-symcc/tests/utils/mod.rs
```diff
@@ -14,10 +14,11 @@
  * limitations under the License.
  */
 
-#![allow(
+#![expect(
```
john-h-kastner-aws, line 17: This is surprising. Is the whole test module `#[cfg(test)]`?

cedar-language-server/src/policy/completion/provider.rs
cdisselkoen: Style, I personally prefer `cfg(test)` first

cedar-policy-symcc/tests/utils/mod.rs
cdisselkoen: After changing this to `#![cfg(test)]`, clippy still flags many of the functions in this file as dead code, at least with my rust-analyzer configuration. I agree this is surprising.

chaluli: I have experienced this before. It was surprising. Rust just doesn't handle compilation of test modules well.

PR #2048, opened 2025-12-17
add new SymCC primitives for comparing the matches-behavior of individual policies

cdisselkoen:
## Description of changes

Rust PR corresponding to https://github.com/cedar-policy/cedar-spec/pull/808 on the Lean side; see notes there. It's useful to compare the matches-behavior of individual policies, particularly in the presence of forbid policies. Previously this would require recompiling the forbid policy as a permit policy.

### DIFF
--- a/cedar-policy-symcc/src/lib.rs
+++ b/cedar-policy-symcc/src/lib.rs
@@ -393,9 +393,10 @@ impl<S: Solver> CedarSymCompiler<S> {
     /// function differs from `check_equivalent` on singleton policysets in how it treats
     /// `forbid` policies -- while `check_equivalent` trivially holds for any pair of
     /// `forbid` policies (as they both always-deny), `check_matches_equivalent` only
-    /// holds if the two policies match exactly the same set of inputs. Also, a nontrivial
-    /// `permit` and nontrivial `forbid` policy can be `check_matches_equivalent`, but can
-    /// never be `check_equivalent`.
+    /// holds if the two policies match exactly the same set of inputs. Also, a nonempty
+    /// `permit` and nonempty `forbid` policy can be `check_matches_equivalent`, but can
+    /// never be `check_equivalent`. (By "nonempty" we mean, matches at least one request
+    /// in the given symbolic environment.)
     ///
     /// Consider using the optimized version `check_matches_equivalent_opt()` instead,
     /// which will allow you to reuse a `CompiledPolicy` across many queries.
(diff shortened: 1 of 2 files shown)


chaluli (approved):
Looks like a pretty clear translation of the lean implementation for these functions.

We should probably add some tests of these new functions.

cdisselkoen:
Added tests (both in this PR and the Lean one).

cedar-policy-symcc/src/symccopt/verifier.rs
```diff
@@ -51,6 +55,37 @@ pub fn verify_evaluate_opt(phi: impl FnOnce(&Term) -> Term, policy: &CompiledPol
...
+    phi: impl FnOnce(&Term, &Term) -> Term,
+    policy1: &CompiledPolicy,
+    policy2: &CompiledPolicy,
+) -> Asserts {
+    assert_eq!(&policy1.symenv, &policy2.symenv);
+    // As an optimization here in `symccopt`:
+    // Our callers only pass relatively simple functions as `phi`.
+    // We expect that `enforce_pair_compiled_policy()` is much more expensive to compute than `phi`.
+    // So, we first compute the assert involving `phi`. If that is
+    // constant-false, we can just reutrn constant-false and not compute
```
john-h-kastner-aws: typo: reutrn

cedar-policy-symcc/src/lib.rs
```diff
@@ -378,6 +384,251 @@ impl<S: Solver> CedarSymCompiler<S> {
...
+    ///
+    /// Compare with `check_equivalent`, which takes two policysets (which could consist
+    /// of a single policy, or more) and determines whether the _authorization behavior_
+    /// of those policysets is equivalent for well-formed inputs in the `symenv`. This
+    /// function differs from `check_equivalent` on singleton policysets in how it treats
+    /// `forbid` policies -- while `check_equivalent` trivially holds for any pair of
+    /// `forbid` policies (as they both always-deny), `check_matches_equivalent` only
+    /// holds if the two policies match exactly the same set of inputs. Also, a nontrivial
+    /// `permit` and nontrivial `forbid` policy can be `check_matches_equivalent`, but can
+    /// never be `check_equivalent`.
```
john-h-kastner-aws: For this and the other new functions: is this saying that the policy effects have bearing on the result of the comparison? I think this should be stated more directly, assuming my understanding is correct.

john-h-kastner-aws (approved):

cedar-policy-symcc/src/lib.rs
cdisselkoen: For the new functions, the effect has _no_ bearing on the result. (If you convert the inputs to singleton policysets and then use the old functions, the effect _does_ have bearing on the result.)

chaluli: By non-trivial, do you mean the permit's guard is not semantically equivalent to `false`?

cdisselkoen: Yes, I mean that it matches at least one request

cdisselkoen: maybe nonempty is a better word

cdisselkoen: Updated this and corresponding comments to say "nonempty" and also provide a definition of nonempty

PR #2054, opened 2025-12-22
Cedar 3.4.x formatter regex fix

lianah:
## Description of changes

Porting constructing formatter regex in lazy block. 

### DIFF
--- a/cedar-policy-formatter/src/pprint/token.rs
+++ b/cedar-policy-formatter/src/pprint/token.rs
@@ -21,14 +21,14 @@ use logos::{Logos, Span};
 use smol_str::SmolStr;
 use std::fmt::{self, Display};
 
-// PANIC SAFETY: These regex patterns are valid
-#[allow(clippy::unwrap_used)]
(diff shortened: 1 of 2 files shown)


cedar-policy-formatter/src/pprint/token.rs
```diff
@@ -18,15 +18,21 @@
...
 use logos::{Logos, Span};
-use regex::Regex;
 use smol_str::SmolStr;
 use std::fmt::{self, Display};
 
+// PANIC SAFETY: These regex patterns are valid
+#[allow(clippy::unwrap_used)]
+pub(crate) mod regex_constants {
+    use regex::Regex;
+    lazy_static::lazy_static! {
```
cdisselkoen: On [main](https://github.com/cedar-policy/cedar/blob/main/cedar-policy-formatter/src/pprint/token.rs), this code has been updated to use the standard library's `LazyLock` instead of the 3rd-party `lazy_static` macro. Might make sense to update this here as well, while we're here?

lianah: Good point. I will update.

PR #2057, opened 2025-12-23
Fix cedar-wasm build errors

chaluli:
## Description of changes

cedar-wasm was failing in CI. This PR updates the cedar-wasm-bindgen version and edits a cedar-ffi type's definition to use it's explicit name rather than `Self` in its recursive definition. This seems to *avoid* the bug in the cedar-wasm generation which fails to properly translate `Self` to `this`.

Unrelated change to `cargo_audit` action, because on investigation of the WASM CI failure, I found that the `cargo_audit` action uses `cargo-deny`. I updated it to use the same version as `build_and_test` action.

### DIFF
--- a/cedar-policy/src/ffi/utils.rs
+++ b/cedar-policy/src/ffi/utils.rs
@@ -30,7 +30,10 @@ pub use cedar_policy_core::jsonvalue::JsonValueWithNoDuplicateKeys;
 #[cfg(feature = "wasm")]
 extern crate tsify;
 
-#[expect(clippy::use_self, reason="cedar-wasm build will fail until wasm-bindgen fixes a bug that doesn't translsate `Self` to `this`.")]
+#[expect(
+    clippy::use_self,
+    reason = "cedar-wasm build will fail until wasm-bindgen fixes a bug that doesn't translsate `Self` to `this`."
+)]
 /// Structure of the JSON output representing one `miette` error
 #[derive(Debug, PartialEq, Eq, Clone, Hash, Deserialize, Serialize, Default)]
 #[cfg_attr(feature = "wasm", derive(tsify::Tsify))]


cedar-policy/src/ffi/utils.rs
```diff
@@ -53,7 +53,7 @@ pub struct DetailedError {
     pub source_locations: Vec<SourceLabel>,
     /// Related errors
     #[serde(default)]
-    pub related: Vec<Self>,
+    pub related: Vec<DetailedError>,
```
john-h-kastner-aws, line 60: Looks like we changed this recently to satisfy clippy, so we should probably `expect` it with comment so we don't break this again

chaluli: Do we know what lint? Clippy doesn't seem to complain, so I don't know what lint one would expect.

john-h-kastner-aws: Might need to run with `--all-features`. I see this in the action logs

```
 warning: unnecessary structure name repetition
  --> cedar-policy/src/ffi/utils.rs:56:22
   |
56 |     pub related: Vec<DetailedError>,
   |                      ^^^^^^^^^^^^^ help: use the applicable keyword: `Self`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.92.0/index.html#use_self
note: the lint level is defined here
  --> cedar-policy/src/lib.rs:23:27
   |
23 | #![warn(clippy::pedantic, clippy::use_self, clippy::option_if_let_else)]
   |                           ^^^^^^^^^^^^^^^^
```

chaluli: Weird, locally I get:

```
warning: this lint expectation is unfulfilled
  --> cedar-policy/src/ffi/utils.rs:56:14
   |
56 |     #[expect(clippy::use_self, reason="cedar-wasm build will fail until wasm-bindgen fixes a bug that doesn't translsate `Self` to `this`.")]
   |              ^^^^^^^^^^^^^^^^
   |
   = note: cedar-wasm build will fail until wasm-bindgen fixes a bug that doesn't translsate `Self` to `this`.
   = note: `#[warn(unfulfilled_lint_expectations)]` on by default
 ```

chaluli: This was after adding the lint. To the definition.

chaluli: I've added the expect, but I dislike that there is different behaviors for that locally vs CI. 🤷

john-h-kastner-aws (approved):

cedar-policy/src/ffi/utils.rs
john-h-kastner-aws: might be a different version?

cdisselkoen (approved):

PR #2058, opened 2025-12-26
Feat/cedar schema to json with resolved types

Swolebrain:
## Description of changes

The schema type resolver feature implements a function in Cedar's FFI crate that converts Cedar schema strings to JSON format while resolving ambiguous "EntityOrCommon" type designations to their specific Entity or CommonType classifications. This function addresses the limitation where Cedar's default schema serialization uses generic "EntityOrCommon" types when the specific classification is ambiguous, making it difficult for downstream tools to work with precise type information.

The solution leverages Cedar's internal ValidatorSchema and AllDefs structures to perform accurate type resolution, ensuring that all output JSON contains only explicit Entity or CommonType designations.

I made the call to allow warnings. Initially, i wanted to not support schemas that parse with warnings. However, some of the warnings involve shadowing extension function names. In the future, if we release a `path` extension function, and this new function to resolve schema types returned `Err` when there were warnings, then the introduction of `path` would potentially be a breaking change in the behavior of this function for some users. 

I don't expect that all those snapshot tests will make it into the repo, but they're useful in order to evaluate that the results are as expected. I can submit a review removing the snapshots and the line of code that generates them if we so desire and that's the only thing blocking the merge.

## Issue #, if available

n/a

### DIFF
--- a/cedar-policy-core/src/validator/json_schema.rs
+++ b/cedar-policy-core/src/validator/json_schema.rs
@@ -235,8 +235,10 @@ impl Fragment<RawName> {
         }
 
         // Add aliases for primitive types in the empty namespace (so "String" resolves to "__cedar::String")
-        for tyname in primitives_as_internal_names {
-            all_defs.mark_as_defined_as_common_type(tyname);
+        for tyname in &primitives_as_internal_names {
+            if !all_defs.is_defined_as_common(tyname) && !all_defs.is_defined_as_entity(tyname) {
+                all_defs.mark_as_defined_as_common_type(tyname.clone());
+            }
         }
 
         // Step 1: Convert Fragment<RawName> to Fragment<ConditionalName>
@@ -1564,12 +1566,12 @@ impl TypeVariant<InternalName> {
         match self {
             TypeVariant::EntityOrCommon { type_name } => {
                 // Check if this is an entity type or common type
-                if all_defs.is_defined_as_entity(&type_name) {
+                if all_defs.is_defined_as_common(&type_name) {
+                    Ok(ResolvedTypeVariant::CommonTypeRef(type_name))
+                } else if all_defs.is_defined_as_entity(&type_name) {
                     Ok(ResolvedTypeVariant::TypeVariant(TypeVariant::Entity {
                         name: type_name,
                     }))
-                } else if all_defs.is_defined_as_common(&type_name) {
-                    Ok(ResolvedTypeVariant::CommonTypeRef(type_name))
                 } else {
                     // If it's neither, this means that the json_schema::Fragment wasn't complete
                     Err(TypeNotDefinedError {
--- a/cedar-policy/src/ffi/snapshots/cedar_policy__ffi__convert__test__schema_resolution_round_trip@shadowed_entity_and_common.snap
+++ b/cedar-policy/src/ffi/snapshots/cedar_policy__ffi__convert__test__schema_resolution_round_trip@shadowed_entity_and_common.snap
@@ -1,0 +1,76 @@
+---
+source: cedar-policy/src/ffi/convert.rs
+expression: resolved_json
+input_file: cedar-policy/src/ffi/test_schemas/shadowed_entity_and_common.cedarschema
+---
+{
+  "": {
+    "actions": {
+      "updateContact": {
+        "appliesTo": {
+          "context": {
+            "attributes": {
+              "addy": {
+                "type": "Address"
+              }
+            },
+            "type": "Record"
+          },
+          "principalTypes": [
+            "User"
+          ],
+          "resourceTypes": [
+            "User"
+          ]
+        }
+      }
+    },
+    "commonTypes": {
+      "Address": {
+        "attributes": {
+          "city": {
+            "type": "String"
+          },
+          "street": {
+            "type": "String"
+          },
+          "zipcode": {
+            "type": "String"
+          }
+        },
+        "type": "Record"
+      }
+    },
+    "entityTypes": {
+      "Address": {
+        "shape": {
+          "attributes": {
+            "e_city": {
+              "type": "String"
+            },
+            "e_street": {
+              "type": "String"
+            },
+            "e_zipcode": {
+              "type": "String"
+            }
+          },
+          "type": "Record"
+        }
+      },
+      "User": {
+        "shape": {
+          "attributes": {
+            "active": {
+              "type": "Bool"
+            },
+            "address": {
+              "type": "Address"
+            }
+          },
+          "type": "Record"
+        }
+      }
+    }
+  }
+}
--- a/cedar-policy/src/ffi/test_schemas/shadowed_entity_and_common.cedarschema
+++ b/cedar-policy/src/ffi/test_schemas/shadowed_entity_and_common.cedarschema
@@ -1,0 +1,24 @@
+type Address = {
+    "street": String,
+    "city": String,
+    "zipcode": String
+};
+
+entity Address = {
+    "e_street": String,
+    "e_city": String,
+    "e_zipcode": String
+};
+
+entity User = {
+    "address": Address,
+    "active": Bool
+};
+
+action updateContact appliesTo {
+    principal: [User],
+    resource: [User],
+    context: {
+        addy: Address
+    }
+};


cedar-policy-core/src/validator/json_schema.rs
```diff
@@ -831,6 +895,26 @@ pub struct ActionType<N> {
...
+            attributes: old.attributes,
+            applies_to: new_apply_spec,
+            member_of: old.member_of,
+            annotations: old.annotations,
+            loc: old.loc,
+            #[cfg(feature = "extended-schema")]
+            defn_loc: old.defn_loc,
+        }
+    }
+}
```
cdisselkoen: could we just use Rust [struct update syntax](https://doc.rust-lang.org/book/ch05-01-defining-structs.html#creating-instances-with-struct-update-syntax) to simplify this? or, probably, avoid the need for this helper function altogether?

cedar-policy-core/src/validator/json_schema.rs
```diff
@@ -1399,6 +1483,176 @@ impl Type<ConditionalName> {
     }
 }
 
+impl Type<InternalName> {
+    /// Resolve EntityOrCommon types to specific Entity or CommonType designations
+    pub fn resolve_type_entity_or_common(
```
cdisselkoen: nit: I find it a little messy that similar functions are named `resolve_type_entity_or_common()` here, `fully_resolve_common_and_entity_types()` on `NamespaceDef`, and `resolve_entity_or_common_types()` on `Fragment`. Is one of these names better than the others? should we use the same name everywhere?

cedar-policy-core/src/validator/schema/action.rs
```diff
@@ -55,6 +55,21 @@ pub struct ValidatorActionId {
...
+/// PartialEq implementation for ValidatorActionId, assumes that location
+/// information is irrelevant to the comparison
+impl PartialEq for ValidatorActionId {
+    fn eq(&self, other: &Self) -> bool {
+        self.name == other.name
+            && self.applies_to == other.applies_to
+            && self.descendants == other.descendants
+            && self.context == other.context
+    }
+}
```
cdisselkoen: For similar purposes elsewhere in `cedar-policy-core` we use `educe` to derive implementations of `PartialEq` automatically that ignore certain fields

cedar-policy-core/src/validator/schema/entity_type.rs
```diff
@@ -50,19 +50,34 @@ pub struct ValidatorEntityType {
...
+/// PartialEq implementation for ValidatorEntityType, assumes that location
+/// information is irrelevant to the comparison
+impl PartialEq for ValidatorEntityType {
+    fn eq(&self, other: &Self) -> bool {
+        self.name == other.name
+            && self.kind == other.kind
+            && self.descendants == other.descendants
+            && self.attributes == other.attributes
+    }
+}
```
cdisselkoen: Same here

cedar-policy/src/ffi/convert.rs
```diff
@@ -178,6 +187,138 @@ pub fn schema_to_json(schema: Schema) -> SchemaToJsonAnswer {
...
+            Err(e) => {
+                return SchemaToJsonWithResolvedTypesAnswer::Failure {
+                    errors: vec![miette::Report::new(e).into()],
+                };
+            }
+        };
+
+    let mut all_defs = AllDefs::single_fragment(&validator_fragment);
+
+    // Add built-in primitive types in the __cedar namespace
```
cdisselkoen: Could/should we reuse the function [here](https://github.com/cedar-policy/cedar/blob/0e33c2d05ac13d75f2c4c497b849774ca5c5e35d/cedar-policy-core/src/validator/schema.rs#L1098), or parts of it, so that we have a single place where the implicit `__cedar` defines are defined?

cedar-policy/src/ffi/convert.rs
```diff
@@ -178,6 +187,138 @@ pub fn schema_to_json(schema: Schema) -> SchemaToJsonAnswer {
     }
 }
 
+/// Convert a Cedar schema string to JSON format with resolved types.
```
john-h-kastner-aws, line 184: I haven't done a full review of the code yet, but my first impression is to wonder if the bulk of this implementation should live somewhere else. The `ffi` module (and `cedar-policy` crate as a whole) typically doesn't contain complex logic.

In a similar vein, IIUC what you've implemented, we'll also want to resolve types for the CLI schema conversion command, so we'll want a non-wasm/ffi function in cedar-policy that can call

Swolebrain: Ah the cli point is a good point. So I agree this function should be moved, and there should be a thin wasm+ffi wrapper around it. Where should I move this main function to?

cedar-policy/src/ffi/convert.rs
Swolebrain: There's also similar functionality in `ValidatorSchema::from_schema_fragments`. I thought about trying to consolidate it but i felt like i couldn't figure out something big enough to cut out. In `ValidatorSchema::from_schema_fragments`, we're only adding the __cedar definitions of primitives and extensions if they are shadowed as commontypes or entitytypes. 

In this function, I wanna add the primitive shadows always, because otherwise i'll fail at converting to `Fragment<InternalName>` if the input contains `__cedar::String` (for example).

cedar-policy-core/src/validator/json_schema.rs
```diff
@@ -1399,6 +1534,176 @@ impl Type<ConditionalName> {
...
+        match self {
+            TypeVariant::EntityOrCommon { type_name } => {
+                // Check if this is an entity type or common type
+                if all_defs.is_defined_as_entity(&type_name) {
+                    Ok(ResolvedTypeVariant::TypeVariant(TypeVariant::Entity {
+                        name: type_name,
+                    }))
+                } else if all_defs.is_defined_as_common(&type_name) {
+                    Ok(ResolvedTypeVariant::CommonTypeRef(type_name))
+                } else {
```
cdisselkoen, line 1575: If a name is defined as both an entity and common type, [RFC 24 specifies](https://cedar-policy.github.io/rfcs/0024-schema-syntax.html#rule-3-resolve-name-references-in-a-priority-order) that the common-type definition "wins". So I think we need to switch the order of these two checks (and maybe add a comment)

Swolebrain: Good catch will swap the order

Swolebrain: What about in a case like the shadowed_primitive test, where there's an entity called String?

If an attribute of another entity (for example `newTypeReason` in that test)  is defined to be `String`, will the built-in String take precedence over the entity type String?

cdisselkoen: No, it will not; to refer to the built-in String in that case you would need to explicitly use `__cedar::String`

cedar-policy-core/src/validator/json_schema.rs
```diff
@@ -195,6 +196,81 @@ impl Fragment<RawName> {
...
+                all_defs.mark_as_defined_as_common_type(ext_type.as_ref().qualify_with(None));
+            }
+        }
+
+        // Add aliases for primitive types in the empty namespace (so "String" resolves to "__cedar::String")
+        for tyname in &primitives_as_internal_names {
+            if !all_defs.is_defined_as_common(tyname) && !all_defs.is_defined_as_entity(tyname) {
+                all_defs.mark_as_defined_as_common_type(tyname.clone());
+            }
+        }
```
john-h-kastner-aws, line 242: Could this be moved into  the first `primitives_as_internal_names` loop so that it has the same structure as the extension types loop? I found it slightly confusing to be adding primitive types again after adding extension types

cedar-policy-core/src/validator/json_schema.rs
```diff
@@ -195,6 +196,81 @@ impl Fragment<RawName> {
         file.read_to_string(&mut src)?;
         Self::from_cedarschema_str(&src, extensions)
     }
+
+    /// Convert this `Fragment<RawName>` to a `Fragment<InternalName>` where all the
+    /// entity or common type references have been resolved
```
john-h-kastner-aws, line 201: Worth explicitly commenting that this errors on any undefined type references

PR #2070, opened 2025-12-31
Optimized compile function computes footprint in one pass

cdisselkoen:
## Description of changes

Rust port of the changes in https://github.com/cedar-policy/cedar-spec/pull/812; see notes there. This new optimized `compile` function computes not only the `Term`, but also the footprint, in one pass. This is a significant optimization compared to the status quo, where the unoptimized `footprint` function invokes `compile` on every _subexpression_, resulting in a huge amount of repeated work compiling.

### DIFF
--- a/cedar-policy-symcc/src/symccopt/compiler.rs
+++ b/cedar-policy-symcc/src/symccopt/compiler.rs
@@ -276,7 +276,7 @@ fn compile_app2(
                 term,
             })
         }
-        (LessEq, Bitvec { n: SIXTY_FOUR  }, Bitvec { n: SIXTY_FOUR }) => {
+        (LessEq, Bitvec { n: SIXTY_FOUR }, Bitvec { n: SIXTY_FOUR }) => {
             let term = some_of(bvsle(t1, t2));
             Ok(CompileResult {
                 // assuming everything is well-typed, the `binary_op_footprint()` here will be `None`,


cedar-policy-symcc/src/symccopt/authorizer.rs
```diff
@@ -0,0 +1,70 @@
...
+use cedar_policy::Effect;
+use cedar_policy_core::ast::{Policy, PolicySet};
+
+use super::compiler::{compile, CompileResult, Result};
+use crate::{
+    term_factory::{and, any_true, eq, not, some_of},
+    SymEnv,
+};
+
+pub fn compile_with_effect(
```
john-h-kastner-aws, line 26: I found this naming slightly confusing at first. I read it as "compiling using this effect" instead of "compile if it has this effect"

cedar-policy-symcc/src/symccopt/compiler.rs
```diff
@@ -0,0 +1,964 @@
...
+    /// of the following form:
+    ///
+    /// * A variable term with an entity type
+    /// * An entity reference literal
+    /// * An attribute access expression with an entity type
+    /// * A binary (`getTag`) expression with an entity type
+    ///
+    /// These are the only basic expressions in Cedar that may evaluate to an entity.
+    /// All other expressions that evaluate to an entity are built up from the above
+    /// basic expressions.
```
john-h-kastner-aws, line 61: not obvious to me what this does with `if` (can be and entity type, not one of these expressions)

cedar-policy-symcc/src/symccopt/authorizer.rs
cdisselkoen: This is fair. I named it to correspond with the function `satisfied_with_effect()` which has existed forever, but of which you could probably make the same complaint. What would be a good alternate name?

cedar-policy-symcc/src/symccopt/compiler.rs
cdisselkoen: `if` cannot evaluate to an entity type unless one (both) of its branches are one of these types. That is what "basic" is intended to mean here.

cdisselkoen: This is just a copy of a similar comment by @emina in the Lean codebase :)

cedar-policy-symcc/src/symccopt/authorizer.rs
john-h-kastner-aws: `satisfied_with_effect` makes sense to  me. I think it just doesn't read well here because `compile` is a verb. `compile_if_effect` might be better, but I could see the original making more sense in context given the parallel structure in `satisfied_with_effect`.

cedar-policy-symcc/src/symcc/op.rs
```diff
@@ -155,9 +155,7 @@ impl ExtOp {
                     xty: ExtType::IpAddr,
                 }] =>
             {
-                Some(TermType::Option {
-                    ty: Arc::new(TermType::Bitvec { n: 5 }),
-                })
+                Some(TermType::option_of(TermType::Bitvec { n: 5 }))
```
john-h-kastner-aws: I was surprised this line wasn't covered, looked coverage report, and this whole file isn't covered. Is it actually used anywhere?

cdisselkoen: The function this line is a part of is called in one place, at line 263 in this file. The function which line 263 is a part of is called in no places. I'll remove these functions in this PR.

cdisselkoen: Those two functions also have no analogue in the Lean file Op.lean.

chaluli (approved):

PR #2073, opened 2025-12-31
optimize decoder using references and SmolStr

cdisselkoen:
## Description of changes

Rust-specific performance optimization to the SymCC decoder -- use references instead of owned types in `IdMaps`, and (somewhat relatedly) use `SmolStr` instead of `String` in some places where we expect typical strings to be small.

### DIFF
--- a/cedar-policy-symcc/src/symcc/decoder.rs
+++ b/cedar-policy-symcc/src/symcc/decoder.rs
@@ -25,8 +25,7 @@ use cedar_policy::{EntityId, EntityUid};
 use itertools::Itertools;
 use miette::Diagnostic;
 use num_bigint::BigUint;
-
-use smol_str::SmolStr;
+use smol_str::{SmolStr, SmolStrBuilder};
 use thiserror::Error;
 
 use crate::symcc::bitvec::BitVecError;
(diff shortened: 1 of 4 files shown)


cedar-policy-symcc/src/symcc/decoder.rs
```diff
@@ -155,9 +154,8 @@ pub enum SExpr {
 /// - The (right) inverse: `encode_string`
 /// - The concrete C++ implementation in cvc5, which this function mimics
 ///   https://github.com/cvc5/cvc5/blob/b78e7ed23348659db52a32765ad181ae0c26bbd5/src/util/string.cpp#L136
-fn decode_string(s: &[u8]) -> Option<String> {
-    // Now handle string-theory-level escape sequences
-    let mut out = String::with_capacity(s.len());
+fn decode_string(s: &[u8]) -> Option<SmolStr> {
+    let mut out = SmolStrBuilder::new();
```
john-h-kastner-aws, line 158: Annoying that SmolStrBuilder doesn't have `with_capacity`

cdisselkoen: yeah. I think that using `SmolStrBuilder` will still be better performance than `String::with_capacity` followed by `.into()`, though. For one thing, if the final string is more than 0 but less than 23 characters (I anticipate by far the most common case), `SmolStrBuilder` doesn't allocate while `String::with_capacity` would (once).

PR #2079, opened 2026-01-02
performance optimization for computing footprints on the optimized path

cdisselkoen:
## Description of changes

Rust-specific optimization for computing footprints on the optimized path. Avoids collecting into `BTreeSet` at every intermediate step, in favor of collecting (and removing duplicates) once at the end. 

This PR also includes a technically-unrelated change to make `ExtType` `Copy`; I can separate that out into a different PR if desired.

### DIFF
--- a/cedar-policy-symcc/src/symcc/compiler.rs
+++ b/cedar-policy-symcc/src/symcc/compiler.rs
@@ -480,7 +480,7 @@ pub fn compile_call2(
     t2: Term,
 ) -> Result<Term> {
     let enc = |t1: Term, t2: Term| -> Term { some_of(enc(t1, t2)) };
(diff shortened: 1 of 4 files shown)


john-h-kastner-aws:
I came across this https://nnethercote.github.io/perf-book/iterators.html#chaining a while ago, which warns chaining iterators itself might be slow. I've not see it cause problems myself, but could be worth thinking if there's a better way to do it

cdisselkoen:
Noted. But (1) this PR barely increasing the amount of iterator-chaining going on, and (2) chaining iterators is surely still much much faster than collecting into a `BTreeSet` and then destructing the `BTreeSet`

PR #2084, opened 2026-01-05
add more entity getter methods

cdisselkoen:
## Description of changes

Adds `attrs()` and `tags()` methods on `Entity`, to iterate over all attributes/tags. The current API only allows getting specific attribute/tag values, not iterating over all attribute/tag values.

### DIFF
--- a/cedar-policy/CHANGELOG.md
+++ b/cedar-policy/CHANGELOG.md
@@ -14,6 +14,10 @@ Starting with version 3.2.4, changes marked with a star (*) are _language breaki
 
 Cedar Language Version: TBD
 
+### Added
+
+- `Entity::attrs()` and `Entity::tags()` to iterate over all attributes/tags of an `Entity` (#2084)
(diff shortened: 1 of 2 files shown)


cdisselkoen:
It's unfortunate that these methods have to be fallible due to the legacy partial-evaluation API. If we waited to add these until the legacy partial-evaluation API was removed in favor of TPE, we could make these methods infallible.

john-h-kastner-aws:
> It's unfortunate that these methods have to be fallible due to the legacy partial-evaluation API. If we waited to add these until the legacy partial-evaluation API was removed in favor of TPE, we could make these methods infallible.

I think this is a problem in enough places to merit `cedar-policy` library version 5.0.0

cdisselkoen:
If that's the route we go, then there's no reason to wait adding these getters in 4.x -- we would just change their signatures to infallible in 5.0.0

PR #2088, opened 2026-01-07
Added workflow for publishing to crates.io

katherine-hough:
## Description of changes
Added a GitHub workflow which can publish a GitHub tag to crates.io. This workflow must be manually dispatched. The workflow is set up to use [trusted publishing](https://crates.io/docs/trusted-publishing). I have not yet configured trusted publishing for our crates. I will do so after this is merged. I added a GitHub environment for the workflow called [release](https://github.com/cedar-policy/cedar/settings/environments). If we had a visible (non-secret) GitHub team, then we could restrict the "Required reviewers" for the environment to that team which I think will limit who can trigger this workflow.

## Issue #, if available
N/A

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,5 +1,7 @@
-# Note: This workflow is intended to always be run from the 'main' branch to
+# Note: This workflow can only be run from the 'main' branch to
 # ensure that the latest version of this workflow is used.
+# However, the release will not necessarily be performed from the 'main' branch 
+# as the input `tag` is used to determine which GitHub tag will be checked out for the release.
 name: Publish base Cedar crates to crates.io
 on:
     # This workflow must be triggered manually.


.github/workflows/publish.yml
```diff
@@ -0,0 +1,70 @@
...
+        runs-on: ubuntu-latest
+        environment: release # See https://github.com/cedar-policy/cedar/settings/environments
+        permissions:
+            id-token: write # Required for OIDC token exchange
+        steps:
+            - name: Validate branch
+              if: github.ref != 'refs/heads/main'
+              uses: actions/github-script@ffc2c79a5b2490bd33e0a41c1de74b877714d736 # v3.2.0
+              with:
+                  script: core.setFailed('This workflow must be triggered from the "main" branch.')
```
john-h-kastner-aws, line 29: from this it sounds like the comment at the top could be "... can only be run from ..."

.github/workflows/publish.yml
```diff
@@ -0,0 +1,70 @@
...
+            - name: Validate tag
+              run: |
+                  REGEX_PATTERN="^v[0-9]+\.[0-9]+\.[0-9]+$"
+                  if [[ ! "$TAG_NAME" =~ $REGEX_PATTERN ]]; then 
+                      echo 'Specified tag must match "^v[0-9]+.[0-9]+.[0-9]$"'
```
cdisselkoen:
```suggestion
                      echo 'Specified tag must match "$REGEX_PATTERN"'
```

cdisselkoen: In particular, the current `echo` message differs from the actual `REGEX_PATTERN` by a `+`

.github/workflows/publish.yml
```diff
@@ -0,0 +1,70 @@
+# Note: This workflow is intended to always be run from the 'main' branch to
+# ensure that the latest version of this workflow is used.
+name: Publish to crates.io
```
cdisselkoen: maybe we should have the name make clear that it's publishing cedar-policy, cedar-policy-cli, etc but not cedar-policy-symcc (which also lives in this Git repo)

cdisselkoen (approved):

.github/workflows/publish.yml
katherine-hough: Good point. Unfortunately if the name is too long the GitHub UI will cut it off. Is there a single word that describes these crates? How about "Publish base Cedar crates to crates.io"?

cdisselkoen: Sure

john-h-kastner-aws (approved):

.github/workflows/publish.yml
katherine-hough: That's accurate. I double checked this when I did some test runs on my fork. I will update the comment.

PR #2089, opened 2026-01-07
rewrite symcc's extractor.rs to better correspond to the Lean

cdisselkoen:
## Description of changes

Rewrites SymCC's extractor.rs to better correspond to Lean's Extractor.lean. In the process, fixes a bug (the added test case in this PR was failing but is now fixed).

### DIFF
--- a/cedar-policy-symcc/src/symcc/concretizer.rs
+++ b/cedar-policy-symcc/src/symcc/concretizer.rs
@@ -31,7 +31,6 @@ use cedar_policy_core::extensions::Extensions;
 use cedar_policy_core::parser::Loc;
 use miette::Diagnostic;
 use num_bigint::{BigInt, TryFromBigIntError};
-use ref_cast::RefCast;
 use smol_str::SmolStr;
 use thiserror::Error;
 


cedar-policy-symcc/src/symcc/extractor.rs
```diff
@@ -26,53 +26,86 @@
...
         Udf {
             table: Arc::new(new_table),
-            default: udf.default.clone(),
-            arg: udf.arg.clone(),
-            out: udf.out,
+            default: match &udf.out {
+                TermType::Set { ty } => factory::set_of([], (**ty).clone()),
+                _ => udf.default,
+            },
+            ..udf
```
cdisselkoen, line 108: I suspect this in particular is the bug fix

cedar-policy-symcc/tests/integration_tests.rs
```diff
@@ -1369,6 +1369,27 @@ async fn transitivity_and_in() {
...
+        r#"
+        permit(principal, action, resource)
+        when {
+            context.dept in context.depts
+        };
+    "#,
+        &validator,
+    );
+    assert_implies(&mut compiler, &pset1, &pset2, &envs).await;
+    assert_does_not_imply(&mut compiler, &pset2, &pset1, &envs).await;
```
john-h-kastner-aws, line 1391: I assume these also assert something about the decoded counter example?

Assuming it does, it might still be useful to have a direct test for decoding a supposed cycle. (easier to debug if it starts failing, can't change depending on cvc5 version or change to the smt query).

cedar-policy-symcc/src/symcc/extractor.rs
lianah: I see, before it used the default value in the model instead of setting it to the empty set for entities not in the footprint. Can line 106 ever happen? The ancestor function always returns a `Set` and this only gets called for ancestors.

cdisselkoen: Yeah it probably can't, so putting anything is sound. We just do what the Lean does, and therefore enjoy the Lean soundness proofs

cedar-policy-symcc/tests/integration_tests.rs
cdisselkoen: Yes, they check that the counterexample's request and entities both validate against the schema (although they do not actually check the authorization behavior of the counterexample).

I'll look into making a unit test for decoding

cdisselkoen: Spent an hour and cannot get a test which succeeds with the right footprint and fails with the wrong footprint -- must be doing something wrong when trying to initialize the `SymEnv` and `Interpretation` that `SymEnv::extract()` requires. Will delay this testing idea to a future PR.

PR #2090, opened 2026-01-08
expose json schema with resolved types functionality in cli

Swolebrain:
## Description of changes

Refactored the ffi functionality a little bit to move some of it to cedar-policy so it can be reused by both the ffi module and the cli. Added cli functionality to convert cedar schemas to json with resolved types.

## Issue #, if available

n/a

- [x] I'm not sure how my change impacts the documentation. (Post your PR anyways, and we'll discuss in the comments.)

### DIFF
--- a/cedar-policy-cli/src/lib.rs
+++ b/cedar-policy-cli/src/lib.rs
@@ -170,7 +170,10 @@ pub enum SchemaTranslationDirection {
     JsonToCedar,
     /// Cedar schema syntax -> JSON
     CedarToJson,
-    /// Cedar schema syntax -> JSON with all types resolved to entity or common
+    /// Cedar schema syntax -> JSON with all types resolved to entity or common.
+    ///
+    /// In contrast to `cedar-to-json`, this option requires that every type
+    /// referenced in the schema is also defined.
     CedarToJsonWithResolvedTypes,
 }
 
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -2144,7 +2144,7 @@ impl Schema {
 /// This is primarily meant to be used when working with schemas programmatically,
 /// for example when creating a schema building UI.
 ///
-/// Returns `Ok((json_value`, warnings)) on success, or Err(error) on failure.
+/// Returns `Ok((json_value, warnings))` on success, or `Err(error)` on failure.
 /// Fails if there are any types in the schema that are unresolved.
 pub fn schema_str_to_json_with_resolved_types(
     schema_str: &str,


cedar-policy/src/api.rs
```diff
@@ -2137,6 +2137,50 @@ impl Schema {
...
+/// Fails if there are any types in the schema that are unresolved.
+pub fn schema_str_to_resolved_fragment(
+    schema_str: &str,
+) -> Result<
+    (
+        json_schema::Fragment<cedar_policy_core::ast::InternalName>,
+        Vec<miette::Report>,
+    ),
+    Vec<miette::Report>,
+> {
```
cdisselkoen: Instead of returning warnings and errors as `miette::Report`, in this `cedar-policy` function can we return them as `SchemaError` and `SchemaWarning`, which are already public types? (Or whatever other specific error types are appropriate here.) This will also allow us to get rid of the code in CLI that detects parse errors just by looking for specific strings like "parse" in the error message. And the code for converting into `miette::Report`s can live in the `ffi` module

Swolebrain: wrap SchemaError and SchemaWarning in an enum for the Err type?

cdisselkoen: isn't it only SchemaError in the Err case? the warnings being in the Ok pair?

Swolebrain: yeah my bad, there's two errors in the Err case though, which was why I didn't do it this way. There's `CedarSchemaParseErrors` from converting the`&str` to `Fragment<RawName>` and `SchemaError` from converting from raw to internal. Should I wrap those two in an enum for this function? Didn't love the idea of creating a new public enum just for this

cedar-policy-cli/src/lib.rs
```diff
@@ -170,6 +170,8 @@ pub enum SchemaTranslationDirection {
     JsonToCedar,
     /// Cedar schema syntax -> JSON
     CedarToJson,
+    /// Cedar schema syntax -> JSON with all types resolved to entity or common
```
john-h-kastner-aws:
```suggestion
    /// Cedar schema syntax -> JSON with all types resolved to entity or common
    /// 
    /// In contrast to `cedar-to-json`, this option requires that every type
    /// referenced in the schema is also defined.
```

cedar-policy/src/api.rs
cdisselkoen: `CedarSchemaError` is an existing public enum covering this

cdisselkoen: (I think)

cedar-policy/src/api.rs
```diff
@@ -2137,6 +2137,50 @@ impl Schema {
...
+/// This is primarily meant to be used when working with schemas programmatically,
+/// for example when creating a schema building UI.
+///
+/// Returns Ok((fragment, warnings)) on success, or Err(errors) on failure.
+/// Fails if there are any types in the schema that are unresolved.
+pub fn schema_str_to_resolved_fragment(
+    schema_str: &str,
+) -> Result<
+    (
+        json_schema::Fragment<cedar_policy_core::ast::InternalName>,
```
Swolebrain: is it ok that this function in `cedar_policy::api` is leaking  `cedar_policy_core::ast::InternalName`? Should i make it return something else, like maybe `serde_json::Value`?

cdisselkoen: It's also leaking `json_schema::Fragment`, if I'm not mistaken.

`serde_json::Value` sounds good to me -- I think both the CLI and FFI don't need anything more detailed than that.

Swolebrain: yeah it's definitely leaking that too

cedar-policy/src/api.rs
```diff
@@ -2137,9 +2137,140 @@ impl Schema {
...
 }
 
+/// Convert a Cedar schema string to JSON format with resolved types.
+///
+/// This function resolves ambiguous "`EntityOrCommon`" types to their specific
+/// Entity or `CommonType` classifications using the schema's type definitions.
+/// This is primarily meant to be used when working with schemas programmatically,
+/// for example when creating a schema building UI.
+///
+/// Returns `Ok((json_value`, warnings)) on success, or Err(error) on failure.
```
john-h-kastner-aws:
```suggestion
/// Returns `Ok((json_value, warnings))` on success, or `Err(error)` on failure.
```

Issue #1895, opened 2025-10-14
Enhanced TPE Simplification for non-erroring expressions


john-h-kastner-aws:
### Category

Other

### Describe the feature you'd like to request

The implementation of TPE (and the concrete evaluator) treats policy scopes the same as any other expression. This causes us to miss some seemingly obvious simplification. If there residual expression in the policy scope, followed by a `false` expression either from the policy scope or condition, TPE will never reduce the whole policy to `false` even though it should be possible.

For example, given a policy 

```cedar
forbid (
  principal in UserGroup::"interns",
  action == Action::"delete",
  resource
);
```

Running tpe with any action other than `Action::"delete"` will give a residual

```cedar
forbid (
  principal,
  action,
  resource
) when {
  principal in UserGroup::"interns" && false
};
```

We cannot reduce `e && false` to false in general, so partial evaluation stops here, but it is unfortunate that a policy with a trivial `false` scope condition cannot be reduced to `false` by partial evaluation. We know that scope conditions always evaluate to a boolean and never error, so we should be able to simplify this to `false`.

To motivate this simplification, consider the new `queryAction` function using tpe. If this function is not given a concrete principal, it will _never_ be able to conclude that the `forbid` policy doesn't apply, even for actions that _aren't_ Action::"delete". All authorized actions will be returned as "possibly allowed" as never definitely allowed. Our main use case fortunately doesn't differentiate between these cases.

Note that the "never errors" condition is stronger than what is guaranteed by typechecking. A well type policy may still have integer overflow or entity-does-not-exist errors, which are just as much a problem as type errors. Some expression, and in fact all  policy scope expressions, are guaranteed to never evaluate to an error.

For non-type-aware partial evaluation, we implemented something similar to this in the [is_projectable](https://github.com/cedar-policy/cedar/blob/562bdfe8639eb8e55ce98769d7fb257bcb82814d/cedar-policy-core/src/ast/expr.rs#L334) function. That function as implemented isn't quite strong enough. For example, it doesn't claim that `in` or `&&` will never error. This was necessary because these expression can have type errors.

For TPE, we know the expressions are well typed, so we can rule out the possibility of type error. We then know that `in` and `&&` can't ever error (assuming their operands don't error), enabling the additional reduction.

luxas:
[...]

Issue #1736, opened 2025-08-01
Return (a superset of) EvaluationError from TPE evaluation


luxas:
### Category

User level API features/changes

### Describe the feature you'd like to request

When evaluating a policy in TPE mode, normal evaluation errors might happen. For example, an arithmetic overflow can happen at any point. Right now, the context on what part of the evaluation failed is not exposed as part of `Residual::Error`. In order to get the same diagnosis possibilities as for normal evaluation, it'd be nice to get the same error causes exposed.

I was discussing with @shaobo-he-aws on Slack about this as well.

### Describe alternatives you've considered

-

### Is this something that you'd be interested in working on?

- [x] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

oflatt:
Once this is done, we should also change batched evaluation to return a full concrete `Response` including `Diagnostics` instead of just a `Decision`.

luxas:
I started working on this in #2162

PR #2091, opened 2026-01-09
TPE: Simplify `<residual> && false` to `false` when `<residual>` cannot error

luxas:
## Description of changes

Adds a method that determines if a residual can produce runtime errors that the validator would not already warn the user about (hence the "well-typed" assumption). See #1895 for further discussion. The function could be made slightly more precise by indicating what extension functions can and cannot error.

The "can error" method is used to simplify `<residual> && false` to `false` and `<residual> || true` to `true` when `<residual>` cannot error. Simplification cannot be done when the residual could error.

## Reviewer Notes

Note that there are lots of typing errors that can be thrown by `Evaluator.interpret` on arbitrarily constructed `Residual`s. So do you prefer

a) Making `Residual` construction private, so it can only be constructed by the typechecker after a successful pass, or
b) Somehow trying to inline what the validator would do in `Evaluator.interpret` every time we need to know if a sub-expr is well-formed or not (complex and seems slow), or
c) Keep the behavior undefined for anyone that constructs semantically invalid `Residual`s (seems dangerous)

I guess I lean towards a), at least until we know if anyone needs those enums to be public.

In addition, I think it'd be a good idea to use separate `Residual::Error` in to `Residual::TypeError` and `Residual::RuntimeError`, where Residual::TypeError` should never happen (if validator is sound), but `Residual::RuntimeError` could indeed. This might be a good time to add the same context to `Residual::RuntimeError` as for normal evaluation (https://github.com/cedar-policy/cedar/issues/1736).

## Issue #, if available

Fixes: #1895 
(or at least gets us some way towards that goal)

The change in this PR is

I confirm that this PR

- [x] Updates the "Unreleased" section of the CHANGELOG with a description of this change.

- [x] Requires updates, @john-h-kastner-aws was looking at those already.

### DIFF
--- a/cedar-policy/CHANGELOG.md
+++ b/cedar-policy/CHANGELOG.md
@@ -23,7 +23,7 @@ the `ffi` module (new impls on `ffi::EntityUid`, `ffi::Context`, `ffi::Entities`
 `ffi::Template`, and `ffi::StaticPolicySet`) (#2085)
 - `schema_to_json_with_resolved_types()` function, which takes in a Cedar schema and returns a json schema without any instances of EntityOrCommon; they're all either Entity or CommonType (#2058)
 - More derives (`PartialEq`, `Clone`, etc) for a number of types in the `ffi` module (#2083)
-
+- TPE: Simplify `<residual> && false` to `false` and `<residual> || true` to `true` when `<residual>` is error-free. (#2091)
 
 ## [4.8.2] - 2025-12-09
 


cedar-policy-core/src/tpe/residual.rs
```diff
@@ -85,6 +85,78 @@ impl Residual {
...
+                    test_expr.can_error_assuming_well_formed()
+                        || then_expr.can_error_assuming_well_formed()
+                        || else_expr.can_error_assuming_well_formed()
+                }
+                ResidualKind::Is { expr, .. } => expr.can_error_assuming_well_formed(),
+                ResidualKind::Like { expr, .. } => expr.can_error_assuming_well_formed(),
+
+                ResidualKind::BinaryApp { op, arg1, arg2 } => match op {
+                    // Arithmetic operations could error due to integer overflow
+                    ast::BinaryOp::Add => true,
```
luxas, line 118: Are there other cases than these, or gaps in the validator coverage that I'd need to know about?

cedar-policy-core/src/tpe/residual.rs
```diff
@@ -85,6 +85,78 @@ impl Residual {
...
+                    | ast::BinaryOp::LessEq => {
+                        arg1.can_error_assuming_well_formed()
+                            || arg2.can_error_assuming_well_formed()
+                    }
+                },
+
+                // Extension function invocations can error at runtime.
+                ResidualKind::ExtensionFunctionApp { .. } => true,
+
+                ResidualKind::GetAttr { expr, .. } => expr.can_error_assuming_well_formed(),
```
john-h-kastner-aws: This and `getTag` are the other important cases. They will error if the target entity isn't in the entity data.

john-h-kastner-aws: These also gets at some tricky cases around optional attributes. The type annotated AST doesn't carry the capability information used to check that optional attributes will be present, so just having a locally well typed resdiual isn't enough to guarantee that there will never be an unknown attribute error.  (the residuals resulting from typechecking will satisfy this property, but it's trickier to show that is holds for the residual resulting from partial eval).

john-h-kastner-aws:
> Note that there are lots of typing errors that can be thrown by Evaluator.interpret on arbitrarily constructed Residuals. So do you prefer
>
> ...
> c) Keep the behavior undefined for anyone that constructs semantically invalid Residuals (seems dangerous)

We're fine with (c) so long as you're talking about the behavior when calling `cedar-policy-core` directly. The `cedar-policy` API should rule out this sort of error, and to my knowledge it does.

In other words, our point of view is that we're doing (a) (keeping the constructor private) as long as it's not exported from `cedar-policy`.

john-h-kastner-aws:
for other reviewers: a partial Lean proof of this simplification is in this branch https://github.com/cedar-policy/cedar-spec/tree/tpe_and_false

The proof is complete (no `sorry`s) for the part of this optimization it implements, but it only implements the simplification for `&&`, and it has a narrower definition of non-erroring expressions. We'll need to extend it to match this PR, but it's enough to be confident that the simplification is correct (up to the issue with `GetAttr` and `GetTag`)

cedar-policy-core/src/tpe/residual.rs
```diff
@@ -85,6 +85,78 @@ impl Residual {
             Residual::Error(ty) => ty,
         }
     }
+
+    /// Whether this residual can result in a runtime error, assuming that self is well-formed, that is, has been validated against a schema.
+    pub fn can_error_assuming_well_formed(&self) -> bool {
+        match self {
+            Residual::Concrete { .. } => false,
+            Residual::Error(_) => true,
```
john-h-kastner-aws, line 93: looks like this line isn't covered by any tests. Seems worth writing one

cedar-policy-core/src/tpe/residual.rs
luxas: This is why I was suggesting that the only way to construct a `Residual` (even in `cedar-policy-core`) would be from an `Expr` that has been validated. For such a Residual, the validator made sure through its capabilities that any `getAttr` or `getTag` present (in any sub-expr of the root) cannot error, right? Even (and especially) if the attribute is optional.

This is why I am suggesting to narrow down how `Residual`s can be constructed (even in `cedar-policy-core`), so that we can uphold the invariant that `getAttr` or `getTag` can never error, as any `Residual` instance had to necessarily pass validation even before becoming constructed.

luxas:
> In other words, our point of view is that we're doing (a) (keeping the constructor private) as long as it's not exported from cedar-policy.

See my comment https://github.com/cedar-policy/cedar/pull/2091#discussion_r2686279947; I was thinking to make the `Residual` constructor even more private (even in `cedar-policy-core`), such that it can only be constructed from an `Expr`, `PartialRequest` and a `Schema`, e.g. `tpe::make_residual(e: Expr, req: &PartialRequest, s: &ValidatorSchema) -> Result<Residual, TpeError>`. The validator thus ensures that `getAttr` and `getTag` can never error at anywhere in the residual constructed this way. Could this be proved to be the case also in the Lean model?

cedar-policy-core/src/tpe/residual.rs
john-h-kastner-aws: The optional attribute point is just an aside. It doesn't impact correctness, just makes it a pain to prove. 

The real issue is that `getTag` and `getAttr` will both error whenever the entity they access doesn't exist. This is slightly subtle for TPE since the missing entity is initially treated as an unknown value (resulting in a non-concrete residual rather than an error), but when we go to concretely evaluate the residual, it's still an error if the data for the entity isn't present.

Validation doesn't do anything to guard against this sort of error. It's just one of the gaps in it's soundness theorem.

luxas: Right, you were thinking about errors from concrete reauthorization too; I was only thinking about whether TPE itself could produce errors, which `getTag` and `getAttr` would not (a missing entity would just yield another residual). Good catch. That limits the usefulness of this rewrite quite a bit, hmm... 

To some extent, that type of error "feels different" though, in that the schema "promised" that all entity references could be followed, but the implementation did not live up to that promise. Maybe one way to find a middle-ground here would be to let the implementation (the TPE `Evaluator`) provide (or not) explicitly whether it guarantees that all referenced entities exist, or if it cannot guarantee this.

However, this PR (even if I add that these operators can error too) gets us some of the way, and most likely the `possible_bool_outcomes` described in https://github.com/cedar-policy/cedar/issues/1895#issuecomment-3720043450 gets us the rest of the way for those users that do not care about the error precision between `false` and `error` for permit policies, for example.

john-h-kastner-aws: The fact that `getAttr` can error definitely limits the usefulness of this optimization since it folding away an attribute access to eliminate the need to fetch that entity data is exactly what I'd ideally want it to do.

I guess this just points us towards looking at the `PossibleBoolOutcome` idea or changing the auth algorithm to account for the possibility of errors.

john-h-kastner-aws: > To some extent, that type of error "feels different" though, in that the schema "promised" that all entity references could be followed, but the implementation did not live up to that promise. Maybe one way to find a middle-ground here would be to let the implementation (the TPE Evaluator) provide (or not) explicitly whether it guarantees that all referenced entities exist, or if it cannot guarantee this.

There's a possibility of something like this, probably by using level-based policy validation and entity slicing as a pre-condition, but I don't think we'll want to make this change too quickly.

luxas: > folding away an attribute access to eliminate the need to fetch that entity data is exactly what I'd ideally want it to do

Yes, this exactly 👍 

> There's a possibility of something like this, probably by using level-based policy validation and entity slicing as a pre-condition, but I don't think we'll want to make this change too quickly.

Cool, let's discuss this separately then in #1895 or a new after we've merged this initial optimization.

I'll make the change for `getAttr` and `getTag` next. Should I also rename the `From<Expr<Option<Type>>> for Residual` "constructor" and make `Residual(Kind)` private (as per option `a` above), and instead add the constructor `tpe::make_residual(e: Expr, req: &PartialRequest, s: &ValidatorSchema) -> Result<Residual, TpeError>` that makes sure `Evaluator::interpret` is always sound? (If not, the `interpret` function would be unsound on `Residual`s that are not validated)

john-h-kastner-aws: >  Should I also rename the From<Expr<Option<Type>>> for Residual "constructor" and make Residual(Kind) private (as per option a above),

I'd rather not make this now. Let's get the simplification merged and then consider this API change separately. I'm leaning towards making it (even if it's not necessary to maintain invariants  from `cedar-policy`) since I don't think it will causes us any problems anywhere else

luxas:
PR is updated now, `getTag` and `getAttr` can also error according to the function. After this change, the following are the only boolean-valued operators one can safely use on the LHS of `&&` or RHS of `||` to get this simplification:

Composite expressions:
- `if`, given the `then` and `else` branch return some boolean of these other error-free exprs
- `||` and `&&` in nested expressions
- `!`

"Leaf" boolean expressions, when neither LHS (or RHS, if applicable) contain any dereferencing operators.
- `==`
- `in`
- `hasTag`
- `has`
- `is`
- `contains` (only with literal sets)
- `containsAll` (only with literal sets)
- `containsAny` (only with literal sets)

This means in short that we can simplify the policy header, but most likely nothing else, as `getAttr` is very likely used in the `when` section, right after `has` or `hasTag`. 

But let's merge this now, and then follow up on further improvement! 😄 I'll prototype the `possible_bool_outcomes` idea, and see what it looks like in practice. 

In a follow-up PR I also want to separate the the possible errors into `Runtime` (e.g. arithmetic overflow), `EntityNotExist` (`getAttr`, `getTag`), and `Validation` (by making the constructor always validate, as we discussed), and thus fix #1736. By having the `EntityNotExist` error type distinct, one could even return two distinct TPE `Response`s in the high level API if one wanted, one which can do concrete reauthorization (and thus cannot fold `EntityNotExist` errors) and one which cannot reauthorize concretely, just reauthorize using TPE (akin to batched evaluation), and thus _can_ fold possible EntityNotExist` errors more aggressively.

I'll take a look at the cedar-lean PR. Do you usually merge the Rust or Lean PRs first?

cedar-policy-core/src/tpe/residual.rs
```diff
@@ -85,6 +85,81 @@ impl Residual {
             Residual::Error(ty) => ty,
         }
     }
+
+    /// Whether this residual can result in a runtime error, assuming that self is well-formed, that is, has been validated against a schema.
+    pub fn can_error_assuming_well_formed(&self) -> bool {
```
john-h-kastner-aws, line 90: nit: `_assuming_well_typed` matches terminology in the Lean model

john-h-kastner-aws (approved):
lgtm

john-h-kastner-aws:
> I'll take a look at the cedar-lean PR. Do you usually merge the Rust or Lean PRs first?

Typically Rust before Lean, the only hard requirement is that both are in before we do a release.  Ideally we'll merge these on the same day to avoid noise in our nightly differential testing.

cdisselkoen (approved):
Looks great, thanks for the thorough tests as well.

john-h-kastner-aws:
Merging this! Please do take a closer look at the Lean PR, we can tweak the Rust/Lean in follow up PRs if you see anything.

cedar-policy-core/src/tpe/residual.rs
luxas: That's cool as well, I can rename for consistency in some follow-up 👍

PR #2096, opened 2026-01-12
Added workflow for publishing cedar-wasm

katherine-hough:
## Description of changes
* Restructured the "publish" GitHub workflow to put validation into a separate job.
* Added "publish_wasm" workflow to publish cedar-wasm to NPM. Note that I have not yet set up trusted publishing, but this workflow is intended to use trusted publishing. 

## Issue #, if available
N/A

### DIFF
--- a/.github/workflows/publish_wasm.yml
+++ b/.github/workflows/publish_wasm.yml
@@ -94,7 +94,7 @@ jobs:
                       console.log(`Latest version found on NPM: ${latestRelease}`);
                       let npmTag;
                       if (compareVersions.compare(version, latestRelease, '<=')) {
-                        npmTag = `v${major}.${minor}-lts`;
+                        npmTag = `v${major}.${minor}-x`;
                       }  else {
                         npmTag = 'latest';
                       }
@@ -146,7 +146,7 @@ jobs:
               # 
               # We explicitly set the tag to "latest", if the version being released is the latest
               # version according to semantic versioning.
-              # Otherwise, we use the tag "v<MAJOR>.<MINOR>-lts".
+              # Otherwise, we use the tag "v<MAJOR>.<MINOR>-x".
               run: npm publish --access=public --tag "$NPM_TAG"
               working-directory: ./cedar-wasm/pkg/
               env:


.github/workflows/publish_wasm.yml
```diff
@@ -0,0 +1,148 @@
...
+              run: TEST_TS=1 ./build-wasm.sh
+            # Check that the tag matches the generated node package version.
+            - name: Check package version
+              run: test "v$(jq --raw-output .version ./cedar-wasm/pkg/package.json)" = "$TAG_NAME"
+              env:
+                  TAG_NAME: ${{ inputs.tag }}
+            - name: Publish to NPM
+              # Note: npm will implicitly apply the "latest" tag if the `--tag` option is not specified,
+              # so for other releases we use the tag v<MAJOR>.<MINORR>-lts
+              run: npm publish --access=public --tag "$NPM_TAG"
```
cdisselkoen: I don't understand this comment. The code here always specifies the `--tag` option

katherine-hough: I was trying to explain why we are bothering to use the tag option at all, since our previous release process did not use it. How about this:

```
              # Note: npm will implicitly apply the "latest" tag if the `--tag` option is not specified.
              # If the version to be released is less than the previously published "latest" version and 
              # implicitly tagged as the "latest", NPM will produce an error.
              # 
              # We explicitly set the tag to "latest", if the version being released is the latest
              # version according to semantic versioning.
              # Otherwise, we use the tag "v<MAJOR>.<MINOR>-lts".
```

katherine-hough: Maybe we should use a different suffix other than lts? I did it because it's [what angular uses](https://www.npmjs.com/package/@angular/core?activeTab=versions), but really we only support the latest minor so it's potentially confusing. NPM will not allow `v<MAJOR>.<MINOR>` because it can conflict with existing version numbers. We could do `v<MAJOR>.<MINOR>-patch` or `v<MAJOR>.<MINOR>-x`?

john-h-kastner-aws: that's a good point, we don't have "long term support" versions.

Both alternative namings sound good to me, so happy to go with the `-x` in the current revision of this PR

PR #2099, opened 2026-01-14
more generic extractor implementation

cdisselkoen:
## Description of changes

Corresponds to https://github.com/cedar-policy/cedar-spec/pull/826 on the Lean side. See notes there.

Open to naming suggestions for the type representing a `CompiledPolicy(ies)`, which on the Lean side I've named `CompiledPolicyₛ` for now, and thus on the Rust side I've just named `CompiledPolicys` for now.

### DIFF
--- a/cedar-policy-symcc/src/symccopt.rs
+++ b/cedar-policy-symcc/src/symccopt.rs
@@ -40,13 +40,12 @@ impl<S: Solver> SymCompiler<S> {
     /// Optimized version of `sat_asserts()`.
     ///
     /// Corresponds to `satAssertsOpt?` in the Lean.
-    pub async fn sat_asserts_opt<'a, I: Iterator<Item = &'a CompiledPolicys<'a>> + Clone>(
+    pub async fn sat_asserts_opt<'a>(
         &mut self,
         asserts: &Asserts,
-        cps: impl IntoIterator<Item = &'a CompiledPolicys<'a>, IntoIter = I>,
+        cps: impl IntoIterator<Item = &'a CompiledPolicys<'a>> + Clone,
     ) -> Result<Option<Env>> {
-        let mut cps = cps.into_iter().peekable();
-        match cps.peek() {
+        match cps.clone().into_iter().next() {
             None => Err(Error::NoPolicies),
             Some(cp_s) => match self.check_sat_asserts(asserts, cp_s.symenv()).await? {
                 None => Ok(None),
--- a/cedar-policy-symcc/src/symccopt/compiled_policies.rs
+++ b/cedar-policy-symcc/src/symccopt/compiled_policies.rs
@@ -186,6 +186,7 @@ impl CompiledPolicies {
 
 /// Represents a `CompiledPolicy` or a `CompiledPolicies`, for APIs that don't care
 /// which one they get.
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum CompiledPolicys<'a> {
     Policy(&'a CompiledPolicy),
     Policies(&'a CompiledPolicies),
--- a/cedar-policy-symcc/src/symccopt/extractor.rs
+++ b/cedar-policy-symcc/src/symccopt/extractor.rs
@@ -31,14 +31,14 @@ use crate::{err::ConcretizeError, Env, Interpretation};
 ///
 /// Caller guarantees that all of the `CompiledPolicys` were compiled for the same `symenv`.
 pub fn extract_opt<'a>(
-    cps: impl Iterator<Item = &'a CompiledPolicys<'a>> + Clone,
+    cps: impl IntoIterator<Item = &'a CompiledPolicys<'a>> + Clone,
     interp: &Interpretation<'_>,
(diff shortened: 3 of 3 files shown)


cedar-policy-symcc/src/symccopt.rs
```diff
@@ -42,36 +43,14 @@ impl<S: Solver> SymCompiler<S> {
...
-    ) -> Result<Option<Env>> {
-        match cps.clone().into_iter().peekable().peek() {
-            None => Err(Error::NoPolicies),
-            Some(CompiledPolicy { symenv, .. }) => {
-                match self.check_sat_asserts(asserts, symenv).await? {
-                    None => Ok(None),
-                    Some(interp) => Ok(Some(CompiledPolicy::extract_opt(cps, &interp)?)),
-                }
-            }
+            Some(cp_s) => match self.check_sat_asserts(asserts, cp_s.symenv()).await? {
```
john-h-kastner-aws, line 50: naming nit: having `cps` and then `cp_s` as an element of `cps` is mildly confusing

cedar-policy-symcc/src/symccopt.rs
```diff
@@ -42,36 +43,14 @@ impl<S: Solver> SymCompiler<S> {
     pub async fn sat_asserts_opt<'a>(
         &mut self,
         asserts: &Asserts,
-        cps: impl IntoIterator<Item = &'a CompiledPolicies> + Clone,
+        cps: impl IntoIterator<Item = &'a CompiledPolicys<'a>> + Clone,
     ) -> Result<Option<Env>> {
-        match cps.clone().into_iter().peekable().peek() {
+        match cps.clone().into_iter().next() {
```
john-h-kastner-aws, line 48: Does changing `peek` to `next` effect the behavior of this function? I see `cps` is used again later.

cedar-policy-symcc/src/symccopt.rs
cdisselkoen: I agree. In Lean I am currently using `cp` for a variable of type `CompiledPolicy`; `cps` for a variable of type `CompiledPolicies`; and `cpₛ` for a variable of type `CompiledPolicyₛ`. In Rust the translation of `cpₛ` to `cp_s` is confusing. Many other places we've translated terse Lean variable names to more readable (if more verbose) Rust variable names. Any suggestions?

Relatedly, if you have a better suggestion for the name of the `CompiledPolicyₛ` Lean type / `CompiledPolicys` Rust type, I'm all ears -- see description of both PRs

cedar-policy-symcc/src/symccopt.rs
cdisselkoen: It does not, because of the `.clone()`. I can leave this as `peek` though if it's more clear.

cdisselkoen: surely the performance is the same, I think?

john-h-kastner-aws: Make sense, doesn't matter too much to me.

Though, now I'm wondering why we `clone` the whole object just to get the first element

cdisselkoen: We are just cloning the iterator itself, which is supposed to be cheap (no way to really guarantee this, but iterators returning `&'a` types are usually going to be just a pointer traversing some `Vec` or etc, ideally)

The reason we can't just `.peek()` without cloning is because we need to continue to use `cp_s` after the later call that actually consumes `cps`

john-h-kastner-aws: Looks like this is actually cloning an `IntoIterator` (which based on the call sites is at most an array with two elements, so still cheap, but I guess it could end up being a large vec somewhere)

cdisselkoen: definitely a possible performance footgun in the future. do you think it would be better to have this function to take `&[&'a CompiledPolicys<'a>]`? this would allow us to get the first element without cloning anything (specifically because the `IntoIterator` in this case, the `&[]`, is `Copy`). but callers might end up having to `collect`/allocate in order to call this function.

john-h-kastner-aws: I haven't thought about what this looks like in any downstream functions, but the other option might be taking the first element as a separate argument.

but this code looks fine for now. If you think if something that's betterm I'm happy to review that change separately.

cedar-policy-symcc/src/symccopt.rs
chaluli: Hmm, what about

CompiledPolicy (single policy)
CompiledPolicySet (policyset)
CompiledPolicies (single policy or policyset)

cp: CompiledPolicy
cpset : CompiledPolicySet
cps: CompiledPolicies

for naming conventions.

Though, you do still have naming conflicts for containers/iterators of CompiledPolicy and CompiledPolicies.

cdisselkoen: Sounds like a good suggestion. Will make the change in a followup PR for clarity.

PR #2100, opened 2026-01-16
Optimize context json parsing

benelser:
## Description of changes

This PR removes redundant deserialization when converting an already-parsed `serde_json::Value` into `CedarValueJson`.

Previously, the `from_json_value` path re-entered Serde via `serde_json::from_value`, incurring a full visitor-based deserialization and allocating intermediate structures. This PR introduces a direct conversion path (`CedarValueJson::from_serde_value_direct`) that operates directly on `serde_json::Value`, preserving semantics while avoiding the extra pass.

**Impact:**
- Eliminates double parsing on the `from_json_value` path
- Reduces memory amplification from **7.27× → 1.02×** (≈ **85% reduction**)
- Improves performance and allocation behavior in security- and policy-critical context parsing

Semantic equivalence with the previous implementation was validated via call-graph analysis, error-behavior review, and targeted edge-case checks (including escape handling and error classification).

---

## Issue #, if available

N/A

---

- [x] Does not update the CHANGELOG because this change does not affect public APIs or user-visible behavior.

- [x] Does not require updates because this change does not affect the Cedar formal model or DRT infrastructure.

- [x] Does not require updates because this change does not affect the Cedar language specification.

### DIFF
--- a/cedar-policy/benches/context_json_parsing.rs
+++ b/cedar-policy/benches/context_json_parsing.rs
@@ -21,12 +21,6 @@
 //! This benchmark measures:
 //! - Parsing performance (wall time) for various input sizes and structures
 //! - Memory usage (RSS) to identify amplification issues
-//!
-//! For detailed heap allocation profiling, use:
-//! ```bash
-//! cargo bench --bench context_json_parsing --features heap-profiling
-//! ```
-//! This will use dhat-rs to track all heap allocations and produce detailed reports.
 
 use std::hint::black_box;
 
@@ -34,13 +28,6 @@ use cedar_policy_core::entities::json::{ContextJsonParser, NullContextSchema};
 use cedar_policy_core::extensions::Extensions;
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
 use serde_json::json;
-
-#[cfg(feature = "heap-profiling")]
-use dhat::DhatAllocator;
-
-#[cfg(feature = "heap-profiling")]
-#[global_allocator]
-static ALLOCATOR: DhatAllocator = DhatAllocator;
 
 /// Generate a large context JSON with a single large string value
 fn generate_large_context_string(size_mb: usize) -> serde_json::Value {
@@ -129,7 +116,6 @@ fn get_memory_rss() -> Option<usize> {
 
 /// Benchmark context parsing with memory measurement
 /// Uses RSS (Resident Set Size) for memory tracking
-/// For more detailed heap allocation tracking, use dhat-rs with --features heap-profiling
 fn bench_context_parsing_memory(c: &mut Criterion) {
     let mut group = c.benchmark_group("context_json_parsing_memory");
 


john-h-kastner-aws:
This optimization makes sense to me and looks promising, but we're very concerned with maintaining the correctness of Cedar and not accidentally introducing any breaking changes. 

Can you please elaborate on what exactly you mean by 

> Semantic equivalence with the previous implementation was validated via call-graph analysis, error-behavior review, and targeted edge-case checks (including escape handling and error classification).

I don't see any additional testing or static analysis included in this PR

Also, you quote some numbers for reduced memory usage, but do you also have any numbers for the speed up (if any) in performance (either in benchmarks or real applications)?  We'd be particularly interested in hearing about applications of Cedar that are running into performance issues.

cedar-policy-core/src/entities/json/value.rs
```diff
@@ -801,7 +927,8 @@ impl<'e> ValueParser<'e> {
                 )),
                 val => {
                     let actual_val = {
-                        let jvalue: CedarValueJson = serde_json::from_value(val)?;
+                        // Use direct conversion for consistency and performance
+                        let jvalue = CedarValueJson::from_serde_value_direct(val, ctx)?;
```
john-h-kastner-aws, line 935: What does "for consistency" mean in this comment? Is it implying there may be a  difference in behavior, or is it simply stating that you use it here because you use it below? 

If the latter, I don't think this comment adds any value

benelser: Hi John! Yes, it was the latter. I went ahead and removed. Good call out.

benelser:
Thanks @john-h-kastner-aws. I agree that correctness is the primary concern.

I have added explicit tests to back up the “semantic equivalence” claim. The new test module compares the old path (`serde_json::from_value::<CedarValueJson>`) and the new direct path (`from_serde_value_direct`) on a corpus of `serde_json::Value` inputs. It asserts identical `CedarValueJson` results on success, and matches error category on failure. It also includes directed cases for `__entity`, `__extn` (single and multi arg), `__expr` (reserved), and multi key objects containing escape keys to ensure they fall back to `Record`. Null is the one intentional difference at this layer. Serde can produce the `CedarValueJson::Null` variant, while the direct path rejects `Value::Null` immediately. Both code paths still reject null when converting to an expression via `into_expr`, and the tests document that.

On performance, the benchmarks are in `cedar-policy/benches/context_json_parsing.rs`. They measure wall time via Criterion for nested contexts and many attributes, and they include a `from_json_value` benchmark that targets the optimized path specifically. The memory amplification numbers in the PR come from the RSS-based measurements in this benchmark suite.

benelser:
Hey @john-h-kastner-aws, just checking in on this one. I’m curious if the new tests and benchmark data I added covered the concerns you had around potential breaking changes? 

Everything is currently green in CI with solid coverage, but let me know if there's any other specific edge case you're wondering about that I should verify.

benelser:
@john-h-kastner-aws anything else you wanted me to take a look at?

victornicolet:
This PR might be a partial solution to https://github.com/cedar-policy/cedar/issues/2265

Issue #2015, opened 2025-12-02
CLI `run-tests` should accept schema and validate policies+entities+requests


john-h-kastner-aws:
### Category

CLI features/changes

### Describe the feature you'd like to request

The `run-tests` command at the CLI takes asserts lets you assert that some request gives the expected authorization decision, but it's easy to write incorrect tests with invalid entity data or request variables.

E.g., this test case might be invalid if `read` only applies to `Document`, not `Dir`, but the test might still pass, depending on the applicable policies. 

```json
[
  {
    "name": "alice_can_read",
    "request": {
      "principal": "User::\"alice\"",
      "action": "Action::\"read\"",
      "resource": "Dir::\"dir1\"",
      "context": {}
    },
    "entities": [],
    "decision": "allow",
    "reason": [],
    "num_errors": 0
  }
]
```

The `run-tests` command should accept a `--schema` argument and use it to validate that requests and entities for each test case before running the test. Invalid requests should be treated as test failures. The `--schema` argument should behave the same as `--schema` for `cedar authorize`.

### Describe alternatives you've considered

.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

repairman29:
Hi! I'm interested in helping with this issue. 

I have experience with index (repairman29/ai-gm-service) and think I can contribute here. Would love to discuss the approach and see if we can make this happen.

*(Connected via [Echeo](https://echeo.io) - a platform for matching developers with open source bounties)*

cdisselkoen:
#2101 already has this issue, it looks like.

PR #2101, opened 2026-01-17
Add schema validation for the run-tests command

jv-garcia:
- **Fix typo in run tests functionality**
- **Remove unneeded clone**
- **Add optional schema args to run tests**
- **Deserialize seed wired**
- **Simplify run_tests function**
- **Wire schema file to actually be used in the tests for run tests**
- **Add policy set validation for the run tests command**
- **Add tests for new schema validation in run tests**
- **Add changelog entry to new feature in run tests command**

## Description of changes

Hi folks, greetings :) I was looking for things to contribute on Rust projects, saw #2015 and decided to take a stab at it. Here are some general observations:

* The main weirdness is perhaps the use of the `[DeserializeSeed` trait](https://docs.rs/serde/latest/serde/de/trait.DeserializeSeed.html) to do schema validation for requests and entities. I didn't see any other API to do validation post-creation, so I thought this would be a decent approach which seems to be somehow `serde` idiomatic. Happy to change to something else if there are better APIs. I felt nevertheless that some way to do Schema checking on JSON deserialization would be nice outside of this PR, happy to work on something for `cedar_policy` for example if you folks think that'd be worth it (and I didn't miss the obvious way to do this).
* Something that might be weird is also the use of an `Option<&Schema>` for the Seed struct. I thought that for the purpose of this PR it actually fits the purpose pretty straightforwardly since it's not a crate API.
* New tests are a bit meh, but well, they're something :)
* The current way to report results for the `run_tests` command is a bit weird since it doesn't really contemplate the possibility of validation failures (so, in case of just a validation failure in the command, you still have to expect Failure). I'm assuming that this is mostly OK as it is, but do let me know if we should rethink that in this or an upcoming PR.

~~PS: the failing test seems unrelated?~~ Nvm, should be fixed.

## Issue #, if available
#2015 

- [X] A bug fix or other functionality change requiring a patch to `cedar-policy`. (I think

### DIFF
--- a/cedar-policy-formatter/tests/cli-snapshots/cedar_policy_formatter__pprint__fmt__tests__format_files@tiny_sandboxes__sample11__invalid_policy.cedar.snap
+++ b/cedar-policy-formatter/tests/cli-snapshots/cedar_policy_formatter__pprint__fmt__tests__format_files@tiny_sandboxes__sample11__invalid_policy.cedar.snap
@@ -1,0 +1,10 @@
+---
+source: cedar-policy-formatter/src/pprint/fmt.rs
+expression: formatted
+input_file: cedar-policy-cli/sample-data/tiny_sandboxes/sample11/invalid_policy.cedar
+---
+permit (
+  principal == User::"Alice",
+  action == Action::"Invalid",
+  resource
+);
--- a/cedar-policy-formatter/tests/cli-snapshots/cedar_policy_formatter__pprint__fmt__tests__format_files@tiny_sandboxes__sample11__valid_policy.cedar.snap
+++ b/cedar-policy-formatter/tests/cli-snapshots/cedar_policy_formatter__pprint__fmt__tests__format_files@tiny_sandboxes__sample11__valid_policy.cedar.snap
@@ -1,0 +1,10 @@
+---
+source: cedar-policy-formatter/src/pprint/fmt.rs
+expression: formatted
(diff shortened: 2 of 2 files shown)


cedar-policy-cli/src/lib.rs
```diff
@@ -1806,15 +1809,29 @@ fn compare_test_decisions(test: &TestCase, ans: &Response) -> TestResult {
...
+fn run_one_test(
+    policies: &PolicySet,
+    test: &serde_json::Value,
+    schema: Option<&Schema>,
+) -> Result<TestResult> {
+    let test = CheckedTestCaseSeed(schema)
+        .deserialize(test.into_deserializer())
+        .into_diagnostic()?;
+    if let Some(schema) = schema {
+        let validator = Validator::new(schema.clone());
```
john-h-kastner-aws, line 1821: can we have `run_one_test` accept a `&Validator` to avoid cloning the schema for every test case?

jv-garcia:
Apologies folks, I’ve been on vacation and not really able to work on this.
I’d be happy to take a look at improvements (or other work 🙂) when I’m
back in about three weeks.

John Kastner ***@***.***> schrieb am Mo. 2. Feb. 2026 um
14:43:

john-h-kastner-aws:
> Apologies folks, I’ve been on vacation and not really able to work on this. I’d be happy to take a look at improvements (or other work 🙂) when I’m back in about three weeks.

not a problem, i asked for a really small tweak, so I figured it was best to merge your work as is

PR #2103, opened 2026-01-20
add two more getters to SymCC public API

cdisselkoen:

### DIFF
--- a/cedar-policy-symcc/src/lib.rs
+++ b/cedar-policy-symcc/src/lib.rs
@@ -157,6 +157,11 @@ impl CompiledPolicy {
         self.policy.effect()
     }
 
+    /// Get the (post-typecheck) `Policy` that this `CompiledPolicy` represents
+    pub fn policy(&self) -> &cedar_policy_core::ast::Policy {
+        &self.policy.policy
+    }
+
     /// Convert a `CompiledPolicy` to a `CompiledPolicies` representing a
     /// singleton policyset with just that policy.
     ///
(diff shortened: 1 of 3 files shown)


cedar-policy-symcc/src/lib.rs
```diff
@@ -157,6 +157,11 @@ impl CompiledPolicy {
         self.policy.effect()
     }
 
+    /// Get the (post-typecheck) `Policy` that this `CompiledPolicy` represents
+    pub fn policy(&self) -> &cedar_policy_core::ast::Policy {
```
john-h-kastner-aws, line 161: not up to speed on the API expectations for cedar-policy-symcc, but I'll assume this is supposed to be returning the `core`?

cdisselkoen: Good point, it probably should not.

cdisselkoen: Actually I propose we leave this for now -- this needs to analogize to `WellTypedPolicy::policy()` in the unoptimized pathway, and that already returns a `core::ast::Policy`. Separately we should figure out how to fix both of these. The reason this is difficult is that these policies are produced via typechecking (so, in Core form), but we are in the symcc crate, and `cedar_policy::Policy::from_ast()` is currently `pub(crate)`, not a public API, so symcc can't use it to convert into a `cedar_policy::Policy`. We could make `from_ast()` a public API, perhaps with `doc(hidden)` like other things related to Core that live in the public API, but let's discuss separately since this problem is pre-existing.

PR #2111, opened 2026-01-27
Add symcc solver pooling

### LINKED ISSUE #1861: Provide a way to pool SymCC solver workers
### Category

When using `LocalSolver`, it would be nice to have a higher-level interface that keeps a set of worker threads of cvc5 running (possibly scaling the set of workers up and down dynamically as needed) in order to not have to spawn a new process every time SymCC needs to evaluate something. For example, to not have to spawn a process (expensive) on every new HTTP request (like I do [here in this PoC](https://github.com/upbound/kubernetes-cedar-authorizer/blob/c1ce0f60ea3656989adb64a3f57475fe7d131fff/src/cedar_authorizer/symcc.rs#L43))

None yet.

### Additional context

_No response_

micahhausler:
## Description of changes

Implement a `SolverPool` that maintains a pool of warm CVC5 solver processes for reuse across queries to avoid spawning processes on each request.

## Issue #, if available

Resolves #1861

### DIFF
--- a/cedar-policy-symcc/src/symcc/solver_pool.rs
+++ b/cedar-policy-symcc/src/symcc/solver_pool.rs
@@ -1,0 +1,552 @@
+/*
+ * Copyright Cedar Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+//! A pool of warm CVC5 solver processes for reusing across queries.
+//!
+//! This module provides [`SolverPool`], which maintains a pool of pre-warmed
+//! solver processes that can be acquired and released efficiently, avoiding
+//! the overhead of spawning a new process for each query.
+//!
+//! # Example
+//!
+//! ```no_run
+//! use cedar_policy_symcc::solver_pool::{SolverPool, SolverPoolConfig};
+//! use cedar_policy_symcc::CedarSymCompiler;
+//!
+//! # async fn example() -> Result<(), Box<dyn std::error::Error>> {
+//! // Create a pool with default configuration
+//! let pool = SolverPool::new(SolverPoolConfig::default()).await?;
+//!
+//! // Acquire a solver from the pool
+//! let solver = pool.acquire().await?;
+//!
+//! // Use the solver with CedarSymCompiler
+//! let mut compiler = CedarSymCompiler::new(solver)?;
+//! // ... perform queries ...
+//!
+//! // Solver is automatically returned to the pool when dropped
+//! # Ok(())
+//! # }
+//! ```
+
+use super::smtlib_script::SmtLibScript;
+use super::solver::{Decision, LocalSolver, Solver, SolverError};
+use miette::Diagnostic;
+use std::io::{self, ErrorKind};
+use std::pin::Pin;
+use std::sync::Arc;
+use std::task::{Context, Poll};
+use std::time::Duration;
+use thiserror::Error;
+use tokio::io::AsyncWrite;
+use tokio::sync::{Mutex, OwnedSemaphorePermit, Semaphore};
+
+/// Errors that can occur when acquiring a solver from a [`SolverPool`].
+#[derive(Debug, Diagnostic, Error)]
+pub enum SolverPoolError {
+    /// The pool's internal semaphore was closed unexpectedly.
+    /// This indicates a bug or that the pool has been corrupted.
+    #[error("solver pool semaphore closed unexpectedly")]
+    SemaphoreClosed,
+
+    /// Timed out waiting for a solver to become available from the pool.
+    /// This occurs when all solvers are in use and `acquire_timeout` is set.
+    #[error("timeout waiting for solver from pool")]
+    AcquireTimeout,
+
+    /// An error occurred while spawning or interacting with the underlying solver.
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    Solver(#[from] SolverError),
+}
+
+/// A writer that always returns an error, used when a solver has been marked as failed.
+#[derive(Debug, Default)]
+struct FailedWriter;
+
+impl FailedWriter {
+    fn error() -> io::Error {
+        io::Error::new(ErrorKind::Other, SolverError::SolverMarkedFailed)
+    }
+}
+
+impl AsyncWrite for FailedWriter {
+    fn poll_write(
+        self: Pin<&mut Self>,
+        _cx: &mut Context<'_>,
+        _buf: &[u8],
+    ) -> Poll<io::Result<usize>> {
+        Poll::Ready(Err(Self::error()))
+    }
+
+    fn poll_flush(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        Poll::Ready(Err(Self::error()))
+    }
+
+    fn poll_shutdown(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        Poll::Ready(Err(Self::error()))
+    }
+}
+
+/// Configuration for a [`SolverPool`].
+#[derive(Clone, Debug)]
+pub struct SolverPoolConfig {
+    /// Minimum number of solvers to pre-warm at pool creation.
+    /// Defaults to 1.
+    pub min_solvers: usize,
+    /// Maximum number of concurrent solvers allowed.
+    /// Defaults to 4.
+    pub max_solvers: usize,
+    /// Timeout for acquiring a solver from the pool.
+    /// `None` means wait indefinitely.
+    /// Defaults to `None`.
+    pub acquire_timeout: Option<Duration>,
+}
+
+impl Default for SolverPoolConfig {
+    fn default() -> Self {
+        Self {
+            min_solvers: 1,
+            max_solvers: 4,
+            acquire_timeout: None,
+        }
+    }
+}
+
+/// A pool of warm CVC5 solver processes.
+///
+/// The pool maintains a set of pre-warmed solver processes that can be
+/// efficiently reused across queries. This avoids the overhead of spawning
+/// a new process for each query.
+///
+/// Solvers are acquired using [`SolverPool::acquire`], which returns a
+/// [`PooledSolver`] that implements the [`Solver`] trait. When the
+/// [`PooledSolver`] is dropped, it attempts to reset and return the solver
+/// to the pool. If the reset fails (e.g., due to a parse error that killed
+/// the solver), the solver is discarded.
+#[derive(Debug)]
+pub struct SolverPool {
+    /// Available solvers ready for use
+    available: Arc<Mutex<Vec<LocalSolver>>>,
+    /// Semaphore to limit maximum concurrent solvers
+    semaphore: Arc<Semaphore>,
+    /// Pool configuration
+    config: SolverPoolConfig,
+}
+
+impl SolverPool {
+    /// Creates a new solver pool with the given configuration.
+    ///
+    /// This will pre-warm `config.min_solvers` solver processes eagerly.
+    /// Returns an error if any of the pre-warmed solvers fail to spawn.
+    pub async fn new(config: SolverPoolConfig) -> Result<Self, SolverError> {
+        let semaphore = Arc::new(Semaphore::new(config.max_solvers));
+        let mut solvers = Vec::with_capacity(config.max_solvers);
+
+        // Eagerly spawn min_solvers
+        for _ in 0..config.min_solvers {
+            solvers.push(LocalSolver::cvc5()?);
+        }
+
+        Ok(Self {
+            available: Arc::new(Mutex::new(solvers)),
+            semaphore,
+            config,
+        })
+    }
+
+    /// Acquires a solver from the pool.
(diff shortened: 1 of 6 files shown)


cedar-policy-symcc/src/symcc/solver_pool.rs
```diff
@@ -0,0 +1,455 @@
...
+    pub async fn acquire(&self) -> Result<PooledSolver, SolverError> {
+        // Acquire a permit from the semaphore (blocks if at max capacity)
+        let permit = match self.config.acquire_timeout {
+            Some(timeout) => {
+                match tokio::time::timeout(timeout, Arc::clone(&self.semaphore).acquire_owned())
+                    .await
+                {
+                    Ok(Ok(permit)) => permit,
+                    Ok(Err(_)) => {
+                        return Err(SolverError::Solver(
```
luxas: Check how Cedar generally uses thiserror to make error enum variants for possible failure modes.

cedar-policy-symcc/src/symcc/solver_pool.rs
```diff
@@ -0,0 +1,455 @@
...
+        self.solver
+            .as_mut()
+            .expect("solver should always be present while PooledSolver is alive")
+    }
+
+    /// Marks this solver as failed, so it will be discarded instead of returned to the pool.
+    ///
+    /// This is useful when you know the solver is in a bad state (e.g., after a parse error)
+    /// and should not be reused.
+    pub fn mark_failed(&mut self) {
```
luxas, line 256: hmm, is there any way we could understand this automatically instead, so the caller wouldn't have to know that you need to call this function on weird errors? Or does that become too "magic" if we automatically discard on any error 🤔

cedar-policy-symcc/src/symcc/solver_pool.rs
```diff
@@ -0,0 +1,455 @@
...
+    pool: Arc<Mutex<Vec<LocalSolver>>>,
+    /// Semaphore permit - released when this PooledSolver is dropped
+    _permit: OwnedSemaphorePermit,
+}
+
+impl PooledSolver {
+    /// Returns a mutable reference to the underlying solver.
+    #[expect(
+        clippy::expect_used,
+        reason = "internal invariant: solver is always Some unless mark_failed() was called, which indicates a bug if the solver is subsequently used"
```
luxas: Isn't this so that a user first could call `mark_failed`, then `check_sat` and get a panic, thus not internal-only? I guess one could return an implementation of `dyn tokio::io::AsyncWrite + Unpin + Send)` that just errors from `smtlib_input`, and the error instantly from `check_sat` and `get_model`, if solver is none.

cedar-policy-symcc/src/symcc/solver_pool.rs
```diff
@@ -0,0 +1,455 @@
...
+            // Pool should now be empty
+            assert_eq!(pool.available_count().await, 0);
+
+            // Basic solver operation should work
+            let decision = solver.check_sat().await.unwrap();
+            assert_eq!(decision, Decision::Sat);
+        }
+
+        // Give the async drop task time to complete
+        tokio::time::sleep(Duration::from_millis(50)).await;
```
luxas: Maybe it could try once per 10ms or sth, but fail the test only after 10-20 unsuccessful retries to reduce flakes?

cedar-policy-symcc/src/symcc/solver_pool.rs
```diff
@@ -0,0 +1,455 @@
...
+
+        // Second acquire - should reuse the same solver (after reset)
+        {
+            let mut solver = pool.acquire().await.unwrap();
+            // The solver was reset, so this should work fresh
+            solver.smtlib_input().assert("false").await.unwrap();
```
luxas: The default, if you provide no assertions, is sat. However, if you assert false, the answer is always unsat, no matter the previous assertions. So swapping these around to first test false/unsat, and then sat/true should actually give confidence that this is testing the right thing

cedar-policy-symcc/src/symcc/solver_pool.rs
```diff
@@ -0,0 +1,455 @@
...
+            .await
+            .unwrap(),
+        );
+
+        // Acquire the only available solver
+        let _solver = pool.acquire().await.unwrap();
+
+        // Trying to acquire another should timeout
+        let result = pool.acquire().await;
+        assert_matches!(result, Err(SolverError::Solver(msg)) if msg.contains("timeout"));
```
luxas: This would be nicer if we followed the Cedar pattern of thiserror enums

cedar-policy-symcc/src/symcc/solver_pool.rs
```diff
@@ -0,0 +1,455 @@
...
+        compiler
+            .solver_mut()
+            .smtlib_input()
+            .set_logic("ALL")
+            .await
+            .unwrap();
+        compiler
+            .solver_mut()
+            .smtlib_input()
+            .assert("false")
```
luxas: could also do something like 1 == 2 to be fancy 😄

cedar-policy-symcc/src/symcc/solver_pool.rs
micahhausler: Great point

cedar-policy-symcc/src/symcc/solver_pool.rs
micahhausler: you're right, will fix this

cedar-policy-symcc/src/symcc/solver_pool.rs
micahhausler: sure

cedar-policy-symcc/src/symcc/solver_pool.rs
micahhausler: ah ok switched the order

cedar-policy-symcc/src/symcc/solver_pool.rs
micahhausler: will do

cedar-policy-symcc/src/symcc/solver_pool.rs
micahhausler: 🎩

cedar-policy-symcc/src/symcc/solver_pool.rs
micahhausler: I feel like this is a little too magical. If you're making symcc statements to a pool of solvers, you probably know what you're doing and might want a little more control. Tell me what you think

cedar-policy-symcc/src/symcc/solver_pool.rs
```diff
@@ -0,0 +1,549 @@
...
+}
+
+impl SolverPool {
+    /// Creates a new solver pool with the given configuration.
+    ///
+    /// This will pre-warm `config.min_solvers` solver processes eagerly.
+    /// Returns an error if any of the pre-warmed solvers fail to spawn.
+    pub async fn new(config: SolverPoolConfig) -> Result<Self, SolverError> {
+        let semaphore = Arc::new(Semaphore::new(config.max_solvers));
+        let mut solvers = Vec::with_capacity(config.min_solvers);
```
cdisselkoen: nit: might make sense to use `max_solvers` as the vec capacity?

cedar-policy-symcc/src/symcc/solver_pool.rs
```diff
@@ -0,0 +1,549 @@
...
+                .acquire_owned()
+                .await
+                .map_err(|_| SolverPoolError::SemaphoreClosed)?,
+        };
+
+        // Try to get an available solver from the pool
+        let solver = {
+            let mut available = self.available.lock().await;
+            available.pop()
+        };
```
cdisselkoen: nit: comment says "Try", but there's nothing here that can fail or return an error. (I suppose deadlock is a failure condition, but if that's what the comment is referring to, it could be more explicit)

cedar-policy-symcc/src/symcc/solver_pool.rs
```diff
@@ -0,0 +1,549 @@
...
+            Some(s) => Ok(s),
+            None => Err(SolverError::SolverMarkedFailed),
+        };
+        async move {
+            match result {
+                Ok(s) => s.check_sat().await,
+                Err(e) => Err(e),
+            }
+        }
+    }
```
cdisselkoen:
```suggestion
    async fn check_sat(&mut self) -> Result<Decision, SolverError> {
        let solver = self.solver.ok_or(SolverError::SolverMarkedFailed)?;
        solver.check_sat().await
    }
```

cedar-policy-symcc/src/symcc/solver_pool.rs
```diff
@@ -0,0 +1,549 @@
...
+            Some(s) => Ok(s),
+            None => Err(SolverError::SolverMarkedFailed),
+        };
+        async move {
+            match result {
+                Ok(s) => s.get_model().await,
+                Err(e) => Err(e),
+            }
+        }
+    }
```
cdisselkoen:
```suggestion
    async fn get_model(&mut self) -> Result<Option<String>, SolverError> {
        let solver = self.solver.ok_or(SolverError::SolverMarkedFailed)?;
        solver.get_model().await
    }
```

cedar-policy-symcc/src/symcc/solver_pool.rs
```diff
@@ -0,0 +1,549 @@
...
+
+            // Spawn a task to reset and return the solver to the pool
+            tokio::spawn(async move {
+                // Try to reset the solver
+                if solver.smtlib_input().reset().await.is_ok() {
+                    // Reset succeeded, return to pool
+                    pool.lock().await.push(solver);
+                }
+                // If reset failed, the solver is discarded (not returned to pool)
+                // The semaphore permit is released automatically when _permit is dropped
```
cdisselkoen: I think this releases the semaphore permit when `drop()` returns, before the async task returns the solver to the pool? Should we instead move the permit into the async closure, so that the permit is not released until the solver is returned?

cedar-policy-symcc/src/symcc/solver_pool.rs
micahhausler: ah will make this more clear

cedar-policy-symcc/src/symcc/solver_pool.rs
micahhausler: I _think_ I've done this properly now

cdisselkoen: Looks right to me

cedar-policy-symcc/src/symcc/solver_pool.rs
```diff
@@ -0,0 +1,552 @@
...
+        // The permit is moved into the task so it's only released after
+        // the solver is returned to the pool (or discarded).
+        tokio::spawn(async move {
+            if let Some(mut solver) = solver {
+                // Try to reset the solver
+                if solver.smtlib_input().reset().await.is_ok() {
+                    // Reset succeeded, return to pool
+                    pool.lock().await.push(solver);
+                }
+                // If reset failed, the solver is discarded (not returned to pool)
```
Swolebrain: Should we create a new solver and put it in the pool here? That way the pool can't get into a runtime state where it's doing nothing other than waiting for a semaphore, then a lock, only to return a `LocalSolver::cvc5()?` (line 209)

cedar-policy-symcc/src/symcc/solver_pool.rs
```diff
@@ -0,0 +1,552 @@
...
+                }
+            }
+            None => Arc::clone(&self.semaphore)
+                .acquire_owned()
+                .await
+                .map_err(|_| SolverPoolError::SemaphoreClosed)?,
+        };
+
+        // Get an available solver from the pool
+        let solver = {
```
Swolebrain: If you changed `available` to use RwLock instead of mutex, would that enable you to first check here if there are any available Solvers first without write contention? You could do something like

```rust
pub async fn available_count(&self) -> usize {
    self.available.read().await.len()
}
```
Then in this closure:
```rust
let solver = {
    if self.available_count().await == 0 {
        Some(LocalSolver::cvc5()?)
    } else {
        let mut available = self.available.lock().await;
        available.pop()
    }
};
```

cedar-policy-symcc/src/symcc/solver_pool.rs
```diff
@@ -0,0 +1,552 @@
...
+    ///
+    /// This will pre-warm `config.min_solvers` solver processes eagerly.
+    /// Returns an error if any of the pre-warmed solvers fail to spawn.
+    pub async fn new(config: SolverPoolConfig) -> Result<Self, SolverError> {
+        let semaphore = Arc::new(Semaphore::new(config.max_solvers));
+        let mut solvers = Vec::with_capacity(config.max_solvers);
+
+        // Eagerly spawn min_solvers
+        for _ in 0..config.min_solvers {
+            solvers.push(LocalSolver::cvc5()?);
```
luxas, line 161: nit: most likely someone will later ask for this constructor to be configurable, so one might put this into the config

cedar-policy-symcc/src/symcc/solver_pool.rs
luxas: Yeah, I agree, was mostly thinking out loud. The user most likely is fine doing this themselves, now it's also better as the other functions error accordingly when marked failed.

john-h-kastner-aws: Just from the context I have from reviewing this PR, it's not clear to me when I would need to call `mark_failed`. If the solver is in a bad state, shouldn't `solver.smtlib_input().reset().await.is_ok()` be `false` when dropping, having the same effect as marking it failed?

cedar-policy-symcc/src/symcc/solver_pool.rs
luxas: This would also give users the ability to calculate the cache miss rate.

cdisselkoen:
Going to merge this since it has sufficient approvals, passing CI, nothing major outstanding. For the remaining small items being discussed, please open followup issues or PRs.

PR #2120, opened 2026-02-02
Added variadic isInRange to symcc under experimental flag.

lianah:
## Description of changes

Added ipaddr variadic isInRange to symcc under experimental flag.

### DIFF
--- a/cedar-policy-symcc/src/symcc/compiler.rs
+++ b/cedar-policy-symcc/src/symcc/compiler.rs
@@ -483,6 +483,7 @@ pub fn compile_call2(
     compile_call2_error(xty, xty, enc, t1, t2)
 }
 
+#[cfg(feature = "variadic-is-in-range")]
 // Use directly for encoding calls that can error with n arguments
 pub fn compile_call_n_error(
     xty: ExtType,
@@ -523,6 +524,7 @@ pub fn compile_call_n_error(
     }
 }
 
+#[cfg(feature = "variadic-is-in-range")]
 // Use directly for encoding calls that cannot error with n arguments
 pub fn compile_call_n(
     xty: ExtType,
--- a/cedar-policy-symcc/src/symccopt/compiler.rs
+++ b/cedar-policy-symcc/src/symccopt/compiler.rs
@@ -798,6 +798,7 @@ pub fn compile_call2(
     compile_call2_error(xty, xty, |t1, t2| some_of(enc(t1, t2)), arg1, arg2)
 }
 
+#[cfg(feature = "variadic-is-in-range")]
 // Use directly for encoding calls that can error with n arguments
 pub fn compile_call_n_error(
     xty: ExtType,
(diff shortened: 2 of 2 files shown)


cedar-policy-symcc/src/symcc/compiler.rs
```diff
@@ -546,9 +617,27 @@ pub fn compile_call(xfn: &cedar_policy_core::ast::Name, ts: Vec<Term>) -> Result
...
         }
-        ("isInRange", 2) => {
-            let (t1, t2) = extract_first2(ts);
-            compile_call2(ExtType::IpAddr, extfun::is_in_range, t1, t2)
+        ("isInRange", n) => {
+            #[cfg(feature = "variadic-is-in-range")]
+            if n < 2 {
+                Err(CompileError::TypeError)
+            } else {
+                let (t, tn) = extract_first_n(ts, n);
```
john-h-kastner-aws: since `n` is `ts.len()` could this be

```
let t = ts.remove(0);
```

lianah: Yup, it requires updating the `compile_call` function to take a `mut` but that does not cause any other issues.

john-h-kastner-aws: Sometimes I like to shadow the parameter with a local like

```
let mut ts = ts;
```

so that it doesn't have to be `mut` in the whole function

cedar-policy-symcc/src/symcc/compiler.rs
```diff
@@ -483,6 +483,59 @@ pub fn compile_call2(
...
+    if t.type_of() != ty {
+        return Err(CompileError::TypeError);
+    }
+    if ts.len() != xtys.len() {
+        return Err(CompileError::TypeError);
+    }
+    let expected_types: Vec<_> = xtys
+        .iter()
+        .map(|xty| TermType::option_of(TermType::Ext { xty: *xty }))
+        .collect();
```
cdisselkoen: Shouldn't need to `.collect()` this? We should be able to leave it an iterator and thus avoid allocating a `Vec` at this intermediate step

cedar-policy-symcc/src/symccopt/compiler.rs
```diff
@@ -798,6 +798,72 @@ pub fn compile_call2(
     compile_call2_error(xty, xty, |t1, t2| some_of(enc(t1, t2)), arg1, arg2)
 }
 
+// Use directly for encoding calls that can error with n arguments
+pub fn compile_call_n_error(
```
john-h-kastner-aws, line 803: I'll just take this opportunity to complain about needing to duplicate these changes in `symcc` and `symccopt`.

lianah: ➕ I did notice a lot of duplication in `symcc` and `symccopt` which is unfortunate.

cdisselkoen: Should be fixed in Lean first imo.

PR #2139, opened 2026-02-06
Updated ordering of symcc terms

katherine-hough:
## Description of changes
Added a manual implementation of `Ord`/`PartialOrd` for `TermType` and changed the ordering of variant declarations for `Term` and `TermPrim`. We derive `Ord` for `Term` and `TermPrim` and when `Ord` is derived on enums, variants are ordered primarily by their discriminants which is determined by default by their declaration order. 

The goal of these changes is to ensure that the order of terms is consistent with the Lean to prevent DRT failures.

## Issue #, if available
N/A

### DIFF
--- a/cedar-policy-symcc/src/symcc/term_type.rs
+++ b/cedar-policy-symcc/src/symcc/term_type.rs
@@ -25,9 +25,9 @@ use std::collections::BTreeMap;
 use std::sync::Arc;
 
 /// Types of the intermediate [`super::term::Term`] representation.
-//  Note: The declaration order of variants for this enum affects the derived definition of `Ord`
+//  Note: The declaration order of variants for this enum affects the derived definitions of `Ord`
 //  and `PartialOrd` which must be consistent with the Lean.
-#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Clone, Debug, PartialEq, Eq, Ord, PartialOrd)]
 #[expect(missing_docs, reason = "fields are self explanatory")]
 pub enum TermType {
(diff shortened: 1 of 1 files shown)


cedar-policy-symcc/src/symcc/term_type.rs
```diff
@@ -192,3 +193,40 @@ impl TermType {
         }
     }
 }
+
+impl Ord for TermType {
```
cdisselkoen: could we avoid the custom `Ord` by just declaring the variants in alphabetical order, like for `Term` and `TermPrim`? or is there something I'm missing?

katherine-hough: It wouldn't be alphabetical it would be Option, then all the primitives, then record, then set because in the Lean have a separate type of TermPrimType.

katherine-hough: I didn't love either option, so if you have a preference I am happy to go with that.

katherine-hough:
> Looks good with two comments
> 
> 1. Have you checked if [educe](https://docs.rs/educe/latest/educe/) could simpilfy the custom ord impl?
> 2. Would it be simpler to change the ordering in Lean? (Probably a question for @cdisselkoen )

Good call on educe. I will use that.

katherine-hough:
educe does not let you rank variants just fields. To change the rank of variants they recommend changing the discriminant which is really not all that different to re-ordering the variants. I opted to reorder the variants of `TermType` and add a comment.

john-h-kastner-aws (approved):

PR #2149, opened 2026-02-12
update SymCC version and dependencies

cdisselkoen:
- SymCC version to 0.3.0
- SymCC depending on crates.io Cedar 4.9.0, not by path dependency
- `cargo update`


cdisselkoen:
The downstream CI failures will have to be fixed in a cedar-spec PR

Issue #1889, opened 2025-10-13
Support extended has operations in EST and AST


shaobo-he-aws:
### Category

Cedar language or syntax features/changes

### Describe the feature you'd like to request

Cedar policy format supports extended has operations, which are desugared into a chain of `has` and get attribute operations when converted into ESTs. This desugaring can cause performance blowups for a long list of attributes in an extended has operation. It'd be better if we can support extended has operations in the EST.

### Describe alternatives you've considered

N/A

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

john-h-kastner-aws:
extending the scope of this issue to also include updating the AST. That's a more performance critical path, and it makes sense to update the EST and AST at the same time

shaobo-he-aws:
Another idea is to extend ASTs with let-bindings, which is also a generalization of generalized templates.

shaobo-he-aws:
Perhaps the real problem is that the expression builder disallows sharing common expressions, whereas the definition of `ExprKind` does. That is, the expression builder has to re-build subexpressions even though they should be easily shared via cloning `Arc`. For instance, `context has a.b.c` desugars to `((context has a && context.a has b) && context.a.b has c)`. We should be able to share subexpressions like `context`, `context.a`. The resulting algorithm should be linear.

john-h-kastner-aws:
> The resulting algorithm should be linear.

Even if the size of the datastructure is linear, I don't think we'll get linear evaluation without doing something more than `Arc` so that we can share evaluation results

john-h-kastner-aws:
But that change sounds easy, so maybe we should just make it

victornicolet:
This https://github.com/cedar-policy/cedar/pull/2560 adds support for extended has in the AST.

Issue #1702, opened 2025-07-14
Cedar schema format does not allow entities attribute defined entirely by a common type


john-h-kastner-aws:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [x] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Bug Category

Schemas and Validation

### Describe the bug

Given

```
type Foo = { bar : Long };
entity Baz = Foo;
```

we get the error

```
jkastner@88665a03aa97 ~ % cedar check-parse -s test.cedarschema
  × failed to parse schema from file test.cedarschema
  ╰─▶ error parsing schema: unexpected token `Foo`
   ╭─[2:14]
 1 │ type Foo = { "bar": Long };
 2 │ entity Baz = Foo;
   ·              ─┬─
   ·               ╰── expected `{`
   ╰────
```

But the equivalent JSON schema parses as expected 

```json
{
  "": {
    "commonTypes": {
      "Foo": {
        "type": "Record",
        "attributes": {
          "bar": {
            "type": "Long"
          }
        }
      }
    },
    "entityTypes": {
      "Baz": {
        "shape": {
          "type": "Foo"
        }
      }
    },
    "actions": {}
  }
}
```

### Expected behavior

.

### Reproduction steps

.

### Code Snippet

```rust
// Put your code below this line.

```

### Log output

```
// Put your output below this line

```

shaobo-he-aws:
I don't think it's a bug per se. We can certainly extend the grammar to incorporate this case.

john-h-kastner-aws:
It's a bug to the extent that every JSON schema should convert to a valid Cedar schema. So maybe more of a bug in the spec for the grammar than it implementation

john-h-kastner-aws:
Or, the better way to state this bug is that the JSON schema above converts to the Cedar schema, but that cedar schema doesn't parse.

shaobo-he-aws:
> Or, the better way to state this bug is that the JSON schema above converts to the Cedar schema, but that cedar schema doesn't parse.

Correct. Maybe update the title?

john-h-kastner-aws:
[...]

PR #2154, opened 2026-02-12
extend the 'has' relation in EST and JSON policy format

victornicolet:
## Description of changes

This add a new variant for the `has` operator in the EST, which translates to an extended `has` in JSON:
```
{"has": {"left": <EXPR>, "attr": ["attribute", "nested",...]}}
```
This is backwards compatible, in the sense that:
-  simple `has` i.e. ```{"has": {"left": <EXPR>, "attr": "attribute"}}``` still results in the same expression as before,
- the extended `has` operator in the Cedar language is still desugared to a conjunction of `has` and `.` when converting to JSON.

We may want to validate that in the extended has, each element in the list is a valid identifier to match the extended has of the Cedar policy language.

The new EST results in changes in the exported types in `cedar-wasm/pkg/*/cedar_wasm.d.ts`. We have a new type:
```diff
+ export type HasAttrRepr = { left: Expr; attr: SmolStr } | { left: Expr; attr: NonEmpty<SmolStr> };
```
And the type of expressions uses this type:
```diff
export type ExprNoExt = { Value: CedarValueJson } 
| { Var: Var } | { Slot: string } | { "!": { arg: Expr } } 
| { neg: { arg: Expr } } | { "==": { left: Expr; right: Expr } } 
| { "!=": { left: Expr; right: Expr } } | { in: { left: Expr; right: Expr } } 
| { "<": { left: Expr; right: Expr } } | { "<=": { left: Expr; right: Expr } } 
| { ">": { left: Expr; right: Expr } } | { ">=": { left: Expr; right: Expr } }
 | { "&&": { left: Expr; right: Expr } } 
| { "||": { left: Expr; right: Expr } } 
| { "+": { left: Expr; right: Expr } } 
| { "-": { left: Expr; right: Expr } } 
| { "*": { left: Expr; right: Expr } } 
| { contains: { left: Expr; right: Expr } } 
| { containsAll: { left: Expr; right: Expr } } | { containsAny: { left: Expr; right: Expr } } 
| { isEmpty: { arg: Expr } } 
| { getTag: { left: Expr; right: Expr } } | { hasTag: { left: Expr; right: Expr } } | { ".": { left: Expr; attr: SmolStr } } 
+ | { has: HasAttrRepr } 
- | { has:  { left: Expr; attr: SmolStr } }
| { like: { left: Expr; pattern: PatternElem[] } } | { is: { left: Expr; entity_type: SmolStr; in?: Expr } } 
| { "if-then-else": { if: Expr; then: Expr; else: Expr } } | { Set: Expr[] } | { Record: Record<string, Expr> };
```
Typescript that was previously type-checking should typecheck with the new types.

## Issue #, if available

#1889 only the EST extension part. Does not touch the AST.

- [x] Requires updates, and I have made / will make these updates myself. (Please include in your description a timeline or link to the relevant PR in [`cedar-docs`](https://github.com/cedar-policy/cedar-docs). PRs should be targeted at a `staging-X.Y` branch, not `main`.) Changes in branch https://github.com/cedar-policy/cedar-docs/tree/json-extended-has-2154

### DIFF
--- a/cedar-policy-core/src/est/expr.rs
+++ b/cedar-policy-core/src/est/expr.rs
@@ -1012,7 +1012,6 @@ impl Expr {
                     Arc::unwrap_or_clone(left).try_into_ast(id)?,
                     attr,
                 )),
-                // Should the try_into_ast use a builder anyway?
                 HasAttrRepr::Extended { left, attr } => Ok(ast::ExprBuilder::new()
                     .extended_has_attr(Arc::unwrap_or_clone(left).try_into_ast(id)?, &attr)),
             },
@@ -1433,7 +1432,8 @@ impl BoundedDisplay for ExprNoExt {
                         write!(f, " has \"{}\"", attr.head.escape_debug())?;
                     }
                     for attr in attr.tail.iter() {
-                        // TODO: validation, we shouldn't have non-idents here
+                        // TODO: validation, we shouldn't have non-idents here because
+                        // `principal has "foo".bar` doesn't parse
                         if is_normalized_ident(attr) {
                             write!(f, ".{}", attr)?;
                         } else {


victornicolet:
Looks like this breaks cedar-wasm because it exposes the EST.

cedar-policy-core/src/est/expr.rs
```diff
@@ -157,6 +158,29 @@ impl From<crate::ast::Pattern> for Vec<PatternElem> {
...
+/// the extended has operation
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(untagged)]
+pub enum HasAttrRepr {
+    /// Simple has test with a single attribute :
+    /// { "has": { "left": ..., "attr": <string>}}
+    Simple {
+        /// Left-hand argument
+        left: Arc<Expr>,
+        /// Atribute tested
```
cdisselkoen:
```suggestion
        /// Attribute tested
```

cedar-policy-core/src/est/expr.rs
```diff
@@ -1735,4 +1792,45 @@ mod test {
...
+        // Extended has - serialize
+        let expr = Expr::ExprNoExt(ExprNoExt::HasAttr(HasAttrRepr::Extended {
+            left: Arc::new(Expr::ExprNoExt(ExprNoExt::Var(ast::Var::Context))),
+            attr: nonempty!["user".into(), "profile".into(), "email".into()],
+        }));
+        let json = serde_json::to_value(&expr).unwrap();
+        assert_eq!(
+            json,
+            serde_json::json!({"has": {"left": {"Var": "context"}, "attr": ["user", "profile", "email"]}})
+        );
```
cdisselkoen: Might be useful to have a test for an Extended with only one attr. (Obviously that could also be represented with Simple, but IIUC, in this PR we will allow the Extended representation with a singleton set as a legal JSON)

victornicolet: Yes, good point. I'll also add a test that shows deserializing `"has": { ... "attr": []}` fails (empty list).

victornicolet:
This results in a change in the exported types in `cedar-wasm/pkg/*/cedar_wasm.d.ts`.
Before:
```ts
export type ExprNoExt = ... | { has: { left: Expr; attr: SmolStr }} | ...
```
Now:
```ts
export type ExprNoExt = ... | { has: HasAttrRepr } | ...
export type HasAttrRepr = { left: Expr; attr: SmolStr } | { left: Expr; attr: NonEmpty<SmolStr> };
```
Expressions that typechecked before still typecheck after the change.

cdisselkoen:
Nice, way to go making the TS export a nonbreaking change

cedar-policy-core/src/est/expr.rs
```diff
@@ -972,10 +1007,15 @@ impl Expr {
...
-            Expr::ExprNoExt(ExprNoExt::HasAttr { left, attr }) => Ok(ast::Expr::has_attr(
-                Arc::unwrap_or_clone(left).try_into_ast(id)?,
-                attr,
-            )),
+            Expr::ExprNoExt(ExprNoExt::HasAttr(repr)) => match repr {
+                HasAttrRepr::Simple { left, attr } => Ok(ast::Expr::has_attr(
+                    Arc::unwrap_or_clone(left).try_into_ast(id)?,
+                    attr,
+                )),
+                // Should the try_into_ast use a builder anyway?
```
john-h-kastner-aws: I believe `try_into_ast` could written entirely in terms of a generic ExprBuilder if that's what you're asking. For the moment, we only convert to AST so it doesn't matter, but we might want that for a future public-syntax tree.

cedar-policy-core/src/est/expr.rs
```diff
@@ -1376,14 +1416,33 @@ impl BoundedDisplay for ExprNoExt {
...
-            }
+                HasAttrRepr::Extended { left, attr } => {
+                    maybe_with_parens(f, left, n)?;
+                    if is_normalized_ident(&attr.head) {
+                        write!(f, " has {}", attr.head)?;
+                    } else {
+                        write!(f, " has \"{}\"", attr.head.escape_debug())?;
+                    }
+                    for attr in attr.tail.iter() {
+                        // TODO: validation, we shouldn't have non-idents here
```
john-h-kastner-aws: I assume this is what you mean here, but it's worth commenting that `principal has "foo".bar` doesn't parse

victornicolet: Yes that's what I meant, I can add the comment.

john-h-kastner-aws (approved):

cedar-policy-core/src/est/expr.rs
```diff
@@ -1376,14 +1415,34 @@ impl BoundedDisplay for ExprNoExt {
...
+                    maybe_with_parens(f, left, n)?;
+                    if is_normalized_ident(&attr.head) {
+                        write!(f, " has {}", attr.head)?;
+                    } else {
+                        write!(f, " has \"{}\"", attr.head.escape_debug())?;
+                    }
+                    for attr in attr.tail.iter() {
+                        // TODO: validation, we shouldn't have non-idents here because
+                        // `principal has "foo".bar` doesn't parse
+                        if is_normalized_ident(attr) {
```
cdisselkoen, line 1437: if this is still a TODO, can we open a new issue for it (and maybe update the comment here to include the issue number)?

victornicolet: Mentioned the lack of validation in https://github.com/cedar-policy/cedar/issues/1702, since I think this can be solved in the same place we need to solve the "JSON policy with zero-argument method style call" problem.

PR #2156, opened 2026-02-16
small perf optimization in Decimal::from_str()

cdisselkoen:
## Description of changes

The `is_match()` check was redundant; we were evaluating the regex twice for no reason


victornicolet (approved):

cdisselkoen:
Pushed a commit that updates deps (particularly `keccak 0.1.5 -> 0.1.6`) to avoid CI failure due to yanked `keccak 0.1.5` in our transitive deps

PR #2162, opened 2026-02-16
PoC: Return the same evaluation errors from TPE as from concrete evaluation

### LINKED ISSUE #1736: Return (a superset of) EvaluationError from TPE evaluation

luxas:
## Description of changes

Draft PR, I need to double-check some of the logic here still and add some more unit tests, but it should be good enough to discuss at least. What do you think about this rewrite? I think it could help make the code between TPE and the concrete evaluator more similar and hopefully comparable, especially when the older `partial-eval` is deprecated and removed from the concrete evaluator.

Discussed this with @john-h-kastner-aws. The idea is to return a concrete error whenever the concrete evaluator would, and just exactly the same error as the concrete evaluator would. However, cases like `<residual> && <error>` do not fold to `<error>`, as `<residual>` could evaluate to another error.

Also noted that the stack overflow check was not added to TPE. Also I'm trying to make the errors nice by preserving the Loc information, but it seems like I need to do some other thing still to produce nice reports in miette to show where the error came from.

Fixes: https://github.com/cedar-policy/cedar/issues/1736

(at least part of, we could add the `Diagnostics` in a follow-up PR)

Not sure, do you want a release note for this one? I can add one.

Would you be interested in updating cedar-spec with a theorem that the errors should be the same between concrete and TPE?

### DIFF
--- a/cedar-policy-core/src/tpe/evaluator.rs
+++ b/cedar-policy-core/src/tpe/evaluator.rs
@@ -22,6 +22,7 @@ use crate::tpe::err::ExprToResidualError;
 use crate::validator::types::Type;
 use crate::{
     ast::{self, BinaryOp, EntityUID, Expr, PartialValue, Set, Value, ValueKind, Var},
+    evaluator::{stack_size_check, EvaluationError},
     extensions::Extensions,
 };
 
@@ -30,6 +31,8 @@ use crate::{
     tpe::request::PartialRequest,
     tpe::residual::{Residual, ResidualKind},
 };
+
+use nonempty::nonempty;
 
 /// The partial evaluator
 #[derive(Debug)]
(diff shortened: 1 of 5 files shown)


cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -50,464 +53,381 @@ impl Evaluator<'_> {
...
-                    (Residual::Error(_), _) => Residual::Error(ty),
-                    (_, Residual::Error(_)) => Residual::Error(ty),
-                    (_, _) => residual(arg1, arg2, ty),
+                    // bubble up any errors (arg1 before arg2, if both error) without modification.
+                    // Note that this is what concrete evaluation also does, if either LHS or RHS errors, that error is returned.
+                    (Residual::Error { err, ty }, _) => Residual::Error {
+                        err: err.clone(),
+                        ty: ty.clone(),
+                    },
+                    (_, Residual::Error { err, ty }) => Residual::Error {
```
luxas, line 426: from live review: the left hand side could return an error, even if it is a residual right now. To make the error exactly as it would be in the concrete case, we need a way to say "we got an error, but we not sure which"

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -675,20 +620,25 @@ impl Evaluator<'_> {
...
+                    })
+                } else if let Some((err, ty)) = rs
+                    .iter()
+                    .flat_map(|r| match r {
+                        Residual::Error { err, ty } => Some((err, ty)),
+                        _ => None,
+                    })
+                    .next()
+                {
+                    // return the first error encountered, like normal eval
```
luxas, line 635: TODO: Again, we need to return "we are sure we are erroring, but do not know which error" here instead.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -701,26 +651,21 @@ impl Evaluator<'_> {
...
                     let mut m = BTreeMap::new();
                     for (a, r) in record {
-                        if matches!(r, Residual::Error(_)) {
-                            return Residual::Error(ty);
-                        } else {
-                            m.insert(a, r);
+                        match r {
+                            Residual::Error { err, ty } => return Residual::Error { err, ty },
+                            _ => {
+                                m.insert(a, r);
```
luxas, line 664: try to align with the Set logic above, and return "error, but not sure yet"

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -50,464 +53,381 @@ impl Evaluator<'_> {
...
-                        } => left,
+                        } => left, // <residual> && true == <residual>
                         Residual::Concrete {
                             value:
                                 Value {
                                     value: ValueKind::Lit(ast::Literal::Bool(false)),
                                     ..
                                 },
-                            ..
+                            ty,
```
luxas, line 139: call this `right_ty`, so we know which type we're talking about, it's easier to reason about the code then.

cedar-policy-core/src/tpe/evaluator.rs
john-h-kastner-aws: I was thinking about this a bit and I don't think it should be too bad. One nice way to do it might be splitting `ResidualError` into `Residual::ConcreteError` and `Residual::PartialError` to represent these two cases. `PartialError` would carry a `ResidualKind` instead of an `EvaluationError`.

Then this functions becomes

```
// arg1 evals either to a concrete or residual error, that error is the result
(r1, _) if r1.is_error() => make_error(r1, ty),
// r1 isn't an error variant, but it could be residual which eventually
// evaluates to an error, so we can't ignore it.
(_, r2) if r2.is_error() => residual_error(arg1, arg2, ty)
```

luxas: > split into `Residual::ConcreteError` and `Residual::PartialError`

yep, that's my plan 👍 If we need later, we could provide a function to collect known errors from the partial error tree (if the user would like to know what definitely did go wrong).

PR #2164, opened 2026-02-17
revert symcc to depend on cedar-policy by path again

cdisselkoen:
Before this PR, cedar-lean-ffi is broken on the release branch; see detailed comments on https://github.com/cedar-policy/cedar-spec/pull/873.


cdisselkoen:
Also added a `cargo update` to fix CI -- notably updating away from `keccak 0.1.5` which was yanked (and which was a transitive dependency for us).

Issue #816, opened 2024-04-25
Concrete public type for the EST


aaronjeline:
### Category

User level API features/changes

### Describe the feature you'd like to request

While the EST provides a public shape for cedar ASTs, we don't provide a public type for manipulating it. 

### Describe alternatives you've considered

The status quo of just manipulating the JSON is plausible, but lacks ergonomics

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

cdisselkoen:
Related, and might make sense to tackle at the same time: #366

victornicolet:
The PST released in [4.11.0 ](https://github.com/cedar-policy/cedar/releases/tag/v4.11.0) provides a programmatic AST.

PR #2170, opened 2026-02-18
add datastructure for public syntax tree, display and ExprBuilder

victornicolet:
## Description of changes

This is a first PR for the implementation of a public AST (motivated by https://github.com/cedar-policy/cedar/issues/816), the "PST". 
This PR defines:
- basic data structures for the PST,
- a basic display for the PST expressions,
- an implementation of `ExprBuilder` for the PST.

The goal in the design of the PST here was to decouple the public representation from internal data structures and data structures meant for serialization/deserialization.
- where types are similar to the types reused in both AST and EST, we implement `From<...>` functions,
- most types define a new representation for policies and expressions.

### DIFF
--- a/cedar-policy-core/src/pst/expr.rs
+++ b/cedar-policy-core/src/pst/expr.rs
@@ -97,6 +97,7 @@ impl Name {
     }
 }
 
+#[doc(hidden)]
 impl From<ast::Name> for Name {
     fn from(name: ast::Name) -> Self {
         let ast::Name {
@@ -105,8 +106,7 @@ impl From<ast::Name> for Name {
         Name {
             id: id.into_smolstr(),
             namespace: Arc::new(
-                Arc::try_unwrap(path)
-                    .unwrap_or_else(|arc| (*arc).clone())
+                Arc::unwrap_or_clone(path)
                     .into_iter()
                     .map(|id| id.to_smolstr())
                     .collect(),
@@ -115,6 +115,7 @@ impl From<ast::Name> for Name {
     }
 }
 
+#[doc(hidden)]
 impl TryFrom<Name> for ast::Name {
     type Error = crate::parser::err::ParseErrors;
 
@@ -152,12 +153,14 @@ impl EntityType {
     }
 }
 
+#[doc(hidden)]
 impl From<ast::EntityType> for EntityType {
     fn from(et: ast::EntityType) -> Self {
         EntityType(et.into_name().into())
     }
 }
 
+#[doc(hidden)]
 impl TryFrom<EntityType> for ast::EntityType {
     type Error = crate::parser::err::ParseErrors;
 
@@ -187,6 +190,7 @@ pub struct EntityUID {
     pub eid: SmolStr,
 }
 
+#[doc(hidden)]
 impl From<ast::EntityUID> for EntityUID {
     fn from(uid: ast::EntityUID) -> Self {
         let (ty, eid) = uid.components();
@@ -436,6 +440,7 @@ pub enum PatternElem {
     Wildcard,
 }
 
+#[doc(hidden)]
 impl From<ast::Pattern> for Vec<PatternElem> {
     fn from(pattern: ast::Pattern) -> Self {
         pattern
@@ -907,7 +912,7 @@ impl std::fmt::Display for Expr {
             Expr::Literal(lit) => match lit {
                 Literal::Bool(b) => write!(f, "{}", b),
                 Literal::Long(i) => write!(f, "{}", i),
-                Literal::String(s) => write!(f, "\"{}\"", s.escape_default()),
+                Literal::String(s) => write!(f, "\"{}\"", s.escape_debug()),
                 Literal::EntityUID(uid) => write!(f, "{}", uid),
             },
             Expr::Var(v) => match v {


cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1164 @@
...
+        }
+    }
+}
+
+impl Display for EntityType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let ast_et: Result<ast::EntityType, _> = self.clone().try_into();
+        match ast_et {
+            Ok(n) => write!(f, "{}", n),
+            Err(_) => write!(f, "<invalid name {}>", self.to_smolstr()),
```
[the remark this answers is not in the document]
victornicolet: I think I trust the compiler on this one.

victornicolet:
Getting coverage for the changes in this PR won't be very meaningful, coverage will increase once we start adding functionality to convert / manipulate the PST.

cedar-policy-core/src/pst/constraints.rs
```diff
@@ -0,0 +1,69 @@
...
+
+/// Action scope constraint
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ActionConstraint {
+    /// Any action
+    Any,
+    /// Equals specific action
+    Eq(EntityUID),
+    /// In set of actions (single action is just length 1)
+    In(nonempty::NonEmpty<EntityUID>),
```
cdisselkoen: Do the AST/EST have the nonempty constraint here? Is `permit(principal, action in [], resource);` a valid Cedar policy?

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1164 @@
...
+/// Slot identifier for template policies
+///
+/// Cedar supports two slot types: `principal` and `resource`
+#[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord, Hash)]
+pub enum SlotId {
+    /// Principal slot
+    Principal,
+    /// Resource slot
+    Resource,
+}
```
cdisselkoen, line 39: Discussion question: do we want to try to make the PST structures such that the modifications that will be needed for RFC 98 are backwards-compatible? is that an achievable goal?

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1164 @@
...
+
+/// Entity type name
+///
+/// Represents the type of an entity in Cedar (e.g., `User`, `Photo`, `Namespace::Resource`)
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum EntityType {
+    /// An entity type with a qualified name
+    EntityType(Name),
+}
```
cdisselkoen: why an `enum` instead of something like `pub struct EntityType(pub Name)`?

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1164 @@
...
+
+impl EntityType {
+    /// Create an entity type from a name
+    pub fn entity_type(name: impl Into<Name>) -> Self {
```
cdisselkoen: maybe this should be called `new()` or `from_name()`

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1164 @@
...
+    Neg(Arc<Expr>),
+    /// Binary operation
+    BinaryOp {
+        /// The operator
+        op: BinaryOp,
+        /// Left operand
+        left: Arc<Expr>,
+        /// Right operand
+        right: Arc<Expr>,
+    },
```
cdisselkoen: Any reason to flatten the unary ops (Not/Neg) into Expr, but not the binary ops?

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1164 @@
...
+    /// Record literal
+    Record(BTreeMap<String, Arc<Expr>>),
+    /// Function call (builtin or extension). Syntactically, this can be either a function-style or
+    /// method-style call depending on the extension.
+    FuncCall {
+        /// Function name
+        name: SmolStr,
+        /// Arguments
+        args: Vec<Arc<Expr>>,
+    },
```
cdisselkoen: The comment says "builtin or extension", but builtin calls like `.containsAny()` or `.getTag()` are actually handled above in `BinaryOp`. Maybe this should be `ExtFuncCall` and specifically for extension functions?

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1164 @@
...
+    /// Check if set/record is empty
+    IsEmpty(Arc<Expr>),
```
cdisselkoen: I could be remembering wrong but I think this is only for sets, not records

cedar-policy-core/src/pst/expr.rs
victornicolet: I think it depends on what we define as "backwards compatible"? We can also force clients to not depend on exhaustively matching on the enums. 

I think we can achieve compile-time backwards compatible changes if we want to extend types, whether it's because we have less restrictive types (have a variant that accepts any name) or because we force non-exhaustive matches (i.e. #[non_exhaustive](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute)). By compile-time backwards compatible, I mean that a change in the reprsentation will not cause clients that previously compiled to fail.

cedar-policy-core/src/pst/constraints.rs
victornicolet: They do not, good catch.

cedar-policy-core/src/pst/expr.rs
victornicolet: Yes, some confusion on my part on what is a builtin and what is an extension from the point of view of a user. Is "extension" mostly an internal concept? Should we actually move the `.containsAny` and `.getTag` to be functions instead of binary operators, or vice-versa?

cedar-policy-core/src/pst/expr.rs
victornicolet: I guess the fact that there are a lot more binary operators, and using "not" is very common and it's marginally more convenient to have it as a top level enum, but we could move it inside a `UnaryOp` with `Neg` and `IsEmpty` as well.

cedar-policy-core/src/pst/expr.rs
victornicolet: No good reason. I must have thought we needed more types of EntityTypes.

cedar-policy-core/src/pst/expr.rs
victornicolet: I think that's worth a small discussion, and from what I can see in the doc everything is an "operation" and it would make sense to treat all those uniformly.

The split mostly followed the AST/EST's shapes.

cedar-policy-core/src/pst/expr.rs
john-h-kastner-aws: I think the design needs to be able to maintain compiler-level backwards compatibility of the PST for backwards compatible changes to the Cedar language. We're not going to want to release new library major version for minor language versions.

At least, that's the attitude we've had so far. I've argued before that we should be comfortable with the library and language version getting out of sync.

john-h-kastner-aws: This is kind of unfortunate since I might like to force consumers of the library to update their code if we add a new operator/slot kind.

cedar-policy-core/src/pst/expr.rs
john-h-kastner-aws: `name: SmolStr` feels like the wrong design here. The actual ought to be part of an enum (`BinOp`). The extension functions could use `SmolStr` as well, but it would also be reasonable to represent them as an enum in the PST (committing to the notion that extensions can't be user defined, but we've already done that for the most part)

cedar-policy-core/src/pst/expr.rs
cdisselkoen: I think RFC 98 will probably require library breaking changes in other places (i.e., a `cedar-policy` 5.0 using language version 4.x). However, it might be nice if consumers of the PST-specifically were not broken, if we have the opportunity to do that. OTOH, if we're breaking consumers of other parts of the public API it's ok to break consumers of the PST too.

victornicolet: Do you agree the non-exhaustive attribute on the public structs and enums would be sufficient to ensure compiler backwards compatibility, and probably the better option over not having enums?

On the other hand, if there are language characteristics that we know would require breaking changes in other places, it would be nice that a library update forces the client to update their code. I can't predict that with my lack of experience here.

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1404 @@
...
+pub struct Name {
+    /// Basename (the final component of the name)
+    pub id: SmolStr,
+    /// Namespace components (empty for unqualified names)
+    pub namespace: Arc<Vec<SmolStr>>,
+}
+
+impl Name {
+    /// Constructs a simple (unqualified) name.
+    pub fn simple(id: impl Into<SmolStr>) -> Self {
```
cdisselkoen: perhaps this should be named `unqualified()`?

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1404 @@
...
+        EntityType(Name {
+            id: et.name().basename().to_smolstr(),
+            namespace: Arc::new(
+                et.name()
+                    .0
+                    .namespace_components()
+                    .map(|id| id.to_smolstr())
+                    .collect(),
+            ),
+        })
```
cdisselkoen: Any way to reuse the implementation of `From<ast::Name> for Name` in this function? It looks very similar

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1404 @@
...
+/// Literal values
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[non_exhaustive]
+pub enum Literal {
+    /// Boolean literal
+    Bool(bool),
+    /// Integer literal
+    Long(i64),
+    /// String literal
+    String(String),
```
cdisselkoen: Should this be `SmolStr`?

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1404 @@
...
+    },
+    /// Set literal
+    Set(Vec<Arc<Expr>>),
+    /// Record literal
+    Record(BTreeMap<String, Arc<Expr>>),
+    /// Representation of an unknown for partial evaluation
+    Unknown {
+        /// Name of the unknown
+        name: SmolStr,
+    },
```
cdisselkoen, line 531: discussion: should we remove this, and just fail PST conversion for ASTs that contain Unknowns? once we fully remove partial evaluation in favor of TPE, we won't have this node in the AST, right?

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1404 @@
...
+                f,
+                "function {} expects {} argument(s), got {}",
+                name, expected, got
+            ),
+        }
+    }
+}
+
+impl std::error::Error for ExprConstructionError {}
+
```
cdisselkoen: nit: In this codebase, typically instead of manually impl'ing `Display` and `Error` for error types, we use `thiserror` to `derive(Error)` and add `#[error]` annotations to each variant. You can find plenty of examples in this repo

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1404 @@
...
+        parse_errors: crate::parser::err::ParseErrors,
+    ) -> Result<Self::Expr, Self::ErrorType> {
+        // PST doesn't support error nodes for now, it will propagate parse errors
+        Err(parse_errors)
+    }
+}
+
+impl std::fmt::Display for Expr {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // This Display implementation is mostly for debugging purposes
```
cdisselkoen: Might be good to call out that this specifically does not print valid Cedar expressions from valid PSTs (I assume that is a non-goal) -- eg, it doesn't print method-style calls correctly, doesn't print get-attr for attributes containing whitespace correctly, etc. Rather than duplicate all that logic here, we can use this comment to point people to converting to EST and printing that, as a better alternative if what you want to do is print a valid Cedar expression.

cedar-policy-core/src/pst/policy.rs
```diff
@@ -0,0 +1,179 @@
...
+        match self {
+            Clause::When(expr) => write!(f, "when {{ {} }}", expr),
+            Clause::Unless(expr) => write!(f, "unless {{ {} }}", expr),
+        }
+    }
+}
+
+impl std::fmt::Display for Policy {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // Display in Cedar syntax
```
cdisselkoen: This won't be valid Cedar syntax for the same reason that the `Display` for `pst::Expr` won't be valid Cedar syntax for expressions. I think that's OK to be a non-goal, but we should call it out in this comment.

cedar-policy-core/src/pst/expr.rs
victornicolet: I actually did call it `unqualified` previously on my staging branch because it's what it is, but then renamed it to `simple` because it felt nicer to use, but I can revert to unqualified.

cedar-policy-core/src/pst/expr.rs
victornicolet: Good point, I'll add a comment. The intention was mostly to have a display before the conversion to/from EST are implemented.

cedar-policy-core/src/pst/expr.rs
victornicolet: Yes, I guess most if not all string syntax entities are SmolStr in the rust code?

cedar-policy-core/src/pst/expr.rs
victornicolet: Is the Unknown something the client should never see and construct?
From a practical standpoint, currently the ExprBuilder interface cannot fail on `unknown` construction (unless we panic of course). 
Is that a change that would require removing Unknowns everywhere (AST, EST, the "unknown" in the extension functions), and will it be breaking?

cdisselkoen: Removing it from the AST should be nonbreaking because the AST is not exposed publicly. If the EST currently has an Unknown node, then removing it is breaking. Does the EST have an Unknown node, or does it represent unknowns via an extension function `unknown()`?

john-h-kastner-aws: From a quick look we should be able to remove the `unknown` in the AST, leaving the extension function representation. IIRC, that's how the EST does it.

It'd be a larger change though.

victornicolet: The EST does represent unknowns with an extension function `unknown()`. Although I think removing the "unknown" extension supported in the PST to represent unknowns would still be a breaking change? And not having the `Unknown` node in the PST would make the language EST and PST can represent different?

victornicolet: I would lean towards having `Unknown` in the PST for now since it's also documented in the [JSON policy format](https://docs.cedarpolicy.com/policies/json-format.html#JsonExpr-Unknown)

victornicolet:
A summary of the important discussion topics in this PR:
- regarding backwards compatibility, some enum types are marked as `#non_exhaustive` (e.g unary and binary operations, the main `Expr` enum, the `Literal` enum, and for RFC 98, the `SlotId`)
- the PST has `Unknown` and `Error` nodes for now, which are there mostly to make conversion from other representations non-faillible. An `Error` cannot be constructed by a client, and a client cannot inspect the contents of an error node. `Unknown` can be constructed for now, we are also discussing whether we should even have it.

@john-h-kastner-aws and @cdisselkoen what do you think? Is there anything that needs resolving before merging?
We can also keep this PR open and I can make PRs to this branch for the rest of the code.

cdisselkoen:
I'm good to merge this as-is, and if necessary we can continue the discussion going forward. I also think it's easier to merge this now and not make a whole bunch of PRs into this branch -- among other things, it will make it easier to ensure that this work doesn't keep diverging from `main` and creating annoying merge conflicts.

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1388 @@
...
+#[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord, Hash)]
+#[non_exhaustive]
+pub enum SlotId {
+    /// Principal slot
+    Principal,
+    /// Resource slot
+    Resource,
+}
+
+impl From<ast::SlotId> for SlotId {
```
john-h-kastner-aws, line 41: nit: we'll probably want `#[doc(hidden)]` on conversions from internal types

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1388 @@
...
+
+impl From<ast::Name> for Name {
+    fn from(name: ast::Name) -> Self {
+        let ast::Name {
+            0: ast::InternalName { id, path, .. },
+        } = name;
+        Name {
+            id: id.into_smolstr(),
+            namespace: Arc::new(
+                Arc::try_unwrap(path)
```
john-h-kastner-aws: Could this be `Arc::unwrap_or_clone`?

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1388 @@
...
+        write!(f, "{}", self.id)?;
+        Ok(())
+    }
+}
+
+/// Entity type name
+///
+/// Represents the type of an entity in Cedar (e.g., `User`, `Photo`, `Namespace::Resource`)
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct EntityType(pub Name);
```
john-h-kastner-aws, line 147: Is there a way this could be the existing `EntityTypeName` struct?

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1388 @@
...
+        let got = args.len();
+
+        if expected != got {
+            return Err(ExprConstructionError::WrongArity {
+                name: name.to_string(),
+                expected,
+                got,
+            });
+        }
+        Ok(match args.len() {
```
john-h-kastner-aws, line 590: Might be able to write this with `match args.as_slice()` and avoid unwraps

cedar-policy-core/src/pst/expr.rs
```diff
@@ -0,0 +1,1388 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // This Display implementation is mostly for debugging purposes, and it does not print
+        // valid Cedar expressions.
+        // If you need to print a valid Cedar expression from a PST expression, you should convert
+        // it to an EST expression first.
+        match self {
+            Expr::Literal(lit) => match lit {
+                Literal::Bool(b) => write!(f, "{}", b),
+                Literal::Long(i) => write!(f, "{}", i),
+                Literal::String(s) => write!(f, "\"{}\"", s.escape_default()),
```
john-h-kastner-aws: We've used `escape_debug` more often in other places. Not sure how they're different

cedar-policy-core/src/pst/expr.rs
victornicolet: I guess it represents the same thing (and `EntityTypeName` wraps `ast::EntityType`) although I don't think you can construct the name, you can only parse it? One goal was also to make the PST's representation completely decoupled from the internal representations, but this of course creates duplication. Not sure what's the best solution in this case.
This solution allows creating a name from a list of strings for the namespace and doesn't require parsing.

cedar-policy-core/src/pst/expr.rs
victornicolet: I don't know if I can without adding a `.clone()` because `.as_slice()` gives a reference. Current solution is a little verbose but doesn't clone

cedar-policy-core/src/pst/expr.rs
victornicolet: Documentation indicates they're handling unicode differently. I'll use `escape_debug` for consistency then.

cedar-policy-core/src/pst/expr.rs
victornicolet: I'll keep it as-is in this PR, we can change our mind later.

PR #2186, opened 2026-02-27
Implement AST to PST conversion

victornicolet:
This is an implementation of conversion functions between ASTs and PSTs. Conversion between ASTs and ESTs will come afterwards, but the EST uses some ast types, so it made more sense to do a PR for AST first. 

All the conversion traits and functions are factored in the `ast_conversions` module. 

Some unrelated refactoring for display PST's operators using already defined constant to avoid multiple definitions across the codebase.

### DIFF
--- a/cedar-policy-core/src/pst/ast_conversions.rs
+++ b/cedar-policy-core/src/pst/ast_conversions.rs
@@ -24,6 +24,7 @@ use super::{
     SlotId, UnaryOp, Var,
 };
 use crate::ast;
+use crate::expr_builder;
 use crate::pst::expr::{ErrorNode, PstBuilder};
 use itertools::Itertools;
 use std::str::FromStr;
@@ -57,8 +58,10 @@ impl TryFrom<Policy> for ast::Template {
             .clauses
             .into_iter()
             .map(|clause| match clause {
-                Clause::When(expr) => (*expr).clone().try_into(),
-                Clause::Unless(expr) => (*expr).clone().try_into().map(|x| builder.clone().not(x)),
+                Clause::When(expr) => Arc::unwrap_or_clone(expr).try_into(),
+                Clause::Unless(expr) => Arc::unwrap_or_clone(expr)
+                    .try_into()
+                    .map(|x| builder.clone().not(x)),
             })
             .rev()
             .collect::<Result<Vec<_>, _>>()?
@@ -73,11 +76,7 @@ impl TryFrom<Policy> for ast::Template {
             .annotations
             .into_iter()
             .map(|(key, val)| {
-                let value = if val.is_empty() {
-                    None
-                } else {
-                    Some(val.to_smolstr())
-                };
+                let value = if val.is_empty() { None } else { Some(val) };
                 (
                     ast::AnyId::new_unchecked(key),
                     ast::Annotation::with_optional_value(value, None),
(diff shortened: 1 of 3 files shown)


cedar-policy-core/src/pst/ast_conversions.rs
```diff
@@ -0,0 +1,621 @@
...
+    ast::Pattern::from_iter(elems)
+}
+
+impl TryFrom<Expr> for ast::Expr {
+    type Error = PstConstructionError;
+
+    fn try_from(expr: Expr) -> Result<Self, PstConstructionError> {
+        expr_to_ast(expr).map_err(|e| {
+            PstConstructionError::InvalidConversion(format!("Failed to convert expr: {:?}", e))
+        })
```
cdisselkoen: I'm not clear why we rewrap the error, which was already a `PstConstructionError` as returned from `expr_to_ast()`? But if we do intend to rewrap the error, we should use `{}` rather than `{:?}` to print the error message itself instead of the Debug representation of the error object

cedar-policy-core/src/pst/ast_conversions.rs
```diff
@@ -0,0 +1,621 @@
...
+            Literal::Long(i) => Ok(builder.val(i)),
+            Literal::String(s) => Ok(builder.val(s)),
+            Literal::EntityUID(uid) => {
+                // Convert PST EntityUID to AST EntityUID using existing TryFrom impl
+                let ast_et: ast::EntityType = uid.ty.try_into().map_err(|e| {
+                    PstConstructionError::InvalidConversion(format!("Invalid entity type: {:?}", e))
+                })?;
+                let ast_eid = ast::Eid::new(uid.eid.as_str());
+                let ast_uid = ast::EntityUID::from_components(ast_et, ast_eid, None);
+                Ok(builder.val(ast_uid))
```
cdisselkoen: This comment doesn't appear to correctly describe what the code is doing

cedar-policy-core/src/pst/ast_conversions.rs
```diff
@@ -0,0 +1,621 @@
...
+                UnaryOp::Not => builder.not(inner),
+                UnaryOp::Neg => builder.neg(inner),
+                UnaryOp::IsEmpty => builder.is_empty(inner),
+                // The other unary operators are extension functions.
+                _ => match op.to_name() {
+                    Some(fn_name) => builder.call_extension_fn(fn_name.clone(), vec![inner]),
+                    None => Err(PstConstructionError::InvalidConversion(format!(
+                        "unknown unary operator: {:?}",
+                        op
+                    )))?,
```
cdisselkoen: Not sure if "InvalidConversion" is a good descriptor of this error, it's more that the input PST was simply invalid. Perhaps `InvalidExpression` or `UnknownFunction`, which are both currently also variants of `PstConstructionError`

cedar-policy-core/src/pst/ast_conversions.rs
```diff
@@ -0,0 +1,621 @@
...
+            elements_into_ast_pattern(pattern),
+        )),
+        Expr::Record(elems) => builder
+            .record(
+                elems
+                    .into_iter()
+                    .map(|(k, v)| Ok((k.into(), expr_to_ast(Arc::unwrap_or_clone(v))?)))
+                    .collect::<Result<Vec<_>, PstConstructionError>>()?,
+            )
+            .map_err(|cstr_err| PstConstructionError::InvalidConversion(format!("{:?}", cstr_err))),
```
cdisselkoen: Here's another place where we should use `{}` instead of `{:?}`. Or, equivalently, just `cstr_err.to_string()`, which uses the `Display` impl like `{}`

cedar-policy-core/src/pst/ast_conversions.rs
```diff
@@ -0,0 +1,621 @@
...
+impl From<ast::Name> for Name {
+    fn from(name: ast::Name) -> Self {
+        let ast::Name {
+            0: ast::InternalName { id, path, .. },
+        } = name;
+        Name {
+            id: id.into_smolstr(),
+            namespace: Arc::new(
+                Arc::try_unwrap(path)
+                    .unwrap_or_else(|arc| (*arc).clone())
```
cdisselkoen: This could be `Arc::unwrap_or_clone()`

cedar-policy-core/src/pst/mod.rs
```diff
@@ -26,13 +26,36 @@
...
+    BinaryOp, EntityType, EntityUID, Expr, Literal, Name, PatternElem, SlotId, UnaryOp, Var,
 };
-pub use policy::{Clause, Effect, Policy};
+pub use policy::{Clause, Effect, Policy, PolicyID};
+
+use crate::ast;
+
+impl Policy {
+    /// Convert this PST policy to an AST policy for evaluation
+    pub fn to_ast(self) -> Result<ast::Policy, PstConstructionError> {
```
cdisselkoen: I believe Rust convention is that methods that consume `self` would be named `into_*` instead of `to_*`. Alternately, it's not clear why we need this method at all when callers could just directly use `try_into()` / `try_from()`. Same below

cedar-policy-core/src/pst/ast_conversions.rs
victornicolet: Thank you, yes there's a few places like this, it should just be the original error (also changed the `{:?}` to `{}` or `to_string()` where possible).

cedar-policy-core/src/pst/ast_conversions.rs
victornicolet: I put an `InvalidExpression`. If we don't make mistakes, a client should never see this particular error.

cedar-policy-core/src/pst/mod.rs
victornicolet: The intent behind this one was to make only the policy and expression level conversions public; I could move the trait implementation here instead (the ast_conversions module is `pub(crate)` so that all the other conversions are only available in the crate. I think this interface will change. For now renamed it with `try_into_ast_policy` and `try_into_ast_expr` to mirror the est's equivalent functions.

cedar-policy-core/src/pst/errors.rs
```diff
@@ -0,0 +1,108 @@
...
+//! - Validating PST structure and semantics
+
+use crate::extensions::ExtensionFunctionLookupError;
+use miette::Diagnostic;
+use smol_str::SmolStr;
+use thiserror::Error;
+
+/// Errors that can occur during PST construction or conversion.
+#[derive(Debug, Clone, PartialEq, Eq, Diagnostic, Error)]
+pub enum PstConstructionError {
```
john-h-kastner-aws, line 32: If this will be a publicly exported error enum it should follow the conventions here https://github.com/cedar-policy/cedar/blob/main/style_guide.md#error-types

cedar-policy-core/src/pst/ast_conversions.rs
```diff
@@ -0,0 +1,1203 @@
...
+    fn try_from(policy: Policy) -> Result<Self, Self::Error> {
+        use crate::expr_builder::ExprBuilder;
+        // Convert clauses - fold them into a single expression (following EST pattern)
+        let builder = ast::ExprBuilder::<()>::new();
+        let mut conds_rev_iter = policy
+            .clauses
+            .into_iter()
+            .map(|clause| match clause {
+                Clause::When(expr) => (*expr).clone().try_into(),
+                Clause::Unless(expr) => (*expr).clone().try_into().map(|x| builder.clone().not(x)),
```
john-h-kastner-aws: Could these be `Arc::unwrap_or_clone`?

cedar-policy-core/src/pst/ast_conversions.rs
```diff
@@ -0,0 +1,1203 @@
...
+
+        // Convert annotations
+        let annotations: ast::Annotations = policy
+            .annotations
+            .into_iter()
+            .map(|(key, val)| {
+                let value = if val.is_empty() {
+                    None
+                } else {
+                    Some(val.to_smolstr())
```
john-h-kastner-aws: Should annotation value be smolstrs in the pst?

cedar-policy-core/src/pst/ast_conversions.rs
```diff
@@ -0,0 +1,1203 @@
...
+impl TryFrom<Expr> for ast::Expr {
+    type Error = PstConstructionError;
+
+    fn try_from(expr: Expr) -> Result<Self, PstConstructionError> {
+        expr_to_ast(expr)
+    }
+}
+
+// Helper to convert PST Expr to AST Expr using the AST builder
+fn expr_to_ast(expr: Expr) -> Result<ast::Expr, PstConstructionError> {
```
john-h-kastner-aws: can/will this be able to use a generic `ExprBuilder`?

cedar-policy-core/src/pst/ast_conversions.rs
```diff
@@ -0,0 +1,1203 @@
...
+        let (ty, eid) = uid.components();
+        EntityUID {
+            ty: ty.into(),
+            eid: eid.into_smolstr(),
+        }
+    }
+}
+
+#[expect(clippy::fallible_impl_from, reason = "the unwrap cannot fail")]
+impl From<ast::Expr> for Expr {
```
john-h-kastner-aws, line 476: If this one can use a generic `ExprBuilder` it should already be covered by `ast::Expr::into_expr`

victornicolet: Indeed it can

cedar-policy-core/src/pst/ast_conversions.rs
victornicolet: It can, I'll revise this

cedar-policy-core/src/pst/errors.rs
victornicolet: Thanks for the pointer to the doc! If that's not too much of an issue, I will refactor the error type in a later PR.

cedar-policy-core/src/pst/ast_conversions.rs
victornicolet: I kept the `From<..>` trait for convenience, but it calls the `into_expr`, which the PST conversion coming next will reuse.

cedar-policy-core/src/pst/mod.rs
```diff
@@ -43,19 +43,24 @@ use crate::ast;
 
 impl Policy {
     /// Convert this PST policy to an AST policy for evaluation
-    pub fn to_ast(self) -> Result<ast::Policy, PstConstructionError> {
+    pub fn try_into_ast_policy(self) -> Result<ast::Policy, PstConstructionError> {
+        // Currently the trait implementations for converting AST to PST are all implemented in
+        // ast_conversions, which is only pub(crate).
+        // This function is the public boundary between PST and AST. We may need more conversion
+        // functions, or we may make ast_conversions pub rather than pub(crate) if needed.
```
cdisselkoen, line 50: If you implement a public trait (like `TryFrom`) for a public type in a `pub(crate)` module, is the trait implementation visible at `pub` or `pub(crate)`?

victornicolet: Great question, I thought the `impl` wouldn't be visible outside the crate. Will look into it by testing it.

PR #2190, opened 2026-03-03
refactor Solver trait for usability

cdisselkoen:
## Description of changes

Refactors the `Solver` trait to include `check_sat_with_model()` instead of `get_model()`, making the API easier to use with fewer implicit ordering constraints. (Previously you were required to call `check_sat()` prior to `get_model()`, and calling `get_model()` first, or after `check_sat()` produced anything other than SAT, was an error. Now, `check_sat_with_model()` will perform all necessary operations for you in the correct sequence.)

- [x] A breaking change requiring a major version bump to `cedar-policy-symcc` (e.g., changes to the signature of an existing API).

### DIFF
--- a/cedar-policy-symcc/src/symcc/solver.rs
+++ b/cedar-policy-symcc/src/symcc/solver.rs
@@ -228,21 +228,14 @@ impl Solver for LocalSolver {
     }
 
     async fn check_sat_with_model(&mut self) -> Result<DecisionWithModel> {
-        self.check_child_process_status().await?;
-        self.smtlib_input().check_sat().await?;
-        self.solver_stdin.flush().await?;
-        let mut output = String::new();
-
-        // Read the first line
-        self.read_line(&mut output).await?;
-        match output.as_str() {
-            "sat\n" => {
+        match self.check_sat().await? {
+            Decision::Sat => {
                 // in the SAT case, we ask the solver for a model, which we expect to be in one of the following forms:
                 // 1. "(\n<the actual model>\n)\n"
                 // 2. "(error ...)\n"
                 self.smtlib_input().get_model().await?;
                 self.solver_stdin.flush().await?;
-                output.clear(); // remove the `sat` line, which is not part of the model
+                let mut output = String::new();
                 self.read_line(&mut output).await?;
                 match output.as_str() {
                     "(\n" => {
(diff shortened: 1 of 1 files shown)


cedar-policy-symcc/src/symcc/solver.rs
```diff
@@ -196,34 +227,45 @@ impl Solver for LocalSolver {
         }
     }
 
-    async fn get_model(&mut self) -> Result<Option<String>> {
+    async fn check_sat_with_model(&mut self) -> Result<DecisionWithModel> {
```
john-h-kastner-aws, line 230: Could this function start with `match self.check_sat().await`?

john-h-kastner-aws (approved):

cedar-policy-symcc/src/symcc/solver.rs
cdisselkoen: Yes. Good call.

cedar-policy-symcc/src/symcc/solver.rs
```diff
@@ -196,34 +227,37 @@ impl Solver for LocalSolver {
...
-                        reason = "`output.len() - len` gives the end index of `output` before the `read_line`"
-                    )]
-                    if &output[output.len() - len..] == ")\n" {
-                        break;
+    async fn check_sat_with_model(&mut self) -> Result<DecisionWithModel> {
+        match self.check_sat().await? {
+            Decision::Sat => {
+                // in the SAT case, we ask the solver for a model, which we expect to be in one of the following forms:
+                // 1. "(\n<the actual model>\n)\n"
+                // 2. "(error ...)\n"
```
victornicolet, line 235: It's not the right PR to solve this, but would a s-epxression parser work better here? 
I guess Z3 and CVC5 follow that specific convention of newlines, but that might not be the case for any SMTLIB solver.

cdisselkoen: That's a good idea for a future PR, and we already have an s-expression parser in this codebase (in decoder.rs). Make an issue?

cdisselkoen: Also, our model-parser explicitly only works with CVC5 models (assumes some CVC5 conventions) and will not work with Z3 models without additional work.

cdisselkoen:
Most of the uncovered lines are error conditions involving invalid/unexpected solver output, and in `WriterSolver` where we already did not have tests. I'll merge this despite only 73% coverage of added/modified lines.

PR #2201, opened 2026-03-04
pst <-> est conversions

victornicolet:
This PR adds the PST to EST, and EST to PST conversion functionality.

Makes some changes in the EST to be able to construct and extract some information, and add an error type in the `ExprBuilder` to allow expressing some fallible builders.

### DIFF
--- a/cedar-policy-core/src/pst/err.rs
+++ b/cedar-policy-core/src/pst/err.rs
@@ -185,7 +185,7 @@ pub mod error_body {
         pub(crate) description: String,
     }
 
-    /// Invalid entity type name
+    /// Invalid entity type error (often failure to parse the name)
     #[derive(Debug, Clone, PartialEq, Eq, Diagnostic, Error)]
     #[error("invalid entity type: `{description}`")]
     pub struct InvalidEntityTypeError {
@@ -193,8 +193,8 @@ pub mod error_body {
     }
 
     impl InvalidEntityTypeError {
-        /// The invalid entity type
-        pub fn entity_type(&self) -> &str {
+        /// The description of the entity type error
+        pub fn description(&self) -> &str {
             &self.description
         }
     }


cedar-policy-core/src/pst/est_conversions.rs
```diff
@@ -0,0 +1,1231 @@
...
+
+    fn try_from(est_policy: est::Policy) -> Result<Self, Self::Error> {
+        let clauses: Result<Vec<_>, _> = est_policy
+            .conditions
+            .into_iter()
+            .map(|c| c.try_into())
+            .collect();
+
+        Ok(Policy {
+            id: PolicyID("policy".into()),
```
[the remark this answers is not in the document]
victornicolet: We will provide proper API working with policy IDs later.

cedar-policy-core/src/est/expr.rs
```diff
@@ -1555,10 +1765,8 @@ impl std::fmt::Display for ExtFuncCall {
 
 impl BoundedDisplay for ExtFuncCall {
     fn fmt(&self, f: &mut impl std::fmt::Write, n: Option<usize>) -> std::fmt::Result {
-        #[expect(clippy::unreachable, reason = "safe due to INVARIANT on `ExtFuncCall`")]
-        let Some((fn_name, args)) = self.call.iter().next() else {
-            unreachable!("invariant violated: empty ExtFuncCall")
-        };
+        #[expect(clippy::unwrap_used, reason = "safe due to INVARIANT on `ExtFuncCall`")]
+        let (fn_name, args) = self.try_components().unwrap();
```
victornicolet, line 1769: This changes the behavior slightly; `try_components` checks for exactly one call, not just at least one.

cedar-policy-core/src/est/expr.rs
```diff
@@ -913,6 +947,192 @@ impl Expr {
...
+                left,
+                entity_type,
+                in_expr,
+            }) => {
+                let entity_type_name = ast::EntityType::from_normalized_str(&entity_type)
+                    .map_err(FromJsonError::InvalidEntityType)?;
+                let left_expr = Arc::unwrap_or_clone(left).try_into_expr::<B>()?;
+                match in_expr {
+                    // unlike try_into_ast, call is_in_entity_type instead of desugaring here
+                    Some(in_expr) => Ok(builder.is_in_entity_type(
```
john-h-kastner-aws, line 1079: could `try_into_ast` reasonably call this too if it simplify that function a little bit

cedar-policy-core/src/pst/err.rs
```diff
@@ -158,15 +187,15 @@ pub mod error_body {
...
     pub struct InvalidEntityTypeError {
-        pub(crate) entity_type: String,
+        pub(crate) description: String,
     }
 
     impl InvalidEntityTypeError {
         /// The invalid entity type
         pub fn entity_type(&self) -> &str {
-            &self.entity_type
+            &self.description
```
cdisselkoen: This is a confusing change

victornicolet: Sorry, I did mean to do that rename but didn't fix the accessor. Initially I thought only entity_type names would go in the error body, but with the conversion from the `FromJsonError` we might get more complete descriptions of why it's an invalid entity type.

cedar-policy-core/src/est/expr.rs
victornicolet: Possibly, and as a result there would be fewer places where the desugaring of "is_in_entity_type" is defined. I'll make a separate PR for this.

PR #2216, opened 2026-03-09
refactor some builder functions into `_arc` version to save on clones

victornicolet:
## Description of changes

Refactors some of `ExprBuilder`'s functions to require implementation with `Arc<..>` and then provide default without `Arc<..>`. The benefits are probably not very significant, but I noticed all implementations were using `Arc<..>`. 
- Adding `has_attr_arc` and `get_attr_arc` allows extended has to clone `Arc<..>`. 
- Adding `is_entity_type_arc` and `is_in_arc` allows `is_in_entity_type` to clone Arc. 
- `ite_arc` was in one of the implementations but not all, so I made it instead the required implementation instead of `ite`, and `ite` is provided by the trait default impl.

This doesn't simplify the `ExprBuilder` interface but does simplify the implementations.

-  [x] Does not require updates because my change does not impact the Cedar language specification.

### DIFF
--- a/cedar-policy-core/src/expr_builder.rs
+++ b/cedar-policy-core/src/expr_builder.rs
@@ -31,6 +31,7 @@ use crate::{
 #[cfg(feature = "tolerant-ast")]
 use crate::parser::err::ParseErrors;
 use std::fmt::Debug;
+use std::sync::Arc;
 
 /// Defines a generic interface for building different expression data
 /// structures.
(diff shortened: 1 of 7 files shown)


cedar-policy-core/src/expr_builder.rs
```diff
@@ -109,8 +110,26 @@ pub trait ExprBuilder: Clone {
     fn slot(self, s: SlotId) -> Self::Expr;
 
     /// Create a ternary (if-then-else) `Expr`.
-    fn ite(self, test_expr: Self::Expr, then_expr: Self::Expr, else_expr: Self::Expr)
-        -> Self::Expr;
+    fn ite(
```
john-h-kastner-aws, line 113: Are the non-arc versions still widely used? If we can mostly call the arc variant directly I'd consider removing these so that we can't accidentally clone an arc to call it when we could be calling the arc function.

victornicolet: I think in most cases the non-arc versions are used, for example in conversions, where we actually have a new expression and need to wrap it in an `Arc` (hence replacing by the `_arc` version we would need to wrap the arguments with `Arc::new(...)`)

PR #2218, opened 2026-03-09
doc: public types in the PST, add hidden doc attributes

victornicolet:
## Description of changes

This PR mostly adds documentation and `#[doc(hidden)]` attributes where appropriate.
Also removes the code in `pst/mod.rs` because there is no need for an extra API for conversions.

### DIFF
--- a/cedar-policy-core/src/pst/mod.rs
+++ b/cedar-policy-core/src/pst/mod.rs
@@ -14,8 +14,12 @@
  * limitations under the License.
  */
 
-//! The PST is a public syntax tree representation of Cedar policies designed for programmatic
-//! manipulation. All the types necessary to build a valid PST are public.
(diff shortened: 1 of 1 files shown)


cedar-policy-core/src/pst/mod.rs
```diff
@@ -14,17 +14,128 @@
  * limitations under the License.
  */
 
-//! Public Syntax Tree (PST) - Ergonomic representation for programmatic policy construction
+//! The PST is a public syntax tree representation of Cedar policies designed for programmatic
+//! manipulation. All the types necessary to build a valid PST are public.
```
john-h-kastner-aws: Probably worth putting this in a more prominent non-doc comment for our own benefit since it goes against the usual pattern in our API

PR #2222, opened 2026-03-09
adding policy level ast to pst conversion and template linking

victornicolet:
## Description of changes

This adds a few pieces to the PST to make it possible to use it as one of the `LosslessPolicy`'s representation:
- converting `ast::Policy` to `pst::Policy`
- link a `pst::Policy`

Will still need to figure out some details related to linking policies (right now, it is supported so that we can represent linked policies from the AST).


cedar-policy-core/src/ast/policy.rs
```diff
@@ -181,6 +181,25 @@ impl Template {
         Template::from(body)
     }
 
+    /// Get the components of the underlying template.
+    ///
+    /// Returns `None` when the underlying template doesn't have any components (i.e., it's an error
+    /// node).
+    #[expect(clippy::type_complexity, reason = "policies just have many components")]
+    pub(crate) fn into_template_components_opt(
```
victornicolet, line 189: This is a little ugly, alternative would be to make nested struct's field pub(crate), but there are also multiple levels of nesting.

cedar-policy-core/src/pst/ast_conversions.rs
```diff
@@ -726,311 +878,137 @@ mod tests {
...
-            ActionConstraint, Clause, Effect, Policy, PolicyID, PrincipalConstraint,
-            ResourceConstraint,
-        };
-        use std::collections::BTreeMap;
+    /// Helper to normalize whitespace for policy string comparison
+    fn normalize(s: &str) -> String {
+        s.replace('\n', " ")
+            .split_whitespace()
+            .collect::<Vec<_>>()
+            .join(" ")
```
cdisselkoen, line 925: nit: if you `use itertools::Itertools`, then you can `.join(" ")` directly on an iterator without collecting

victornicolet: Thanks, good tip, I'll put this small refactor in my next PR refactoring tests: https://github.com/cedar-policy/cedar/commit/0a09d449cce583ce1e1835d9f077fc5eeb39843c#diff-8e0b37e9be8039797442d11e7c68d71fbd84c7f41a3bc72ae4d014f0383e6c5fL648

cedar-policy-core/src/pst/expr.rs
```diff
@@ -495,6 +498,71 @@ pub(crate) struct ErrorNode {
 }
 
 impl Expr {
+    /// Fill in any slots in this expression using the values in `vals`.
+    /// Returns an error if a slot is encountered that has no mapping in `vals`.
+    pub fn link(self, vals: &HashMap<SlotId, EntityUID>) -> Result<Self, LinkingError> {
```
john-h-kastner-aws: I don't think I understand why we need a `link` function that does the actual substitution on the PST. In evaluation we treat the slots as variables.

john-h-kastner-aws: > it is supported so that we can represent linked policies from the AST

Is the point that a template linked policy should converted to PST from AST should have the substitution done internally?

victornicolet: I guess there's something I misunderstood, since the AST and EST support slots in the expressions, I assumed slots were supported in when/unless clauses. But they're rejected at the level of the parser?
Why do we have slots in the AST/EST in that case?

victornicolet: I think slots in expression would still be linked to the slot values in the env:
https://github.com/cedar-policy/cedar/blob/f5677f71cca5716e760a623ee5842f90f49e2e88/cedar-policy-core/src/evaluator.rs#L456-L457
But I guess it's done at evaluation time here, as opposed to when the template is linked?

victornicolet: I mentioned linking policies is something I'm still unsure about, and it should probably behave like for the ast, so I might be wrong here. 
It seems like this would make the PST more expressive than the other representations (in terms of what templates the user can express).

victornicolet:
I refactored the policy types significantly now you have:
- a `Template` that is basically the syntax body of a policy, template or linked policy.
- two separate types `StaticPolicy` and `LinkedPolicy` representing either. 
- a `Policy` enum representing both in a type.
At the top level, you might only manipulate and inspect a policy, but you need to explicitly construct either. You cannot build a `StaticPolicy` without using a method that checks the absence of slots.
I will keep refining on the checks, but I hope this is a better design than what was there before.

cedar-policy-core/src/ast/policy.rs
```diff
@@ -480,6 +499,12 @@ impl Policy {
         Self::new(Arc::new(t), None, SlotEnv::new())
     }
 
+    /// Get the owned template, link id (if this is a template-linked policy)
+    /// and slot environment.
+    pub(crate) fn into_components(self) -> (Arc<Template>, Option<PolicyID>, SlotEnv) {
+        (self.template, self.link, self.values)
+    }
```
cdisselkoen, line 506: My instinct is that it might be better to just have these fields be `pub(crate)`. but open to discussion

victornicolet: I think it could simplify away some struct methods too, and there are some places (AST to EST conversion) where there's extra cloning. Downside is that now we would be able to break invariants on the policy inside the crate by accessing those fields directly.

cdisselkoen: I'd argue that the caller already needs to be aware of those invariants (eg on `self.link`) because of this function, but it's true that the caller could do "even more" bad things with pub(crate) fields -- particularly, constructing invalid instances of `Self`.

victornicolet: I'll make a PR that proposes to changes these fields to `pub(crate)` and we can decide then based on those changes.

victornicolet: #2232

PR #2224, opened 2026-03-11
normalize changelogs on 4.10.x release branch

cdisselkoen:


victornicolet:
I'm not sure if I am missing something, if we merge this into branch then the link will point to itself? Which won't contain the changelog? Or should this merge on 4.10.x?

cdisselkoen:
yes you are right lol, the base branch should be 4.10.x. Changed now.

PR #2227, opened 2026-03-12
Cherry-picks to 4.10.x

cdisselkoen:
Cherry-picks two PRs to 4.10.x:

- #2218 
- #2225

### DIFF
--- a/cedar-policy-core/src/pst/ast_conversions.rs
+++ b/cedar-policy-core/src/pst/ast_conversions.rs
@@ -33,6 +33,7 @@ use itertools::Itertools;
 use std::str::FromStr;
 use std::sync::Arc;
 
+#[doc(hidden)]
 impl TryFrom<Policy> for ast::Policy {
     type Error = PstConstructionError;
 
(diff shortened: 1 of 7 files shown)


john-h-kastner-aws:
Is the PST coming to the public API with this release?

cdisselkoen:
> Is the PST coming to the public API with this release?

I just looked over the diff in the `cedar-policy` crate from v4.9.1 to this PR branch, and the only changes I see to `cedar-policy` are in the `proto` module (the ownership stuff), tests (eg for the new extended-has-in-policy-json feature), and the version number bumps.

victornicolet:
I don't think there is anything related to the PST in the `cedar-policy` crate, but this would bring documentation for the PST in `cedar-policy-core`. Is `cedar-policy-core` considered not public API?

cdisselkoen:
> Is cedar-policy-core considered not public API?

Correct, it's considered internal and we actively discourage anyone other than us from depending on it. It has frequent breaking changes, even in patch releases. `cedar-policy` is the stable API we commit to maintaining long-term.

PR #2233, opened 2026-03-16
add pst to api, revamp lossless policy repr

victornicolet:
## Description of changes

Adds public API methods to construct and extract PST representations from cedar_policy::Policy and cedar_policy::Template, and re-exports the pst module from the cedar-policy crate.

### Public API additions

Re-export `pub use cedar_policy_core::pst` — PST types accessible as `cedar_policy::pst::*`.

Methods for `Template`:
- `from_pst(pst::Template) -> Result<Self, PstConstructionError>` — construct from a PST template (rejects static policies with no slots)
- `to_pst(&self) -> Result<pst::Template, PstConstructionError>` — get PST representation
- `try_into_pst(self) -> Result<pst::Template, PstConstructionError>` — consuming version

Policy:
- `from_pst(pst::Policy) -> Result<Self, PstConstructionError>` — construct from a PST policy (accepts both static and linked)
- `to_pst(&self) -> Result<pst::Policy, PstConstructionError>` — get PST representation
- `try_into_pst(self) -> Result<pst::Policy, PstConstructionError>` — consuming version

Also added some error conversion functionality for `pst::PstConstructionError`: `invalid_conversion(description) -> Self` public constructor for conversion errors EmptyPolicy` variant added. 

### Internal changes

Split `LosslessPolicy` into two types to properly support PST as a lossless representation:
- `LosslessTemplate` — used by cedar_policy::Template, stores pst::Template in its Pst variant
- `LosslessPolicy` — used by cedar_policy::Policy, stores pst::Policy in its Pst variant

This split was necessary because PST has distinct types for templates vs policies, unlike EST which uses a single type for both. `LosslessTemplate::link()` produces a `LosslessPolicy`, bridging the two when `PolicySet::link()` instantiates a template.

### DIFF
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -3683,6 +3683,7 @@ pub struct Policy {
     /// We can't just derive this on-demand from `ast`, because the AST is lossy:
     /// we can't reconstruct an accurate CST/EST/policy-text from the AST, but
     /// we can from the EST (modulo whitespace and a few other things like the
+    /// we can from the EST or PST (modulo whitespace and a few other things like the
     /// order of annotations).
     pub(crate) lossless: LosslessPolicy,
 }


cedar-policy/src/api.rs
cdisselkoen:
```suggestion
   /// we can from the EST or PST (modulo whitespace and a few other things like the
   /// order of annotations).
```

cedar-policy-core/src/pst/err.rs
```diff
@@ -31,6 +31,10 @@ use crate::est;
 #[derive(Debug, Clone, PartialEq, Eq, Diagnostic, Error)]
 #[non_exhaustive]
 pub enum PstConstructionError {
+    /// Trying to construct an empty policy
+    #[error("empty policy")]
+    #[diagnostic(code(pst::empty_policy))]
+    EmptyPolicy,
```
john-h-kastner-aws, line 37: this error doesn't make sense without more context. We should improve it if it's part of the public API

victornicolet: I'll make a note of this one in particular; in general, I wanted to do a pass over the errors once we have a more complete picture.

PR #2241, opened 2026-03-19
Eliminate error variant of PST expression and use Infallible trait

victornicolet:
## Description of changes

Eliminates the `Error` variant of the `pst::Expr`, which was there because we needed to handle the possibility to fail in the construction of function calls. We may introduce an `Error` variant for other purposes later, but the current reason it is there wasn't very good.

Instead, the builder is allowed to fail, but in most places the error is `Infallible`. This his handled by new traits `IsInfallible` and `UnwrapInfallible` that allows safely unwrapping for implementations that we not expected to fail for function creation.
In the future, we may want some builders to implement validation in function construction (e.g. to avoid converting method calls without arguments).

### DIFF
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -505,10 +505,7 @@ impl<T> Expr<T> {
                 .record(
                     Arc::unwrap_or_clone(map)
                         .into_iter()
-                        .map(|k_v| -> Result<_, B::BuildError> {
-                            let (k, v) = k_v;
-                            Ok((k, v.try_into_expr::<B>()?))
-                        })
+                        .map(|(k, v)| Ok((k, v.try_into_expr::<B>()?)))
                         .collect::<Result<Vec<_>, _>>()?,
                 )
                 .unwrap()),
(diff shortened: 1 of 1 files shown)


cedar-policy-core/src/ast/expr.rs
```diff
@@ -434,89 +434,106 @@ impl<T> Expr<T> {
...
                 clippy::unwrap_used,
                 reason = "`map` is a map, so it will not have duplicate keys, so the `.record()` constructor cannot error"
             )]
-            ExprKind::Record(map) => builder
+            ExprKind::Record(map) => Ok(builder
                 .record(
                     Arc::unwrap_or_clone(map)
                         .into_iter()
-                        .map(|(k, v)| (k, v.into_expr::<B>())),
+                        .map(|k_v| -> Result<_, B::BuildError> {
```
katherine-hough: This isn't an issue, but I'm curious why you changed this to no longer destructure the closure parameters?

cedar-policy-core/src/ast/expr.rs
```diff
@@ -434,89 +434,106 @@ impl<T> Expr<T> {
...
                 reason = "error type is Infallible so can never happen"
             )]
-            ExprKind::Error { .. } => builder
+            ExprKind::Error { .. } => Ok(builder
                 .error(ParseErrors::singleton(ToASTError::new(
                     ToASTErrorKind::ASTErrorNode,
                     Some(Loc::new(0..1, "AST_ERROR_NODE".into())),
                 )))
-                .unwrap(),
+                .unwrap()), // we could have unwrap_infallible + trait bound  but attribtutes in where clauses are unstable
```
katherine-hough: Nit: attribtutes -> attributes

cedar-policy-core/src/ast/expr.rs
victornicolet: Good catch, probably a result of multiple attempts to get the right type to `try_collect`, but totally unnecessary here. I will clean up

PR #2248, opened 2026-03-23
Making sure we fail for Error nodes instead of converting to Any.

lianah:

### DIFF
--- a/cedar-policy/src/test/test.rs
+++ b/cedar-policy/src/test/test.rs
@@ -12427,7 +12427,6 @@ mod pst_api {
 mod tolerant_ast_tests {
     use super::*;
     use cedar_policy_core::ast;
-    use std::sync::Arc;
 
     /// Helper: build a TemplateBody with ActionConstraint::ErrorConstraint
     fn template_body_with_error_action() -> ast::TemplateBody {


cedar-policy/src/api.rs
```diff
@@ -3581,12 +3581,11 @@ impl Template {
...
             ast::ActionConstraint::Eq(id) => ActionConstraint::Eq(id.as_ref().clone().into()),
             #[cfg(feature = "tolerant-ast")]
+            #[expect(clippy::unimplemented, reason = "experimental feature")]
             ast::ActionConstraint::ErrorConstraint => {
                 // We will only have an ErrorConstraint if we are using a parser that allows Error nodes
                 // It is not recommended to evaluate an AST that allows error nodes
-                // If somehow someone tries to evaluate an AST that includes an Action constraint error, we will
-                // treat it as `Any`
-                ActionConstraint::Any
+                unimplemented!("It is not safe to evaluate a tolerant-ast with Error nodes.")
```
john-h-kastner-aws: This doesn't sound quite right. Evaluation (relatively) safely treats it as an error, and this function has nothing to do with evaluation.

victornicolet: To be clear, you're suggesting the message in the `unimplemented` should change, not that we should return an error here?

john-h-kastner-aws: yeah, just from a documentation perspective I don't the message correctly explains why this is unimplemented.

lianah: Good point. I updated it to "internal ErrorConstraint cannot be represented in the public API"

Issue #2237, opened 2026-03-18
Improving the PST error APIs


victornicolet:
We should improve the types of errors and how errors are propagated in the PST related APIs.

Currently, there might be some error types that should not be public, and some error conversions that might lose information.

> this error doesn't make sense without more context. We should improve it if it's part of the public API

_Originally posted by @john-h-kastner-aws in https://github.com/cedar-policy/cedar/pull/2233#discussion_r2947764902_

victornicolet:
#2316 eliminates an additional error type.

PR #2249, opened 2026-03-24
Cleanup pst error types

victornicolet:
## Description of changes

Refactors the `PstConstructionError` type to be more precise:
- removes the `InvalidConversion` generic type,
- removes the `NotImplemented` error which was here for development purposes, but is not used,
- removes the `FunctionLookup` error which talks about internal representation of extension in favor of the `UnknownFunction` error, which tells the user enough about the error,
- removed `InvalidAttributePath / InvalidAttributePathError` and `EmptyAttributePath / EmptyAttributePathError` which were never used
- remove `InvalidRecord / InvalidRecordError`, which were never constructed (we have a duplicate key error).
- adds the `WrongSlotPosition` error for wrong slots in a template. (e.g. `?resource` where principal is expected).

Also makes some changes to use the semantically richer error (e.g. `InvalidEntityType` instead of `ParsingFailed` when parsing an entity type fails).

## Issue #, if available

#2237 

### DIFF
--- a/cedar-policy/src/test/test.rs
+++ b/cedar-policy/src/test/test.rs
@@ -12422,3 +12422,45 @@ mod pst_api {
         assert_pst_sets_eq(&pst_set, &recovered, "both_slots_linked roundtrip");
     }
 }
+
+#[cfg(feature = "tolerant-ast")]
+mod tolerant_ast_tests {
+    use super::*;
+    use cedar_policy_core::ast;
+
+    /// Helper: build a TemplateBody with ActionConstraint::ErrorConstraint
+    fn template_body_with_error_action() -> ast::TemplateBody {
+        let pc = ast::PrincipalConstraint::new(ast::PrincipalOrResourceConstraint::any());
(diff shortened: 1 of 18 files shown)


cedar-policy/src/test/test.rs
```diff
@@ -11430,10 +11430,10 @@ mod pst_api {
             p.clone().to_pst().unwrap().to_string(),
             pst_template_with_slot().to_string()
         );
-        assert_eq!(
+        assert!(matches!(
```
john-h-kastner-aws, line 11433: you can use `assert_matches!` from `cool_asserts`, applies to some other tests too

victornicolet: It would look nicer; I see a lot of `assert!(matches!(...` in also unrelated tests, will do a quick follow up PR for just that refactor.

victornicolet: https://github.com/cedar-policy/cedar/pull/2268

PR #2251, opened 2026-03-24
document features and add warnings in cedar-policy-core crate doc

victornicolet:
## Description of changes

This PR adds documentation to `cedar-policy-core`, which would make the usage of the crate clearer for consumers that only read the crate's doc.

### DIFF
--- a/cedar-policy-core/src/lib.rs
+++ b/cedar-policy-core/src/lib.rs
@@ -53,7 +53,7 @@
 //!
 //! - `experimental` — Enables all experimental features listed below.
 //! - `variadic-is-in-range` — Variadic overload for the `isInRange` function.
-//! - `tpe` — Total policy evaluation / batched authorization. Enables the
+//! - `tpe` — Type-aware partial evaluation / batched authorization. Enables the
 //!   [`batched_evaluator`] and [`tpe`] modules.
 //! - `partial-eval` — Partial evaluation of Cedar policies. You should prefer `tpe` above.
 //! - `partial-validate` — Partial validation of Cedar policies.


cedar-policy-core/src/lib.rs
```diff
@@ -15,7 +15,58 @@
...
+//! in an authorization path. This includes, but is not limited to, parsing,
+//! serializing, and deserializing policies. Use them only for development,
+//! testing, or prototyping purposes.
+//!
+//! - `experimental` — Enables all experimental features listed below.
+//! - `partial-eval` — Partial evaluation of Cedar policies.
+//! - `partial-validate` — Partial validation of Cedar policies.
+//! - `tolerant-ast` — Error-tolerant parsing that produces a (possibly
+//!   incomplete) AST even when the input contains syntax errors. This should
+//!   only be used for providing helpful error handling in language servers.
```
john-h-kastner-aws, line 77: I would move this to a different section as "internal" or something

cedar-policy-core/src/lib.rs
```diff
@@ -15,7 +15,58 @@
...
+//! - `entity-manifest` — Entity manifest computation for entity slicing.
+//! - `extended-schema` — Extended schema features.
```
john-h-kastner-aws: This should have similar remarks to `tolerant-ast` (internal LSP only)

cedar-policy-core/src/lib.rs
```diff
@@ -15,7 +15,58 @@
...
+//! serializing, and deserializing policies. Use them only for development,
+//! testing, or prototyping purposes.
+//!
+//! - `experimental` — Enables all experimental features listed below.
+//! - `partial-eval` — Partial evaluation of Cedar policies.
+//! - `partial-validate` — Partial validation of Cedar policies.
+//! - `tolerant-ast` — Error-tolerant parsing that produces a (possibly
+//!   incomplete) AST even when the input contains syntax errors. This should
+//!   only be used for providing helpful error handling in language servers.
+//! - `entity-manifest` — Entity manifest computation for entity slicing.
```
john-h-kastner-aws: can mention it's deprecated (should use TPE)

cedar-policy-core/src/lib.rs
```diff
@@ -15,7 +15,58 @@
...
+//! ## Experimental features
+//!
+//! **WARNING:** Experimental features are unstable and subject to breaking
+//! changes in any release, including patch releases. They **must not** be used
+//! in an authorization path. This includes, but is not limited to, parsing,
+//! serializing, and deserializing policies. Use them only for development,
+//! testing, or prototyping purposes.
+//!
+//! - `experimental` — Enables all experimental features listed below.
+//! - `partial-eval` — Partial evaluation of Cedar policies.
```
john-h-kastner-aws: Maybe move below `tpe` and add a mention that TPE should be prefered

cedar-policy-core/src/lib.rs
victornicolet: In that case, I would argue for also moving it out of the set of experimental features in `Cargo.toml` and have a new set of `internal` features?

cedar-policy-core/src/lib.rs
```diff
@@ -15,7 +15,58 @@
...
+//!   for several types in this crate. Useful for fuzzing.
+//! - `test-util` — Exposes the [`test_utils`] module with helpers for testing.
+//! - `wasm` — Enables WebAssembly bindings via `wasm-bindgen` and `tsify`.
+//!
+//! ## Experimental features
+//!
+//! **WARNING:** Experimental features are unstable and subject to breaking
+//! changes in any release, including patch releases. They **must not** be used
+//! in an authorization path. This includes, but is not limited to, parsing,
+//! serializing, and deserializing policies. Use them only for development,
```
john-h-kastner-aws: This statement is too strong. `tpe` or `variadic-is-in-range` would be fine to use in authorization, just at-your-own-risk due to lack of proofs and fuzzing

cedar-policy-core/src/lib.rs
```diff
@@ -15,7 +15,58 @@
...
+//!   dependency.
+//!
+//! ## Optional features
+//!
+//! - `arbitrary` — Enables [`Arbitrary`](https://docs.rs/arbitrary) implementations
+//!   for several types in this crate. Useful for fuzzing.
+//! - `test-util` — Exposes the [`test_utils`] module with helpers for testing.
+//! - `wasm` — Enables WebAssembly bindings via `wasm-bindgen` and `tsify`.
+//!
+//! ## Experimental features
```
john-h-kastner-aws, line 48: Do we already have a similar discussion of features on the `cedar-policy` crate?

cedar-policy-core/src/lib.rs
john-h-kastner-aws: Yeah, it was probably only there because it made CI easier

cedar-policy-core/src/lib.rs
victornicolet: Not that I can see, there are callouts for experimental features and a link to https://github.com/cedar-policy/rfcs/blob/main/README.md#experimental-features for tpe.

cedar-policy-core/src/lib.rs
victornicolet: Someone who wants `tpe` or `variadic-is-in-range` will probably use it through `cedar-policy` where we can be less strong about the wording on the feature in that crate. The blanket statement in `cedar-policy-core` might be redundant with the wording earlier in the doc though, so happy to change.

cedar-policy-core/src/lib.rs
john-h-kastner-aws: Probably worth saying a bit more

cedar-policy-core/src/lib.rs
john-h-kastner-aws: I think with the upfront warning about using the crate at all, the `-core` README can omit most of this if we instead covered it in `cedar-policy`.

Probably still worth having a general comment on experimental features, and then the harsher warning against `toleran-ast`

cedar-policy-core/src/lib.rs
```diff
@@ -48,22 +48,35 @@
...
-//! in an authorization path. This includes, but is not limited to, parsing,
-//! serializing, and deserializing policies. Use them only for development,
-//! testing, or prototyping purposes.
+//! changes in any release, including patch releases. Use those features at your
+//! own risk.
 //!
 //! - `experimental` — Enables all experimental features listed below.
-//! - `partial-eval` — Partial evaluation of Cedar policies.
+//! - `variadic-is-in-range` — Variadic overload for the `isInRange` function.
+//! - `tpe` — Total policy evaluation / batched authorization. Enables the
```
john-h-kastner-aws: Type-aware partial evaluation

cedar-policy-core/src/lib.rs
```diff
@@ -48,22 +48,35 @@
...
+//!   [`batched_evaluator`] and [`tpe`] modules.
+//! - `partial-eval` — Partial evaluation of Cedar policies. You should prefer `tpe` above.
 //! - `partial-validate` — Partial validation of Cedar policies.
+//! - (deprecated) `entity-manifest` — Entity manifest computation for entity slicing.
+//!   This feature is deprecated; you should use `tpe` instead.
+//!
+//! ## Unstable tooling features
```
john-h-kastner-aws, line 63: "Internal tooling features" maybe? "unstable" and "experimental" are semantically too close together for the stark distinction I want to create between these sets of options

victornicolet: My intent was to highlight features that will always be unstable and they are for tooling - a consumer should never expect this to stabilize because it's tied to the low-level representation. 
An experimental feature on the other hand signals that we're experimenting with something that may become a feature later.

I was trying to avoid internal because I don't think it says much for someone who doesn't necessarily see the set of cedar-policy crates as a whole.

cedar-policy-core/src/lib.rs
victornicolet: The new warning is more general with "use at your own risk" and the harsher warning is on `unstable-tooling` features (which might be a little over-general since it includes the `extended-schema`).

cedar-policy-core/src/lib.rs
victornicolet: Yes, I can start a separate PR on the cedar-policy doc update (if we have more discussions on the main doc update, we can still merge this one).

cedar-policy-core/src/lib.rs
john-h-kastner-aws: Make sense. I want to send the message that nobody should use these, but the docs do cover that. We should go verbose with something like `unstable-lsp-implementation-details` but that's kind of excessive

cedar-policy-core/src/lib.rs
victornicolet: Following up with https://github.com/cedar-policy/cedar/pull/2253

PR #2252, opened 2026-03-24
fix(smt): support Windows CRLF in SMT-LIB output parsing

jiangshibiao:
## Description of changes

Windows-based SMT solvers (e.g., cvc5) typically output lines ending with `\r\n` (CRLF). The current implementation used strict string matching for solver responses like `sat\n`, `unsat\n`, `unknown\n`.

This PR introduces `.trim()` when matching the solver's stdout lines to ensure that trailing carriage returns do not cause parsing failures. Specifically:

+ Updated `check_sat()` to handle trimmed status strings.
+ Updated `check_sat_with_model()` to handle trimmed delimiters.

## Issue #, if available

N/A

- [x] [UPD] For the consistency, I apply the same logic on [cedar-spec#914](https://github.com/cedar-policy/cedar-spec/pull/914)

### DIFF
--- a/cedar-policy-symcc/src/symcc/solver.rs
+++ b/cedar-policy-symcc/src/symcc/solver.rs
@@ -480,4 +480,40 @@ mod test {
         let status = my_solver.child.try_wait().unwrap();
         assert!(status.is_some());
     }
+
+    #[tokio::test]
+    async fn check_sat_crlf_test() {
+        let mut cmd = Command::new("sh");
+        cmd.args(["-c", "read line && printf 'sat\r\n'"]);
+        let mut solver = LocalSolver::from_command(&mut cmd).unwrap();
+        let decision = solver.check_sat().await.unwrap();
+        assert_eq!(decision, Decision::Sat);
+    }
+
+    #[tokio::test]
+    async fn check_sat_with_model_crlf_test() {
+        let mut cmd = Command::new("sh");
+        cmd.args([
+            "-c",
+            "read line && printf 'sat\\r\\n' && read line && printf '(\\r\\n  define-fun x () Int 0\\r\\n)\\r\\n'",
+        ]);
+        let mut solver = LocalSolver::from_command(&mut cmd).unwrap();
+
+        let decision = solver.check_sat_with_model().await.unwrap();
+
+        assert_matches!(decision, DecisionWithModel::Sat { model } => {
+            assert_eq!(model, "(\r\n  define-fun x () Int 0\r\n)\r\n");
+        });
+    }
+
+    #[tokio::test]
+    async fn process_error_output_parse_error_no_line_ending_test() {
+        let mut solver = LocalSolver::cvc5().unwrap();
+        let res = solver
+            .process_error_output("(error \"Parse Error: mock windows error\")")
+            .await;
+        assert_matches!(res, SolverError::Solver(msg) => {
+            assert_eq!(msg, "Parse Error: mock windows error");
+        });
+    }
 }


victornicolet:
I think there is at least one more place where an update is required (the `process_error_output` method of `LocalSolver`).

john-h-kastner-aws (commented):
Changes looks reasonable (modulo failing unit tests).

We also do much more complete parsing of the sovler output in `decoder.rs`. Are there similar changes we should make there? If not, we can at least add some unit tests showing we correctly decode s-expressions containing `\r`. 

These changes should also be made in the Lean model. We don't do any proofs on this code, and I don't believe any of our differential testing would expose this discrepancies, but the code should be the same regardless.

jiangshibiao:
Hi @victornicolet  and @john-h-kastner-aws , thanks for the review!

@victornicolet , good catch! I've now applied the `.trim()` logic to the `process_error_output()` method in LocalSolver. I previously focused on the successful scenario (sat/unsat) during my local Windows testing and overlooked the error output scenarios, so that the CI tests were failing. Now CI tests are all passed.

@john-h-kastner-aws : I believe further changes aren't strictly necessary here in `decoder.rs`, because the S-expression parsing already operates on a line-by-line basis.  I considered adding a cross-platform test in `solver.rs`, but simulating Windows-specific SMT output in a Unix-based CI environment would require a fairly complex mock of the LocalSolver process. As for the Lean Model, I agree that the Lean model should ideally match the Rust implementation, but this CRLF handling seems primarily an engineering implementation detail related to OS-specific I/O, rather than a logic flaw that affects the formal correctness or the SMT-LIB encoding itself? Apologize for not being familiar enough with the Lean4 proof details.

I’ll continue to monitor the behavior on Windows as I use this for my local development. If any subtle edge cases arise, I will definitely follow up with a refined PR.

john-h-kastner-aws:
>  I considered adding a cross-platform test in solver.rs, but simulating Windows-specific SMT output in a Unix-based CI environment would require a fairly complex mock of the LocalSolver process. 

I'm suggesting some very simple unit tests testing the parsing function given strings containing `\r\n`. I'm not worried about testing against the exact behavior on windows, just something that will catch a regression if we remove the `trim` for some reason in the future. 

> As for the Lean Model, I agree that the Lean model should ideally match the Rust implementation, but this CRLF handling seems primarily an engineering implementation detail...

The Lean model is also executable, so the same implementation details will be relevant is some ever tries to run it on windows. Like I said, there isn't any proof on this part of the code, so the update should be trivial.

E.g., this function https://github.com/cedar-policy/cedar-spec/blob/ddcec57af0acee3d4b72f8b35dfa6358021f9605/cedar-lean/Cedar/SymCC/Solver.lean#L173

jiangshibiao:
Hi @john-h-kastner-aws, 

Thanks for the suggestions and quick reactions(it takes some times to align windows/unix environments and pass the online CI tests)! I have updated this PR and opened a new one for the [cedar-spec#914](https://github.com/cedar-policy/cedar-spec/pull/914).

Here are the updates:
1. **Added unit tests in Rust**: I've added simple unit tests for `check_sat`, `get_model`, and `process_error_output`. To test the `\r\n` behavior without modifying the core logic or introducing complex abstractions, I used a lightweight mock via `Command::new("sh")` with `printf` to accurately simulate the solver's CRLF output. 
2. **Updated the Lean model**: I have applied the equivalent fix (`trimAsciiEnd`) to `Solver.lean` and added the corresponding CRLF tests in the `cedar-spec` repository.

PR #2253, opened 2026-03-24
add features doc to cedar-policy

victornicolet:
## Description of changes

Adds documentation about features in `cedar-policy`.

### DIFF
--- a/cedar-policy/src/lib.rs
+++ b/cedar-policy/src/lib.rs
@@ -20,6 +20,50 @@
 // to add a separate test specifically for README examples by introducing a
 // private, empty, and unused function with `#[doc = include_str!("../README.md")]`.
 #![doc = include_str!("../README.md")]
+//!
+//! # Feature flags
+//!
+//! ## Default features
+//!
+//! The following features are enabled by default and provide the built-in Cedar
+//! extension functions:
+//!
+//! - `ipaddr` — IP address extension functions (`ip`, `isIpv4`, `isIpv6`,
+//!   `isLoopback`, `isMulticast`, `isInRange`).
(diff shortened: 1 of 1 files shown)


cedar-policy/src/lib.rs
```diff
@@ -20,6 +20,50 @@
...
+//! - `heap-profiling` — Enables heap profiling via `dhat`.
+//! - `corpus-timing` — Enables corpus timing instrumentation.
+//! - `wasm` — Enables WebAssembly bindings via `wasm-bindgen` and `tsify`.
+//!
+//! ## Experimental features
+//!
+//! **WARNING:** Experimental features are unstable and subject to breaking
+//! changes in any release, including patch releases. Use those features at your
+//! own risk.
+//!
```
victornicolet, line 50: Unlike for `cedar-policy-core`, I think changing feature sets would be a backwards incompatible change, so all features are listed as `experimental` like in the `Cargo.toml`

victornicolet: We can debate whether we should be more precise an apply different warnings to each of the features.

john-h-kastner-aws: > I think changing feature sets would be a backwards incompatible change 

Not sure I understand. Are you saying it could incompatible for non-experimental users?

john-h-kastner-aws (approved):
Looks good for now. Definitely think we should apply the same split of features though.

cedar-policy/src/lib.rs
victornicolet: I guess not, only for users that have the "experimental" feature set on, and amongst those users, the ones who would use the features we'd remove from "experimental".

katherine-hough (approved):

PR #2256, opened 2026-03-25
Unified publish workflow

victornicolet:
## Description of changes

Merges workflow for publishing core crates and symcc into one workflow with different targets:
- `core` for all core crates (i.e. not symcc and cli)
- `symcc` 
- `cli` for only the cli.
- `all` for publishing all crates at once. 
The workflow requires a tag to decide where to publish from and at least the version of the core crates to check what is being published, or validate what core version symcc is being published on.

It requires explicit versions for what is being published depending on the target, on top of the tag that specifies what code to checkout for publishing.

- [x] A change "invisible" to users (e.g., documentation, changes to "internal" crates like `cedar-policy-core`, `cedar`


.github/workflows/publish.yml
```diff
@@ -2,87 +2,156 @@
...
+                    - symcc
+                    - cli
+                    - all
             tag:
                 required: true
                 type: string
-                description: "The GitHub tag to be released. Must be of the form 'v<MAJOR>.<MINOR>.<PATCH>'"
-
+                description: "The core SDK tag. Must be of the form 'v<MAJOR>.<MINOR>.<PATCH>'. Required for core, cli, and all."
+            symcc_tag:
```
katherine-hough: I think it makes more sense to have one tag and all of the crates to be published come from checking out that tag. If publishing core or all the tag should be of the form `v<MAJOR>.<MINOR>.<PATCH>`. If publishing just one crate (symcc or the cli) the tag should be `<CRATE>-v<MAJOR>.<MINOR>.<PATCH>`.

Then, if the target is "all" you need to specify the version of symcc and cli since they can't be extracted from the tag.

.github/workflows/publish.yml
```diff
@@ -2,87 +2,156 @@
...
+                required: true
+                type: choice
+                description: "Which crate group to publish."
+                options:
+                    - core
+                    - symcc
+                    - cli
+                    - all
             tag:
                 required: true
```
katherine-hough, line 20: If this isn't need when the target is "symcc" it should not be marked as required.

.github/workflows/publish.yml
```diff
@@ -2,87 +2,156 @@
...
-                        -p cedar-policy \
-                        -p cedar-policy-cli
-                  else 
-                    cargo publish \
-                        -p cedar-policy-core \
-                        -p cedar-policy-formatter \
-                        -p cedar-policy \
-                        -p cedar-policy-cli
-                  fi
+              # Order matters: dependencies must be listed before dependents.
```
katherine-hough, line 190: I don't think the order actually matters if you specify packages in a single cargo publish command.

.github/workflows/publish.yml
```diff
@@ -2,87 +2,156 @@
...
+                SYMCC_TAG: ${{ inputs.symcc_tag }}
+    publish-core:
+        if: inputs.target == 'core'
+        needs: [validate-core]
         runs-on: ubuntu-latest
-        environment: release # See https://github.com/cedar-policy/cedar/settings/environments
+        environment: release
         permissions:
-            id-token: write # Required for OIDC token exchange
+            id-token: write
```
katherine-hough: We should keep the original comment.

.github/workflows/publish.yml
```diff
@@ -2,87 +2,156 @@
...
-    publish-crates:
-        # Only publish if validation succeeds.
-        needs: [validate]
+                SYMCC_TAG: ${{ inputs.symcc_tag }}
+    publish-core:
+        if: inputs.target == 'core'
+        needs: [validate-core]
         runs-on: ubuntu-latest
-        environment: release # See https://github.com/cedar-policy/cedar/settings/environments
+        environment: release
```
katherine-hough: We should keep the original comment.

.github/workflows/publish.yml
```diff
@@ -2,87 +2,156 @@
...
         inputs:
+            target:
+                required: true
+                type: choice
+                description: "Which crate group to publish."
+                options:
+                    - core
+                    - symcc
+                    - cli
+                    - all
```
katherine-hough, line 18: It may make sense to have "all" be a separate workflow for all and put the shared logic in a composite GitHub action.

.github/workflows/publish.yml
victornicolet: The current workflow will validate the core tag, even if that validation is not doing much. We could remove this and only validate the symcc tag for a symcc release. 

(edit: I am wrong about this, please ignore)

.github/workflows/publish.yml
victornicolet: I agree the current version is a little cumbersome because it forces you to always give the tag for core and sometimes the tag to symcc (if you're publishing symcc or all of them). 

In the current workflow, all jobs will do at least a `validate-core` step. We could make the validation for the core tag more meaningful in symcc publishing by making `symcc` validate that the provided core tag is the version of the core on the provided symcc tag, according to the Cargo.toml.

I think having less implicit things (i.e. having to specify explicit tags, and two of them in some cases where you might only need one) is not necessarily a bad thing. But happy to change and make it easier to run

.github/workflows/publish.yml
victornicolet: Would that mean each main `publish` workflow is a composite Github action triggered by `workflow_dispatch`? Otherwise I don't know how well composite actions and `workflow_dispatch` mix.

katherine-hough: AFAIK composite actions can be called in the same way any action can. https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action
The advantage of this over a reusable workflow is that it will run in the same job using the same runner as the caller. So you can have it do things like configure the environment.

.github/workflows/publish.yml
victornicolet: Resolving, latest proposal uses the tag to only mean tag not version.

.github/workflows/publish.yml
victornicolet: I cannot find clear documentation that says that it does or it doesn't, and I can find some suggestions that it does. The previous workflow had the crates specified in the dependency order, so unless I'm proven wrong I'll leave it here.

.github/workflows/publish.yml
victornicolet: Pushed another "proposal" that still has all targets, each target runs 2 jobs, no reusable workflow. We could use composite action to remove the copy-pasted parts between each publish job?

john-h-kastner-aws (approved):
Looks good. I'm hoping to use this (with `targets: all`) for the upcoming release of 4.10.0

katherine-hough (approved):

.github/workflows/publish.yml
```diff
@@ -25,19 +45,81 @@ jobs:
               uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
               with:
                   script: core.setFailed('This workflow must be triggered from the "main" branch.')
-
-            # Check that the input tag is welformed.
+            # Check that the input tag is well-formed.
+            # For core, cli and all, we expect a tag "v<MAJOR>.<MINOR>.<PATCH>"
```
katherine-hough, line 49: For the cli the tag should be of the form `cedar-policy-cli-v<MAJOR>.<MINOR>.<PATCH>` to distinguish between the release of the cli for that version and the release of the base cedar crates for that version

.github/workflows/publish.yml
```diff
@@ -69,20 +187,91 @@ jobs:
...
+            - name: Configure rust toolchain
+              run: rustup update stable && rustup default stable
+            - name: Authenticate with crates.io
+              id: auth
+              uses: rust-lang/crates-io-auth-action@bbd81622f20ce9e2dd9622e3218b975523e45bbe # v1.0.4
+            - name: Publish to crates.io
+              run: cargo publish -p cedar-policy-cli
+              env:
+                  CARGO_REGISTRY_TOKEN: ${{ steps.auth.outputs.token }}
+    publish-all:
```
katherine-hough, line 251: Will the procedure for patching old cedar versions be that we run publish core and then optionally publish cli?

victornicolet: Yes, I think so?

.github/workflows/publish.yml
victornicolet: Yes,I think that once the patch versions of CLI and Core start diverging, we'll need to make that change.

katherine-hough (approved):

PR #2269, opened 2026-03-31
eliminate error cases for pst->ast Expr conversion with new pst::Id

victornicolet:
## Description of changes

This PR mainly makes the conversion of expressions from PST to AST infallible. This is the conversion path were users would have been most likely to introduce errors, and also conversion errors to an internal type would be bad user experience.
To do so, in this PR, we:
- inline the operator to function conversion to avoid having code for an unreachable case,
- unwrap the record conversion, and give the reason why we can (from a map to a map, no duplicate keys)
- replace `SmolStr` type with a new validated `pst::Id` type in `pst::Name`. This is the only non-transparent type of the pst that enforces validation on creation with `pst::Id::new(..)`. This means `Id`s are validated at construction rather than conversion. I think this is a better user experience. The doc has been updated to reflect this.

## Issue #, if available

Related to https://github.com/cedar-policy/cedar/issues/2237

### DIFF
--- a/cedar-policy-core/src/pst/expr.rs
+++ b/cedar-policy-core/src/pst/expr.rs
@@ -155,7 +155,11 @@ pub struct Name {
 }
 
 impl Name {
-    /// Constructs an unqualified name. Validates that `id` is a legal Cedar identifier.
+    /// Constructs an unqualified name. This is a convenience constructor that validates
+    /// that `id` is a legal Cedar identifier.
+    ///
+    /// If you have an `Id` (which is `AsRef<str>`), you can infallibly construct the name
+    /// yourself.
     pub fn unqualified(id: impl AsRef<str>) -> Result<Self, PstConstructionError> {
         Ok(Name {
             id: Id::new(id)?,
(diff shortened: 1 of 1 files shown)


cedar-policy-core/src/pst/expr.rs
```diff
@@ -88,30 +149,31 @@ impl Display for SlotId {
...
+    pub namespace: Arc<Vec<Id>>,
 }
 
 impl Name {
-    /// Constructs an unqualified name.
-    pub fn unqualified(id: impl Into<SmolStr>) -> Self {
-        Name {
-            id: id.into(),
+    /// Constructs an unqualified name. Validates that `id` is a legal Cedar identifier.
+    pub fn unqualified(id: impl AsRef<str>) -> Result<Self, PstConstructionError> {
```
john-h-kastner-aws, line 163: Maybe add a note in the docs that if you have an `Id` (which is `AsRef<str>`) you can infallibly construct the name yourself

john-h-kastner-aws (approved):
looks great

PR #2272, opened 2026-04-01
add residual errors in pst for tpe support

victornicolet:
## Description of changes

This adds a new `ResidualError` variant in `pst::Expr` to enable support for `pst` in TPE.

Given a `residual : Policy`, `residual.to_pst()`  gives you the PST representation of the residual, including the error nodes.

### DIFF
--- a/cedar-policy-core/src/pst/expr.rs
+++ b/cedar-policy-core/src/pst/expr.rs
@@ -743,7 +743,7 @@ pub enum Expr {
     ///
     /// This is distinct from tolerant-ast parse errors — `ResidualError`
     /// represents a semantically meaningful result from type-aware partial
-    /// evaluation (e.g., accessing a missing attribute on a concrete entity).
+    /// evaluation (e.g., arithmetic overflow).
     #[cfg(feature = "tpe")]
     ResidualError,
 }


cedar-policy-core/src/pst/expr.rs
```diff
@@ -738,6 +738,14 @@ pub enum Expr {
         /// Name of the unknown
         name: SmolStr,
     },
+    /// A TPE residual error node: indicates that this subexpression would
+    /// produce an evaluation error if reached at runtime.
+    ///
+    /// This is distinct from tolerant-ast parse errors — `ResidualError`
+    /// represents a semantically meaningful result from type-aware partial
+    /// evaluation (e.g., accessing a missing attribute on a concrete entity).
```
john-h-kastner-aws: a better example would be arithmetic overflow, since missing attribute errors shouldn't happen for well-typed polices

cedar-policy-core/src/tpe/residual.rs
```diff
@@ -522,10 +528,9 @@ impl From<Residual> for Expr {
             Residual::Concrete { value, .. } => value.into(),
             Residual::Error(_) => {
```
john-h-kastner-aws, line 529: This whole residual to expression conversion could use the expression builder trait if we want.

victornicolet: One difference is the residual error node which might not match the builder's `error()`, which usually used for the tolerant ast.

john-h-kastner-aws: right, I guess the builder would need an `residual_error` function on the trait. Possible future refactoring

PR #2273, opened 2026-04-01
Fixed issue with SymCC silently ignoring template-linked polices

katherine-hough:
## Description of changes
Updated `well_type_policies` to return an `Err` if the input policies contain a template-linked policy instead of silently ignoring them ([slots are not supported by the symbolic compiler](https://github.com/cedar-policy/cedar/blob/036757e1313045670af8516cba4ff97ca0798f93/cedar-policy-symcc/src/symccopt/compiler.rs#L1050)).

I will make updates to the symcc CLI subcommand in another PR. I wanted this one clean so that it's easier to backport.

## Issue #, if available
N/A

### DIFF
--- a/cedar-policy-symcc/src/err.rs
+++ b/cedar-policy-symcc/src/err.rs
@@ -65,9 +65,6 @@ pub enum Error {
     /// Empty-list of policies was passed to a function that expects at least one policy
     #[error("expected to have at least one policy")]
     NoPolicies,
-    /// Unsupported features.
-    #[error("unsupported feature in SymCC: {0}")]
-    UnsupportedFeature(String),
 }
 
 /// A result type that potentially returns a SymCC [`enum@Error`].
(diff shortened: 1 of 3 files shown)


cedar-policy-symcc/src/err.rs
```diff
@@ -65,6 +65,9 @@ pub enum Error {
     /// Empty-list of policies was passed to a function that expects at least one policy
     #[error("expected to have at least one policy")]
     NoPolicies,
+    /// Unsupported features.
+    #[error("unsupported feature in SymCC: {0}")]
+    UnsupportedFeature(String),
```
victornicolet: nit naming: I don't know if we'd want this reserved for "features" like tpe, tolerant-ast, and this just be "TemplateNotSupported" or similar.

katherine-hough: I went with this to be consistent with language used by the error the symbolic compiler gives if you try to compile a slot: https://github.com/cedar-policy/cedar/blob/main/cedar-policy-symcc/src/symccopt/compiler.rs#L1050.

john-h-kastner-aws: Could also reuse `CompileError::UnsupportedFeature` if making this a "compile error" is fine

katherine-hough: That sounds reasonable to me.`well_typed_policies` is public so technically it could be called outside symbolic compilation, but since it's intended as a pre-compilation step what would be the point of that.

PR #2274, opened 2026-04-01
Removed the "template_linked_file" options for symcc CLI commands

katherine-hough:
## Description of changes
* Removed the "template_linked_file" options for symcc CLI commands.
* Updated symcc cli commands to warn users if an input policy set contains templates.

## Issue #, if available
N/A

### DIFF
--- a/cedar-policy-cli/tests/symcc.rs
+++ b/cedar-policy-cli/tests/symcc.rs
@@ -975,6 +975,101 @@ fn test_always_denies_does_not_hold_no_counterexample() {
 }
 
 #[test]
+fn test_warn_if_contains_templates_single_policy_set() {
+    let schema = write_temp(SAMPLE_SCHEMA);
+    // A policy set with one template (uses ?principal slot) and one static policy
+    let policy = write_temp(
+        r#"
+        permit(principal, action, resource);
+        forbid(principal == ?principal, action, resource);
+        "#,
+    );
+
+    cargo::cargo_bin_cmd!("cedar")
+        .arg("symcc")
+        .arg("--principal-type")
+        .arg("Identity")
+        .arg("--action")
+        .arg(r#"Action::"view""#)
+        .arg("--resource-type")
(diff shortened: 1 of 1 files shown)


cedar-policy-cli/src/lib.rs
```diff
@@ -890,11 +890,11 @@ pub struct SymccArgs {
...
     // --- Single-policy primitives ---
     /// Verify that a policy never produces runtime errors
-    NeverErrors(PoliciesArgs),
+    NeverErrors(SymccPoliciesArgs),
     /// Verify that a policy always matches (is always true)
-    AlwaysMatches(PoliciesArgs),
+    AlwaysMatches(SymccPoliciesArgs),
     /// Verify that a policy never matches (is always false)
-    NeverMatches(PoliciesArgs),
+    NeverMatches(SymccPoliciesArgs),
```
victornicolet, line 897: Do the other primitives that use `PoliciesArgs` support template-linked policies? `AlwaysAllows` and `AlwaysDenies`

katherine-hough: The [symbolic compiler itself doesn't support template slots](https://github.com/cedar-policy/cedar/blob/036757e1313045670af8516cba4ff97ca0798f93/cedar-policy-symcc/src/symccopt/compiler.rs#L1050). So, none of the primitives will work out of the box with template-linked policies.

victornicolet: I think it's probably safer the way you corrected it, I was wondering whether any of the primitives was doing the linking before compiling (hence compiling only static policies).

katherine-hough: I tried looking into binding all the template slots before running the symbolic compiler, but I didn't really see a good way of doing it. I figured if someone asks for analysis over linked-templates in the future we can revisit this, but for now it's better to just leave it as unsupported.

PR #2291, opened 2026-04-05
feat: Add help text for function argument validation errors

### LINKED ISSUE #834: Add help text for  `FunctionArgumentValidationError`
### Category

Error message improvements

This errors is reported when the validator finds an error in the arguments to an extension function. It could include help text for example to describing the format of a valid decimal or ipaddress. 

.

### Additional context

_No response_

Vedthakar:
## Description of changes

Closes #834

This PR adds help text to `FunctionArgumentValidationError` so validator errors for invalid extension function arguments include human-readable format guidance when available.

Previously, these errors only surfaced a generic parse failure message. With this change, invalid arguments to extension functions such as `decimal`, `ip`, `datetime`, and `duration` can now also include help text describing the expected input format.

### What changed
- Updated extension argument validation to return structured validation errors instead of only raw strings
- Added optional help text propagation into `FunctionArgumentValidationError`
- Surfaced upstream parser help text when available, with fallback help text for cases where none is provided
- Added coverage for:
  - `ip`
  - `decimal`
  - `datetime`
  - `duration`
  - public API behavior
  - FFI behavior

This directly addresses the request in #834 to make these validator errors more actionable by explaining what valid extension function inputs should look like.

## Issue #, if available

Closes #834

The change in this PR is:

I confirm that this PR:

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec):

I confirm that [`docs.cedarpolicy.com`](https://docs.cedarpolicy.com/):

### DIFF
--- a/cedar-policy-core/src/validator/diagnostics/validation_errors.rs
+++ b/cedar-policy-core/src/validator/diagnostics/validation_errors.rs
@@ -465,10 +465,16 @@ pub struct FunctionArgumentValidation {
     pub policy_id: PolicyID,
     /// Error message
     pub msg: String,
+    /// Optional help for resolving the error
+    pub help: Option<String>,
 }
 
 impl Diagnostic for FunctionArgumentValidation {
     impl_diagnostic_from_source_loc_opt_field!(source_loc);
+
+    fn help<'a>(&'a self) -> Option<Box<dyn Display + 'a>> {
+        self.help.as_ref().map(|h| Box::new(h) as Box<dyn Display>)
+    }
 }
 
 /// Structure containing details about a hierarchy not respected error
--- a/cedar-policy-core/src/validator/extensions/datetime.rs
+++ b/cedar-policy-core/src/validator/extensions/datetime.rs
@@ -20,9 +20,12 @@
 
 use crate::ast::{Expr, ExprKind, Literal, Name};
 use crate::extensions::datetime;
-use crate::validator::extension_schema::{ArgumentCheckFn, ExtensionFunctionType, ExtensionSchema};
+use crate::validator::extension_schema::{
+    ArgumentCheckFn, ArgumentValidationError, ExtensionFunctionType, ExtensionSchema,
+};
 use crate::validator::types::{self, Type};
 use itertools::Itertools;
+use miette::Diagnostic;
 
 use super::eval_extension_constructor;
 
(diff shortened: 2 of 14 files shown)


cedar-policy-core/src/validator/extensions/datetime.rs
```diff
@@ -120,12 +126,23 @@ pub fn extension_schema() -> ExtensionSchema {
...
-                .map(|_| ())
-                .map_err(|_| format!("Failed to parse as a datetime value: `{lit_arg}`"))
+            match eval_extension_constructor(datetime_constructor_name, s.clone()) {
+                Ok(_) => Ok(()),
+                Err(err) => Err(ArgumentValidationError::new(
+                    format!("failed to parse as a datetime value: `{lit_arg}`"),
+                    Some(
+                        err.help()
+                            .map(|h| h.to_string())
+                            .unwrap_or_else(|| VALID_DATETIME_HELP.to_string()),
```
john-h-kastner-aws: It might make sense to return these help messages from the main extension parsing functions. This also might let us give the nicer errors for entity parsing failures as well.

We don't want to introduce any unexpected overhead in evaluation, but this should only be only be on the error path.

cedar-policy/src/test/test.rs
```diff
@@ -5980,6 +5980,60 @@ permit(principal, action, resource) when { {"\n": 0} };"#,
...
+mod function_argument_validation_help_tests {
+    use crate::{Policy, PolicySet, ValidationError, ValidationMode, Validator};
+    use cool_asserts::assert_matches;
+    use miette::Diagnostic;
+    use serde_json::json;
+
+    use super::Schema;
+
+    #[test]
+    fn validator_surfaces_help_for_invalid_extension_constructor_arguments() {
```
john-h-kastner-aws, line 5993: Might be able to use [test_utils](https://github.com/cedar-policy/cedar/blob/main/cedar-policy-core/src/test_utils.rs) to tidy up this test case

john-h-kastner-aws (commented):
help message text looks reasonable. might consider pushing it up into the main extension parsing code

cedar-policy-core/src/validator/extensions/datetime.rs
Vedthakar: Good suggestion — I updated this to push the help text generation into the main extension parsing path instead of handling it in the validator layer.

The four format help messages now live with the corresponding extension parsers (`decimal`, `ip`, `datetime`, and `duration`). On parse failure, those parsers now attach advice directly to the `EvaluationError`, using the specific parse error help text when available and otherwise falling back to the general format guidance.

That simplifies the validator-side code to just forwarding `err.help()`, and it also means non-validator paths, including entity parsing failures, now benefit from the same improved help text.

As part of that cleanup, I also updated the two existing `datetime.rs` tests that were still asserting `None` for advice despite the help annotations. They now assert the expected `Some(...)` value instead.

cedar-policy/src/test/test.rs
Vedthakar: Nice catch — I refactored the test to use `cedar_policy_core::test_utils` so it is a bit cleaner and closer to the surrounding test style.

It now uses `expect_err` together with `ExpectedErrorMessageBuilder`, which lets the test check the rendered error message, help text, and source underline in one place instead of chaining several lower-level assertions.

cedar-policy-core/src/validator/diagnostics/validation_errors.rs
```diff
@@ -465,10 +465,16 @@ pub struct FunctionArgumentValidation {
     pub policy_id: PolicyID,
     /// Error message
     pub msg: String,
+    /// Optional help for resolving the error
+    pub help: Option<String>,
```
john-h-kastner-aws, line 469: Since we're forwarding this from the evaluation errors, I think `help` could be an `Option<Box<dyn Display>>` and then `help` just returns `self.help().clone()`. Then we just pass through `.help()` from the eval error without needing any intermediate processing.

john-h-kastner-aws (approved):
Looks good. Just one more possible tweak (if it works) and then you'll need to run `cargo fmt`

cedar-policy-core/src/validator/diagnostics/validation_errors.rs
victornicolet: I think this might be a lot of work if the `Debug,Clone,Hash,Eq` traits of the struct were kept, but I am not sure they are needed.

john-h-kastner-aws: ah yeah, that makes sense. The typechecker does hash to deduplicate errors beteween request environments, so `Option<String>` is probably the way to do it

victornicolet (approved):

john-h-kastner-aws:
Hi! I think this PR is good-to-go after once you push a quick `cargo fmt`

Vedthakar:
Hi, I fixed it thank you for the review and sorry for the delay!

victornicolet (approved):

PR #2316, opened 2026-04-15
pst api adjustments, additional doc

victornicolet:
## Description of changes

This makes a few adjustments in the PST API:
- `try_into_pst` now suceeds even if there is no lossless representation. This allows using `try_into_pst` for policies that are residuals from TPE (prior implementation would return an error). Eliminates an error type.
- adds documentation and examples.
- adds `PartialEq` and `Eq` to the PST types that were missing it,
- re-exports some types from dependencies in the `pst` module, so that crates depending on Cedar don't have to add dependencies, or manage dependency version to match Cedar's.

The last change is probably the one most worth discussing, but I think it makes the API cleaner.


john-h-kastner-aws (approved):

john-h-kastner-aws:
I didn't make any changes to the existing TPE code, so you can just overwrite `tpe.rs` with your updated version of the module

katherine-hough (approved):

PR #2318, opened 2026-04-15
Support template links in TPE

### LINKED ISSUE #2314: Support for template-linked policies in TPE
### Category

Other

 .

.

### Additional context

_No response_

john-h-kastner-aws:
## Description of changes

Resolves #2314 by adding support for type-aware partial evaluation with template links.

I decided to implement this by substituting the link values into the expression (rather than passing along the slot environment) because TPE already needs to translate the `Expr` into a `Residual`, so it is simpler to handle the bindings there and then not need to add a slot variant to residuals.

This changes the public API for TPE by removing the `SlotNotSupported` error variant and adding `UnlinkedSlotError`.

### DIFF
--- a/cedar-policy-core/src/tpe.rs
+++ b/cedar-policy-core/src/tpe.rs
@@ -40,7 +40,7 @@ pub(crate) fn policy_residual_map<'a>(
     ps: &'a PolicySet,
     schema: &ValidatorSchema,
 ) -> std::result::Result<HashMap<&'a PolicyID, Residual>, TpeError> {
-    let mut exprs = HashMap::new();
+    let mut residuals = HashMap::new();
     let tc = Typechecker::new(schema, crate::validator::ValidationMode::Strict);
     let env = request.find_request_env(schema)?;
     for p in ps.policies() {
@@ -56,21 +56,21 @@ pub(crate) fn policy_residual_map<'a>(
         let env = env.clone().link_slot_env(p.env());
         match tc.typecheck_by_single_request_env(t, &env) {
             PolicyCheck::Success(expr) => {
-                exprs.insert(p.id(), Residual::try_from_typed_expr(&expr, p.env())?);
+                residuals.insert(p.id(), Residual::try_from_typed_expr(&expr, p.env())?);
             }
             PolicyCheck::Fail(errs) => {
                 return Err(TpeError::Validation(errs));
             }
             PolicyCheck::Irrelevant(errs, expr) => {
                 if errs.is_empty() {
-                    exprs.insert(p.id(), Residual::try_from_typed_expr(&expr, p.env())?);
+                    residuals.insert(p.id(), Residual::try_from_typed_expr(&expr, p.env())?);
                 } else {
                     return Err(TpeError::Validation(errs));
                 }
             }
         }
     }
-    Ok(exprs)
+    Ok(residuals)
 }
 
 /// Type-aware partial-evaluation on a `PolicySet`.
--- a/cedar-policy-core/src/tpe.rs
+++ b/cedar-policy-core/src/tpe.rs
@@ -83,13 +83,12 @@ pub fn is_authorized<'a>(
     entities: &'a PartialEntities,
     schema: &'a ValidatorSchema,
 ) -> std::result::Result<Response<'a>, TpeError> {
-    let exprs = policy_residual_map(request, ps, schema)?;
     let evaluator = Evaluator {
         request,
         entities,
         extensions: Extensions::all_available(),
     };
-    let residuals: Vec<_> = exprs
+    let residuals = policy_residual_map(request, ps, schema)?
         .into_iter()
         .map(|(id, residual)| {
             let residual = evaluator.interpret(&residual);
@@ -98,15 +97,9 @@ pub fn is_authorized<'a>(
                 reason = "exprs and policy set contain the same policy ids"
             )]
             ResidualPolicy::new(Arc::new(residual), Arc::new(ps.get(id).unwrap().clone()))
-        })
-        .collect();
+        });
 
-    Ok(Response::new(
-        residuals.into_iter(),
-        request,
-        entities,
-        schema,
-    ))
+    Ok(Response::new(residuals, request, entities, schema))
 }
 
 #[cfg(test)]
--- a/cedar-policy-core/src/tpe.rs
+++ b/cedar-policy-core/src/tpe.rs
@@ -25,29 +25,25 @@ pub mod response;
 
 use std::{collections::HashMap, sync::Arc};
 
-use crate::ast::{Expr, PolicyID};
-use crate::tpe::err::{NonstaticPolicyError, TpeError};
+use crate::ast::PolicyID;
+use crate::tpe::err::TpeError;
+use crate::tpe::residual::Residual;
 use crate::tpe::response::{ResidualPolicy, Response};
-use crate::validator::types::Type;
 use crate::validator::Validator;
 use crate::validator::{typecheck::PolicyCheck, typecheck::Typechecker, ValidatorSchema};
 use crate::{ast::PolicySet, extensions::Extensions};
 
 use crate::tpe::{entities::PartialEntities, evaluator::Evaluator, request::PartialRequest};
 
-pub(crate) fn policy_expr_map<'a>(
+pub(crate) fn policy_residual_map<'a>(
     request: &'a PartialRequest,
     ps: &'a PolicySet,
     schema: &ValidatorSchema,
-) -> std::result::Result<HashMap<&'a PolicyID, Expr<Option<Type>>>, TpeError> {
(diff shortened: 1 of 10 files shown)


cedar-policy/CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ Starting with version 3.2.4, changes marked with a star (*) are _language breaki
 ### Added
 
 - Public syntax tree representation for `Policy`, `Template` and `PolicySet` allowing programmatic manipulation of Cedar syntax (#816, #366).
+- The Type-aware partial evaluation (TPE) experimental feature now supports template-linked policies. This would previously return a `SlotNotSupportedError` error.
+  This error variant is removed and replaced with `UnlinkedSlotError` (#2314).
```
victornicolet: nit: not exactly a 1:1 replacement.

cedar-policy-core/src/tpe/residual.rs
```diff
@@ -254,8 +264,21 @@ impl TryFrom<&Expr<Option<Type>>> for Residual {
...
+            ast::ExprKind::Slot(slot) => match env.get(slot) {
+                Some(euid) => {
+                    let value = Value::from(euid.clone());
+                    return Ok(Residual::Concrete { value, ty });
+                }
+                None => {
+                    // Any slot not bound in the env is an error now rather than waiting for evaluation
+                    return Err(evaluation_errors::UnlinkedSlotError {
+                        slot: *slot,
+                        source_loc: None,
```
luxas: nit: doesn't the expr have the loc?

cedar-policy-core/src/tpe/residual.rs
```diff
@@ -180,9 +178,16 @@ impl TryFrom<Residual> for Value {
     }
 }
 
-impl TryFrom<&Expr<Option<Type>>> for Residual {
```
luxas, line 183: do we want to keep this with an empty slotenv as the default?

cedar-policy-core/src/tpe.rs
```diff
@@ -25,46 +25,45 @@ pub mod response;
...
 use crate::tpe::{entities::PartialEntities, evaluator::Evaluator, request::PartialRequest};
 
-pub(crate) fn policy_expr_map<'a>(
+pub(crate) fn policy_residual_map<'a>(
     request: &'a PartialRequest,
     ps: &'a PolicySet,
     schema: &ValidatorSchema,
-) -> std::result::Result<HashMap<&'a PolicyID, Expr<Option<Type>>>, TpeError> {
+) -> std::result::Result<HashMap<&'a PolicyID, Residual>, TpeError> {
     let mut exprs = HashMap::new();
```
luxas: nit: can call this `residuals` now

cedar-policy-core/src/tpe.rs
```diff
@@ -84,29 +83,26 @@ pub fn is_authorized<'a>(
     entities: &'a PartialEntities,
     schema: &'a ValidatorSchema,
 ) -> std::result::Result<Response<'a>, TpeError> {
-    let exprs = policy_expr_map(request, ps, schema)?;
+    let exprs = policy_residual_map(request, ps, schema)?;
```
luxas: nit: I guess we can inline this directly onto line 92 and avoid this temporary var which could be confusing now that these are not exprs anymore?

cedar-policy-core/src/tpe.rs
```diff
@@ -84,29 +83,26 @@ pub fn is_authorized<'a>(
...
-                Arc::new(residual),
-                Arc::new(ps.get(id).unwrap().clone()),
-            ))
+            ResidualPolicy::new(Arc::new(residual), Arc::new(ps.get(id).unwrap().clone()))
         })
         .collect();
 
     Ok(Response::new(
-        residuals?.into_iter(),
+        residuals.into_iter(),
```
luxas: can we skip the `collect` and `into_iter` steps now that we don't need to unwrap the result?

cedar-policy/src/api/tpe.rs
```diff
@@ -2781,4 +2781,137 @@ when { principal in resource.admins };
...
+                )
+                .unwrap();
+
+            let request = partial_req();
+            let es = PartialEntities::empty();
+            let response = policies.tpe(&request, &es, &schema).unwrap();
+
+            assert_eq!(response.decision(), None);
+
+            let expected = pst::Template::new(
```
luxas: nit: this would be easier to read if just parsed from a string, then converted to pst and compared.

cedar-policy-core/src/tpe/residual.rs
```diff
@@ -180,9 +178,16 @@ impl TryFrom<Residual> for Value {
...
-    type Error = ExprToResidualError;
-    fn try_from(expr: &Expr<Option<Type>>) -> std::result::Result<Self, ExprToResidualError> {
+impl Residual {
+    /// Convert a typed expression to a residual.
+    ///
+    /// Takes a `SlotEnv` used to resolve template slots to their values. Each
+    /// template slot expression becomes an entityUID literal for the binding of
+    /// that slot in the environment.
+    pub fn try_from_typed_expr(
```
luxas, line 187: this reminds me of the discussion whether we should allow users to construct their own residuals in the first place, or just provide constructors that enforce the invariant that every constructed `Residual` is guaranteed to be typechecked (and thus can we rely on type errors "never happening", assuming the validator is sound).

luxas (approved):
Thanks, looks good to me 👍 
I like it better that we substitute up front, over keeping this state in the interpreter.

cedar-policy-core/src/tpe/residual.rs
john-h-kastner-aws: I'm personally not a fan of using `TryFrom`/`From` impls in general, and in this case I think the possibility of implicitly converting a template without binding slots makes a `TryFrom` definitively the wrong choice.

cedar-policy-core/src/tpe.rs
```diff
@@ -25,46 +25,45 @@ pub mod response;
...
         let t = p.template();
 
         let errs: Vec<_> = Validator::validate_entity_types_and_literals(schema, t).collect();
         if !errs.is_empty() {
             return Err(TpeError::Validation(errs));
         }
 
+        // Get an environment using the actual types of the entities linked with
+        // this template. If static, the slot env is empty and this is a no-op.
+        let env = env.clone().link_slot_env(p.env());
```
victornicolet, line 56: Initially I thought one solution would be to use the slot env of the policy to replace the slots with the linked entities, and then treat the body of the resulting slot-free "template" as before. Would this approach result in different residuals?

cedar-policy-core/src/tpe/residual.rs
victornicolet: Maybe this function could be `pub(crate)`, can users get to an `ast::Expr`?

cedar-policy-core/src/tpe.rs
john-h-kastner-aws: yeah, nice catch

cedar-policy-core/src/tpe.rs
john-h-kastner-aws: I think substitution before typechecking would give the same behavior, at least with the current limitations on templates.

The current implementation matches how we validate templates and linked policies outside of TPE. That's a little complicated, because we first validate the templates under all possible slot types, and then check that the slots of each linked policy match a valid type. Here it's simpler because we know exactly the type environment we care about. 

Typechecking could be slightly more precise if we validated links by substitution, but it won't actually any different given the current restrictions on slots. The difference that comes to mind is `User::"alice" == ?principal` which could be `true` or `false` depending on the slot binding. But, that only matters for a literal `User::"alice"`, and doesn't apply to the `principal` variable in the same position.  Probably a discussion worth having if we implement generalized templates.

cedar-policy/CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ Starting with version 3.2.4, changes marked with a star (*) are _language breaki
 ### Added
 
 - Public syntax tree (`pst`) module for programmatic construction, inspection, and manipulation of Cedar policies. Accessible via `to_pst()` / `try_into_pst()` / `from_pst()` on `Policy`, `Template`, and `PolicySet`. `try_into_pst()` consumes the value to avoid cloning. TPE residual policies can be converted to PST for structured inspection of residual expressions. Third-party types used in PST fields (`SmolStr`, `LinkedHashMap`, `NonEmpty`) are re-exported from the `pst` module. (#816, #366)
+- The Type-aware partial evaluation (TPE) experimental feature now supports template-linked policies. This would previously return a `SlotNotSupportedError` error.
+  This error variant is removed and replaced with `UnlinkedSlotError`, occurring only when slot in a linked policy is not bound. (#2314).
```
victornicolet, line 19: nit: missing "a" before slot.

victornicolet (approved):

PR #2323, opened 2026-04-22
TPE: Use `value.get_as_x` methods when available

luxas:
## Description of changes

Broken out from #2162

Using the `get_as_x` methods in TPE minimizes the delta to the concrete evaluator (which uses them), and gives us automatically (later) the right error types for type mismatches (which should not happen if the Residual is validated, but anyways).

### DIFF
--- a/cedar-policy-core/src/tpe/evaluator.rs
+++ b/cedar-policy-core/src/tpe/evaluator.rs
@@ -98,57 +98,44 @@ impl Evaluator<'_> {
             ResidualKind::And { left, right } => {
                 let left = self.interpret(left);
                 match &left {
-                    Residual::Concrete {
-                        value:
-                            Value {
-                                value: ValueKind::Lit(ast::Literal::Bool(false)),
-                                ..
-                            },
-                        ..
-                    } => mk_concrete(false.into()),
-                    Residual::Concrete {
-                        value:
-                            Value {
-                                value: ValueKind::Lit(ast::Literal::Bool(true)),
-                                ..
-                            },
-                        ..
-                    } => self.interpret(right),
-                    Residual::Concrete { .. } => mk_error(),
+                    Residual::Concrete { value, .. } => match value.get_as_bool() {
+                        Ok(false) => mk_concrete(false.into()), // false && <right> => false
+                        Ok(true) => self.interpret(right),      // true && <right> => <right>
+                        Err(_) => mk_error(),                   // <error> && <right> => <error>
+                    },
                     Residual::Partial { .. } => {
                         let right = self.interpret(right);
                         match &right {
-                            Residual::Concrete {
-                                value:
-                                    Value {
-                                        value: ValueKind::Lit(ast::Literal::Bool(true)),
-                                        ..
-                                    },
-                                ..
-                            } => left,
-                            Residual::Concrete {
-                                value:
-                                    Value {
-                                        value: ValueKind::Lit(ast::Literal::Bool(false)),
-                                        ..
-                                    },
-                                ..
-                            } => {
-                                if !left.can_error_assuming_well_formed() {
-                                    // simplify <error-free> && false == false
-                                    mk_concrete(false.into())
-                                } else {
-                                    // cannot simplify <non-error-free> && false
-                                    mk_residual(ResidualKind::And {
-                                        left: Arc::new(left),
-                                        right: Arc::new(mk_concrete(false.into())),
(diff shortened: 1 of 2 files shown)


cedar-policy-core/src/tpe/evaluator.rs
john-h-kastner-aws: Any reason not to make the same change here?

john-h-kastner-aws (approved):
Change is good, but I feel like it could probably be more aggressive

cedar-policy-core/src/tpe/evaluator.rs
luxas: I held off at first, as it actually introduced a small change: before, `<residual> && <nonbool>` was left as-is, now `<residual> && <nonbool>` => `<residual> && <typeerror>`. However, this is kinda expected anyways, and should never happen on validated residuals.

Left are only the `Residual::Concrete` matches against `HasAttr` and `GetAttr`, which either match an entity or record, I think it makes the most sense to leave those as-is, as the concrete evaluator also uses similar match statements, and then if none of them match, return a custom `type_error`

luxas:
Answered in https://github.com/cedar-policy/cedar/pull/2323#discussion_r3160152844:

> I held off at first, as it actually introduced a small change: before, <residual> && <nonbool> was left as-is, now <residual> && <nonbool> => <residual> && <typeerror>. However, this is kinda expected anyways, and should never happen on validated residuals.
>
> Left are only the Residual::Concrete matches against HasAttr and GetAttr, which either match an entity or record, I think it makes the most sense to leave those as-is, as the concrete evaluator also uses similar match statements, and then if none of them match, return a custom type_error

However, now I updated the RHS matching for `And` and `Or` to use `get_as_bool` as well, and improved the and and or tests to follow the same order as the match statements, such that we cover all cases exhaustively (some small edge cases were not covered before, I checked the coverage report). Also using insta to assert inner structure of the residual, this could be done for other TPE tests as well in a follow-up.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -97,57 +97,44 @@ impl Evaluator<'_> {
...
+                                // "<left-residual> && <nonbool>" => "<left-residual> && <error>"
+                                // TODO(luxas): Introduce a Residual::PartialError variant that says "the expression definitely errors, but with an unknown error"
+                                Err(_) => mk_residual(ResidualKind::And {
+                                    left: Arc::new(left),
+                                    right: Arc::new(mk_error()),
+                                }),
+                            },
+                            // Cannot simplify "<left-residual> && <right-residual>" or "<left-residual> && <error>"
+                            // The latter expression could become a Residual::PartialError later.
+                            Residual::Partial { .. } | Residual::Error(_) => {
```
luxas, line 133: using the explicit notion here instead of `_` for reader clarity.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -830,9 +794,9 @@ mod tests {
                 ..
             },
         );
-        // <non-error-free> && false cannot be simplified, e.g.
-        // principal.foo + 1 == 100 && 41 == 42
-        assert_matches!(
+        // "<non-error-free> && false" cannot be fully simplified, e.g.
+        // "principal.foo + 1 == 100 && 41 == 42" => "principal.foo + 1 == 100 && false"
+        assert_debug_snapshot!(
```
victornicolet: nit: I'm not a big fan of using debug-print string equality testing on internal representations, because it can add test maintenance burden when the representation changes even a little bit. I'm also saying this based on a very high-level understanding of insta and I don't know if insta makes that easier.

john-h-kastner-aws: Insta does make this a _lot_ easier. The test update is just running `cargo insta review` which will show you the diff and replace the expected output if you're happy with it.

That said, using the debug repr for expressions does make it difficult to read the tests, so asserting the snapshot on the cedar expression string might be nicer if there isn't anything lost in that conversion (e.g., type information) that we want to assert equality for.

luxas: Totally, strings are nicer to look at, so asserting those now instead, using insta 😄 
While I was at it, I also converted to use typed string exprs as input, to make it easier for the reader.
I can convert the other tests as well, but preferably in another PR.

victornicolet (approved):

PR #2331, opened 2026-05-01
Proto decode shall not panic

victornicolet:
## Description of changes

Removes places where the protobuf decode implementation could panic when converting from the protobuf model to the internal type. The goal is to eliminate panics in protobuf decoding in the API. 
(will update changelog because this changes some of the proto API, from -> try_from)

### DIFF
--- a/cedar-policy/src/proto/api.rs
+++ b/cedar-policy/src/proto/api.rs
@@ -85,45 +85,6 @@ impl TryFrom<models::PolicySet> for api::PolicySet {
     }
 }
 
-#[expect(clippy::use_self, reason = "readability")]
-impl From<&api::ValidationMode> for models::ValidationMode {
-    fn from(v: &api::ValidationMode) -> Self {
-        match v {
-            api::ValidationMode::Strict => models::ValidationMode::Strict,
-            #[cfg(feature = "permissive-validate")]
-            api::ValidationMode::Permissive => models::ValidationMode::Permissive,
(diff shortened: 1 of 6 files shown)


cedar-policy/src/proto/api.rs
```diff
@@ -399,4 +399,38 @@ mod test {
             let _ = crate::PolicySet::decode(*input);
         }
     }
+
+    #[test]
+    fn validation_mode_strict_roundtrip() {
```
john-h-kastner-aws: Do you see why `VaildationMode` is included in the protobuf definitions?  It feels to me like it shouldn't need to exist.

john-h-kastner-aws (approved):

john-h-kastner-aws (approved):

cedar-policy/src/proto/api.rs
victornicolet: Not sure. The only place I can see where the model is being used is in the Lean FFI. I think we could clean this up in the proto api of cedar-policy

john-h-kastner-aws: yeah, should clean this up before stable release if we don't find any reason not to

katherine-hough (approved):

cedar-policy/src/proto/api.rs
victornicolet: Cleaned up in this PR, looks like there isn't actual dependencies on the api part.

PR #2335, opened 2026-05-05
add requirements on having a release tag on the cli

victornicolet:
Now publish requires an additional tag to publish the cli.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -33,6 +33,10 @@ on:
                 required: false
                 type: string
                 description: "The CLI crate version. Must be of the form '<MAJOR>.<MINOR>.<PATCH>'. Required for cli and all."
+            cli_tag:
+                required: false
+                type: string
+                description: "The CLI tag (e.g., cedar-policy-cli-v4.11.0). Required for cli and all targets. Used to identify the CLI release."
 # Declare default permissions as read only.
 permissions: read-all
 jobs:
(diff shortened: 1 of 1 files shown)


katherine-hough (approved):

.github/workflows/publish.yml
```diff
@@ -123,7 +150,20 @@ jobs:
             - name: Checkout tag
               uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
               with:
-                  ref: refs/tags/${{ inputs.tag }}
+                  ref: refs/tags/${{ inputs.target == 'cli' && inputs.cli_tag || inputs.tag }}
+                  fetch-depth: 0
+            - name: Validate cli_tag points to same commit as tag
```
john-h-kastner-aws, line 155: Cam you explain this a bit more? If we force the commits to be the same, why introduce this second tag?

I guess it identifies the version, but that's also in `cargo.toml`, so I'm not sure I understand the value.

victornicolet: Yes the information would be in the `cargo.toml` (and meta-understanding on how we release things), this just makes it more explicit. If I need to checkout the repository for a specific release of the CLI, I just need to check out that tag. Otherwise, you need to first check if that tag exists (it may exist if the cli was released on its own), otherwise check the version of core the cli was released with and then check out the core tag.

john-h-kastner-aws: make sense. so for now we always release `cedar-policy` if we're releasing the CLI, but that could change

john-h-kastner-aws (approved):

PR #2338, opened 2026-05-06
adding support for linked policies in symcc

victornicolet:
This adds support for templated-linked policies in symcc.

DRT changes: https://github.com/cedar-policy/cedar-spec/pull/948

### DIFF
--- a/cedar-policy-cli/src/command/symcc.rs
+++ b/cedar-policy-cli/src/command/symcc.rs
@@ -21,7 +21,7 @@ use miette::{miette, Result};
 use std::path::PathBuf;
 
 use crate::CedarExitCode;
-use crate::{read_cedar_policy_set, read_json_policy_set, PoliciesArgs};
+use crate::{add_template_links_to_set, read_cedar_policy_set, read_json_policy_set, PoliciesArgs};
 use crate::{PolicyFormat, SchemaArgs};
 
 #[derive(Args, Debug)]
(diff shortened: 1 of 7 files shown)


cedar-policy-cli/src/command/symcc.rs
```diff
@@ -96,15 +96,22 @@ pub struct SymccPoliciesArgs {
     /// Format of policies in the `--policies` file
     #[arg(long = "policy-format", default_value_t, value_enum)]
     pub policy_format: PolicyFormat,
+    /// File containing template-linked policies
+    #[arg(short = 'k', long = "template-linked", value_name = "FILE")]
+    pub template_linked_file: Option<String>,
```
john-h-kastner-aws, line 101: Do we need a separate `SymccPoliciesArg` now that it accepts templates links? I think I remember @katherine-hough  duplicated in order to not accept templates here.

cedar-policy-symcc/src/symcc.rs
```diff
@@ -655,16 +658,35 @@ fn well_typed_policy_inner(
     env: &RequestEnv<'_>,
     schema: &Schema,
 ) -> Result<Policy> {
+    // For linked policies, create a static equivalent by resolving slots in the
```
john-h-kastner-aws, line 661: Could we change the model to match this? IIRC, the lean FFI layer currently does this, but it could make sense to handle this is the corresponding lean function 

here,  I think:

https://github.com/cedar-policy/cedar-spec/blob/92de50d691a8f59bc7bdfce37dd070e6d50a0b58/cedar-lean/Cedar/SymCC.lean#L58

might be a larger update than i'm thinking though

cedar-policy-cli/src/command/symcc.rs
katherine-hough: John is correct see: https://github.com/cedar-policy/cedar/commit/afae911e4c3acaf318fafa84ba7d8f1774921ca8

cedar-policy-symcc/src/symcc.rs
victornicolet: Yes the Lean FFI layer is currently doing the slot replacement. 
I attempted changing that function in Lean, but it would be a large update. We can decide that it's necessary though.

Another difference between the Rust and Lean is that the Lean env doesn't get the extra information that the linked env in Rust passes. However, trying to integrate that alone in Lean doesn't seem trivial either.

cedar-policy-cli/src/command/symcc.rs
victornicolet: Thanks, I should have remembered too.

cedar-policy-symcc/src/symcc.rs
john-h-kastner-aws: If you have a decent idea of what would be needed, can you type up an issue on cedar-spec?

PR #2339, opened 2026-05-07
More error handling in proto decode

victornicolet:
Adds more error handling in the proto decode path.

Replaces panic behavior with error behavior, but you can still decode invalid schemas.

### DIFF
--- a/cedar-policy/src/proto/validator.rs
+++ b/cedar-policy/src/proto/validator.rs
@@ -190,14 +190,24 @@ impl TryFrom<models::EntityDecl> for cedar_policy_core::validator::ValidatorEnti
             Some(enum_choices) => {
                 // `enum_choices` is not empty, so `v` represents an enumerated entity type.
                 // enumerated entity types must have no attributes or tags.
-                assert_eq!(v.attributes, HashMap::new());
-                assert_eq!(v.tags, None);
-                Ok(Self::new_enum(
-                    name,
-                    descendants,
-                    enum_choices.map(Eid::new),
-                    None,
-                ))
+                if !v.attributes.is_empty() {
+                    Err(ProtobufConversionError::InvalidValue(format!(
+                        "enum type {} should not have attributes",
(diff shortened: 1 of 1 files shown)


cedar-policy/src/proto/validator.rs
```diff
@@ -190,14 +190,24 @@ impl TryFrom<models::EntityDecl> for cedar_policy_core::validator::ValidatorEnti
...
+                    )))
+                } else if v.tags.is_some() {
+                    Err(ProtobufConversionError::InvalidValue(format!(
+                        "enum type {} should not have tags",
+                        name
+                    )))
+                } else {
+                    Ok(Self::new_enum(
+                        name,
+                        descendants,
```
john-h-kastner-aws, line 206: Should add some tests around descendants as well:

```
entity E enum ["0"];
entity D in E;
```

is allowed, and looks like it should be allowed here too.

The other case is

```
entity E enum ["0"] in D;
entity D;
```

Which fails to parse, but I don't see where this will get rejected in protobuf parsing

victornicolet: Do we need additional validation in the protobuf parsing to reject that case? Since the protobuf has `descendants` and not membership like the Cedar syntax, the thing we're trying to reject here requires looking at the entire schema.

john-h-kastner-aws: yeah, it probably does

victornicolet: Do you mind if we merge this PR (I'll add the two tests) and create an issue for additional validation, so that we can get more coverage from the panic-testing DRT in the meantime?

john-h-kastner-aws: yeah, fine for now. Needs to change for protobuf stabalization

john-h-kastner-aws (approved):

katherine-hough (approved):

PR #2368, opened 2026-05-28
fix model decoding to work zith z3

victornicolet:
## Description of changes
Skips extra `(define-fun ...` in solver model output because some solver may include more declaration than the expected model values for the `(define-const`'d variables. 

Skipping extra definitions in the model should have no impact on backwards compatibility; an extra definition is not an error. Later on, we may want to validate models further (e.g., that those extra definitions are declarations in the input of the solver). 


katherine-hough (approved):

john-h-kastner-aws (approved):
Fix looks reasonable

Issue #2380, opened 2026-06-03
try_into_pst() fails with WrongArityError on variadic isInRange calls


tomlikestorock:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [x] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Describe the bug

When the `variadic-is-in-range` feature is enabled, `isInRange` calls with 3+ arguments are legal per RFC 99 (https://github.com/cedar-policy/rfcs/pull/99). They parse, validate, and evaluate correctly through the AST path. However, calling `try_into_pst()` on a policy containing such a call fails with:

`WrongArityError { name: "isInRange", expected: 2, got: 3 }`

The root cause is two stacked issues in `cedar-policy-core/src/pst/expr.rs`:
1. `from_function_names_and_args` checks `extension.arg_types().len()` against the actual arg count without consulting `extension.is_variadic()`. For `isInRange, arg_types().len()` is always 2 regardless of the feature flag.
2. The PST `Expr` enum has no variant that can hold more than 2 arguments for an extension call — `BinaryOp { op, left, right }` is strictly binary, and the fallthrough arm `_ => UnknownFunctionError` rejects anything else.

This means `try_into_pst()` is lossy for any legal AST containing a variadic extension function call, making PST unusable for codebases that enable `variadic-is-in-range`.

### Reproduction steps

With `cedar-policy = { version = "4.11.0", features = ["variadic-is-in-range"] }`:

```rust
// Put your code below this line.
use cedar_policy::{Policy, PolicyId};
use std::str::FromStr;
let json = serde_json::json!({
    "effect": "permit",
    "principal": { "op": "All" },
    "action": { "op": "All" },
    "resource": { "op": "All" },
    "conditions": [{
        "kind": "when",
        "body": {
            "isInRange": [
                {"ip": [{"Value": "10.0.0.1"}]},
                {"ip": [{"Value": "10.0.0.0/24"}]},
                {"ip": [{"Value": "192.168.0.0/16"}]}
            ]
        }
    }]
});

let policy_id = PolicyId::from_str("test").unwrap();

// This succeeds — the AST accepts variadic isInRange
[...]

PR #2381, opened 2026-06-03
Add variadic extension function support to PST

tomlikestorock:
## Description of changes
The PST module (introduced in 4.11.0) cannot represent variadic extension function calls such as `isInRange` with 3+ arguments (legal under the `variadic-is-in-range` feature flag per [RFC 99](https://github.com/cedar-policy/rfcs/pull/99)). `try_into_pst()` fails with `WrongArityError` because:

1. `from_function_names_and_args` compares `arg_types().len()` (always 2 for `isInRange`) against the actual arg count without consulting `is_variadic()`
2. The `Expr` enum has no variant capable of holding more than 2 args for an extension call — only `UnaryOp` and `BinaryOp`

This PR:

- Adds `Expr::ExtensionCall { name, style, args }` variant for variadic calls, storing `CallStyle` for future display/pretty-printing use
- Fixes the arity gate to consult `extension.is_variadic()`
- Routes 2-arg calls through existing `BinaryOp::IsInRange` (unchanged)
- Routes 3+ arg calls through the new `ExtensionCall` variant
- Adds PST↔AST and PST↔EST conversion support
- Adds round-trip tests for 2, 3, and 4-arg `isInRange` calls
- Adds error case test for too-few args on variadic functions
- Adds Display round-trip and unknowns detection tests
- Documents feature-flag requirement and invariant caveats on the variant

The design is generic — future variadic extension functions will route through `ExtensionCall` automatically without additional PST changes.

## Issue #, if available
Issue #2380

### DIFF
--- a/cedar-policy-core/src/pst/ast_conversions.rs
+++ b/cedar-policy-core/src/pst/ast_conversions.rs
@@ -501,6 +501,18 @@ impl Expr {
                 name,
                 type_annotation: None,
             }),
+            #[cfg(feature = "variadic-is-in-range")]
+            Expr::VariadicIsInRange { left, rights } => {
+                use crate::extensions::ipaddr::names::IS_IN_RANGE;
+                let mut ast_args = Vec::with_capacity(1 + rights.len());
+                ast_args.push(Arc::unwrap_or_clone(left).into_expr::<B>());
+                for right in rights {
+                    ast_args.push(Arc::unwrap_or_clone(right).into_expr::<B>());
+                }
+                builder
+                    .call_extension_fn(IS_IN_RANGE.clone(), ast_args)
+                    .unwrap_infallible()
+            }
             #[cfg(feature = "tpe")]
             Expr::ResidualError => builder
                 .call_extension_fn(crate::tpe::residual::ERROR_NAME.clone(), std::iter::empty())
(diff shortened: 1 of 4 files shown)


victornicolet:
Thank you for the contribution!
I think we need to be a little careful if we want to add support for the `variadic-is-in-range` feature for the PST. One of the high-level properties we'd like to maintain is: matching the PST should not require you to match operator names, we have statically defined enums that represent all possible operators.

tomlikestorock:
@victornicolet Oh, thanks for the feedback, wasn't entirely sure how specific to make this. Would you prefer a VariadicOp enum (parallel to UnaryOp/BinaryOp) with VariadicOp::IsInRange as its first variant, or a dedicated IsInRange { subject, targets } variant on Expr that separates the receiver from the variadic range arguments?

Totally fine to do either version.

victornicolet:
Thank you, I put up an example PR with a slightly different design: https://github.com/cedar-policy/cedar/pull/2382/changes I'll expect other contributors to have an opinion and we can choose either design.

Can you make the tests follow the general shape of tests in either test file when relevant, similar to that other PR?
Tests in the `ast_conversions.rs` use the `assert_expr_roundtrip` helper for example, and simply parse expressions as opposed to building them.

PR #2383, opened 2026-06-03
A optional depth-limit when parsing policies 

john-h-kastner-aws:
## Description of changes

Adds new functions to the public API to configure an optional depth limit for policies when parsing. These function will return an error result for any policy that exceeds the limit, avoiding any possible stack overflow.

As an additional complication, I also had to add a `Drop` implementation for the CST so that we would not stackoverflow when dropping that structure after returning an error. The drop impl is from Claude. IIUC, Rust will always full free the structure regardless of what we do in our Drop, so there's no risk of memory leaks. 

### DIFF
--- a/cedar-policy/src/api.rs
+++ b/cedar-policy/src/api.rs
@@ -2545,40 +2545,78 @@ impl FromStr for PolicySet {
     /// See [`Policy`] for more.
     fn from_str(policies: &str) -> Result<Self, Self::Err> {
         let (texts, pset) = parser::parse_policyset_and_also_return_policy_text(policies)?;
-        #[expect(clippy::expect_used, reason = "By the invariant on `parse_policyset_and_also_return_policy_text(policies)`, every `PolicyId` in `pset.policies()` occurs as a key in `text`.")]
-        let policies = pset.policies().map(|p|
-            (
-                PolicyId::new(p.id().clone()),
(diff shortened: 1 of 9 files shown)


cedar-policy/src/api.rs
```diff
@@ -2545,40 +2545,78 @@ impl FromStr for PolicySet {
...
+impl PolicySet {
+    /// Create a policy set as in [`PolicySet::from_str`], but rejecting any
+    /// policy whose expression depth exceeds `depth_limit`.
+    ///
+    /// This function can be used to limit the maximum recursion depth, avoiding
+    /// the stackoverflows when parsing policies with unknown depth.
+    pub fn from_str_with_depth_limit(
+        policies: &str,
+        depth_limit: usize,
+    ) -> Result<Self, ParseErrors> {
```
victornicolet, line 2561: Would users also want to know the depth of the policies they parsed? I guess it's not really useful and we expect them to test parsing with enough depth that it accepts their policies, but doesn't stack overflow. 
In terms of API, what does backwards compatibility mean, does it imply if parsing X with depth D suceeds in one version, then it should succeed in a non-breaking version?

victornicolet:
A note that there is another place where we are hitting depth limits that are even more shallow: `Protobuf::decode` will hit a `RecursionLimitReached` error based on `prost`'s fixed limit of 100.

victornicolet:
I think this could still be useful, at least when doing differential testing, when we are parsing random bytes and we don't want to have a stack overflow.

PR #2401, opened 2026-06-10
Add functions for inspecting TPE determining policies

john-h-kastner-aws:
Adds `TpeResponse::reason`, `TpeResponse::satisfied_permits`, and  `TpeResponse::satisfied_forbids` to the public TPE API to enable callers to find determining policies for partial authorization responses. 

### DIFF
--- a/cedar-policy-core/src/tpe/response.rs
+++ b/cedar-policy-core/src/tpe/response.rs
@@ -59,8 +59,8 @@ impl ResidualPolicy {
     }
 
     /// Get the [`PolicyID`]
-    pub fn get_policy_id(&self) -> PolicyID {
-        self.policy.id().clone()
+    pub fn get_policy_id(&self) -> &PolicyID {
+        self.policy.id()
     }
 
     /// All literal uids referenced by this residual
@@ -130,20 +130,20 @@ impl<'a> Response<'a> {
             match rp.get_effect() {
                 Effect::Forbid => {
                     if r.is_true() {
-                        satisfied_forbids.insert(id);
+                        satisfied_forbids.insert(id.clone());
                     } else if r.is_false() || r.is_error() {
-                        false_forbids.insert(id);
+                        false_forbids.insert(id.clone());
                     } else {
-                        residual_forbids.insert(id);
+                        residual_forbids.insert(id.clone());
                     }
                 }
                 Effect::Permit => {
                     if r.is_true() {
-                        satisfied_permits.insert(id);
+                        satisfied_permits.insert(id.clone());
                     } else if r.is_false() || r.is_error() {
-                        false_permits.insert(id);
+                        false_permits.insert(id.clone());
                     } else {
-                        residual_permits.insert(id);
+                        residual_permits.insert(id.clone());
                     }
                 }
             }
--- a/cedar-policy-core/src/tpe/response.rs
+++ b/cedar-policy-core/src/tpe/response.rs
@@ -88,13 +88,13 @@ pub struct Response<'a> {
     decision: Option<Decision>,
     residuals: HashMap<PolicyID, ResidualPolicy>,
     // All of the [`Effect::Permit`] policies that were satisfied
-    satisfied_permits: HashSet<PolicyID>,
+    true_permits: HashSet<PolicyID>,
     // All of the [`Effect::Permit`] policies that were not satisfied
     false_permits: HashSet<PolicyID>,
     // All of the [`Effect::Permit`] policies that evaluated to a residual
     residual_permits: HashSet<PolicyID>,
     // All of the [`Effect::Forbid`] policies that were satisfied
-    satisfied_forbids: HashSet<PolicyID>,
+    true_forbids: HashSet<PolicyID>,
     // All of the [`Effect::Forbid`] policies that were not satisfied
     false_forbids: HashSet<PolicyID>,
     // All of the [`Effect::Forbid`] policies that evaluated to a residual
@@ -117,10 +117,10 @@ impl<'a> Response<'a> {
         schema: &'a ValidatorSchema,
     ) -> Self {
         let mut residual_map = HashMap::new();
-        let mut satisfied_permits = HashSet::new();
+        let mut true_permits = HashSet::new();
         let mut false_permits = HashSet::new();
         let mut residual_permits = HashSet::new();
-        let mut satisfied_forbids = HashSet::new();
+        let mut true_forbids = HashSet::new();
         let mut false_forbids = HashSet::new();
         let mut residual_forbids = HashSet::new();
         for rp in residuals {
@@ -130,7 +130,7 @@ impl<'a> Response<'a> {
             match rp.get_effect() {
                 Effect::Forbid => {
                     if r.is_true() {
-                        satisfied_forbids.insert(id.clone());
+                        true_forbids.insert(id.clone());
                     } else if r.is_false() || r.is_error() {
                         false_forbids.insert(id.clone());
                     } else {
@@ -139,7 +139,7 @@ impl<'a> Response<'a> {
                 }
                 Effect::Permit => {
                     if r.is_true() {
-                        satisfied_permits.insert(id.clone());
+                        true_permits.insert(id.clone());
                     } else if r.is_false() || r.is_error() {
                         false_permits.insert(id.clone());
                     } else {
@@ -150,8 +150,8 @@ impl<'a> Response<'a> {
         }
 
         let decision = match (
-            !satisfied_forbids.is_empty(),
-            !satisfied_permits.is_empty(),
+            !true_forbids.is_empty(),
+            !true_permits.is_empty(),
             !residual_permits.is_empty(),
             !residual_forbids.is_empty(),
         ) {
@@ -170,10 +170,10 @@ impl<'a> Response<'a> {
         Self {
             decision,
             residuals: residual_map,
-            satisfied_permits,
+            true_permits,
             false_permits,
             residual_permits,
-            satisfied_forbids,
+            true_forbids,
             false_forbids,
             residual_forbids,
             request,
@@ -188,7 +188,7 @@ impl<'a> Response<'a> {
             clippy::unwrap_used,
             reason = "we know that the policy ids are in the residuals map"
         )]
-        self.satisfied_permits
+        self.true_permits
             .iter()
             .map(|id| self.residuals.get(id).unwrap())
     }
@@ -199,7 +199,7 @@ impl<'a> Response<'a> {
             clippy::unwrap_used,
             reason = "we know that the policy ids are in the residuals map"
         )]
-        self.satisfied_forbids
+        self.true_forbids
             .iter()
             .map(|id| self.residuals.get(id).unwrap())
     }
@@ -256,6 +256,14 @@ impl<'a> Response<'a> {
     /// Attempt to get the authorization decision
     pub fn decision(&self) -> Option<Decision> {
         self.decision
+    }
+
+    /// Get the determining policies for the authorization decision
+    pub fn reason(&self) -> Option<impl Iterator<Item = &PolicyID>> {
+        match self.decision? {
+            Decision::Allow => Some(self.true_permits.iter()),
+            Decision::Deny => Some(self.true_forbids.iter()),
+        }
     }
 
     /// Perform reauthorization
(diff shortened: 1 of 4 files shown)


cedar-policy/src/api/tpe.rs
```diff
@@ -384,6 +385,44 @@ impl TpeResponse<'_> {
...
+            Some(Decision::Allow) => self.satisfied_permits().collect(),
+            Some(Decision::Deny) => self.satisfied_forbids().collect(),
+        }
+    }
+
+    /// Get the permit policies that are concretely satisfied by the partial request.
+    ///
+    /// If the eventual concrete authorization decision is allow, then these
+    /// will be a subset of the determining policies. However, they may still be
+    /// overridden by a non-trivial residual forbid policy.
```
victornicolet: "they may still be overriden" in what cases? Not when the eventual decision is allow? Should it be something like:
```
/// If the eventual concrete authorization decision is allow, then these will be a subset of the determining policies.
/// In general, when TPE did not reach a concrete decision, they may still be overriden ...
```

cedar-policy/src/api/tpe.rs
```diff
@@ -384,6 +385,44 @@ impl TpeResponse<'_> {
...
     }
 
+    /// Get the determining policies for the partial authorization decision.
+    /// These are a subset of the determining policies for any subsequent
+    /// concrete reauthorization.
+    ///
+    /// When [`TpeResponse::decision`] returns a concrete allow or deny, the
+    /// determining policies are the satisfied permits or forbids respectively.
+    /// If TPE did not reach a concrete decision, then this will always return
+    /// an empty set.
```
luxas: We could note here that an empty set might also be returned legitimately when there are no policies matching at all.

cedar-policy/src/api/tpe.rs
```diff
@@ -384,6 +385,44 @@ impl TpeResponse<'_> {
...
+    pub fn reason(&self) -> HashSet<&PolicyId> {
```
luxas: Was there a specific reason for making this a HashSet over e.g. an iterator?

cedar-policy/src/api/tpe.rs
```diff
@@ -384,6 +385,44 @@ impl TpeResponse<'_> {
...
+        match self.0.decision() {
+            None => HashSet::new(),
```
luxas: Another option, which I guess you considered, would be to return `Option<HashSet...>` and `None` here to clearly indicate the unknown status, as an empty map is otherwise ambiguous if you look at that alone (no matching policies vs not sure yet)

cedar-policy/src/api/tpe.rs
```diff
@@ -384,6 +385,44 @@ impl TpeResponse<'_> {
         self.0.decision()
     }
 
+    /// Get the determining policies for the partial authorization decision.
+    /// These are a subset of the determining policies for any subsequent
+    /// concrete reauthorization.
```
luxas: I'm not sure I understand the `for any subsequent concrete reauthorization` part of the comment, the returned policies already evaluated to true, so there is no need for reauthorization?

cedar-policy/src/api/tpe.rs
```diff
@@ -384,6 +385,44 @@ impl TpeResponse<'_> {
...
+            .satisfied_permits()
+            .map(|rp| PolicyId::ref_cast(rp.get_policy_id()))
+    }
+
+    /// Get the forbid policies that are concretely satisfied by the partial request.
+    ///
+    /// Presence of any satisfied forbids guarantees that they are exactly the
+    /// policies returned by [`TpeResponse::reason`] and that [`TpeResponse::decision`]
+    /// must return `Deny`.
+    pub fn satisfied_forbids(&self) -> impl Iterator<Item = &PolicyId> {
```
luxas: On naming, I think satisfied_ is an ok name, but it's a little bit confusing in cedar-policy-core today where true ones are called `satisfied_` but false ones are called `false_`. Not sure if it'd be better to just call them `true_` and `false_` respectively; however, we don't expose the `false_` methods in concrete diagnostics either.

cedar-policy/src/api/tpe.rs
```diff
@@ -384,6 +385,44 @@ impl TpeResponse<'_> {
...
+    /// If TPE did not reach a concrete decision, then this will always return
+    /// an empty set.
+    pub fn reason(&self) -> HashSet<&PolicyId> {
+        match self.0.decision() {
+            None => HashSet::new(),
+            Some(Decision::Allow) => self.satisfied_permits().collect(),
+            Some(Decision::Deny) => self.satisfied_forbids().collect(),
+        }
+    }
+
```
luxas, line 403: While we're at it, should we consider surfacing `erroring_policies` as well (it can be a separate PR)? Right now we put them in the `false_` bucket.

luxas (commented):
Thanks, this is a good addition to the API surface 👍

What the Kubernetes integration in particular is interested in, is understanding that when given `decision=Deny`, did that originate from a matched deny policy or just no matching policies? In the latter case, one could treat that as a `NoOpinion` mode, and let other subsequent authorizers (if many) have their say.

cedar-policy/src/api/tpe.rs
john-h-kastner-aws: mirrors the current API for `reason` on a concrete request. might be worth changing

cedar-policy/src/api/tpe.rs
john-h-kastner-aws: I thought about this, but I kind of like the property that this always returns a subset of determining policies which we know are determining. Callers need to inspect `decision` anyways to know how to interpret this, and adding an extra option on this call felt unnecessary.

cedar-policy/src/api/tpe.rs
john-h-kastner-aws: The idea is that if you were to do concrete authorization, there might be _more_ determining policies that aren't represented here. If you want to take some action based on which policies apply to a request, this might be important.

cedar-policy/src/api/tpe.rs
john-h-kastner-aws: yeah, may as well expose the full true/false/error API here.

cedar-policy/src/api/tpe.rs
john-h-kastner-aws: expanded comments substantially. let me know if it makes sense

cedar-policy/src/api/tpe.rs
luxas: Let's distinguish the erroring policies here now then already (and similarly in `cedar_policy_core`), as preparation for #2162? (I'll pick up the PR first after you've done the partial attributes change, to not make your proofs harder for now as we agreed)

cedar-policy/src/api/tpe.rs
```diff
@@ -384,6 +385,54 @@ impl TpeResponse<'_> {
...
+    ///
+    /// If partial authorization does not reach a decision, then this function
+    /// returns `None`. It's reasonable to treat this response as "no known
+    /// determining policies", in which case you can call this function as
+    /// `response.reason().into_iter().flatten()`.
+    pub fn reason(&self) -> Option<impl Iterator<Item = &PolicyId>> {
+        Some(self.0.reason()?.map(PolicyId::ref_cast))
+    }
+
+    /// Get the permit policies that are concretely satisfied by the partial request.
```
luxas, line 419: I like it 👍

cedar-policy/src/api/tpe.rs
luxas: Maybe `determining policies for any subsequent concrete reauthorization` could be reworded as `determining policies of an concrete authorization request with a complete Entity store`? I guess that's what you mean?

luxas (commented):
Thanks, I like it 👍 
Should we make the error change in this or another PR? It could be done in another.

cedar-policy-core/src/tpe/response.rs
```diff
@@ -130,28 +136,32 @@ impl<'a> Response<'a> {
             match rp.get_effect() {
                 Effect::Forbid => {
                     if r.is_true() {
```
luxas, line 138: nit: feels like this could be a good candidate to have an enum / match statement for in the future.

cedar-policy-core/src/tpe/response.rs
```diff
@@ -215,6 +227,17 @@ impl<'a> Response<'a> {
             .map(|id| self.residuals.get(id).unwrap())
     }
 
+    /// Get trivially erroring permit residual policies
```
luxas, line 230:
```suggestion
    /// Get permit policies that errored during evaluation
```

nit: not sure I understand the word trivially in this context, maybe just drop it?

cedar-policy-core/src/tpe/response.rs
```diff
@@ -226,6 +249,17 @@ impl<'a> Response<'a> {
             .map(|id| self.residuals.get(id).unwrap())
     }
 
+    /// Get trivially erroring forbid residual policies
```
luxas, line 252:
```suggestion
    /// Get forbid policies that errored during evaluation
```

cedar-policy/src/api/tpe.rs
```diff
@@ -384,7 +385,136 @@ impl TpeResponse<'_> {
...
+    pub fn error_permits(&self) -> impl Iterator<Item = &PolicyId> {
+        self.0
+            .error_permits()
+            .map(|rp| PolicyId::ref_cast(rp.get_policy_id()))
+    }
+
+    /// Get the forbid policies that did not reach a concrete value or error for the partial request.
+    ///
+    /// This function only returns the `PolicyId`s for residual policies.
+    /// To access the residual policy conditions, use [`TpeResponse::nontrivial_residual_policies`].
```
luxas, line 466: this reminds me that we should probably rename `TpeResponse::nontrivial_residual_policies` to just `TpeResponse::residual_policies` now, as we've anyways factored out the "trivial" ones to their own constructors and thus don't need that trivial/nontrivial distinction. That'd then be consistent with calling the other getters here "just" `residual_forbids/permits`

cedar-policy/src/api/tpe.rs
```diff
@@ -384,7 +385,136 @@ impl TpeResponse<'_> {
...
+    }
+
+    /// Get the forbid policies that are concretely satisfied by the partial request.
+    ///
+    /// Presence of any satisfied forbids guarantees that they are exactly the
+    /// policies returned by [`TpeResponse::reason`] and that [`TpeResponse::decision`]
+    /// must return a concrete `Deny`.
+    pub fn true_forbids(&self) -> impl Iterator<Item = &PolicyId> {
+        self.0
+            .satisfied_forbids()
```
luxas, line 484: seems like one missing getter rename

cedar-policy/src/api/tpe.rs
```diff
@@ -384,7 +385,136 @@ impl TpeResponse<'_> {
...
+    ///   iterator may be empty if no permits were satisfied, or it may contain
+    ///   satisfied permits which have been overridden by at least one satisfied
+    ///   forbid policy.
+    /// * For an unknown decision, these will be a subset of the determining
+    ///   policies _if_ the eventual concrete authorization decision is `Allow`,
+    ///   but they may still be overridden by any non-trivial residual forbid
+    ///   policy.
+    pub fn true_permits(&self) -> impl Iterator<Item = &PolicyId> {
+        self.0
+            .satisfied_permits()
```
luxas, line 436: one missing getter rename, or don't we want to update all of these at the same time?

luxas (approved):
Approving with some small renames that could be followed up upon if we like.
Thanks, this should give folks quite a lot of data to play with to deepen the understanding of the decision.

victornicolet (approved):

cedar-policy-core/src/tpe/response.rs
john-h-kastner-aws: We could write this as a match on the residual, but I don't think I see how it would make the code any nicer

cedar-policy/src/api/tpe.rs
john-h-kastner-aws: I need to think more about the naming here. In some places "residual" refers to _any_ output of partial evaluation (even if it reaches a concrete value). In other place "residual" specifically means a policy which didn't reach a  concrete value.

PR #2403, opened 2026-06-11
schema validation

victornicolet:
## Description of changes

Implements the `try_validate` function for `ValidatorSchemas` to enforce that `Schema::decode` validates just as much as schema parsing or deserialization from JSON. 

### DIFF
--- a/cedar-policy-core/src/validator/schema.rs
+++ b/cedar-policy-core/src/validator/schema.rs
@@ -28,6 +28,7 @@ use crate::{
     transitive_closure::compute_tc,
 };
 use educe::Educe;
+use itertools::Itertools;
 use namespace_def::EntityTypeFragment;
 use nonempty::NonEmpty;
 #[cfg(feature = "extended-schema")]
@@ -1079,6 +1080,8 @@ impl ValidatorSchema {
     /// Returns an iterator over all leaf types reachable from entity attributes, tags, and
     /// action contexts. Leaf types are all types except `Set` and `Record`, which are
     /// traversed into automatically.
+    /// Note that a type is traversed multiple times if it is referenced multiple times; consumers
+    /// should use [`Itertools::unique`] when they care about not visiting a type multiple times.
     fn leaf_types(&self) -> TypeIterator<'_> {
         let mut stack: Vec<&Type> = Vec::new();
         for ety in self.entity_types.values() {
@@ -1161,7 +1164,7 @@ impl ValidatorSchema {
     /// Check that all entity types referenced in attribute/tag/context types are declared
     fn check_references_wf(&self) -> std::result::Result<(), SchemaError> {
         let mut undeclared = Vec::new();
-        for ty in self.leaf_types() {
+        for ty in self.leaf_types().unique() {
             if let Type::Entity(EntityKind::Entity(lub)) = ty {
                 for e in lub.iter() {
                     if !self.entity_types.contains_key(e) {
@@ -1179,7 +1182,7 @@ impl ValidatorSchema {
     fn check_extension_types_wf(&self) -> std::result::Result<(), SchemaError> {
         let extensions = Extensions::all_available();
         let valid_ext_types: HashSet<_> = extensions.ext_types().collect();
-        for ty in self.leaf_types() {
+        for ty in self.leaf_types().unique() {
             if let Type::ExtensionType { name } = ty {
                 if !valid_ext_types.contains(name) {
                     return Err(SchemaError::UnknownExtensionType(
--- a/cedar-policy-core/src/validator/schema.rs
+++ b/cedar-policy-core/src/validator/schema.rs
@@ -41,7 +41,7 @@ use crate::validator::{
     cedar_schema::SchemaWarning,
     json_schema,
     partition_nonempty::PartitionNonEmpty,
-    types::{Attributes, EntityKind, OpenTag, RequestEnv, Type},
+    types::{Attributes, EntityKind, OpenTag, RequestEnv, Type, TypeIterator},
     ValidationMode,
 };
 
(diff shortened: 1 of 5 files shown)


cedar-policy-core/src/validator/schema/err.rs
```diff
@@ -257,6 +257,15 @@ pub enum SchemaError {
...
     ActionInvariantViolation(#[from] schema_errors::ActionInvariantViolationError),
+    /// An action has an entity type whose basename is not `Action`.
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    InvalidActionType(#[from] schema_errors::InvalidActionTypeError),
+    /// An enum entity type appears as a descendant of another entity type,
+    /// which is not allowed.
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    EnumEntityInHierarchy(#[from] schema_errors::EnumEntityInHierarchyError),
```
victornicolet, line 275: Note this is new API. We could have a new error type for what `try_validate`, but those enums do represent "schema errors" so I think they fit here.

cedar-policy-core/src/validator/schema.rs
```diff
@@ -1086,6 +1086,41 @@ impl ValidatorSchema {
         })
     }
 
+    /// Iterate over all leaf types reachable from entity attributes, tags, and action contexts,
+    /// calling `visit` on each leaf type. Recursion through sets and records is handled
+    /// automatically.
+    fn for_each_type(
```
john-h-kastner-aws: It doesn't matter too much, but I like the pattern used by `ExprIterator` to return an iterator instead of passing in a closure

john-h-kastner-aws (approved):

cedar-policy-core/src/validator/schema.rs
```diff
@@ -1097,13 +1132,128 @@ impl ValidatorSchema {
...
+                    undeclared_entities.push(ety.clone());
+                }
+            }
+            for descendant in action.descendants() {
+                if !self.action_ids.contains_key(descendant) {
+                    return Err(ActionInvariantViolationError {
+                        euids: NonEmpty::singleton(descendant.clone()),
+                    }
+                    .into());
+                }
```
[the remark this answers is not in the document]
victornicolet: Will add a new enum variant, we shouldn't suggest filing an issue when protobuf decode fails.

cedar-policy-core/src/validator/schema/err.rs
```diff
@@ -161,6 +161,11 @@ pub enum SchemaError {
     #[error(transparent)]
     #[diagnostic(transparent)]
     UndeclaredEntityTypes(#[from] schema_errors::UndeclaredEntityTypesError),
+    /// Undeclared action(s) referenced in the `memberOf` (descendants) field
+    /// of an action.
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    UndeclaredActions(#[from] schema_errors::UndeclaredActionsError),
```
lianah: How were these issues handled before in schema validation if we don't have a schema error for them?

cedar-policy-core/src/validator/schema/err.rs
```diff
@@ -421,6 +437,41 @@ pub mod schema_errors {
         impl_diagnostic_from_method_on_nonempty_field!(types, loc);
     }
 
+    /// Undeclared actions error: an action's descendants list references
```
lianah: Minor typo? "list references actions" -> "references actions"

cedar-policy-core/src/validator/schema/err.rs
victornicolet: There is another error type `ActionNotDefined`, but it's built with types available during the translation of `Fragment`s where names are resolved on the normal parsing path. 
Similarly there's already existing `TypesNotDefined` and `UndeclaredEntityTypes` variants for errors. 
Another solution would be to change the underlying type of `ActionNotDefined` to support different ways of constructing it.

victornicolet: A moral justification is that there is a difference between the two: `ActionNotDefined` occurs when definitions are being resolved while building the schema, while `UndeclaredActions` occurs when some action is used in an already built schema, but it's not declared (by this schema).

cedar-policy-core/src/validator/schema.rs
```diff
@@ -1086,6 +1086,25 @@ impl ValidatorSchema {
         })
     }
 
+    /// Returns an iterator over all leaf types reachable from entity attributes, tags, and
+    /// action contexts. Leaf types are all types except `Set` and `Record`, which are
+    /// traversed into automatically.
+    fn leaf_types(&self) -> TypeIterator<'_> {
```
lianah, line 1085: If a type is referenced from multiple contexts, it will be iterated over it multiple times, right? I doubt it will be an issue since computing TC is probably the real bottleneck, but if performance for validation becomes an issue maybe this is an optimization we could do?

lianah (approved):

cedar-policy-core/src/validator/schema.rs
victornicolet: That's correct; it's easy enough to add `Itertools::unique` when calling `leaf_types` so I'll add that and a doc comment.

cedar-policy-core/src/validator/schema/err.rs
john-h-kastner-aws: The types should all be opaque to the public API, so supporting both kinds in ActionNotDefined sound reasonable

lianah: Is there a DRT opportunity between the two validations: validating an already build schema vs building the same schema from fragments

victornicolet: @john-h-kastner-aws I think I'm leaning towards keeping them separate, and adding documentation as to when the new error can occur as opposed to mixing the two error types. I didn't find an elegant way to do so.

john-h-kastner-aws (approved):

cedar-policy-core/src/validator/schema/err.rs
john-h-kastner-aws: > Is there a DRT opportunity between the two validations: validating an already build schema vs building the same schema from fragments

I think the roundtrip properties will end up testing something close to this

PR #2416, opened 2026-06-19
More protobuf regression tests

victornicolet:
Adds more protobuf regression tests consisting in decoding a previously encoded protobuf and checking it is equal to the original data that was used to encode it. 
Also includes a helper to generate such tests.

Note:
- There are only two `.rs` files in this PR.
- The protobufs that were there previously have only been *moved* and not modified.

Missing from this PR is more tests for sets of policies.


cedar-policy/tests/proto_generate_test_files.rs
john-h-kastner-aws, line 1: Could we use insta to manage updating the snapshots? I guess it would make it easier to accidentally commit a breaking change without realizing, but it'd be simpler to add new tests.

victornicolet: I guess this would mean replacing the generator of this PR by an insta test that tests the other direction (serialized input format results in same protobuf). 
I think we can do it, but doesn't feel natural because 1) it's not the direction being tested, 2) encoding is non-deterministic and 3) comparing the bytes of the encoding is mostly useless.

john-h-kastner-aws: ah, forgot about the non-determinism. IMO it could be worth it for the convenience otherwise, but this is good as is

victornicolet: Tried to think about it a little more and I didn't see a nice solution - would the insta based generation interfere with other insta tests? To update, you would also need to run insta without review (or, reviewing the snapshots is meaningless here). But maybe you had a precise idea of something that's more convenient? 

Currently, you'd only need to run`cargo test -p cedar-policy --features protobufs --test proto_generate_test_files -- --ignored` , imo the inconvenience mostly being that you need to know this command exists (it's in the doc comment).

katherine-hough (approved):

john-h-kastner-aws (approved):

PR #2418, opened 2026-06-19
Rename `nontrivial_residual_policies` to `residual_policies`; `residual_policies` to `policies` 

john-h-kastner-aws:
## Description of changes

* Deprecate `nontrivial_residual_policies`
* Change `residual_policies` to return only non-trivial residual policies
* Add `policies` matching original behavior of `residual_policies`, plus `policy_set` to get these policy collected into a `PolicySet`
* Add `get_policy` to lookup a partially evaluated policy by id, which may be useful when using one of the policy id iterators

Not attached to these exact names if there are any strong opinions, but I do think the current combination of `nontrivial_residual_policies` and `residual_policies` is bad.  

Also contain an unrelated change to document some slightly subtle invariant on `query_resource` and `query_principal`. 

### DIFF
--- a/cedar-policy-core/src/tpe.rs
+++ b/cedar-policy-core/src/tpe.rs
@@ -126,7 +126,7 @@ mod tests {
 
     use super::is_authorized;
 
-    fn collect_policy_ids<'a>(
+    pub(super) fn collect_policy_ids<'a>(
         iter: impl Iterator<Item = &'a super::response::ResidualPolicy>,
     ) -> HashSet<&'a PolicyID> {
         iter.map(|p| p.get_policy_id()).collect()
--- a/cedar-policy-core/src/tpe.rs
+++ b/cedar-policy-core/src/tpe.rs
@@ -106,11 +106,11 @@ pub fn is_authorized<'a>(
 mod tests {
     use cool_asserts::assert_matches;
 
-    use crate::ast::{Annotation, AnyId, BinaryOp, Literal, Value, ValueKind, Var};
+    use crate::ast::{Annotation, AnyId, BinaryOp, Literal, PolicyID, Value, ValueKind, Var};
     use crate::tpe::residual::{Residual, ResidualKind};
     use crate::validator::ValidatorSchema;
     use crate::{
-        ast::{Eid, EntityUID, PolicySet},
+        ast::{EntityUID, PolicySet},
         extensions::Extensions,
         parser::parse_policyset,
     };
@@ -125,6 +125,73 @@ mod tests {
     };
 
     use super::is_authorized;
+
+    fn collect_policy_ids<'a>(
+        iter: impl Iterator<Item = &'a super::response::ResidualPolicy>,
+    ) -> HashSet<&'a PolicyID> {
+        iter.map(|p| p.get_policy_id()).collect()
+    }
+
+    #[track_caller]
+    pub(super) fn get_policy_by_annotation_id<'a>(
+        ps: &'a PolicySet,
+        annotation_id: &str,
+    ) -> &'a crate::ast::Policy {
+        let id_key = AnyId::new_unchecked("id");
+        ps.static_policies()
+            .find(|p| {
+                matches!(p.annotation(&id_key), Some(Annotation { val, .. }) if val == annotation_id)
+            })
+            .unwrap()
+    }
+
+    #[track_caller]
+    pub(super) fn assert_resource_get_attr(residual: &Residual, expected_attr: &str) {
+        assert_matches!(
+            residual,
+            Residual::Partial {
+                kind: ResidualKind::GetAttr { expr, attr },
+                ..
+            } => {
+                assert_matches!(
+                    expr.as_ref(),
+                    Residual::Partial { kind: ResidualKind::Var(Var::Resource), .. }
+                );
+                assert_eq!(attr, expected_attr);
+            }
+        );
+    }
+
+    #[track_caller]
+    pub(super) fn assert_in_resource_attr(
+        residual: &Residual,
+        expected_uid: &EntityUID,
+        expected_attr: &str,
+    ) {
+        assert_matches!(
+            residual,
+            Residual::Partial {
+                kind: ResidualKind::BinaryApp { op: BinaryOp::In, arg1, arg2 },
+                ..
+            } => {
+                assert_entity_uid_literal(arg1.as_ref(), expected_uid);
+                assert_resource_get_attr(arg2.as_ref(), expected_attr);
+            }
+        );
+    }
+
+    #[track_caller]
+    pub(super) fn assert_entity_uid_literal(residual: &Residual, expected_uid: &EntityUID) {
+        assert_matches!(
+            residual,
+            Residual::Concrete {
+                value: Value { value: ValueKind::Lit(Literal::EntityUID(uid)), .. },
+                ..
+            } => {
+                assert_eq!(uid.as_ref(), expected_uid);
+            }
+        );
+    }
 
     fn rfc_policies() -> PolicySet {
         parse_policyset(
@@ -202,11 +269,8 @@ action Delete appliesTo {
 
     fn rfc_request() -> PartialRequest {
         PartialRequest {
-            principal: PartialEntityUID {
-                ty: "User".parse().unwrap(),
-                eid: Some(Eid::new("Alice")),
-            },
-            action: EntityUID::from_components("Action".parse().unwrap(), Eid::new("View"), None),
+            principal: r#"User::"Alice""#.parse::<EntityUID>().unwrap().into(),
+            action: r#"Action::"View""#.parse().unwrap(),
             resource: PartialEntityUID {
                 ty: "Document".parse().unwrap(),
                 eid: None,
@@ -219,7 +283,7 @@ action Delete appliesTo {
     }
 
     fn rfc_entities() -> PartialEntities {
-        let uid = EntityUID::from_components("User".parse().unwrap(), Eid::new("Alice"), None);
+        let uid: EntityUID = r#"User::"Alice""#.parse().unwrap();
         PartialEntities::from_entities_unchecked(
             [(
                 uid.clone(),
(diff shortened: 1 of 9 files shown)


john-h-kastner-aws:
DRT build failure is expected due to API change.

semver-checks failure is warning that adding `#[deprected]` should be on a minor because it introduces new compile warnings. Since this is an experimental API, we're fine with making the change on a patch.

It's also worth calling out that the behavior change in `residual_policies` could be breaking for some callers. If they depend on getting _all_ policies from this function, then they need to update to `policies` or risk being broken. This is also acceptable for an experimental API and is documented in the changelog

cedar-policy-core/src/tpe/response.rs
```diff
@@ -195,7 +195,7 @@ impl<'a> Response<'a> {
     }
 
     /// Get satisfied permit residual policies
-    pub fn satisfied_permits(&self) -> impl Iterator<Item = &ResidualPolicy> {
+    pub fn true_permits(&self) -> impl Iterator<Item = &ResidualPolicy> {
         #[expect(
             clippy::unwrap_used,
             reason = "we know that the policy ids are in the residuals map"
```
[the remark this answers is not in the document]
john-h-kastner-aws: I'd add this for the `cedar-policy` API, but not on `cedar-policy-core`

victornicolet (approved):

cedar-policy-core/src/tpe/response.rs
```diff
@@ -283,8 +283,8 @@ impl<'a> Response<'a> {
     }
 
     /// Look up the [`Residual`] by [`PolicyID`]
```
luxas:
```suggestion
    /// Look up the [`ResidualPolicy`] by [`PolicyID`]
```

cedar-policy-core/src/tpe/response.rs
```diff
@@ -319,15 +319,24 @@ impl<'a> Response<'a> {
...
 
-    /// Get residual policies
-    pub fn residual_policies(&self) -> impl Iterator<Item = &ResidualPolicy> {
+    /// Get all policies (including concrete true/false/error residuals)
+    pub fn policies(&self) -> impl Iterator<Item = &ResidualPolicy> {
         self.residuals.values()
     }
+
+    /// Get all policies (including concrete true/false/error residuals) as a `PolicySet`
+    pub fn poliy_set(&self) -> PolicySet {
```
luxas:
```suggestion
    pub fn policy_set(&self) -> PolicySet {
```

cedar-policy/src/api/tpe.rs
```diff
@@ -130,38 +132,56 @@ impl ResourceQueryRequest {
         .map(Self)
     }
 
-    /// Convert [`ResourceQueryRequest`] to a [`Request`] by providing the resource [`EntityId`]
+    fn principal(&self) -> EntityUid {
+        #[expect(
+            clippy::unwrap_used,
+            reason = "constructor requires concrete principal"
+        )]
+        EntityUid(self.0 .0.get_principal().try_into().unwrap())
```
luxas, line 140:
```suggestion
        EntityUid(self.0.0.get_principal().try_into().unwrap())
```
or is it deliberate?

cedar-policy/src/api/tpe.rs
```diff
@@ -526,50 +559,74 @@ impl TpeResponse<'_> {
...
+    ///
+    /// Each returned [`Policy`] inherits its [`PolicyId`](crate::PolicyId) and
+    /// annotations from the corresponding input policy. Its scope is
+    /// unconstrained and its condition is a single `when` clause containing
+    /// the residual expression.
+    ///
+    /// Call [`Policy::to_pst()`] on each result to get a [`crate::pst::Policy`]
+    /// for structured inspection. Residual expressions may contain
+    /// [`crate::pst::Expr::ResidualError`] nodes indicating subexpressions that
+    /// would error at runtime; use [`crate::pst::Expr::has_error()`] to check.
```
luxas: would certainly or could possibly error? I guess that those that did concretely error already should not be shown here, but those that certainly will error should be here, as we don't know yet which exact error it will be, and thus it is non-trivial.

cedar-policy/src/api/tpe.rs
```diff
@@ -526,50 +559,74 @@ impl TpeResponse<'_> {
             .map_err(Into::into)
     }
 
-    /// Return residuals as [`Policy`]s.
+    /// Returns an iterator of non-trivial (meaning more than just `true`
+    /// or `false`) residuals as [`Policy`]s.
```
luxas: or concrete error?

cedar-policy/src/api/tpe.rs
```diff
@@ -657,7 +712,7 @@ impl PolicySet {
         Ok(TpeResponse(res))
     }
 
-    /// Like [`Authorizer::is_authorized`] but uses an [`EntityLoader`] to load
+    /// Like [`Authorizer::is_authorized`](crate::Authorizer::is_authorized) but uses an [`EntityLoader`] to load
```
luxas: curious, didn't this render correctly before?

cedar-policy/src/api/tpe.rs
```diff
@@ -744,45 +792,38 @@ impl PolicySet {
         schema: &Schema,
     ) -> Result<impl Iterator<Item = EntityUid>, PermissionQueryError> {
         let partial_entities = PartialEntities::from_concrete(entities.clone(), schema)?;
-        let residuals = self.tpe(&request.0, &partial_entities, schema)?;
-        #[expect(
-            clippy::unwrap_used,
-            reason = "policy set construction should succeed because there shouldn't be any policy id conflicts"
-        )]
-        let policies = &Self::from_policies(
```
luxas, line 752: nice deduplication and more power to the user 👍

cedar-policy-core/src/tpe.rs
```diff
@@ -254,12 +254,12 @@ action Delete appliesTo {
             .find(|p| matches!(p.annotation(&id), Some(Annotation {val, ..}) if val == "2"))
             .unwrap();
         // resource["isPublic"]
-        assert_matches!(residuals.get_residual(policy0.id()), Some(Residual::Partial{kind: ResidualKind::GetAttr { expr, attr }, ..}) => {
+        assert_matches!(residuals.get_residual(policy0.id()).unwrap().get_residual().as_ref(), Residual::Partial{kind: ResidualKind::GetAttr { expr, attr }, ..} => {
```
luxas: this slightly stutters with `get_residual` twice, but probably not worth changing. we maybe could call the latter just `.residual().as_ref()`?

luxas (commented):
Thanks for the followup, looks nice 👍

cedar-policy/src/api/tpe.rs
john-h-kastner-aws: `cargo fmt` being a bit funny. It shows up in a couple other places

cedar-policy/src/api/tpe.rs
john-h-kastner-aws: There was an intermediate commit where I had removed the `use crate::Authorizer` from the imports, so I needed this change for the docs to build. In the current state I should be able to undo it.

cedar-policy-core/src/tpe.rs
john-h-kastner-aws: pushing some test helper to make these asserts less of a mess

cedar-policy/src/api/tpe.rs
john-h-kastner-aws: hmmm, the whole `has_error` function is questionably useful. It says only that there is some subexpression which, if evaluated, will evaluate to an error. But, the subexpression might never be evaluated, or some other expression might error regardless of what `has_error` returns.

I think it's only useful when you want to represent a residual as cedar policy text since error residuals aren't directly representable

john-h-kastner-aws: Updated comment here and on `has_error` to be clearer

luxas: yeah, it feels like this is one of the things that will get easier to reason about once I add the `possible_bool_outcomes` after my vacation 👍

cedar-policy-core/src/tpe.rs
```diff
@@ -126,6 +126,73 @@ mod tests {
...
+        let id_key = AnyId::new_unchecked("id");
+        ps.static_policies()
+            .find(|p| {
+                matches!(p.annotation(&id_key), Some(Annotation { val, .. }) if val == annotation_id)
+            })
+            .unwrap()
+    }
+
+    #[track_caller]
+    pub(super) fn assert_resource_get_attr(residual: &Residual, expected_attr: &str) {
```
luxas, line 149: nit: I felt it was quite nice to use `assert_snapshot` and `interpret_typed_str_to_str` in https://github.com/cedar-policy/cedar/pull/2323. I'm thinking to convert most TPE tests to that style down the road.

cedar-policy-core/src/tpe.rs
```diff
@@ -444,88 +504,58 @@ when { principal in resource.editors };
...
-            assert_matches!(arg1.as_ref(), Residual::Partial { kind: ResidualKind::GetAttr { expr, attr }, .. } => {
-                assert_matches!(expr.as_ref(), Residual::Partial { kind: ResidualKind::Var(Var::Resource), .. });
-                assert_eq!(attr, "owner");
-            });
-            assert_matches!(arg2.as_ref(), Residual::Concrete { value: Value { value: ValueKind::Lit(Literal::EntityUID(uid)), ..}, .. } => {
-                assert_eq!(uid.as_ref(), &EntityUID::from_components("User".parse().unwrap(), Eid::new("aaron"), None));
-            });
-        });
-        // (User::"aaron" in (resource["readers"])) || (User::"aaron" in (resource["editors"]))
-        assert_matches!(residuals.get_residual(policy2.id()).unwrap().get_residual().as_ref(), Residual::Partial { kind: ResidualKind::Or{ left, right }, .. } => {
```
luxas, line 510: yeah this was for sure a pain to read, thanks for making it better!

luxas (approved):
Great, LGTM 👍

cedar-policy-core/src/tpe.rs
john-h-kastner-aws: Yeah, insta is really nice. I'd be happy to review PRs adopting it where applicable.

Issue #2439, opened 2026-06-26
Validator::validate() reconstructs Typechecker for every policy, causing redundant O(A×P×R) work


tomlikestorock:
### Before opening, please confirm:

- [x] I have [searched for duplicate or closed issues](https://github.com/cedar-policy/cedar/issues?q=is%3Aissue+).
- [x] I have read the guide for [submitting bug reports](https://github.com/cedar-policy/cedar/blob/main/CONTRIBUTING.md#bug-reports).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Describe the bug

Validator::validate() iterates over all policy templates and calls validate_policy() for each one. validate_policy() calls typecheck_policy(), which constructs a new Typechecker via Typechecker::new(). The constructor computes the cross-product of all actions, principal types, and resource types to build request environments — an O(|actions| × |principals| × |resources|) operation.

For a PolicySet with N policies, this means the same request environments are computed N times. With large schemas (thousands of actions/entity types) and many policies, this redundant work dominates both validation time and memory usage, since each Typechecker instance holds its own copy of the request environments.

The Typechecker is stateless with respect to individual policies — it only depends on the schema and validation mode, both of which are constant across a single validate() call. It should be constructed once and reused.
The same issue applies to validate_with_level().

### Reproduction steps

```rust
use cedar_policy::{PolicySet, Schema, Validator, ValidationMode};

// 1. Build a schema with a large action × principal × resource cross-product.
//    Even a modest schema (e.g. 50 actions × 20 principal types × 20 resource types)
//    makes the per-policy cost visible.
let schema_json = serde_json::json!({
    "": {
        "entityTypes": {
            // Generate many principal/resource entity types
            "P0": { "memberOfTypes": [] },
            "P1": { "memberOfTypes": [] },
            // ... up to P19
            "R0": { "memberOfTypes": [] },
            "R1": { "memberOfTypes": [] },
            // ... up to R19
        },
        "actions": {
            // Generate many actions, each applying to all principal/resource pairs
            "action0": {
                "appliesTo": {
                    "principalTypes": ["P0", "P1", /* ... */],
                    "resourceTypes": ["R0", "R1", /* ... */],
                }
            },
            // ... up to action49
        }
    }
});

let schema = Schema::from_json_value(schema_json).unwrap();

// 2. Create a PolicySet with many policies.
let mut policies = PolicySet::new();
for i in 0..500 {
    let src = format!(
[...]

PR #2440, opened 2026-06-26
Reuse Typechecker across policies in Validator::validate() (#2439)

tomlikestorock:
## Description of changes
The Validator::validate() and Validator::validate_with_level() methods previously constructed a new Typechecker for every policy template in the PolicySet. The Typechecker::new() constructor computes the cross-product of actions × principals × resources to build request environments, which is O(|actions| × |principals| × |resources|). For large schemas with many policies, this redundant work dominates validation time and memory.

This PR constructs the Typechecker once per validate()/validate_with_level() call and shares it across all policy templates:

- Extract validate_policy_with_typechecker to accept a pre-built Typechecker
- Extract validate_policy_with_level_and_typechecker similarly for level validation
- Consolidate typecheck_policy to take a Typechecker reference directly (removing a dead wrapper)
- Add Typechecker::mode() accessor so validate_policy_with_typechecker can read the validation mode from the typechecker rather than accepting it as a separate parameter

The single-policy entry points (validate_policy, validate_policy_with_level) are preserved and construct their own Typechecker internally.

I feel like this approach invites a conversation about further refactors of hoisting typechecker even higher up to be reused even more, but since it's self referential against the validator that houses it, it would be a bit more of a larger change and I'm not sure about the appetite for that or the approach that would be desired here. So comments/rejection reasons are welcome here, happy to rework this.

## Issue #, if available
#2439 is the issue I filed for this.

### DIFF
--- a/cedar-policy-core/src/validator.rs
+++ b/cedar-policy-core/src/validator.rs
@@ -117,9 +117,13 @@ impl Validator {
     /// Validate all templates, links, and static policies in a policy set.
     /// Return a `ValidationResult`.
     pub fn validate(&self, policies: &PolicySet, mode: ValidationMode) -> ValidationResult {
+        // Create the Typechecker once and reuse across all policies.
+        // This avoids recomputing the O(actions × principals × resources)
+        // request environments for every policy template.
+        let typechecker = Typechecker::new(&self.schema, mode);
         let validate_policy_results: (Vec<_>, Vec<_>) = policies
             .all_templates()
-            .map(|p| self.validate_policy(p, mode))
+            .map(|p| self.validate_policy_with_typechecker(p, &typechecker))
             .unzip();
         let template_and_static_policy_errs = validate_policy_results.0.into_iter().flatten();
         let template_and_static_policy_warnings = validate_policy_results.1.into_iter().flatten();
@@ -144,9 +148,13 @@ impl Validator {
         mode: ValidationMode,
         max_deref_level: u32,
     ) -> ValidationResult {
+        // Create the Typechecker once and reuse across all policies.
+        let typechecker = Typechecker::new(&self.schema, mode);
         let validate_policy_results: (Vec<_>, Vec<_>) = policies
             .all_templates()
-            .map(|p| self.validate_policy_with_level(p, mode, max_deref_level))
+            .map(|p| {
+                self.validate_policy_with_level_and_typechecker(p, max_deref_level, &typechecker)
+            })
             .unzip();
         let template_and_static_policy_errs = validate_policy_results.0.into_iter().flatten();
         let template_and_static_policy_warnings = validate_policy_results.1.into_iter().flatten();
@@ -164,6 +172,7 @@ impl Validator {
     /// Run all validations against a single static policy or template (note
     /// that Core `Template` includes static policies as well), gathering all
     /// validation errors and warnings in the returned iterators.
+    #[cfg(test)]
     fn validate_policy<'a>(
         &'a self,
         p: &'a Template,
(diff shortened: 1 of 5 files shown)


cedar-policy-core/src/validator.rs
```diff
@@ -235,20 +224,44 @@ impl Validator {
...
-    /// the argument static policy or template (note that Core `Template`
-    /// includes static policies as well) in the context of the schema for this
-    /// validator. Any detected type errors are wrapped and returned as
-    /// `ValidationErrorKind`s.
-    fn typecheck_policy<'a>(
+    /// Like [`validate_policy`] but reuses an existing [`Typechecker`] instead
+    /// of constructing a new one for each policy. This avoids recomputing
+    /// request environments when validating multiple policies against the same
+    /// schema.
+    fn validate_policy_with_typechecker<'a>(
```
john-h-kastner-aws, line 231: Can this drop the `self` parameter? It should still be able to get the schema out of the typechecker where it needs that.

cedar-policy-core/src/validator.rs
```diff
@@ -235,20 +224,44 @@ impl Validator {
...
+            )
+        }
+        .into_iter()
+        .flatten();
+        let (errors, warnings) = self.typecheck_policy_with_typechecker(typechecker, p);
+        (validation_errors.chain(errors), warnings)
+    }
+
+    /// Use a [`Typechecker`] to detect type errors in a policy.
+    fn typecheck_policy_with_typechecker<'a>(
```
john-h-kastner-aws, line 257: This could definitely drop `self`

cedar-policy-core/src/validator.rs
```diff
@@ -164,6 +172,7 @@ impl Validator {
     /// Run all validations against a single static policy or template (note
     /// that Core `Template` includes static policies as well), gathering all
     /// validation errors and warnings in the returned iterators.
+    #[cfg(test)]
```
john-h-kastner-aws, line 175: probably should comment why this is `#[cfg(test)]`

cedar-policy-core/src/validator/level_validate.rs
```diff
@@ -68,29 +69,81 @@ impl Validator {
...
-        let (errors, warnings) = self.validate_policy(p, mode);
-
         let typechecker = Typechecker::new(&self.schema, mode);
-        let type_annotated_asts = typechecker.typecheck_by_request_env(p);
+        self.validate_policy_with_level_and_typechecker(p, max_deref_level, &typechecker)
+    }
+
+    /// Like [`validate_policy_with_level`] but reuses an existing
+    /// [`Typechecker`] instead of constructing a new one for each policy.
+    pub(super) fn validate_policy_with_level_and_typechecker<'a>(
```
john-h-kastner-aws, line 78: Where's the extra code coming from here?

john-h-kastner-aws:
This approach seems reasonable enough merge, but I agree that there's a larger refactoring to do here. If you're up for it, I say go for the whole things, but I'm also fine taking this optimization and leaving an issue for the rest.

cedar-policy-core/src/validator.rs
victornicolet: Would it also make sense to rewrite the doc comment to document what this function does instead of leaving it on `validate_policy`, which just calls this function? Since this is where the logic lives now.

cedar-policy-core/src/validator.rs
```diff
@@ -235,20 +224,44 @@ impl Validator {
...
         &'a self,
+        p: &'a Template,
+        typechecker: &Typechecker<'a>,
+    ) -> (
+        impl Iterator<Item = ValidationError> + 'a,
+        impl Iterator<Item = ValidationWarning> + 'a,
+    ) {
+        let mode = typechecker.mode();
+        let validation_errors = if mode.is_partial() {
+            None
```
victornicolet, line 241: I think it'd be nice to leave the comment that was in the `validate_policy` function here, same for `else` branch below.

cedar-policy-core/src/validator/level_validate.rs
victornicolet: Should this extra code be replaced by a call to `validate_policy_with_typechecker` ?

PR #2448, opened 2026-07-02
Rewrite TPE evaluator tests to use `insta` inline snapshots

john-h-kastner-aws:
Instead of manually constructing expression and using `assert_matches` on residuals, we can parse expression and then assert on the string representation of the residual. This makes the tests substantially more compact, and much easier to update and review for future changes to the algorithm. 

Total LOC doesn't change much because I've also add new tests where I felt coverage with incomplete. 

Given the size of this diff I'm mostly asking you to trust that I've not been deleting tests and just review for maintainability and effectiveness of the new tests. Rather than reviewing the diff it's probably easier to just treat the whole test module as new code and review that. 

### DIFF
--- a/cedar-policy-core/src/tpe/evaluator.rs
+++ b/cedar-policy-core/src/tpe/evaluator.rs
@@ -728,12 +728,12 @@ mod tests {
             extensions: Extensions::all_available(),
         };
         let interpret_typed_str_to_str = |e| interpret_typed_str_to_str(&eval, e, &schema());
-        // principal -> principal because its eid is unknown
+        // principal -> User::"foo"
         assert_snapshot!(
             interpret_typed_str_to_str("principal"),
             @r#"User::"foo""#
         );
-        // resource -> E::""
+        // resource -> resource because its eid is unknown
         assert_snapshot!(
             interpret_typed_str_to_str("resource"),
             @"resource"
--- a/cedar-policy-core/src/tpe/evaluator.rs
+++ b/cedar-policy-core/src/tpe/evaluator.rs
@@ -644,61 +644,99 @@ fn normalize_ext_value_inner(value: &Value) -> Option<Value> {
 #[cfg(test)]
 mod tests {
     use std::collections::{BTreeMap, HashSet};
+    use std::str::FromStr;
 
-    use crate::ast::{Expr, SlotEnv, UnwrapInfallible};
+    use crate::ast::{Expr, SlotEnv};
+    use crate::parser::parse_expr;
     use crate::tpe::err::ExprToResidualError;
+    use crate::validator::typecheck::Typechecker;
     use crate::validator::types::Type;
+    use crate::validator::{ValidationMode, ValidatorSchema};
     use crate::{
-        ast::{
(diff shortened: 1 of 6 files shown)


cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -912,1392 +915,1326 @@ mod tests {
...
+
+        // `E` entities can't have tags, but `hasTag` is still well-typed. These could all reduce to
+        // `false`, but only the explicit empty tag case does atm.
+        assert_snapshot!(
+            interpret_typed_str_to_str(r#"E::"empty_tags".hasTag("s") && E::"empty_tags".getTag("s") == "bar" "#),
+            @"false"
+        );
+        assert_snapshot!(
+            interpret_typed_str_to_str(r#"E::"none_tags".hasTag("s") && E::"none_tags".getTag("s") == "bar" "#),
+            @r#"E::"none_tags".hasTag("s")"#
```
john-h-kastner-aws, line 1923: This is a new test case I added, and the current output doesn't make sense to me. `E` entities can't have tags, so even without explicitly saying the tags are empty, we could reduce this to false. We don't, which is imprecise but fine, but then we somehow eliminate the whole right operand to the `&&`, which doesn't make sense.

This reduction is what I'd expect if we had an RHS reducing to `true`, but that shouldn't be the case here, so I don't see how we get to this residual.

The original expression and the residual are both always `false`, so this is sound as-is, but I don't see how we get to this exact residual.

luxas:
Thanks for this PR! I'm now on vacation, so I'll take a look first when I'm back. If you need to get it merged faster, other maintainers can check it out, otherwise I'm happy to take a look later.

victornicolet (approved):
Definitely more compact and readable.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -912,1392 +915,1377 @@ mod tests {
...
-            Residual::Concrete {
-                value: Value {
-                    value: ValueKind::Lit(Literal::Long(2)),
-                    ..
-                },
-                ..
-            }
+        assert_snapshot!(
+            interpret_typed_str_to_str(r#"if (resource == User::"alice") then Document::"A" else Document::"B""#),
+            @r#"if (resource == User::"alice") then Document::"B" else Document::"B""#
```
john-h-kastner-aws, line 951: This is residual odd, but it happens as a result of how typechecking work. It short-circuits typechecking the `then` branch, but populates it with the expression from the `else` branch, so the residual we have here is actually the input to TPE coming from the typechecker.

The typechecker can't reduce this to `Document::"B"` itself because this would break other downstream algorithms (level checking and entity manifests, at least), but TPE should be able to conclude that the `==` between entities of different types is always `false` to get this reduction. I haven't thought this through fully, but it likely runs into some subtly with error semantics which we can solve using the current `can_error` function, or there may be a more precises solution using the "possible boolean values" idea.

whonore (approved):

victornicolet (approved):

PR #2453, opened 2026-07-07
Simplify `in` set with unknown ancestors

john-h-kastner-aws:
## Description of changes

Minor improvement to TPE for `in` set, allowing simplification of `User::"alice" in [Group::"admins",User::"alice"]` when ancestors of alice are unknown. 

Not expecting this to be useful anywhere, but the early return path just looked wrong in the code.

Lean at https://github.com/cedar-policy/cedar-spec/pull/980

I'll add a test case covering the changed behavior once #2448 is merged 

### DIFF
--- a/cedar-policy-core/src/tpe/evaluator.rs
+++ b/cedar-policy-core/src/tpe/evaluator.rs
@@ -281,10 +281,22 @@ impl Evaluator<'_> {
                                 if let Ok(uid2) = v2.get_as_entity() {
                                     if uid1 == uid2 {
                                         return mk_concrete(true.into());
+<<<<<<< HEAD
                                     } else if let Some(entity) = self.entities.get(uid1) {
                                         if let Some(ancestors) = entity.ancestors() {
                                             return mk_concrete(ancestors.contains(uid2).into());
                                         }
+||||||| parent of 4d4582f9 (Simplify `in` set with unknown ancestors)
+                                    } else if let Some(entity) = self.entities.get(uid1) {
+                                        if let Some(ancestors) = &entity.ancestors {
+                                            return mk_concrete(ancestors.contains(uid2).into());
+                                        }
+=======
+                                    } else if let Some(ancestors) =
+                                        self.entities.get_ancestors(uid1)
+                                    {
+                                        return mk_concrete(ancestors.contains(uid2).into());
+>>>>>>> 4d4582f9 (Simplify `in` set with unknown ancestors)
                                     }
                                     binapp_residual(arg1, arg2)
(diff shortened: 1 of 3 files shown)


cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -1807,6 +1803,81 @@ mod tests {
...
+            @r#"true"#
+        );
+        assert_snapshot!(
+            interpret_typed_str_to_str(r#"E::"undefined" in [E::"other", E::"undefined"]"#),
+            @r#"true"#
+        );
+
+        assert_snapshot!(
+            interpret_typed_str_to_str(r#"E::"undefined" in [E::"undefined", context.e]"#),
+            @r#"E::"undefined" in [E::"undefined", context.e]"#
```
victornicolet, line 1826: Why does that not reduce to `true`?

john-h-kastner-aws: All binary operators evaluate to a residual if either operand is a residual.

Possible room for further reduction in the future, but it would need to be careful to handle error semantics correctly.

whonore (approved):

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -293,22 +293,18 @@ impl Evaluator<'_> {
...
+                                        if uids.contains(&uid1)
+                                            || ancestors.is_some_and(|ancestors| {
+                                                uids.iter().any(|uid2| ancestors.contains(uid2))
+                                            })
+                                        {
+                                            mk_concrete(true.into())
+                                        } else if !uids.is_empty() && ancestors.is_none() {
+                                            binapp_residual(arg1, arg2)
+                                        } else {
+                                            mk_concrete(false.into())
```
victornicolet, line 306: This logic is a little hard to read through maybe a few comments on what holds in each branch would help (especially the last branch, which I think is "ancestors is none and uids empty or ancestors doesn't intersect with uids"?). 
Alternatively rewriting it like:
```
if uids.contains(&uid1) {
    mk_concrete(true.into())
} else if uids.is_empty() {
    // x in [] is always false
    mk_concrete(false.into())
} else if let Some(ancestors) = ancestors {
    // Known ancestors -> always concrete result from intersection check
    mk_concrete(uids.iter().any(|uid2| ancestors.contains(uid2)).into())
} else {
    // Unknown ancestors case
    binapp_residual(arg1, arg2)
}
```
more conditions but requires less thinking about the negation of the previous conditions.

victornicolet (approved):

PR #2455, opened 2026-07-07
validator: cache unlinked request envs (#2439) + stream typecheck_policy fold

### LINKED ISSUE #2439: Validator::validate() reconstructs Typechecker for every policy, causing redundant O(A×P×R) work

tomlikestorock:
## Description of changes
Two related changes that together reduce validator peak memory (and redundant work) when validating against schemas with large action sets, especially under concurrent load.

1. Cache the schema's unlinked request-env set (fixes #2439)
Validator::validate() reconstructed a Typechecker per policy, and each one recomputed the schema's full set of unlinked request environments — the redundant O(A×P×R) work described in #2439. The env set is now computed once and cached on ValidatorSchema, then reused across policies and requests. This removes the per-call, per-thread reconstruction of the env-set Vec.

2. Stream the typecheck_policy fold
typecheck_policy previously collected every request env's typed PolicyCheck into a vec (via typecheck_by_request_env) before folding to (all_false, all_succ). It now iterates the cached envs, typechecks each, folds its PolicyCheck into the accumulator, and drops it (and its typed condition) before computing the next. Peak live memory is now O(one condition) rather than O(envs × condition).

Sourcing the streamed loop from the cached env set also avoids a per-env inner collect that the cache change would otherwise introduce on its own.

Benchmark (production schema, 2816 actions; one reused Validator; peak = max simultaneously-live bytes; 8 threads × 8 calls, shared Validator):

## Issue #, if available
Fixes #2439.

### DIFF
--- a/cedar-policy-core/src/validator/typecheck.rs
+++ b/cedar-policy-core/src/validator/typecheck.rs
@@ -104,7 +104,7 @@ impl<'a> Typechecker<'a> {
         let mut all_false = true;
         let mut all_succ = true;
         for unlinked_e in self.schema.unlinked_request_envs(self.mode) {
-            for linked_e in self.link_request_env(&unlinked_e, t) {
+            for linked_e in self.link_request_env(unlinked_e, t) {
                 match self.single_env_typechecking(&linked_e, t.id(), &cond) {
                     PolicyCheck::Success(_) => all_false = false,
                     PolicyCheck::Irrelevant(err, _) => {
@@ -144,7 +144,18 @@ impl<'a> Typechecker<'a> {
         &'b self,
         t: &'b Template,
     ) -> Vec<(RequestEnv<'b>, PolicyCheck)> {
-        self.apply_typecheck_fn_by_request_env(t, |request_env, policy_id, expr| {
+        self.typecheck_by_request_env_iter(t).collect()
+    }
+
+    /// Like [`Self::typecheck_by_request_env`] but yields the per-environment
+    /// results lazily instead of collecting them into a `Vec`. Crate-internal
+    /// callers that consume the results in a single pass should prefer this to
+    /// avoid holding every environment's typed condition live at once.
+    pub(crate) fn typecheck_by_request_env_iter<'b>(
+        &'b self,
+        t: &'b Template,
+    ) -> impl Iterator<Item = (RequestEnv<'b>, PolicyCheck)> + 'b {
+        self.apply_typecheck_fn_by_request_env(t, move |request_env, policy_id, expr| {
             self.single_env_typechecking(request_env, policy_id, expr)
         })
     }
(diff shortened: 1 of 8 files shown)


cedar-policy-core/src/validator/typecheck.rs
```diff
@@ -214,17 +215,21 @@ impl<'a> Typechecker<'a> {
...
-                    let check = typecheck_fn(&linked_e, t.id(), &cond);
-                    (linked_e, check)
-                })
-            })
-            .collect()
+        // for the corresponding action. Use nested loops with a single output
+        // Vec: `link_request_env` yields envs borrowing the schema (not the
+        // local `unlinked_e`), so they are consumed immediately while
+        // `unlinked_e` is in scope, avoiding a per-env intermediate Vec.
+        let mut results = Vec::new();
```
john-h-kastner-aws: Did you try returning an iterator here instead of collecting? I think I originally collected because of some lifetime issues arising from the borrowed request envs, but maybe we can use the owned variant now and avoid the allocation

tomlikestorock: Not in this section if I recall correctly, but I'm trying it out now to see what happens.

tomlikestorock: Done in e1c331db. Made a new internal (pub crate) method that does the lazy evaluation that the public facing version that returns a vec wraps and collects over, and is now used here.

cedar-policy/CHANGELOG.md
```diff
@@ -31,6 +31,8 @@ Starting with version 3.2.4, changes marked with a star (*) are _language breaki
 - For the experimental `tpe` feature, `TpeResponse::residual_policies` is updated to return only _non-trivial_ residuals and
   `TpeResponse::nontrivial_residual_policies` is deprecated. The previous behavior (iterating all residuals including trivial ones)
   is available via `TpeResponse::policies`.
+- The validator now computes the schema's set of unlinked request environments once and caches it on the `ValidatorSchema`, reusing it across policies and requests instead of reconstructing it for every policy. This reduces redundant work and per-request allocation when validating large schemas. (#2439)
```
john-h-kastner-aws: These entries should refer to the `cedar-policy` public API types (`Schema` and `Validator`). We also don't usually refer to implementation details in the changelog.

cedar-policy-core/src/validator/typecheck.rs
john-h-kastner-aws, line 146: I think I'd prefer to make the iter variant the only version (breaking the cedar-policy-core API is fine)

cedar-policy-core/src/validator/typecheck.rs
```diff
@@ -200,64 +212,71 @@ impl<'a> Typechecker<'a> {
...
-        // Validate each (principal, resource) pair with the substituted policy
-        // for the corresponding action.
-        self.unlinked_envs
-            .iter()
-            .flat_map(|unlinked_e| {
-                self.link_request_env(unlinked_e, t).map(|linked_e| {
+        // `Arc` so the returned iterator owns the condition and fn (cloned per
+        // env) rather than borrowing a local, letting it stream lazily.
+        let cond = Arc::new(t.condition());
+        let typecheck_fn = Arc::new(typecheck_fn);
```
john-h-kastner-aws: `cond` make sense, but I'm surprised we need this trick for `typecheck_fn`

john-h-kastner-aws (approved):
looks pretty good

cedar-policy/CHANGELOG.md
tomlikestorock: I've updated the comment to mention the speed up users will get, I can remove even that if you like.

cedar-policy-core/src/validator/typecheck.rs
tomlikestorock: Oh! Done! Thanks for the guidance here, I was worried about changing the api here.

cedar-policy-core/src/validator/typecheck.rs
tomlikestorock: Right, made it Arc-less, with a type guard.

john-h-kastner-aws (approved):

victornicolet (approved):
Looks good to me.
Did you mean to include some benchmark data in your PR description?
>Benchmark (production schema, 2816 actions; one reused Validator; peak = max simultaneously-live bytes; 8 threads × 8 calls, shared Validator):

tomlikestorock:
> Looks good to me. Did you mean to include some benchmark data in your PR description?
> 
> > Benchmark (production schema, 2816 actions; one reused Validator; peak = max simultaneously-live bytes; 8 threads × 8 calls, shared Validator):

Oh, I was trying to paste in the metrics to represent the improvement, but it got borked and hit submit. I thought it would be interesting to see the differences. For the record we were seeing massive mallocs over and over on a scale that was oomkilling our policy validation pods after gigabytes were getting alloc'd and tossed over and over and over. With this streaming version we're seeing a massive reduction in memory usage in my benchmark, from gigabytes to small megs instead, which is great.

PR #2479, opened 2026-07-21
make proto encode faillible on message depth

victornicolet:
## Description of changes

Makes the protobuf experimental feature's `encode` API faillible based on encoded message depth. This prevents problems where you could encode a structure successfully and then encounter an error while decoding it.

The depth check is on the `models::` type rather than API type because it is easier to predict whether decode will fail at that level.

The changes are:
- a new trait `EncodeCheck` in `traits.rs`
- implementations of the trait in `traits.rs` for all the `models::` types,
- tests in `api.rs` testing the depth bounds, and importantly, that decode doesn't panic under the boundary

Matching PR for fuzzing targets:
https://github.com/cedar-policy/cedar-spec/pull/987

Screenshot 2026-07-21 at 5 30 50 PM

### DIFF
--- a/cedar-policy/src/proto/traits.rs
+++ b/cedar-policy/src/proto/traits.rs
@@ -48,6 +48,46 @@ impl From<ProtobufConversionError> for DecodeError {
     }
 }
 
+/// Error type for protobuf encoding failures
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+pub enum EncodeError {
+    /// The input data contains too many recursion levels to be encoded
+    #[error("data structure depth exceeds maximum encodable depth of {MAX_ENCODE_DEPTH}")]
+    MaxDepthExceeded,
+    /// The protobuf message could not be encoded (e.g., buffer too small)
+    #[error(transparent)]
+    Proto(#[from] prost::EncodeError),
+}
+
+/// Maximum allowed protobuf recursion depth for encoding.
+///
+/// Prost's decoder has a hardcoded recursion limit of 100, where each nested
+/// message entry counts as one level. We use this limit when calculating the
+/// depth of a `prost::Message` (i.e. the `models::...`), not the depth of
+/// syntax tree (e.g. type or expression).
+/// Typically, the depth of a model is twice the one of the internal representation's
+/// tree.
+///
+/// We set this to 90 to leave a small margin (10 levels) for outer wrappers
+/// like `TemplateBody` or `Entity` that contain an `Expr` field.
+pub const MAX_ENCODE_DEPTH: usize = 90;
+
+/// A trait for protobuf model types that require validation before encoding.
+///
+/// Model types implement this to perform structural checks (such as depth limits)
+/// before the protobuf encoding step. Types that need no pre-encode validation
+/// implement this as a no-op returning `Ok(())`.
+pub trait EncodeCheck {
+    /// Validate that this model is safe to encode.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`EncodeError`] if the model violates encoding constraints
+    /// (e.g., exceeds the maximum nesting depth).
+    fn check_for_encode(&self) -> Result<(), EncodeError>;
+}
+
 /// A trait for objects that have a `try_validate` method returning `self` if the object is
 /// valid.
 pub trait TryValidate: Sized {
@@ -79,7 +119,12 @@ mod private {
 /// Trait allowing serializing and deserializing in protobuf format.
 pub trait Protobuf: Sized + TryValidate + private::Sealed {
     /// Encode into protobuf format. Returns a freshly-allocated buffer containing binary data.
-    fn encode(&self) -> Vec<u8>;
+    ///
+    /// # Errors
+    ///
+    /// Returns [`EncodeError::MaxDepthExceeded`] if the data structure has too many
+    /// recursion levels to be safely encoded and decoded by prost.
+    fn encode(&self) -> Result<Vec<u8>, EncodeError>;
     /// Decode the binary data in `buf`, producing something of type `Self`
     ///
     /// # Errors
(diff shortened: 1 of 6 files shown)


cedar-policy/src/proto/traits.rs
```diff
@@ -79,7 +119,13 @@ mod private {
 /// Trait allowing serializing and deserializing in protobuf format.
 pub trait Protobuf: Sized + TryValidate + private::Sealed {
     /// Encode into protobuf format. Returns a freshly-allocated buffer containing binary data.
-    fn encode(&self) -> Vec<u8>;
+    ///
+    /// # Errors
+    ///
+    /// Returns [`EncodeError::MaxDepthExceeded`] if the data structure has too many
+    /// recursion levels to be safely encoded and decoded by prost.
+    fn encode(&self) -> Result<Vec<u8>, EncodeError>;
```
victornicolet, line 138: This is the main change to the API of the experimental protobufs feature.

cedar-policy/src/proto/traits.rs
```diff
@@ -48,6 +48,57 @@ impl From<ProtobufConversionError> for DecodeError {
...
+/// Maximum allowed protobuf recursion depth for encoding.
+///
+/// Prost's decoder has a hardcoded recursion limit of 100, where each nested
+/// message entry counts as one level. We use this limit when calculating the
+/// depth of a `prost::Message` (i.e. the `models::...`), not the depth of
+/// syntax tree (e.g. type or expression).
+/// Typically, the depth of a model is twice the one of the internal representation's
+/// tree.
+///
+/// We set this to 90 to leave a small margin (10 levels) for outer wrappers
```
katherine-hough, line 72: Is this enough of a margin?

cedar-policy/src/proto/traits.rs
```diff
@@ -108,23 +165,55 @@ pub trait Protobuf: Sized + TryValidate + private::Sealed {
...
+/// (e.g., expression depth exceeds [`MAX_ENCODE_DEPTH`]) or the
+/// user-provided buffer does not have enough capacity.
 #[expect(
     dead_code,
     reason = "experimental feature, we might have use for this one in the future"
 )]
-pub(crate) fn encode<M: prost::Message>(
-    thing: impl Into<M>,
+#[expect(
+    clippy::multiple_bound_locations,
```
katherine-hough: Wouldn't removing the inline bound and changing the where to
`M: prost::Message + EncodeCheck + for<'a> From<&'a T>` fix these lints?

cedar-policy/src/proto/traits.rs
```diff
@@ -223,3 +316,241 @@ impl TryValidate for api::EntityNamespace {
...
+        }
+        Ok(())
+    }
+}
+
+impl EncodeCheck for models::Entity {
+    fn check_for_encode_from_depth(&self, init: usize) -> Result<(), EncodeError> {
+        // Validate all attribute and tag expressions
+        for expr in self.attrs.values().chain(self.tags.values()) {
+            expr.check_for_encode_from_depth(init + 2)?;
```
katherine-hough, line 456: Why is this +2?

katherine-hough:
It seems like this is coupled implementation specifics of prost. Is this the case?

cedar-policy/src/proto/traits.rs
victornicolet: Yes it would, I guess removing the lint is better than trying to have a `where`

cedar-policy/src/proto/traits.rs
victornicolet: +1 for entering recursing into the map, and then +1 for recursing into the container of a map entry.

victornicolet: For a list, it will be just +1 for entering the list container, but the list entry is the object itself (no wrapper for key-value pair).

cedar-policy/src/proto/traits.rs
victornicolet: I think so, but we might adjust before making this stable. We are already accounting for most wrappers, this is an additional margin. I can add a test that shows and expression that is slightly deeper than what the check accepts is still encodes then decodes.

victornicolet:
This is coupled to prost's implementation,and also the trait is on the types that are generated by prost. A change in how prost recurses through the structures would cause issues for this, but tests and fuzzing should catch those.
When we stabilize this, we'll need to be cautious about the promise we make about roundtrip, and we may want to increase the margin.

katherine-hough:
> This is coupled to prost's implementation,and also the trait is on the types that are generated by prost. A change in how prost recurses through the structures would cause issues for this, but tests and fuzzing should catch those. When we stabilize this, we'll need to be cautious about the promise we make about roundtrip, and we may want to increase the margin.

What happen if the prost change happens in a patch version? The change could then breaked already released version of Cedar. Do we need to lock the prost version?

victornicolet:
> What happen if the prost change happens in a patch version? The change could then breaked already released version of Cedar. Do we need to lock the prost version?

I would hope that such changes will not happen, otherwise any prost user that encoded something close to the recursion limit, and that previously decoded, would suddenly see decoding fail when hitting the limit. 

But if we want to have better guarantees, then yes we can do this.

cedar-policy/src/proto/traits.rs
```diff
@@ -223,3 +303,250 @@ impl TryValidate for api::EntityNamespace {
...
+                        if let Some(ref e) = get.expr {
+                            stack.push((e, child_depth));
+                        }
+                    }
+                    ExprKind::HasAttr(has) => {
+                        if let Some(ref e) = has.expr {
+                            stack.push((e, child_depth));
+                        }
+                    }
+                    ExprKind::Like(like) => {
```
john-h-kastner-aws, line 407: The `like` pattern is kind of complicated, so potentially adds meaningful depth. Do we have enough of a buffer to handle that? Maybe here and elsewhere (`is`, possibly) we should require some constant depth remaining based on the encoding structure.

john-h-kastner-aws (approved):

katherine-hough (approved):

cedar-policy/src/proto/traits.rs
victornicolet: I'll add some checks to be more precise, but those patterns should add only 1-3 levels of recursion.

katherine-hough (approved):

PR #2484, opened 2026-07-23
Added script to automate minor version releases

katherine-hough:
## Description of changes
Add a utility script to automate bumping versions and updating changelogs for minor version releases.

Commit 2: shows the change generated when this script is run.


scripts/new_minor.sh
```diff
@@ -0,0 +1,44 @@
...
+#
+# Performs pre-release steps for a release in which a minor version
+# of core, symcc, and cli are performed together.
+#
+# Must be run from the root directory of the cedar repository.
+
+set -euo pipefail
+
+# Bump crate versions.
+(yes || true) | cargo release version minor \
```
john-h-kastner-aws, line 15: whats `|| true` doing here?

scripts/new_minor.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# Update cedar-policy/src/test/test.rs to fill in the new crate version:
+# ```
+#     #[test]
+#     fn test_sdk_version() {
+#         assert_eq!(get_sdk_version().to_string(), "<CEDAR_POLICY_CRATE_VERSION>");
+#     }
+# ```
+sed -i -E \
+    "s|(get_sdk_version\(\)\.to_string\(\), \")[^\"]*(\")|\1${cedar_policy_version}\2|" \
```
john-h-kastner-aws: sed nits:

* I think it might be nicer to single quote the string to avoid some escaping, and then close/reopen the quotes around the `cedar_policy_version` var.
* IIRC sed doesn't error on no matches, if the regex becomes wrong it'll be annoying to diagnose

john-h-kastner-aws (approved):

scripts/new_minor.sh
katherine-hough: When cargo release exits it closes the read end of the pipe, yes will then try to write into a closed channel causing it to fail. Since `set -o pipefail` is set this will cause the whole pipeline to fail and the script to exit. `|| true` ignores the failure from yes

victornicolet (approved):

PR #2487, opened 2026-07-24
typecheck: cache env-independent subtrees across request envs

tomlikestorock:
## Description of changes
Caches the typechecking result of environment-independent sub-expressions across request environments when validating a policy. Behavior-preserving optimization internal to cedar-policy-core's validator (continuation of #2439).

A policy condition is typechecked once per request environment in the principal × action × resource cross-product. An up-front tree walk classifies each condition node: a node whose subtree references the request environment (Var, Slot, or Unknown) is environment-dependent and re-typechecked in every environment; every other node produces the same type in every environment, so it's typechecked once, cached, and reused across the remaining environments instead of being re-typechecked from scratch.

I don't believe that this requires spec or DRT changes since this is about an efficiency gain in processing without changing how things are processed procedurally.

I'm open to refactors or any design considerations you may have on this one. Local benchmarks show policies with many static values (e.g. hundreds of IP CIDRs) validate much faster against multi-environment schemas, since constant subtrees are typechecked once rather than per environment.

There could be a version of this that is more intertwined with the typecheck pass, but that mixed the caching determination with the typechecker code, which is sensitive to edit imo.

## Issue #, if available
This is a continuation of #2439 

### DIFF
--- a/cedar-policy-core/src/ast/expr.rs
+++ b/cedar-policy-core/src/ast/expr.rs
@@ -300,6 +300,40 @@ impl<T> Expr<T> {
         matches!(&self.expr_kind, ExprKind::Slot(_))
     }
 
+    /// Whether this node *directly* reads the request environment — i.e. it is
+    /// a `Var`, a template `Slot`, or an `Unknown`. Such a node can typecheck
+    /// to a different type in each request environment.
+    ///
+    /// A node whose whole subtree contains no such leaf is *env-independent*:
+    /// it typechecks identically in every environment, so the validator
+    /// typechecks it once and reuses the result across the request-env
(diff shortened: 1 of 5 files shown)


john-h-kastner-aws:
I think there's an easy optimization we should take from this.

The extension constructor parsing check is relativly expensive and we know it can never depend on the environment. We can split that out into a separate pass that gets called once instead of for every environment. This should be basically a free optimization, and it doesn't raise any of the correctness concerns that broader memoization might.

john-h-kastner-aws:
Actually, the unfortunate thing there is that it would be less permissive when the constructor is in a dead branch. So, a breaking change we'd have to wait on, but one i'd like to make eventually.

Issue #2122, opened 2026-02-03
Add `Residual.possible_bool_outcomes()` method


luxas:
### Category

User level API features/changes

### Describe the feature you'd like to request

As discussed in https://github.com/cedar-policy/cedar/issues/1895#issuecomment-3720043450

Add a function `Residual::possible_bool_outcomes(&self) -> Option<HashSet<PossibleBoolOutcome>>`, where the `PossibleBoolOutcome `enum has `True`, `False`, `Error` variants.

`None` is returned if the residual does not return a boolean, or alternatively we just name this "possible outcomes", and include `NonBool` as one of the variants.

This would mean that even for those Residuals whose LHS is not error-free (and thus automatically couldn't be folded to true or false), possible_bool_outcomes for residuals would the following:

- never-true `resource.whatever + 1 == 2 && false` => `HashSet::new(False, Error)`,
- never-false `resource.whatever + 1 == 2 || true` => `HashSet::new(True, Error)`,
- error-free expressions like `resource in Group::"foo"` => `HashSet::new(True, False)`, and
- arbitrary expressions like `resource.whatever + 1 == 2` => `HashSet::new(True, False, Error)`.

If `Residual::possible_bool_outcomes` returns only a single item, it is known to be a concrete value (true, false, error). The `HashSet` can never be empty.

This functionality then allows advanced users to filter never-true policies, if the specific use-case of the caller anyways would ignore any error that would arise from concrete evaluation.

I'm planning to submit a PR for this as discussed with @john-h-kastner-aws in #2091 

### Describe alternatives you've considered

Keep this just downstream, however, this seems to be a commonly-occurring request, based on the number of people that have asked why TPE doesn't fold `<possibly-erroring-residual> && false` to `false`

### Is this something that you'd be interested in working on?

- [x] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

luxas:
I'm planning to implement this, after https://github.com/cedar-policy/cedar/pull/2162 is merged

PR #2503, opened 2026-07-29
TPE: reduce has/hasTag/==/in/is using schema type information

### LINKED ISSUE #2500: TPE Fails to Eliminate Irrelevant Policies

### Describe the bug

When the typechecker determines that a policy is irrelevant, it still gets added to the residual set

https://github.com/cedar-policy/cedar/blob/main/cedar-policy-core/src/tpe.rs#L69

I'm not entirely sure why this would be, given we've already proven that the policy is irrelevant? Is it something to do with error propagation? 

### Reproduction steps

```rust
use cedar_policy::{
    Entities, EntityTypeName, EntityUid, PartialEntities, PartialEntityUid, PartialRequest, Policy,
    PolicyId, PolicySet, Schema,
};
use std::str::FromStr;

fn main() {
    let (schema, _) = Schema::from_cedarschema_str(
        r#"
        entity Group;
        entity Role;
        entity User in [Group];
        entity Service;
        action "AssumeRole" appliesTo {
          principal: [User, Service],
          resource: Role,
        };
        "#,
    )
    .unwrap();

    let policies = PolicySet::from_policies([
        Policy::parse(
            Some(PolicyId::new("eq_service")),
            r#"permit(principal == Service::"Automation", action, resource);"#,
        )
        .unwrap(),
        Policy::parse(
            Some(PolicyId::new("in_group")),
            r#"permit(principal, action, resource) when { principal in Group::"1" };"#,
        )
        .unwrap(),
    ])
    .unwrap();

    let entities = PartialEntities::from_concrete(Entities::empty(), &schema).unwrap();
    let action = EntityUid::from_str(r#"Action::"AssumeRole""#).unwrap();

    for ty in ["User", "Service"] {
        let request = PartialRequest::new(
            PartialEntityUid::new(EntityTypeName::from_str(ty).unwrap(), None),
            action.clone(),
            PartialEntityUid::new(EntityTypeName::from_str("Role").unwrap(), None),
            None,
            &schema,
        )
        .unwrap();

        let response = policies.tpe(&request, &entities, &schema).unwrap();
        println!("=== principal: {ty}  decision: {:?}", response.decision());
        for p in response.nontrivial_residual_policies() {
            println!("{}", p.to_cedar().unwrap());
        }
    }
}
```

Both policies are returned for both, even though only a user can be in a group

john-h-kastner-aws:
## Description of changes

Fixes #2500 

Lean model and proofs in https://github.com/cedar-policy/cedar-spec/pull/997

### DIFF
--- a/cedar-policy-core/src/tpe/evaluator.rs
+++ b/cedar-policy-core/src/tpe/evaluator.rs
@@ -19,11 +19,13 @@
 use std::{collections::BTreeMap, sync::Arc};
 
 use crate::{
-    ast::{self, BinaryOp, EntityUID, PartialValue, Set, Value, ValueKind, Var},
+    ast::{self, BinaryOp, EntityType, EntityUID, PartialValue, Set, Value, ValueKind, Var},
     evaluator::stack_size_check,
     extensions::Extensions,
 };
 
+use crate::validator::types::Type;
+use crate::validator::ValidatorSchema;
 use crate::{
     tpe::entities::PartialEntities,
     tpe::request::PartialRequest,
@@ -35,6 +37,7 @@ use crate::{
 pub struct Evaluator<'e> {
     pub(crate) request: &'e PartialRequest,
     pub(crate) entities: &'e PartialEntities,
+    pub(crate) schema: &'e ValidatorSchema,
     pub(crate) extensions: &'e Extensions<'e>,
 }
 
@@ -209,19 +212,14 @@ impl Evaluator<'_> {
             }
             ResidualKind::Is { expr, entity_type } => {
                 let expr = self.interpret(expr);
+                if let Some(v) = try_decide_is_residual(&expr, entity_type) {
+                    return mk_concrete(v);
+                }
                 match &expr {
                     Residual::Concrete { value, .. } => match value.get_as_entity() {
                         Ok(uid) => mk_concrete((uid.entity_type() == entity_type).into()),
                         Err(_) => mk_error(), // <error> is <entity_type> => <error>
                     },
-                    Residual::Partial {
-                        kind: ResidualKind::Var(Var::Principal),
-                        ..
-                    } => mk_concrete((entity_type == self.request.principal_type()).into()),
-                    Residual::Partial {
-                        kind: ResidualKind::Var(Var::Resource),
-                        ..
-                    } => mk_concrete((entity_type == self.request.resource_type()).into()),
                     Residual::Partial { .. } => mk_residual(ResidualKind::Is {
                         expr: Arc::new(expr),
                         entity_type: entity_type.clone(),
@@ -246,6 +244,9 @@ impl Evaluator<'_> {
             ResidualKind::BinaryApp { op, arg1, arg2 } => {
                 let arg1 = self.interpret(arg1);
                 let arg2 = self.interpret(arg2);
+                if let Some(v) = try_decide_binary_residual(self.schema, *op, &arg1, &arg2) {
+                    return mk_concrete(v);
+                }
                 let binapp_residual = |arg1, arg2| {
                     mk_residual(ResidualKind::BinaryApp {
                         op: *op,
@@ -391,6 +392,9 @@ impl Evaluator<'_> {
             }
             ResidualKind::HasAttr { expr, attr } => {
                 let expr = self.interpret(expr);
+                if let Some(v) = try_decide_has_residual(self.schema, &expr, attr) {
+                    return mk_concrete(v);
+                }
                 match &expr {
                     Residual::Concrete { value, .. } => {
                         if let Ok(r) = value.get_as_record() {
@@ -501,6 +505,68 @@ impl Evaluator<'_> {
             }
         }
     }
+}
+
+/// Given an `is` expression applied to a residual, reduce it to a value if
+/// possible, using the residual's type.
+fn try_decide_is_residual(expr: &Residual, entity_type: &EntityType) -> Option<Value> {
+    if expr.can_error_assuming_well_formed() {
+        return None;
+    }
+    let lub = expr.ty().as_entity_lub()?;
+    if !lub.contains_entity_type(entity_type) {
+        // None of the possible entity types so, `is` is `false`.
+        Some(false.into())
+    } else if lub.get_single_entity() == Some(entity_type) {
+        // The only possible entity type, so `is` is `true`.
+        Some(true.into())
+    } else {
+        // Not possible in strict validation
+        None
+    }
+}
+
+/// Given a binary operation applied to residuals, reduce it to a value if
+/// possible, using the residual types and `schema`.
+fn try_decide_binary_residual(
+    schema: &ValidatorSchema,
+    op: BinaryOp,
+    arg1: &Residual,
+    arg2: &Residual,
+) -> Option<Value> {
+    if arg1.can_error_assuming_well_formed() || arg2.can_error_assuming_well_formed() {
+        return None;
+    }
+    let must_be_false = match (op, arg1.ty(), arg2.ty()) {
+        // `in` must be false if `arg1` cannot have an ancestor with the type of `arg2`
+        (BinaryOp::In, ty1, ty2) => match (ty1.as_entity_lub(), ty2.as_set_or_entity_lub()) {
+            (Some(lhs_lub), Some(rhs_lub)) => !schema.any_descendent_of(lhs_lub, rhs_lub),
+            _ => return None,
+        },
+        (BinaryOp::Eq, ty1, ty2) => Type::are_types_disjoint(ty1, ty2),
+        (BinaryOp::HasTag, ty1, Type::String) => {
+            Type::has_declared_entity_types(schema, ty1) && !Type::may_have_tags(schema, ty1)
+        }
+        _ => return None,
+    };
+    must_be_false.then(|| false.into())
+}
+
+/// Given a `has` expression applied to a residual, reduce it to a value if
+/// possible, using only the residual's type and `schema`.
+fn try_decide_has_residual(schema: &ValidatorSchema, expr: &Residual, attr: &str) -> Option<Value> {
+    if expr.can_error_assuming_well_formed() {
+        return None;
+    }
+    if !matches!(expr.ty(), Type::Record { .. } | Type::Entity(_)) {
+        return None;
+    }
+    // The residual can't error and cannot have `attr`, so `has` is always `false`.
+    // We can't have an analogous reduction to `true` because the concrete semantics
+    // for `has` is `false` when the entity isn't present.
+    (Type::has_declared_entity_types(schema, expr.ty())
+        && !Type::may_have_attr(schema, expr.ty(), attr))
+    .then(|| false.into())
 }
 
 /// If the value is an extension value, rebuild the
@@ -666,6 +732,7 @@ mod tests {
         let eval = Evaluator {
             request: &concrete_user_req(),
             entities: &PartialEntities::new(),
+            schema: &schema(),
             extensions: Extensions::all_available(),
         };
         let interpret_typed_str_to_str = |e| interpret_typed_str_to_str(&eval, e, &schema());
@@ -709,6 +776,7 @@ mod tests {
             )
             .unwrap(),
             entities: &PartialEntities::new(),
+            schema: &schema,
             extensions: Extensions::all_available(),
         };
         let interpret_typed_str_to_str = |e| interpret_typed_str_to_str(&eval, e, &schema);
@@ -732,6 +800,7 @@ mod tests {
         let eval = Evaluator {
             request: &concrete_user_req(),
             entities: &PartialEntities::new(),
+            schema: &schema(),
             extensions: Extensions::all_available(),
         };
         let interpret_typed_str_to_str = |e| interpret_typed_str_to_str(&eval, e, &schema());
@@ -804,6 +873,7 @@ mod tests {
         let eval = Evaluator {
             request: &concrete_user_req(),
             entities: &PartialEntities::new(),
+            schema: &schema(),
             extensions: Extensions::all_available(),
         };
         let interpret_typed_str_to_str = |e| interpret_typed_str_to_str(&eval, e, &schema());
@@ -876,6 +946,7 @@ mod tests {
         let eval = Evaluator {
             request: &concrete_user_req(),
             entities: &PartialEntities::new(),
+            schema: &schema(),
             extensions: Extensions::all_available(),
         };
         let interpret_typed_str_to_str = |e| interpret_typed_str_to_str(&eval, e, &schema());
@@ -888,8 +959,12 @@ mod tests {
             @"2"
         );
         assert_snapshot!(
+            interpret_typed_str_to_str(r#"if (resource == Document::"C") then Document::"A" else Document::"B""#),
+            @r#"if resource == Document::"C" then Document::"A" else Document::"B""#
+        );
+        assert_snapshot!(
             interpret_typed_str_to_str(r#"if (resource == User::"alice") then Document::"A" else Document::"B""#),
-            @r#"if resource == User::"alice" then Document::"B" else Document::"B""#
+            @r#"Document::"B""#
         );
         assert_snapshot!(
             interpret_typed_str_to_str(&r#"if (9223372036854775807 * 2) == 0 then resource else Document::"A""#),
@@ -929,6 +1004,7 @@ mod tests {
         let eval = Evaluator {
             request: &req,
             entities: &PartialEntities::new(),
+            schema: &schema,
             extensions: Extensions::all_available(),
         };
         let interpret_typed_str_to_str = |e| interpret_typed_str_to_str(&eval, e, &schema);
@@ -956,6 +1032,14 @@ mod tests {
         assert_snapshot!(
             interpret_typed_str_to_str("principal.baz is Document"),
             @"principal.baz is Document"
+        );
+        assert_snapshot!(
+            interpret_typed_str_to_str(r#"(if principal == User::"alice" then User::"bob" else User::"jane") is Document"#),
+            @"false"
+        );
+        assert_snapshot!(
+            interpret_typed_str_to_str(r#"(if principal == User::"alice" then User::"bob" else User::"jane") is User"#),
+            @"true"
         );
         assert_snapshot!(
             interpret_typed_str_to_str("User::\"alice\" is User"),
@@ -987,6 +1071,7 @@ mod tests {
         let eval = Evaluator {
             request: &req,
             entities: &PartialEntities::new(),
+            schema: &schema,
             extensions: Extensions::all_available(),
         };
         let interpret_typed_str_to_str = |e| interpret_typed_str_to_str(&eval, e, &schema);
@@ -1029,6 +1114,7 @@ mod tests {
         let eval = Evaluator {
             request: &req,
             entities: &PartialEntities::new(),
+            schema: &schema,
             extensions: Extensions::all_available(),
         };
         let interpret_typed_str_to_str = |e| interpret_typed_str_to_str(&eval, e, &schema);
@@ -1083,6 +1169,7 @@ mod tests {
         let eval = Evaluator {
             request: &req,
             entities: &PartialEntities::new(),
+            schema: &schema,
             extensions: Extensions::all_available(),
         };
         let interpret_typed_str_to_str = |e| interpret_typed_str_to_str(&eval, e, &schema);
@@ -1137,6 +1224,7 @@ mod tests {
         let eval = Evaluator {
             request: &req,
             entities: &entities,
+            schema: &schema,
             extensions: Extensions::all_available(),
         };
         let interpret_typed_str_to_str = |e| interpret_typed_str_to_str(&eval, e, &schema);
@@ -1204,6 +1292,7 @@ mod tests {
         let eval = Evaluator {
             request: &req,
             entities: &entities,
+            schema: &schema,
             extensions: Extensions::all_available(),
         };
         let interpret_typed_str_to_str = |e| interpret_typed_str_to_str(&eval, e, &schema);
@@ -1264,6 +1353,7 @@ mod tests {
         let eval = Evaluator {
             request: &req,
             entities: &entities,
+            schema: &schema,
             extensions: Extensions::all_available(),
         };
         let interpret_typed_str_to_str = |e| interpret_typed_str_to_str(&eval, e, &schema);
@@ -1281,7 +1371,7 @@ mod tests {
         );
         assert_snapshot!(
             interpret_typed_str_to_str(r#"principal has other"#),
-            @"principal has other"
+            @"false"
         );
         assert_snapshot!(
             interpret_typed_str_to_str(r#"E::"f" has s"#),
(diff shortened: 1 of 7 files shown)


cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -209,19 +212,20 @@ impl Evaluator<'_> {
             }
             ResidualKind::Is { expr, entity_type } => {
                 let expr = self.interpret(expr);
+                if let Some(expr_ety) = expr
+                    .ty()
+                    .as_entity_lub()
+                    .and_then(|ety| ety.get_single_entity())
+                {
+                    if !expr.can_error_assuming_well_formed() {
```
luxas: After a second thought, I realized that maybe what we care about most cases is "is error free", so I'll probably negate the method in a bit. IIRC, the lean model had the "is error free" variant as well.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -246,6 +250,26 @@ impl Evaluator<'_> {
             ResidualKind::BinaryApp { op, arg1, arg2 } => {
                 let arg1 = self.interpret(arg1);
                 let arg2 = self.interpret(arg2);
+                let must_be_false = match op {
```
luxas: This starts looking a lot like the abstract interpretation I was planning in #2122

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -246,6 +250,26 @@ impl Evaluator<'_> {
...
+                            (Some(lhs_lub), Some(rhs_lub)) => {
+                                !self.schema.any_descendent_of(lhs_lub, rhs_lub)
+                            }
+                            _ => false,
+                        }
+                    }
+                    _ => false,
+                };
+                if must_be_false
+                    && !arg1.can_error_assuming_well_formed()
```
luxas: One optimization that comes to mind here, would be to not interpret arg1 and arg2, but just check their `can_error_assuming_well_formed` in the original form, to potentially save some time.

It came to mind now, so I'll ponder about the soundness a bit more, but especially if I rename this to `is_error_free_assuming_well_typed` or similar, I believe it should be such that if the residual was error-free before interpretation (which is what we care about), then it necessarily also has to be error-free after interpretation (and thus interpretation is unnecessary if we can deduce this). 

In other words, `is_error_free_assuming_well_typed` should be monotonic. This would be a pretty good theorem to add to the Lean code, unless we already have it.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -391,6 +415,14 @@ impl Evaluator<'_> {
             }
             ResidualKind::HasAttr { expr, attr } => {
                 let expr = self.interpret(expr);
+                if !Type::may_have_attr(self.schema, expr.ty(), attr)
```
luxas: Same here, I need to take a closer look at what guarantees the Lean model gives us now (or could be extended to give us), but it'd be nice with a theorem that says the type of the residual is constant during interpretation, and thus could we maybe move interpretation later down.

Or right now, I believe the type is constant during evaluation, but is there a possibility of type refinement as evaluation proceeds? I see that the validator's strict mode that is used in TPE disallows the following:
```
(if principal.foo then User::"A" else Document::"B").str == "foo"
```
even though both `User` and `Document` would have the `str` attribute.

Reading `EntityKind::least_upper_bound` it seems like during strict mode, both types must be equal, and thus the union is a single-member set. Or alternatively, one might get `AnyEntity` from the type of `principal` and `resource`, but that doesn't happen in TPE as those are fully-qualified. Do you know of a case in strict mode where the LUB could contain multiple values?

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -391,6 +415,14 @@ impl Evaluator<'_> {
...
+                    && !expr.can_error_assuming_well_formed()
+                {
+                    // The residual can't error and cannot have `attr`, so `has` is alway `false`.
```
luxas:
```suggestion
                    // The residual can't error and cannot have `attr`, so `has` is always `false`.
```

cedar-policy-core/src/validator/types.rs
```diff
@@ -464,6 +464,49 @@ impl Type {
...
+            _ => None,
+        }
+    }
+
+    /// Get `self` as a specific entity type, or, if `self` is a set, then the
+    /// element type as an entity type.  Returns `None` if `self` is neither an
+    /// entity or set of entity, and for `AnyEntity`.
+    pub(crate) fn as_set_or_entity_lub(&self) -> Option<&EntityLUB> {
+        match self {
+            Type::Entity(_) => self.as_entity_lub(),
```
luxas: nit: I had kind of expected either:
```
Type::Entity(entity_kind) => entity_kind.as_entity_lub(),
```
here, like the explicit, duplicated form that was in `typecheck.rs` before, or
```
            Type::Set {
                element_type: Some(element_type),
            } => element_type.as_entity_lub(),
            _ => self.as_entity_lub(),
```
using the function as a default case, just adding the extra set semantics here.

(doesn't matter really, but just noting that I had to read it twice to make sure I understood it correctly)

cedar-policy-core/src/validator/typecheck.rs
```diff
@@ -1740,32 +1740,6 @@ impl<'a> SingleEnvTypechecker<'a> {
         }
     }
 
-    /// Checks if `lhs_ety` may be a descendant of `rhs_ety` in the action hierarchy.
-    /// We assume that `lhs_ety` is an action entity type, but `rhs_ety` can be any entity type.
-    /// Lean counterpart: <https://github.com/cedar-policy/cedar-spec/blob/7e231a68b0e0eb1b8ce1362e81de4568671a668a/cedar-lean/Cedar/Validation/Types.lean#L202>
-    fn check_action_in_entity_type(&self, lhs_ety: &EntityType, rhs_ety: &EntityType) -> bool {
-        lhs_ety == rhs_ety
```
luxas, line 1747: Was omission of this case above in `maybe_action_descendent_of` intentional?
For 
```
entity Document;
entity User;
action "write" appliesTo {
    principal: User,
    resource: Document
};
action "create" in ["write"] appliesTo {
    principal: User,
    resource: Document
};
```
and
```
permit (principal, action in Action::"create", resource)
unless { action in Action::"create" };
```
the validator today correctly concludes that the policy is impossible. Is that the case after this omission as well?

Is it the case that `self.schema.get_entity_types_in(rhs).contains(&lhs)` is covering action types too? `ValidatorActionId::descendants` doesn't add `self`, `get_entity_types_in` does add `rhs` to the returned list.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -891,9 +928,13 @@ mod tests {
             interpret_typed_str_to_str("if false then resource else 2"),
             @"2"
         );
+        assert_snapshot!(
+            interpret_typed_str_to_str(r#"if (resource == Document::"C") then Document::"A" else Document::"B""#),
+            @r#"if (resource == Document::"C") then Document::"A" else Document::"B""#
+        );
         assert_snapshot!(
             interpret_typed_str_to_str(r#"if (resource == User::"alice") then Document::"A" else Document::"B""#),
-            @r#"if (resource == User::"alice") then Document::"B" else Document::"B""#
```
luxas: Wait, how did the pre-existing code here turn `Document::"A"` -> `Document::"B"` in the then branch during interpretation? 🤔

Very interesting, after some debugging I found that `SingleEnvTypechecker::typecheck` deduces that `test_expr` is `False`, and thus assigns `then_expr'=else_expr` and `else_expr'=else_expr`. This was for me unexpected behavior, but I guess it makes sense in a way. Was the reasoning that if the original exprs were kept, the one which isn't necessary could be ill-typed as the validator short-circuited, and thus pruned?

Do you know if the typechecker does other similar rewrites we should track and tell users about as well?

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -209,19 +212,20 @@ impl Evaluator<'_> {
             }
             ResidualKind::Is { expr, entity_type } => {
                 let expr = self.interpret(expr);
+                if let Some(expr_ety) = expr
+                    .ty()
+                    .as_entity_lub()
+                    .and_then(|ety| ety.get_single_entity())
```
luxas: I was gonna ask about whether it makes sense to write this logic as:
- whenever expr is error free AND
  - if all of the possibly entity types in the LUB match `entity_type` => `true` (which in practice due to it being a set indeed means, "if there is only one element and it matches `entity_type`"), or
  - if none of the possibly entity types in the LUB match `entity_type` => `false`, or
  - otherwise undecided

However, looking more into the LUB in strict mode, I don't see yet how it could contain multiple values (as discussed below), so then it all folds into what you have here indeed.

cedar-policy-core/src/tpe/evaluator.rs
luxas, line 405: Add a comment here to follow-up on resolving the presence or non-presence of required attributes from the schema?

luxas (commented):
Thanks for the PR! 

I should have time pretty soon to resume the `possible_bool_outcomes` feature and I could also rename `can_error_assuming_well_formed` as we discussed earlier.
I'll take a look at the cedar-spec change next.

cedar-policy-core/src/tpe/evaluator.rs
john-h-kastner-aws: There's a trade off here: partial evaluation should very often give us back an error free expression where we didn't have one before. So, by checking on the residual we get to apply the reduction in more cases.

I suppose we could check it before and after reduction if that's a meaningful optimization.

john-h-kastner-aws: The monotonicity theorem, I think there would be something new to prove

We have (omitting welltypedness hypothesis)  

* `∀ r req es, r.ErrorFree → (r.evaluate req es).isOk`
* `∀ r req es, (r.evaluate req es).toOption = ((TPE.evaluate r preq pes).evaluate req es).toOption`

Given `r.ErrorFree`, we know `∀ r req es, ((TPE.evaluate r preq pes).evaluate req es).isOk`. But `ErrorFree` isn't complete, so we'd need to show that `ErrorFree` actually holds for it.

Really, what we care about is the `isOk` result. That's what makes the reduction sound, so I don't think adding this theorem is particularly useful, but it could be fun if you want to try it out.

cedar-policy-core/src/tpe/evaluator.rs
john-h-kastner-aws: Yes, I plan to remove the permissive validation mode at some point, replacing it with a simpler policy linter. It's been an experimental feature forever and we don't plan to stabilize it. 

This will simplify some type handling code here and in symcc

cedar-policy-core/src/tpe/evaluator.rs
john-h-kastner-aws: > I believe the type is constant during evaluation

Correct, this is one of the important TPE lemmas we need to the full correctness proof

cedar-policy-core/src/tpe/evaluator.rs
john-h-kastner-aws: Yeah, it's kind of odd. The typechecker decides the equality cannot be true, so it doesn't check the `then` branch. But, it needs something with the same type as the `else` branch to put there if we want it to construct a well typed expression. The easiest answer _is_ the else branch. 

Some other things that don't work:

* Use a junk value like `1`. Well typed per the typechecker, but downstream analysis erases singleton bool types, so isn't well-typed according to that stricter definition
* Reduce to `Document::"B"` immediately. The guard might error, so it's semantically different
* If we synthesis an arbitrary, expression of the type, we might introduce a entity UID that doesn't exist in the original policy, messing with things like batched evaluation 

So in the end this is kind of the best option (though I'm open to better ideas, because it is odd). 


The two other places to look at are `&&` and `||` since those also short circuit. E.g., the type annotated AST for `x1 && x2` given `x1 => (tx1 : False)` is `tx1`, dropping the `&&` and `x2` entirely.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -1999,23 +2152,23 @@ mod tests {
...
-        // `E` entities can't have tags, but `hasTag` is still well-typed. These could all reduce to
-        // `false`, but only the explicit empty tag case does atm.
+        // `E` declares no tags in the schema, so `hasTag` on any `E` operand reduces to `false`
+        // regardless of whether concrete tag data is present.
         assert_snapshot!(
             interpret_typed_str_to_str(r#"E::"empty_tags".hasTag("s") && E::"empty_tags".getTag("s") == "bar" "#),
             @"false"
         );
         assert_snapshot!(
             interpret_typed_str_to_str(r#"E::"none_tags".hasTag("s") && E::"none_tags".getTag("s") == "bar" "#),
```
john-h-kastner-aws, line 2310: @luxas This demonstrates the `&&` case for typechecking I commented on above.

`E::"none_tags".hasTag("s")` has type `False`, so we skip the RHS. The typechecker won't reduce it to `false` (due to error semantics, but also that's probably not the expected behavior for a typecheker), so we're left with the LHS, dropping the `&& RHS`

cedar-policy-core/src/validator/typecheck.rs
john-h-kastner-aws: I'll look closer. In moving these functions around I've also aligned their implementation to be closer to the Lean model's corresponding functions, but I can check if there's a behavior change here.

cedar-policy-core/src/validator/types.rs
john-h-kastner-aws: Your second option looks nicest to me

cedar-policy-core/src/tpe/evaluator.rs
luxas: #2509

cedar-policy-core/src/tpe/evaluator.rs
john-h-kastner-aws: I think the difference is that here we determine that it must be a single value, allowing us to reduce to a concrete value instead of a residual. For #2122, we'll still be carrying around a residual so we can materialize the specific value of we need to

cedar-policy-core/src/validator/typecheck.rs
john-h-kastner-aws: Confirmed: the Lean model doesn't have an explicit type equality check in it's `maybe_action_descendent_of` either. This is covered by `self.schema.get_entity_types_in(rhs)` including `rhs`.

cedar-policy-core/src/tpe/evaluator.rs
luxas: Right, I was mostly thinking that `residual.possible_bool_outcomes() == [false]` is a generalization of `residual.must_be_false()` which this effectively is. I'll see how closely it fits when I have time to start working on `possible_bool_outcomes` hopefully within the coming month or so.

cedar-policy-core/src/tpe/evaluator.rs
luxas: Thinking more about it, the cases where we'd actually get into the `must_be_false` branch are the cases where the previous residual was very short, so yeah, probably there wouldn't be much of a speedup.

Now that I looked at the existing proofs, the monotonicity is probably subsumed by the other theorems indeed, and we don't need it when we don't need that optimization for now at least. But yeah, maybe I'll try it out for fun at some point 😄

cedar-policy-core/src/tpe/evaluator.rs
luxas: That would be a nice cleanup indeed, should we make an issue to track it?

cedar-policy-core/src/tpe/evaluator.rs
luxas: Today I learned, thanks 😄

cedar-policy-core/src/tpe/evaluator.rs
john-h-kastner-aws: Not sure what exactly you're asking for here. There's a related comment above I think.

cedar-policy-core/src/tpe/evaluator.rs
luxas: https://github.com/cedar-policy/cedar/issues/2528

cedar-policy-core/src/tpe/evaluator.rs
luxas: I meant the dual of `<error-free> has requiredattr` ==> `true` when the arg is error-free and the attr is required. If the entity wouldn't have the attribute on construction, the validator would fail before reaching interpretation, right?

Edit: I saw afterwards that `<nonexistent> has requiredattr` is `false` by the concrete semantics, didn't know that, so then this is disregarded.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -1847,6 +1993,162 @@ mod tests {
...
+            parse_partial_euid("App::User"),
+            r#"App::Action::"view""#.parse().unwrap(),
+            parse_partial_euid("App::Doc"),
+            None,
+            &schema,
+        )
+        .unwrap();
+        let eval = Evaluator {
+            request: &req,
+            entities: &PartialEntities::new(),
```
luxas, line 2121: Now that we anyways take the schema when doing TPE, can we just always consume the schema here and thus always be able to reduce these?

luxas (approved):
Thanks! Seems like we might need to re-run the integration tests.

cedar-policy-core/src/tpe/evaluator.rs
john-h-kastner-aws: Reading concrete actions from the schema felt like the wrong thing to be doing in the evaluator. The current state is that any `PartialEntities` constructor that takes a schema will populate actions for you, giving us the reduction without extra code in the evaluator. You can still construct an empty `PartialEntities`, and so this test exercises that behavior.

cedar-policy-core/src/validator/schema.rs
```diff
@@ -1054,6 +1057,37 @@ impl ValidatorSchema {
         descendants
     }
 
+    /// Checks if an action of entity type `lhs_ety` may be a descendant of an
+    /// action of entity type `rhs_ety` in the action hierarchy.
+    /// Lean counterpart: <https://github.com/cedar-policy/cedar-spec/blob/7e231a68b0e0eb1b8ce1362e81de4568671a668a/cedar-lean/Cedar/Validation/Types.lean#L202>
+    fn maybe_action_descendent_of(&self, lhs_ety: &EntityType, rhs_ety: &EntityType) -> bool {
```
victornicolet, line 1063: nit: maybe the function should just be called `action_descendant_of`, but that may just me reading maybe functions as option-returning functions (the maybe reads when in the context of the evaluation, but just in the context of the schema I don't get why it's a maybe).

cedar-policy-core/src/validator/types.rs
```diff
@@ -464,6 +464,66 @@ impl Type {
...
+        match &self {
+            Type::Entity(entity_kind) => entity_kind.as_entity_lub(),
+            _ => None,
+        }
+    }
+
+    /// Get `self` as a specific entity type, or, if `self` is a set, then the
+    /// element type as an entity type.  Returns `None` if `self` is neither an
+    /// entity or set of entity, and for `AnyEntity`.
+    pub(crate) fn as_set_or_entity_lub(&self) -> Option<&EntityLUB> {
```
victornicolet, line 479: nit: maybe `as_self_or_element_entity_lub` since it return self or the set element's entity lub.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -1847,6 +1993,162 @@ mod tests {
...
+            schema: &schema,
+            extensions: Extensions::all_available(),
+        };
+        let interpret_typed_str_to_str = |e| interpret_typed_str_to_str(&eval, e, &schema);
+
+        // We could reduce to a value in these because the schema includes action ancestors, but
+        // calling a constructor other then `PartialEntities::new` will add them and just give us
+        // the concrete evaluation path, so there's no reason to implement that case. Here we show
+        // that it's not decidable at the type level because `App::Action` can be in an `Groups::Action`.
+        assert_snapshot!(
```
victornicolet, line 2068: I wonder how users will understand this; since the action hierarchy is provided as part of the schema (and so in the evaluator and the partial request) you'd expect it to be used whenever the information from the schema is used, so at the type level. 

I'm not proposing anything actionable here, just remarking that this may be surprising, and I understand it's justified.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -503,6 +507,68 @@ impl Evaluator<'_> {
...
+    if arg1.can_error_assuming_well_formed() || arg2.can_error_assuming_well_formed() {
+        return None;
+    }
+    let must_be_false = match (op, arg1.ty(), arg2.ty()) {
+        // `in` must be false if `arg1` cannot have an ancestor with the type of `arg2`
+        (BinaryOp::In, ty1, ty2) => match (ty1.as_entity_lub(), ty2.as_set_or_entity_lub()) {
+            (Some(lhs_lub), Some(rhs_lub)) => !schema.any_descendent_of(lhs_lub, rhs_lub),
+            _ => return None,
+        },
+        (BinaryOp::Eq, ty1, ty2) => Type::are_types_disjoint(ty1, ty2),
```
victornicolet, line 546: Just curious, any unsoundness in the type disjointness definition in the equivalent Lean would manifest in the proofs? The comment on the function definition in Rust is very defensive.

victornicolet (approved):

cedar-policy-core/src/tpe/evaluator.rs
john-h-kastner-aws: It might be worth changing `PartialEntities::new` in the public API to require a schema. Then it can populate it the actions, making this case unreachable from the public API.

cedar-policy-core/src/tpe/evaluator.rs
john-h-kastner-aws: Right, we this decides types are disjoint when they're not, some term will get type `false` when that not sound, manifesting in the soundness proofs. Similarly in the TPE proofs, though those need typechecking soundness in the first place.

The confusing thing here is Rust's more expressive types, which it would be nice to untangle from the ain type system to better match Lean. See https://github.com/cedar-policy/cedar/issues/2528

cedar-policy-core/src/tpe/evaluator.rs
luxas: Right, I meant always taking a schema in the `PartialEntities` constructors to make this case unreachable, this could be a nice invariant. But could be done in a followup too.

cedar-policy-core/src/tpe/evaluator.rs
```diff
@@ -1359,6 +1455,18 @@ mod tests {
...
         );
+        assert_snapshot!(
+            interpret_typed_str_to_str(r#"resource has hat.origin.other"#),
+            @"false"
+        );
+        assert_snapshot!(
+            interpret_typed_str_to_str(r#"principal has other.name"#),
+            @"false"
+        );
+        assert_snapshot!(
```
john-h-kastner-aws, line 1466: A further extension could simplify this to `false`, but I want to get this PR merged and implement further extension later

victornicolet (approved):

PR #2510, opened 2026-08-09
Pretty expressive formatter

shaobo-he:
## Why `pretty-expressive`?

  `pretty-expressive` is a better fit for Cedar because formatting policies involves semantic constraints, not just
  choosing whether a group fits on one line. Its `full` primitive can express that no code may follow a `//` comment on
  the same line, while `reset` preserves the original indentation of multiline strings and entity identifiers. Its cost-
  based layout selection and arbitrary alternatives also provide a foundation for improving difficult cases such as long
  attribute chains and keeping method names attached to their argument lists.

  The migration uses a small compatibility layer so Cedar’s existing document builders and formatting style remain
  largely unchanged. That layer also guards comment constraints during grouping and preserves the old printer’s newline-
  indentation behavior. As a result, Cedar gains a more expressive layout engine without coupling comment attachment to
  rendering or requiring an immediate formatter rewrite.

### DIFF
--- a/cedar-policy-formatter/src/pprint/doc.rs
+++ b/cedar-policy-formatter/src/pprint/doc.rs
@@ -265,35 +265,70 @@ impl Doc for Node<Option<And>> {
     }
 }
 
+/// Return whether this additive expression is a member chain containing a
+/// function call, without any surrounding arithmetic or unary operator.
+fn is_member_call_expression(expression: &Node<Option<Add>>) -> bool {
+    let Some(addition) = expression.as_inner() else {
+        return false;
+    };
+    if !addition.extended.is_empty() {
+        return false;
+    }
+    let Some(multiplication) = addition.initial.as_inner() else {
+        return false;
+    };
+    if !multiplication.extended.is_empty() {
+        return false;
+    }
+    let Some(unary) = multiplication.initial.as_inner() else {
+        return false;
+    };
+    if unary.op.is_some() {
+        return false;
+    }
+    let Some(member) = unary.item.as_inner() else {
+        return false;
+    };
+    member
+        .access
+        .iter()
+        .any(|access| matches!(access.as_inner(), Some(MemAccess::Call(_))))
+}
+
 impl Doc for Node<Option<Relation>> {
     fn to_doc<'src>(&self, context: &mut Context<'_, 'src>) -> Option<RcDoc<'src>> {
         let e = self.as_inner()?;
         match e {
-            Relation::Common { initial, extended } => {
-                match extended.as_slice() {
-                    [] => initial.to_doc(context),
-                    [(op, n)] =>
-                    // do not group due to the limitation of current design
-                    {
-                        Some(
-                            initial
-                                .to_doc(context)?
-                                .append(RcDoc::space())
-                                .append(add_comment(
-                                    RcDoc::as_string(op),
-                                    get_comment_after_end(
-                                        initial.loc.as_ref().map(|loc| loc.span),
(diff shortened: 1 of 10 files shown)


shaobo-he:
CI failure is actually an improvement in formatting.

After,

```cedar
// Forbid Bedtime Access to Kid Profile
@id("forbid-bedtime-watch-kid-profile")
forbid (
  principal is Subscriber,
  action == Action::"watch",
  resource
)
when { principal.profile.isKid }
unless
{
  // `toTime()` returns the duration modulo one day (i.e., it ignores the "date"
  // component). Here, we use it to calculate the subscriber's local time and
  // compare the result against durations that represent 6:00AM and 9:00PM.
  duration("6h") <=
    context.now.datetime.offset(context.now.localTimeOffset).toTime() &&
  context.now.datetime.offset(context.now.localTimeOffset).toTime() <=
    duration("21h")
};
```

Before,
```
// Forbid Bedtime Access to Kid Profile
@id("forbid-bedtime-watch-kid-profile")
forbid (
  principal is Subscriber,
  action == Action::"watch",
  resource
)
when { principal.profile.isKid }
unless
{
  // `toTime()` returns the duration modulo one day (i.e., it ignores the "date"
  // component). Here, we use it to calculate the subscriber's local time and
  // compare the result against durations that represent 6:00AM and 9:00PM.
  duration("6h") <= context.now
    .datetime
    .offset
    (
      context.now.localTimeOffset
    )
    .toTime
    () &&
  context.now.datetime.offset(context.now.localTimeOffset).toTime() <= duration(
      "21h"
    )
};
```

cedar-policy-formatter/src/pprint/doc.rs
```diff
@@ -265,35 +265,70 @@ impl Doc for Node<Option<And>> {
...
+    let Some(multiplication) = addition.initial.as_inner() else {
+        return false;
+    };
+    if !multiplication.extended.is_empty() {
+        return false;
+    }
+    let Some(unary) = multiplication.initial.as_inner() else {
+        return false;
+    };
+    if unary.op.is_some() {
```
john-h-kastner-aws, line 286: nit: can this fold into the prior pattern?

cedar-policy-formatter/src/pprint/fmt.rs
```diff
@@ -189,6 +203,135 @@ mod tests {
...
+            assert_eq!(
+                policies_str_to_pretty(&formatted, &config).unwrap(),
+                expected,
+                "layout is not idempotent at width {line_width}"
+            );
+        }
+    }
+
+    #[test]
+    fn test_comparison_breaks_before_either_operand_shape() {
```
john-h-kastner-aws, line 279: I think these two tests could be `insta` snapshots (prior tests look like they're doing more). Should just need to add new cedar files in the test data dir.

john-h-kastner-aws (commented):
Seems reasonable, but I don't think I understand the benefits of the new engine.

Also looking at the crate, it's new and we'd be the pretty much the first users of it, which isn't terrible given that it's a small crate, but worth mentioning.

shaobo-he:
> Seems reasonable, but I don't think I understand the benefits of the new engine.

The new formatter is supposedly more expressive than `pretty`. The latter can't handle comments very smoothly. That being said, I discovered serious performance regressions using this crate. Close this PR until we have a good answer to the regression.

john-h-kastner-aws:
> The latter can't handle comments very smoothly.

Ah, so could we remove a lot of our comment handling code? That would be nice

PR #2521, opened 2026-08-18
Updated schema validation for protobuf when no actions are defined

lianah:
## Description of changes
Updated schema validation in protobuf to report an error if the Action type is uninhabited i.e. if not actions are defined.

### DIFF
--- a/cedar-policy-core/src/validator/schema.rs
+++ b/cedar-policy-core/src/validator/schema.rs
@@ -2421,9 +2421,8 @@ pub(crate) mod test {
 
         // User has an attribute referencing Foo::Action, and an action of type
         // Foo::Action is declared.
-        let attr_type = AttributeType::required_attribute(Arc::new(Type::named_entity_reference(
-            action_type,
-        )));
+        let attr_type =
+            AttributeType::required_attribute(Arc::new(Type::named_entity_reference(action_type)));
         let context = Type::record_with_required_attributes([], OpenTag::ClosedAttributes);
         let schema = ValidatorSchema::new(
             [ValidatorEntityType::new_standard(
@@ -2434,7 +2433,14 @@ pub(crate) mod test {
                 None,
                 None,
             )],
-            [ValidatorActionId::new(action_uid, [], [], [], context, None)],
+            [ValidatorActionId::new(
+                action_uid,
+                [],
+                [],
+                [],
+                context,
+                None,
+            )],
         );
 
         assert_matches!(schema.try_validate(), Ok(_));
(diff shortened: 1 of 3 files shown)


john-h-kastner-aws (approved):

cedar-policy-core/src/validator/schema.rs
```diff
@@ -1192,11 +1192,20 @@ impl ValidatorSchema {
...
+            .action_ids
+            .keys()
+            .map(|uid| uid.entity_type())
+            .collect();
         let mut undeclared = Vec::new();
         for ty in self.leaf_types().unique() {
             if let Type::Entity(EntityKind::Entity(lub)) = ty {
                 for e in lub.iter() {
-                    if !self.is_known_entity_type(e) {
+                    if !self.entity_types.contains_key(e) && !action_entity_types.contains(e) {
```
victornicolet, line 1216: Are the other `is_known_entity_type` usages valid?  I mean, should the fix be in that predicate?

cedar-policy-core/src/validator/schema.rs
```diff
@@ -2393,10 +2404,106 @@ pub(crate) mod test {
...
+                None,
+            )],
+        );
+
         assert_matches!(schema.try_validate(), Ok(_));
     }
 
+    /// Regression test for the `protobuf-decode-bytes-schema` fuzz failure
+    /// (nightly 2026-08-09): a schema like `namespace gz { entity sX tags Action; }` has an
+    /// uninhabited `Action` tag type with no declared action, so it must be rejected (see
```
victornicolet, line 2458: nit: I don't know if we should ref fuzzing failures here.

cedar-policy-core/src/validator/schema.rs
lianah: Good question. I played around with moving the check here and while the tests pass, it results in a change in validation behavior for the following policy:
```
permit(principal, action == Action::"foo", resource);
``` 
and schema:
```
namespace NS {
    entity E;
    action "foo" appliesTo { principal: E, resource: E };
}
```
Currently we validate actions in `validate_action_ids` so this results in a `UnrecognizedActionId` with a hint for `NS::Action::"foo"`. If we update `is_known_entity_type` to check actions we also get an additional `UnrecognizedEntityType`. This seems like a less desirable error reporting behavior.

victornicolet: Ok, yes this seems less desirable. 
Maybe a comment on `is_known_entity_type` would be enough so that we don't repeat the mistake of using it as an oracle of "this is a known entity type" like I did.

lianah: Also added a test for the error behavior I mentioned, since there wasn't one.

victornicolet (approved):

victornicolet (approved):

PR #1939, opened 2025-11-07
Fixed bug in identifier escaping not handling whitespaces.

lianah:
## Checklist for requesting a review

The change in this PR is (choose one, and delete the other options):

- [ ] A breaking change requiring a major version bump to `cedar-policy` (e.g., changes to the signature of an existing API).
- [ ] A backwards-compatible change requiring a minor version bump to `cedar-policy` (e.g., addition of a new API).
- [ ] A bug fix or other functionality change requiring a patch to `cedar-policy`.
- [x] A change "invisible" to users (e.g., documentation, changes to "internal" crates like `cedar-policy-core`, `cedar-validator`, etc.)
- [ ] A change (breaking or otherwise) that only impacts unreleased or experimental code.

I confirm that this PR (choose one, and delete the other options):

- [ ] Updates the "Unreleased" section of the CHANGELOG with a description of my change (required for major/minor version bumps).
- [x] Does not update the CHANGELOG because my change does not significantly impact released code.

I confirm that [`cedar-spec`](https://github.com/cedar-policy/cedar-spec) (choose one, and delete the other options):

- [x] Does not require updates because my change does not impact the Cedar formal model or DRT infrastructure.
- [ ] Requires updates, and I have made / will make these updates myself. (Please include in your description a timeline or link to the relevant PR in `cedar-spec`, and how you have tested that your updates are correct.)
- [ ] Requires updates, but I do not plan to make them in the near future. (Make sure that your changes are hidden behind a feature flag to mark them as experimental.)
- [ ] I'm not sure how my change impacts `cedar-spec`. (Post your PR anyways, and we'll discuss in the comments.)

I confirm that [`docs.cedarpolicy.com`](https://docs.cedarpolicy.com/) (choose one, and delete the other options):

- [x] Does not require updates because my change does not impact the Cedar language specification.
- [ ] Requires updates, and I have made / will make these updates myself. (Please include in your description a timeline or link to the relevant PR in [`cedar-docs`](https://github.com/cedar-policy/cedar-docs). PRs should be targeted at a `staging-X.Y` branch, not `main`.)
- [ ] I'm not sure how my change impacts the documentation. (Post your PR anyways, and we'll discuss in the comments.)

PR #2564, opened 2026-09-12
Specialize `from_normalized_str` for `InternalName`, `UnreservedId`, `AnyId` and `EntityUID`

oleksii-udovychenko:
### Problem

Checking that a string is already normalized costs a full parse plus a full render:

- [`default_from_normalized_str`](https://github.com/cedar-policy/cedar/blob/2f4019fd645cc8d4a4c0c1f8bd0280c77d754e28/cedar-policy-core/src/from_normalized_str.rs#L51) calls `T::from_str(s)`, renders the result back out with `to_string()`, and compares the two strings.
- `Name` (#1456) and `Id` (#1939) avoid that with specialized implementations. Four are left in the list @john-h-kastner-aws posted on the issue: `InternalName`, `UnreservedId`, `AnyId` and `EntityUID`.

### Before the Change

`EntityUID::from_str` on `User::"alice"`, 13 bytes, pays:

- a lalrpop parse into a CST, then a second pass in `cst.to_ref()`
- `to_string()` to render the result back out
- a byte comparison, the only part that was actually asked for
- three copies of the string plus the CST nodes, 7.19 µs total

### After the Change

Each of the four scans the input, falling back on `default_from_normalized_str` when it can't prove the input is normalized, so error messages are unchanged.

- `EntityUID` matches the `Type::"eid"` shape: strip the trailing quote, split on `::"`, check the eid survives escaping unchanged. [`Eid::escaped()`](https://github.com/cedar-policy/cedar/blob/2f4019fd645cc8d4a4c0c1f8bd0280c77d754e28/cedar-policy-core/src/ast/entity.rs#L437-L440) is `str::escape_debug`, so an eid it leaves alone is the one `Display` would have written. That also makes splitting on the first `::"` safe, since it rules out an unescaped `"` closing the literal early.
- `UnreservedId` and `AnyId` reuse `Id`'s identifier check. The reserved-word set splits in two, since an `InternalName` may contain `__cedar` where a `Name` may not, and an `AnyId` may be any keyword.
- `InternalName` duplicates about eight lines rather than sharing a helper with `Name`. Sharing one cost 6 to 8% on short `EntityTypeName::from_str`, outside the noise floor below, and `#[inline]` didn't recover it. `Name` is untouched byte for byte.

### Benchmarks

How the numbers below were measured:

- New `EntityUid parsing` group in `entity_parsing.rs`, beside the `EntityTypeName parsing` one from #1456.
- `before` is that benchmark file applied alone on the merge-base, so it measures the same six inputs on the old path.
- criterion 0.8.2 at defaults, 100 samples, means read from `target/criterion/*/estimates.json`, on a Xeon 8488C.

```
cargo bench -p cedar-policy --bench entity_parsing -- --save-baseline before   # merge-base + bench file
cargo bench -p cedar-policy --bench entity_parsing -- --baseline before        # this branch
```

| input | before | after | |
|---|---|---|---|
| `User::"alice"` | 7.19 µs | 242 ns | 29.7x |
| `Namespace::User::"alice"` | 8.39 µs | 299 ns | 28.1x |
| six namespaces | 9.49 µs | 550 ns | 17.3x |
| uuid eid | 8.48 µs | 399 ns | 21.3x |
| 90-char eid | 9.08 µs | 608 ns | 14.9x |
| `User::"alice\nbob"` | 7.89 µs | 7.94 µs | 1.0x |

Two things the table doesn't show on its own:

- The last row needs escaping, so it takes the fallback and prices it at +0.7%.
- `EntityTypeName parsing` is flat across all five sizes (-3.1% to +2.3%). Two builds of byte-identical source move that group by -3.3% to +2.7% here, so that spread is build noise.

### Testing

Each module gets a test asserting the specialized path and `default_from_normalized_str` agree on the same inputs, both on acceptance and on the parsed value.

- `normalized_ids_match_default` (19 inputs), `normalized_names_match_default` (19), `normalized_euid_matches_default` (22).
- Error messages match by construction, since every rejection delegates to the default.
- The euid inputs target the new scan: `Namespace::Type::"a::b::c"` and `A::"A::B::\"quoted\""` for an eid holding `::` and an escaped quote, `A::"a"b"` for an unescaped one, plus `A::"é"`, `A::""`, `A::"\u{0041}"`, and non-UIDs like `A::B`.
- `fmt`, `clippy --all-features` and `doc` are clean, and `cargo test -p cedar-policy-core -p cedar-policy` is 2047 passed.
- No `cvc5` installed locally, so `cedar-policy-symcc` and most of `cedar-policy-cli` are down to CI.

## Issue #, if available

#1411


cedar-policy-core/src/ast/name.rs
```diff
@@ -665,6 +701,44 @@ mod test {
...
+            "foo :: bar",
+            "foo::",
+            "::foo",
+            "foo:bar",
+            "",
+        ] {
+            assert_eq!(
+                InternalName::from_normalized_str(s).ok(),
+                default_from_normalized_str::<InternalName>(s, InternalName::describe_self).ok(),
+                "specialized and default `from_normalized_str()` disagree on {s:?}"
```
victornicolet, line 734: nit: `from_normalized_str`'s default will be to call the function we are testing against, so the test could be vacuous. Right now only the coverage indicates we are testing the proper branches. Maybe having an explicit function that does the parsing could help, but it also makes it more complex for no good reason.

victornicolet (approved):

cedar-policy-core/src/ast/name.rs
oleksii-udovychenko: Thanks! Will keep this in mind for the next change or in case I need to revise this one.

john-h-kastner-aws (approved):

oleksii-udovychenko:
Rebased on the latest due to conflicts.

victornicolet (approved):

